PR #2, opened 2015-01-15
Introduce the minimum API for the AMD development in ckeditor5 (#1)

### LINKED ISSUE #1: Introduce the minimum API for the AMD development in ckeditor5
There is a minimum API necessary for the development of the AMD API in the ckeditor5 repo:

 * `ckeditor` module (CKEDITOR namespace).
 * `utils` module with the `extend()` method.
 * `plugin` module (Plugin class).

fredck:
Fixes #1.

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -17,6 +17,9 @@ module.exports = function( grunt ) {
 
 		jshint: {
 			options: {
+				globals: {
+					'CKEDITOR': false
+				},
 				ignores: ignoreFiles
 			}
 		},
--- a/src/utils.js
+++ b/src/utils.js
@@ -23,7 +23,7 @@ CKEDITOR.define( function() {
 		 * @returns {Object} The `target` object.
 		 */
 		extend: function( target, source ) {
-			if ( !this.isObject( source ) && !this.isFunction( source ) ) {
+			if ( !this.isObject( source ) ) {
 				return target;
 			}
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,0 +1,72 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/**
+ * A utilities library.
+ *
+ * @class utils
+ * @singleton
+ */
+
+CKEDITOR.define( function() {
+	return {
+		/**
+		 * Extends one JavaScript object with the properties defined in one or more objects. Existing properties are
+		 * overridden.
+		 *
+		 * @param {Object} target The object to be extended.
+		 * @param {Object} source One or more objects which properties will be copied (by reference) to `target`.
+		 * @returns {Object} The `target` object.
+		 */
+		extend: function( target, source ) {
+			if ( !this.isObject( source ) && !this.isFunction( source ) ) {
+				return target;
+			}
+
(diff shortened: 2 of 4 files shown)


src/utils.js
```diff
@@ -0,0 +1,72 @@
...
+'use strict';
+
+/**
+ * A utilities library.
+ *
+ * @class utils
+ * @singleton
+ */
+
+CKEDITOR.define( function() {
```
Reinmar: Just to remember - we need to make a decision about https://github.com/ckeditor/ckeditor5-design/issues/13 and implement it sooner than later to have clear situation what tools and how can we use. It doesn't of course affect this review, but we should keep this in mind.

gruntfile.js
```diff
@@ -17,6 +17,9 @@ module.exports = function( grunt ) {
 
 		jshint: {
 			options: {
+				globals: {
+					'CKEDITOR': false
```
Reinmar: Why `false`? And why `CKEDITOR` in quotes?

src/utils.js
```diff
@@ -0,0 +1,72 @@
...
+		/**
+		 * Extends one JavaScript object with the properties defined in one or more objects. Existing properties are
+		 * overridden.
+		 *
+		 * @param {Object} target The object to be extended.
+		 * @param {Object} source One or more objects which properties will be copied (by reference) to `target`.
+		 * @returns {Object} The `target` object.
+		 */
+		extend: function( target, source ) {
+			if ( !this.isObject( source ) && !this.isFunction( source ) ) {
```
Reinmar: What is `isFunction` supposed to check here?

gregpabian: As the author of this code of code I'd say it doesn't make any sense. I must have been tired back then ;)

gruntfile.js
fredck: According to the Uglify docs:

> If value is false (default), JSHint will consider that variable as read-only.

As for the quotes, if you really made a question, just because it sounded like a list of "strings", not really properties of an object. :)

Reinmar: > If value is false (default), JSHint will consider that variable as read-only.

Good one... but absolutely unintuitive :D.

> As for the quotes, if you really made a question, just because it sounded like a list of "strings", not really properties of an object. :)

Makes sense. @pjasiun will like it ;> We can mention this in our style guide if you think it makes sense there.

src/utils.js
fredck: Fixed with 0c5bc1a.

gruntfile.js
mlewand: Assigning `false` makes a perfect sense.

But as for using quotes for keys, I would second to @Reinmar.

I see that you'd like to give a special treatment to `dict` objects, but lets keep it simple. Object is an object, and the same rules apply everywhere.

Keeping consistency makes searching properties initializations much easier - you just make ctrl + f on whole directory with `<propName>:` and you hit all the occurences.

fredck: > Keeping consistency makes searching properties initializations much easier - you just make ctrl + f on whole directory with <propName>: and you hit all the occurences.

The only thing is that you'll never work on such properties directly (like retrieving `globals.CKEDITOR`). You'll just work on the list as a whole, so quick access to the properties is usually useless.

Still I don't see a big deal for having it without quotes. I was just justifying why quotes ended up there. You guys may decide to remove them.

PR #5, opened 2015-01-20
t/4: Added tests for "utils"

### LINKED ISSUE #4: Introduce tests for the currently developed features
Tests for current code must be introduced after/together with the following ckeditor5 issue:
https://github.com/ckeditor/ckeditor5/issues/1

fredck:
Fixes #4.

This PR follows depends on https://github.com/ckeditor/ckeditor5/issues/1.

### DIFF
--- a/tests/utils/utils.js
+++ b/tests/utils/utils.js
@@ -7,9 +7,9 @@
 
 'use strict';
 
-describe( 'extend()', function() {
-	var modules = bender.amd.require( 'utils' );
+var modules = bender.amd.require( 'utils' );
 
+describe( 'extend()', function() {
 	it( 'should extend and override', function() {
 		var utils = modules.utils;
 
@@ -97,8 +97,6 @@ describe( 'extend()', function() {
 } );
 
 describe( 'isFunction()', function() {
-	var modules = bender.amd.require( 'utils' );
-
 	it( 'should be true for functions only', function() {
 		var utils = modules.utils;
 
@@ -123,8 +121,6 @@ describe( 'isFunction()', function() {
 } );
 
 describe( 'isObject()', function() {
-	var modules = bender.amd.require( 'utils' );
-
 	it( 'should be true for pure objects only', function() {
 		var utils = modules.utils;
 
--- a/tests/utils/utils.js
+++ b/tests/utils/utils.js
@@ -7,130 +7,148 @@
 
 'use strict';
 
-bender.amd.require( 'utils' );
+describe( 'extend()', function() {
+	it( 'should extend and override', function( done ) {
+		CKEDITOR.require( [ 'utils' ], function( utils ) {
+			var target = {
+				a: 1,
+				b: 2
+			};
 
-describe( 'extend()', function( utils ) {
-	it( 'should extend and override', function() {
-		var target = {
-			a: 1,
-			b: 2
-		};
+			var extensions = {
+				a: 'A',
+				c: 3,
 
-		var extensions = {
-			a: 'A',
-			c: 3,
+				// Extend by reference (no deep-copy).
+				obj: {
+					a: 1
+				},
 
-			// Extend by reference (no deep-copy).
-			obj: {
+				// Extend by reference (no deep-copy).
+				arr: [ 1, 2 ],
+
+				// Extend by reference.
+				fn: function() {}
+			};
+
+			var ret = utils.extend( target, extensions );
+
+			expect( target.a ).to.equal( 'A' );
+			expect( target.b ).to.equal( 2 );
+			expect( target ).to.have.property( 'c' ).to.equal( 3 );
+
+			expect( target ).to.have.property( 'obj' ).to.equal( extensions.obj );
+			expect( target ).to.have.property( 'arr' ).to.equal( extensions.arr );
+			expect( target ).to.have.property( 'fn' ).to.equal( extensions.fn );
+
+			// "target" should be the return value.
+			expect( ret ).to.equal( target );
+
+			done();
+		} );
+	} );
+
+	it( 'should not be touched by non-objects', function( done ) {
+		CKEDITOR.require( [ 'utils' ], function( utils ) {
+			var target = {
 				a: 1
-			},
+			};
 
-			// Extend by reference (no deep-copy).
-			arr: [ 1, 2 ],
+			expect( utils.extend( target, function() {} ) ).to.equal( target );
+			expect( utils.extend( target, 1 ) ).to.equal( target );
+			expect( utils.extend( target, 'a' ) ).to.equal( target );
+			expect( utils.extend( target, true ) ).to.equal( target );
+			expect( utils.extend( target, undefined ) ).to.equal( target );
+			expect( utils.extend( target, [] ) ).to.equal( target );
+			expect( utils.extend( target, Date.now() ) ).to.equal( target );
+			expect( utils.extend( target ) ).to.equal( target );
 
-			// Extend by reference.
-			fn: function() {}
-		};
+			// None of the above calls should have touched "target".
+			expect( target ).to.have.property( 'a' ).to.equal( 1 );
+			expect( Object.getOwnPropertyNames( target ).length ).to.equal( 1 );
 
-		var ret = utils.extend( target, extensions );
-
-		expect( target.a ).to.equal( 'A' );
-		expect( target.b ).to.equal( 2 );
-		expect( target ).to.have.property( 'c' ).to.equal( 3 );
-
-		expect( target ).to.have.property( 'obj' ).to.equal( extensions.obj );
-		expect( target ).to.have.property( 'arr' ).to.equal( extensions.arr );
-		expect( target ).to.have.property( 'fn' ).to.equal( extensions.fn );
-
-		// "target" should be the return value.
-		expect( ret ).to.equal( target );
+			done();
+		} );
 	} );
 
-	it( 'should not be touched by non-objects', function() {
-		var target = {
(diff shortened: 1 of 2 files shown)


tests/utils/utils.js
```diff
@@ -0,0 +1,153 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals describe, it, expect */
+
+'use strict';
+
+describe( 'extend()', function() {
```
gregpabian: I'd wrap all of these describes with one more:
```
describe('Utils', ...
```

Then it would produce clearer results, like:
"Utils extend() should extend and override"

tests/utils/utils.js
```diff
@@ -0,0 +1,153 @@
...
+			expect( utils.isObject( [] ) ).to.be.true();
+
+			expect( utils.isObject( 1 ) ).to.be.false();
+			expect( utils.isObject( 'a' ) ).to.be.false();
+			expect( utils.isObject( true ) ).to.be.false();
+			expect( utils.isObject( undefined ) ).to.be.false();
+			expect( utils.isObject( Date.now() ) ).to.be.false();
+			expect( utils.isObject( f1 ) ).to.be.false();
+			expect( utils.isObject( f2 ) ).to.be.false();
+
```
gregpabian: I think you should add a test for `null` :trollface:

tests/utils/utils.js
fredck: Do we really need to make the file more bloated, considering that we're already inside the "utils" test page?

Then, the correct name is "utils" (lowercased) and this would make things ackward because there is no dot between "utils" and "extend()": 
"utils extend() should extend and override"

So I think that we should either change the current describe to "utils.extend()" or leave it as it is  now (my option).

tests/utils/utils.js
fredck: True :) just pushed it.

gregpabian:
I've been thinking about all those `CKEDITOR.require` calls and in this case we can use mocha's hooks to simplify the code:
```js
describe( 'extend()', function() {
	var utils;

	// hooks can be asynchronous so we can do the following
	beforeEach( function( done ) {
		CKEDITOR.require( [ 'utils' ], function( u ) {
			utils = u;
			done();
		} );
	} );

	// not async so no need for done callback, yay!
	it( 'should extend and override', function() {
		var target = {
			a: 1,
			b: 2
		};

		var extensions = {
			a: 'A',
			c: 3
		};

		var ret = utils.extend( target, extensions );

		expect( target.a ).to.equal( 'A' );
		...
	} );
} );
```

Besides, I don't think we have to require the library for every test, so we can replace `beforeEach()` with `before()`.

What do you think about this?

Reinmar:
> What do you think about this?

Looks good, but I wonder if we can simplify this even further by creating some helper. It could work this way:

```
var modules;

before( function( done ) {
    doSomeStuffThatYouNeed();
    bender.amd.require( done, modules, [ 'Promise', 'utils' ] );
} );

// modules.Promise;
// modules.utils;
```

And even simpler, if one does not need the `before()` callback:

```
var modules = bender.amd.require( [ 'Promise', 'utils' ] );
```

And the `require()` function, if only one argument was passed, would use `before()` internally.

Reinmar:
One more thing that we'll need to understand - how can we inject some dependencies by overriding what `CKEDITOR.require` returns when loading a module. Most likely it could look like this:

```
it( 'test Foo with mocked Bar', function() {
    bender.amd.inject( 'Bar', ourMockModule );

   CKEDITOR.require( 'Foo', function( Foo ) {
      // Foo internally requires Bar and since we injected ourMockModule
      // it should now use our mock instead of the real Bar.
      done(); 
   } );
} );
```

And the injection should be reverted in `afterEach/beforeEach`, so this TC does not affect the next one.

Does it make sense?

PS. I guess there must be some library that has these features :D.

gregpabian:
> bender.amd.require( done, modules, [ 'Promise', 'utils' ] );

You couldn't think of a worse argument order, could you? :trollface:

Reinmar:
> You couldn't think of a worse argument order, could you?

My apologies :D. I was changing it too many times. Actually, to unify with the 2nd version I would do:

```
modules = bender.amd.require( [ 'Promise', 'utils' ], done );
```

gregpabian:
> how can we inject some dependencies

I think the most common way is to use [Require's mappings](http://requirejs.org/docs/api.html#config-map), but if we decide to come out with our own helper for this purpose, then I'm OK with that.

fredck:
What do you guys think about this?

```
bender.injectModules( 'ckeditor', 'utils' );

describe( 'test', function( ckeditor, utils ) {
    ...
}
```

Basically, the ckeditor5 plugin for bender will add `injectModules` and override `describe` to have them as parameters.

Reinmar:
> I think the most common way is to use Require's mappings, but if we decide to come out with our own helper for this purpose, then I'm OK with that.

I only briefly scanned the mappings doc, but it does not look like a big help, because it'd need to be per file. I would love to have something:

1. Automated.
2. Local.
3. Temporary.

Reinmar:
> Basically, the ckeditor5 plugin for bender will add injectModules and override describe to have them as parameters.

The API looks cool. If it's possible to code it this way (`injectModules` is outside of `describe`), then it'd be a way to go. The other thing is the method name - injecting relates to dependency injection and that's the second feature we need, so I would simply call it `bender.amd.require()`.

fredck:
@Reinmar's [comment](https://github.com/ckeditor/ckeditor5-core/pull/5#issuecomment-70990318):

> Temporary.

I think it is not possible.

Once RequireJS returns `Foo`, from your example, it'll always return that same object. It'll not "recreate" it every time require() is called, so you cannot replace `Bar` inside it so dynamically.

We may figure out a way to make this module override happen once for the whole test page... or even check if there is not really a way to force RequireJS to "forget" the previous module definition and recreate it.

fredck:
> so I would simply call it bender.amd.require().

Cool, I'll give it a try and possibly update the PRs.

fredck:
Btw guys, we're dirtying up this PR too much with side talks. The discussion about module overriding should go somewhere else in an issue.

fredck:
Just pushed `bender.amd.require()` in [the relative ckeditor5 PR](https://github.com/ckeditor/ckeditor5/pull/11), updating this one accordingly.

You gonna love this :) (I hope!)

fredck:
Ok, updated ckeditor5 with the new strategy for `bender.amd.require()` and updated the tests accordingly.

Now the remaining topic... as you can see, every `describe()` in the same test page has its own call to `bender.amd.require()`, repetitively. This is done to not pollute the global namespace and to avoid making things uglier with a big closure.

Options:

 1. Leave it this way.
 2. Pollute the global namespace with this one and only variable named `modules`.
 3. Use a big and ugly closure only when we have several describe()s that share the same modules.
 4. Use a big and ugly closure always.

I would be definitely ok with "2" as it is simple and it should not be a big deal to have the global a bit dirty (but under control) in our tests.

Reinmar:
AFAIR Bender wraps a test file with a closure so the global namespace should not be polluted. But even if not, I'm still ok with 2.

gregpabian:
> Bender wraps a test file with a closure so the global namespace should not be polluted

That's true - the test script lands in a closure so it won't be a problem.

fredck:
> That's true - the test script lands in a closure so it won't be a problem.

Wonderful! Just changed this. We should have the final version ready for review now.

tests/utils/utils.js
```diff
@@ -0,0 +1,146 @@
...
+			a: 1
+		};
+
+		expect( utils.extend( target, function() {} ) ).to.equal( target );
+		expect( utils.extend( target, 1 ) ).to.equal( target );
+		expect( utils.extend( target, 'a' ) ).to.equal( target );
+		expect( utils.extend( target, true ) ).to.equal( target );
+		expect( utils.extend( target, undefined ) ).to.equal( target );
+		expect( utils.extend( target, [] ) ).to.equal( target );
+		expect( utils.extend( target, Date.now() ) ).to.equal( target );
```
gregpabian: `Date.now()` returns a number, so what's the point of this case since the number is being tested few lines earlier?

tests/utils/utils.js
```diff
@@ -0,0 +1,146 @@
...
+		expect( utils.isFunction( f1 ) ).to.be.true();
+		expect( utils.isFunction( f2 ) ).to.be.true();
+
+		expect( utils.isFunction( 1 ) ).to.be.false();
+		expect( utils.isFunction( 'a' ) ).to.be.false();
+		expect( utils.isFunction( true ) ).to.be.false();
+		expect( utils.isFunction( undefined ) ).to.be.false();
+		expect( utils.isFunction( [] ) ).to.be.false();
+		expect( utils.isFunction( {} ) ).to.be.false();
+		expect( utils.isFunction( Date.now() ) ).to.be.false();
```
gregpabian: Again, you're testing a number which has already been tested few lines earlier.

fredck:
@gregpabian, I've fixed your last comments. My mistake.

Reinmar:
> Once RequireJS returns Foo, from your example, it'll always return that same object. It'll not "recreate" it every time require() is called, so you cannot replace Bar inside it so dynamically.

Just found this:

> There is a global function, requirejs.undef(), that allows undefining a module. It will reset the loader's internal state to forget about the previous definition of the module.

http://requirejs.org/docs/api.html#undef

PR #8, opened 2015-01-21
t/6: Introduced the Promise class

### LINKED ISSUE #6: Introduce promises lib
Will close ckeditor/ckeditor5-design#43

fredck:
Closes #6.

### DIFF
--- a/src/promise.js
+++ b/src/promise.js
@@ -8,13 +8,13 @@
 'use strict';
 
 /**
- * A ES6 compatible Promise class, used for deferred and asynchronous computations.
+ * An ES6 compatible Promise class, used for deferred and asynchronous computations.
  *
  * @class Promise
  */
 
 CKEDITOR.define( function() {
-	// For now we're using the native browser implementation of Promise, a ES6 feature. Just IE is not supporting it so
+	// For now we're using the native browser implementation of Promise, an ES6 feature. Just IE is not supporting it so
 	// a polyfill will have to be developed for it.
 	//
 	// http://caniuse.com/#feat=promises
--- a/src/promise.js
+++ b/src/promise.js
@@ -1,0 +1,27 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals window */
+
+'use strict';
+
+/**
+ * A ES6 compatible Promise class, used for deferred and asynchronous computations.
+ *
+ * @class Promise
+ */
+
+CKEDITOR.define( function() {
+	// For now we're using the native browser implementation of Promise, a ES6 feature. Just IE is not supporting it so
+	// a polyfill will have to be developed for it.
+	//
+	// http://caniuse.com/#feat=promises
+
+	if ( !window.Promise ) {
+		throw 'The Promise class is not available natively. CKEditor is not compatible with this browser.';
+	}
+
+	return window.Promise;
+} );
--- a/tests/promise/promise.js
+++ b/tests/promise/promise.js
@@ -1,0 +1,29 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals describe, it, expect, CKEDITOR, window */
+
+'use strict';
+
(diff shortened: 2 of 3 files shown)


tests/promise/promise.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals describe, it, expect, CKEDITOR, window */
+
+'use strict';
+
+describe( 'Promise', function() {
```
Reinmar: This test is... weird and IMO overcomplicated. Just check that the API exists and does the most basic thing (no timeouts).

Besides that I don't like `CKEDITOR.require` call inside the test function - it should be as simple as possible. I don't know the best practices of testing with AMD (and we should check them), but I would for instance move the require() call outside at least of `it()` if not of `describe()`.

BTW. I've stumbled upon http://rangle.io/blog/the-art-of-promise-based-architecture/ last week. The 2nd part of the video is great and touches testing (actually - one thing - that you can return a promise and Mocha will resolve it).

fredck: Well, the test is doing the most basic thing: it creates a Promise and consumes it. I've made it async just to make it more realistic and the setTimeout inside then() is mandatory.

As for "CKEDITOR.require call inside the test function", this is a broader topic that doesn't touch this PR. I've been talking with @gregpabian and we'll be investigating how to change this because this is touching every single test. Still it should not be a blocker here.

Then, when it comes to the returning Promises to Mocha, this is to consider and it requires [Chai as Promised](https://github.com/domenic/chai-as-promised/). I talked with @gregpabian about this as well. Still here we want to test the basics of Promise so this kind of magix is out of scope.

Reinmar: > Then, when it comes to the returning Promises to Mocha, this is to consider and it requires Chai as Promised. I talked with @gregpabian about this as well. Still here we want to test the basics of Promise so this kind of magix is out of scope.

See: https://github.com/mochajs/mocha/pull/329 (especially the code sample in https://github.com/mochajs/mocha/pull/329#issuecomment-4546922). You don't need anything else here - just make the promise fail:

```
return promise.then( function( value ) {
    expect( value ).to.equal( 1 );
} );
```

Reinmar: > Well, the test is doing the most basic thing: it creates a Promise and consumes it. I've made it async just to make it more realistic and the setTimeout inside then() is mandatory.

We don't need more realism here - we're not testing our code - we're testing either native implementations (which we have to believe work correctly) or polyfills (which we have to believe work correctly too).

Besides that, I don't think we should use timeouts if we don't need to. Here you can do:

```
return new Promise( function( resolve ) {
    resolve( 1 );
} ).then( function( value ) {
    expect( value ).to.equal( 1 );
} );
```

Or even:

```
return Promise.resolve( 1 ).then( function( value ) {
    expect( value ).to.equal( 1 );
} );
```

Reinmar: Of course, the above is valid only if `CKEDITOR.require` is used outside `it()` - that's why I mentioned this aspect too.

fredck: Pushed a commit that simplified things. Still a bit more verbose than your super-compact option :)

New things learned ;)

src/promise.js
```diff
@@ -0,0 +1,27 @@
...
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals window */
+
+'use strict';
+
+/**
+ * A ES6 compatible Promise class, used for deferred and asynchronous computations.
```
gregpabian: "A**n** ES6 compatible Promise (...)"

src/promise.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @class Promise
+ */
+
+CKEDITOR.define( function() {
+	// For now we're using the native browser implementation of Promise, a ES6 feature. Just IE is not supporting it so
```
gregpabian: "(...) a**n** ES6 feature."

fredck: Fixed with 882f73b.

src/promise.js
fredck: Fixed with 882f73b.

Issue #7, opened 2015-01-21
Introduce utils lib


fredck:
Will close ckeditor/ckeditor5-design#13

fredck:
I've just pushed https://github.com/cksource/ckeditor5-core/tree/t/7 with a proposal for this. Waiting to close #4 to start a PR for it.

PR #12, opened 2015-02-06
t/11: Introduced the Emitter mixin with the events API

### LINKED ISSUE #11: Introduce the Events API
According to the following talk: https://github.com/ckeditor/ckeditor5-design/issues/7

fredck:
Fixes #11.

We still may have discussion around this topic, so changes may (or may not still happen).

Tests included. As of today, to run them, the `ckeditor5` repo must be a t/1.

### DIFF
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -107,8 +107,8 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 		 * @param {String} event The name of the event.
 		 * @param {Function} callback The function to be called on event.
 		 * @param {Object} [ctx] The object that represents `this` in the callback. Defaults to `emitter`.
-		 * @param {Number} [priority] The priority of this callback in relation to other callbacks to that same event.
-		 * Lower values are called first. Defaults to `10`.
+		 * @param {Number} [priority=10] The priority of this callback in relation to other callbacks to that same event.
+		 * Lower values are called first.
 		 */
 		listenTo: function( emitter, event, callback, ctx, priority ) {
 			var emitters, emitterId, emitterInfo, eventCallbacks;
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -79,8 +79,10 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 		 *
 		 * @param {String} event The name of the event.
 		 * @param {Function} callback The function to stop being called.
+		 * @param {Object} [ctx] The context object to be removed, pared with the given callback. To handle cases where
+		 * the same callback is used several times with different contexts.
 		 */
-		off: function( event, callback ) {
+		off: function( event, callback, ctx ) {
 			var callbacks = getCallbacksIfAny( this, event );
 
 			if ( !callbacks ) {
@@ -89,9 +91,11 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 
 			for ( var i = 0; i < callbacks.length; i++ ) {
 				if ( callbacks[ i ].callback == callback ) {
-					// Remove the callback from the list (fixing the next index).
-					callbacks.splice( i, 1 );
-					i--;
+					if ( !ctx || ctx == callbacks[ i ].ctx ) {
+						// Remove the callback from the list (fixing the next index).
+						callbacks.splice( i, 1 );
+						i--;
+					}
 				}
 			}
 		},
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -177,14 +177,14 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 			if ( callback ) {
 				emitter.off( event, callback );
 			}
-			// Only emitter and event provided. off() all callbacks for that event.
+			// Only the emitter and event provided. off() all callbacks for that event.
 			else if ( eventCallbacks ) {
 				while ( ( callback = eventCallbacks.pop() ) ) {
 					emitter.off( event, callback );
 				}
 				delete emitterInfo.callbacks[ event ];
 			}
-			// Only emitter provided. off() all events for that emitter.
+			// Only the emitter provided. off() all events for that emitter.
 			else if ( emitterInfo ) {
 				for ( event in emitterInfo.callbacks ) {
 					this.stopListening( emitter, event );
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -39,15 +39,15 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 			};
 
 			// Add the callback to the list in the right priority position.
-			for ( var i = 0; i < callbacks.length; i++ ) {
-				if ( callbacks[ i ].priority > priority ) {
-					callbacks.splice( i, 0, callback );
+			for ( var i = callbacks.length - 1; i >= 0; i-- ) {
+				if ( callbacks[ i ].priority <= priority ) {
+					callbacks.splice( i + 1, 0, callback );
 
 					return;
 				}
 			}
 
-			callbacks.push( callback );
+			callbacks.unshift( callback );
 		},
 
 		/**
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -26,7 +26,6 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 		 */
 		on: function( event, callback, ctx, priority ) {
 			var callbacks = getCallbacks( this, event );
-			var wasAdded;
 
 			// Priority defaults to 10.
 			if ( typeof priority != 'number' ) {
@@ -43,14 +42,12 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 			for ( var i = 0; i < callbacks.length; i++ ) {
 				if ( callbacks[ i ].priority > priority ) {
 					callbacks.splice( i, 0, callback );
-					wasAdded = true;
-					break;
+
+					return;
 				}
 			}
 
-			if ( !wasAdded ) {
-				callbacks.push( callback );
-			}
+			callbacks.push( callback );
 		},
 
 		/**
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -225,6 +225,7 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 			for ( var i = 0; i < callbacks.length; i++ ) {
 				callbacks[ i ].callback.apply( callbacks[ i ].ctx, args );
 
+				// Remove the callback from future requests if off() has been called.
 				if ( eventInfo.off.called ) {
 					// Remove the called mark for the next calls.
 					delete eventInfo.off.called;
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -225,11 +225,6 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 			for ( var i = 0; i < callbacks.length; i++ ) {
 				callbacks[ i ].callback.apply( callbacks[ i ].ctx, args );
 
-				// Do not execute next callbacks if stop() was called.
-				if ( eventInfo.stop.called ) {
-					break;
-				}
-
 				if ( eventInfo.off.called ) {
 					// Remove the called mark for the next calls.
 					delete eventInfo.off.called;
@@ -237,6 +232,11 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 					// Remove the callback from the list (fixing the next index).
 					callbacks.splice( i, 1 );
 					i--;
+				}
+
+				// Do not execute next callbacks if stop() was called.
+				if ( eventInfo.stop.called ) {
+					break;
 				}
 			}
 		}
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -21,13 +21,13 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 		 * @param {Function} callback The function to be called on event.
 		 * @param {Object} [ctx] The object that represents `this` in the callback. Defaults to the object firing the
 		 * event.
-		 * @param {Number} [priority] The priority of this callback in relation to other callbacks to that same event.
-		 * Lower values are called first. Defaults to `10`.
+		 * @param {Number} [priority=10] The priority of this callback in relation to other callbacks to that same event.
+		 * Lower values are called first.
 		 */
 		on: function( event, callback, ctx, priority ) {
 			var callbacks = getCallbacks( this, event );
 
-			// Priority defaults to 10.
+			// Set the priority defaults.
 			if ( typeof priority != 'number' ) {
 				priority = 10;
 			}
@@ -58,8 +58,8 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 		 * @param {Function} callback The function to be called on event.
 		 * @param {Object} [ctx] The object that represents `this` in the callback. Defaults to the object firing the
 		 * event.
-		 * @param {Number} [priority] The priority of this callback in relation to other callbacks to that same event.
-		 * Lower values are called first. Defaults to `10`.
+		 * @param {Number} [priority=10] The priority of this callback in relation to other callbacks to that same event.
+		 * Lower values are called first.
 		 */
 		once: function( event, callback, ctx, priority ) {
 			var onceCallback = function( event ) {
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -9,7 +9,7 @@
  * Mixin that injects the events API into its host.
  *
  * @class Emitter
- * @static
+ * @singleton
  */
 
 CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
@@ -26,7 +26,6 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 		 */
 		on: function( event, callback, ctx, priority ) {
 			var callbacks = getCallbacks( this, event );
-
 			var wasAdded;
 
 			// Priority defaults to 10.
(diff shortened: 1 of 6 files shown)


src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+		 * @param {Number} [priority] The priority of this callback in relation to other callbacks to that same event.
+		 * Lower values are called first. Defaults to `10`.
+		 */
+		on: function( event, callback, ctx, priority ) {
+			var callbacks = getCallbacks( this, event );
+
+			var wasAdded;
+
+			// Priority defaults to 10.
+			if ( typeof priority != 'number' ) {
```
gregpabian: Right nowe we have `utils.isObject()` and `utils.isFunction()`, so why not introducing `isNumber()` and similar and use it here?

fredck: I was wondering if this is worth for such kind of basic checks.

Btw, we have isObject() and isFunction() not because we need them in general, just because utils itself needs it.

In fact, we're suppose to remove such functions from utils from now with #7. I'm just waiting for #4 to get merged to propose this change there (it's already pushed in t/7).

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+			this.on( event, onceCallback, ctx, priority );
+		},
+
+		/**
+		 * Stops executing the callback on the given event.
+		 *
+		 * @param {String} event The name of the event.
+		 * @param {Function} callback The function to stop being called.
+		 */
+		off: function( event, callback ) {
```
gregpabian: What if we had the same callback but executed in different contexts? This would remove all those handlers. I think we need an optional context to be passed.

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+			var eventCallbacks = emitterInfo && event && emitterInfo.callbacks[ event ];
+
+			// Stop if nothing has been listened.
+			if ( !emitters || ( emitter && !emitterInfo ) || ( event && !eventCallbacks ) ) {
+				return;
+			}
+
+			if ( callback ) {
+				// All params provided. off() that single callback.
+				emitter.off( event, callback );
```
gregpabian: Refers to my previous comment - this will remove all matching callbacks - no matter what's the context.

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+			} else if ( eventCallbacks ) {
+				// Only emitter and event provided. off() all callbacks for that event.
```
gregpabian: Missing articles.

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+				while ( ( callback = eventCallbacks.pop() ) ) {
+					emitter.off( event, callback );
+				}
+				delete emitterInfo.callbacks[ event ];
+			} else if ( emitterInfo ) {
+				// Only emitter provided. off() all events for that emitter.
```
gregpabian: Missing articles.

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+				// No params provided. off() all emitters.
+				for ( emitterId in emitters ) {
+					this.stopListening( emitters[ emitterId ].emitter );
+				}
+				delete this._listeningTo;
+			}
+		},
+
+		/**
+		 * Fires and event, executing all callbacks registered for it.
```
gregpabian: "Fires an~~d~~ event (...)"

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+		 *
+		 * The first parameter passed to callbacks is a {EventInfo} object, followed by the optional `args` provided in
```
gregpabian: "a**n** {EventInfo} (...)"

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+			var eventInfo = new EventInfo( event );
+
+			// Take the list of arguments to pass to the callbacks.
+			args = Array.prototype.slice.call( arguments, 1 );
+			args.unshift( eventInfo );
+
+			for ( var i = 0; i < callbacks.length; i++ ) {
+				callbacks[ i ].callback.apply( callbacks[ i ].ctx, args );
+
+				if ( eventInfo.stop.called ) {
```
gregpabian: Add a comment this one please.

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+					break;
+				}
+
+				if ( eventInfo.off.called ) {
```
gregpabian: Add a comment this one please.

tests/emitter/emitter.js
```diff
@@ -0,0 +1,356 @@
...
+
+		emitter.fire( 'test' );
+		emitter.fire( 'test' );
+
+		sinon.assert.calledThrice( spy1 );
+		sinon.assert.calledOnce( spy2 );
+		sinon.assert.calledThrice( spy3 );
+	} );
+
+	it( 'should no fail with unknown events', function() {
```
gregpabian: "should no**t** fail (...)"

tests/emitter/emitter.js
```diff
@@ -0,0 +1,356 @@
...
+		emitter.off( 'test', function() {} );
+	} );
+} );
+
+describe( 'listenTo', function() {
```
gregpabian: I miss a test for calling `listenTo` on non-emitter target.

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+					emitter.off( event, callback );
+				}
+				delete emitterInfo.callbacks[ event ];
+			} else if ( emitterInfo ) {
+				// Only emitter provided. off() all events for that emitter.
+				for ( event in emitterInfo.callbacks ) {
+					this.stopListening( emitter, event );
+				}
+				delete emitters[ emitterId ];
+			} else {
```
gregpabian: And what if we do the following:
1. Call listenTo with emitterA
2. Call stopListening with emitterB that we've never listened to

:trollface: ??

As a result we'll stop listening to the emitterA and every other emitter.

tests/emitter/emitter.js
```diff
@@ -0,0 +1,356 @@
...
+		emitter.fire( 'test' );
+
+		sinon.assert.called( spy );
+	} );
+} );
+
+describe( 'stopListening', function() {
+	beforeEach( refreshListener );
+
+	it( 'should stop listening callback on event', function() {
```
gregpabian: I'd rephrase this one:
"should stop listening to a specific event calback"

And besides I'm missing "listening **to**" in every other test case.

tests/emitter/emitter.js
```diff
@@ -0,0 +1,356 @@
...
+
+		listener.listenTo( emitter, 'test', spy );
+
+		emitter.fire( 'test' );
+
+		sinon.assert.called( spy );
+	} );
+} );
+
+describe( 'stopListening', function() {
```
gregpabian: Needs a test case for my trolly scenario: https://github.com/ckeditor/ckeditor5-core/pull/12/files#r24569343

tests/eventinfo/eventinfo.js
```diff
@@ -0,0 +1,48 @@
...
+		var EventInfo = modules.eventinfo;
+
+		var event = new EventInfo( 'test' );
+
+		expect( event.name ).to.equals( 'test' );
+		expect( event.stop.called ).to.not.be.true();
+		expect( event.off.called ).to.not.be.true();
+	} );
+
+	it( 'should have cancel() and off() marked', function() {
```
gregpabian: I don't see `cancel` anywhere. Can you update this description?

tests/utils/utils.js
```diff
@@ -0,0 +1,38 @@
...
+
+var modules = bender.amd.require( 'utils' );
+
+describe( 'spy', function() {
+	it( 'should register calls', function() {
+		var utils = modules.utils;
+
+		var fn1 = utils.spy();
+		var fn2 = utils.spy();
+
```
gregpabian: I think you miss a check if the `fn1` had the `called` property set to `false` after initialization.

src/emitter.js
Reinmar: I don't see a reason why we should replace `typeof` with a method call doing exactly the same.  Tools functions make sense only if a check is complicated or isn't straightforward (like the `Object.prototype.toString.call` hack). Checking if something is a number is a basic operation and we don't care about `new Number( 5 )` or `NaN`. Replacing everything with method calls is a road to having `assign( left, right )` ;).

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+	var EmmitterMixin = {
+		/**
+		 * Registers a callback function to be executed when an event is fired.
+		 *
+		 * @param {String} event The name of the event.
+		 * @param {Function} callback The function to be called on event.
+		 * @param {Object} [ctx] The object that represents `this` in the callback. Defaults to the object firing the
+		 * event.
+		 * @param {Number} [priority] The priority of this callback in relation to other callbacks to that same event.
+		 * Lower values are called first. Defaults to `10`.
```
Reinmar: I never liked the default == 10. It's like you have a space 0-9 to do something earlier, and a lot more space after that. But then you end up with -1, -2, etc., because with time it turns out that you need more... I feel that it will be cleaner to have default == 0.

src/emitter.js
gregpabian: So you want to do `obj !== null && typeof obj == 'object'` every time? Your call... I just want to stick to one convention, cause mix of typeofs and utils methods doesn't look so great (at least for me).

src/emitter.js
Reinmar: Totally +1. Please also see https://github.com/ckeditor/ckeditor5-design/issues/7#issuecomment-74497143.

src/emitter.js
Reinmar: The object case is a different case. I don't think that it makes sense to use `isNumber` and `isBoolean` everywhere just because `isObject` and `isRegExp` exist. It can really lead to `isNull` functions. But on the other hand I agree that having everything consistent is cool too, so actually part of me wants the `isNumber` method :D.

src/emitter.js
```diff
@@ -0,0 +1,275 @@
...
+			}
+
+			callback = {
+				callback: callback,
+				ctx: ctx || this,
+				priority: priority
+			};
+
+			// Add the callback to the list in the right priority position.
+			for ( var i = 0; i < callbacks.length; i++ ) {
```
Reinmar: The reverse priority search from https://github.com/ckeditor/ckeditor-dev/blob/master/core/event.js#L174 actually makes sense, because most listeners have the default priority.

Reinmar: Fixed with 939f5d2.

src/emitter.js
Reinmar: I'm for making the `off()` method more precise, but I wouldn't do that with `stopListening()`. Actually, I would even remove the `callback` (and maybe even the `event`) param and made it a function to "stop listening to: a) anything, b) given emitter)". If you need to be more precise just use `off()`. This should reduce `listenTo()` and `stopListening()` complexity which currently is pretty high and I'm not sure if we know that we need those features.

So the question is - are there any clear cases when `event` and `callback` may be useful and `off()` is not a solution?

src/emitter.js
Reinmar: Fixed with 61a8cfb.

src/emitter.js
Reinmar: Fixed with 61a8cfb.

src/emitter.js
fredck: I think it's more a backwards alignment, to not have to change everything in the API (even principles) for the sake of reviewing every single thing.

Ofc, if there is a general consensus we can change it, because it's the last occasion for it.

src/emitter.js
Reinmar: 41ad6dd

src/emitter.js
Reinmar: 41ad6dd

src/emitter.js
Reinmar: I tried really hard but it is pretty straightforward :P 41ad6dd.

src/emitter.js
Reinmar: Thing that worries me is that if `stop()` was called, then `off()` call will be ignored. It's not what I would expect.

src/emitter.js
Reinmar: I think that there's already enough API changes to force us to review every single line :D. But I don't also have a strong opinion about the default priority, so I'm curious if anyone else would like to see this change.

mlewand: As for me default priority 10 is fine. Listener priority is a kind of our thing so there is so we are free to use CKE4 value. It will be less confusing for developers to switch from 4 to 5.

src/emitter.js
```diff
@@ -0,0 +1,279 @@
...
+		 * @param {Function} callback The function to be called on event.
+		 * @param {Object} [ctx] The object that represents `this` in the callback. Defaults to the object firing the
+		 * event.
+		 * @param {Number} [priority] The priority of this callback in relation to other callbacks to that same event.
+		 * Lower values are called first. Defaults to `10`.
+		 */
+		on: function( event, callback, ctx, priority ) {
+			var callbacks = getCallbacks( this, event );
+
+			// Priority defaults to 10.
```
mlewand: This is just super minor thing, but couldn't we just write: "Priority default value."? That way you reduce the DRY factor and if by any means default value would change, you reduce the risk that the comment will be outdated.

I do belive that we're very unlikely to change this one, but it's a general suggestion.

src/emitter.js
gregpabian: BTW isn't there a syntax for the default value? Something like:
```
 @param {Number} [priority=10] The priority of this callback in relation to other callbacks to that same event.
```

That's how it looks in JSDoc at least.

Reinmar: > BTW isn't there a syntax for the default value? Something like:

This is also how it works in JSDuck.

tests/emitter/emitter.js
fredck: What's the point of this test? Ensure that it fails?

Reinmar: It's like a nanny for kids - to ensure they don't do something stupid :D.

gregpabian: :D

tests/emitter/emitter.js
fredck: Is this what you meant for this?

```js
it( 'should not stop other emitters when a non-listened emitter is provided', function() {
	var spy = sinon.spy();

	var emitter1 = getEmitterInstance();
	var emitter2 = getEmitterInstance();

	listener.listenTo( emitter1, 'test', spy );

	listener.stopListening( emitter2 );

	emitter1.fire( 'test' );

	sinon.assert.called( spy );
} );
```

It goes green for me, but from your comments I had the impression that you expected it to be red. Maybe you want a different test. Lemme know.

gregpabian: I was thinking about a similar case, but now I see that this part actually covers it.
```js
if ( !emitters || ( emitter && !emitterInfo ) || ( event && !eventCallbacks ) ) {
	return;
}
```

So I think we don't need this case.

tests/emitter/emitter.js
fredck: Nah... we would have to create test for failure for every single feature so, which is pointless. :P

src/emitter.js
fredck: Well, for my understanding, the only difference between `listenTo()` and `on()` is who's taking the responsibility of removing the callbacks later. One is the listener, the other is the emitter. Therefore I think that they fit for the very same scenarios.

Anyway, those methods are already in place, paired with tests. They're not that complex... it's more about a series of if's checking parameters. I would take them as good for now.

tests/emitter/emitter.js
fredck: Anyway, I've ended up pushing this tests anyway, as it was already in place. It's not a big deal.

fredck:
All comments addressed. Ready for review again.

src/emitter.js
Reinmar: > Therefore I think that they fit for the very same scenarios.

And what those scenarios are? I haven't worked with Backbone, but AFAIK 90% of cases are `obj.stopListening()`. If we are coding stuff which is not needed we are making a big mistake.

Besides that, I still don't like passing all the arguments to `off()` and even more to `stopListening()`, because not only this is long, but also problematic to do and maintain. If we make `on()` returning a listener instance (which is basically this [object](https://github.com/ckeditor/ckeditor5-core/pull/12/files#diff-d28ecdbfa507067c55c71c8d7c12167aR35) on steroids) we can IMO greatly improve the API.

> Anyway, those methods are already in place, paired with tests.

TBH, that should not be an argument. This way we will accept everything if it's already coded and working.

> They're not that complex... it's more about a series of if's checking parameters.

And these series are IMO unnecessarily complex. Plus, there's that big private object `this._listeningTo` which wouldn't be necessary if we resigned from the listeningTo's params and made `on()` return listener instance. We can make this cleaner and faster with little effort.

Reinmar: PS. To not stop this PR any more we can accept it and I'll propose my ideas in the next one. For the time being the stopListening() params should not be used.

fredck: > And what those scenarios are?  

Just like a button click. You can do button.on( ‘click’ ) or me.listenTo( button, ‘click’ ). Then I want to listen to other stuff from button, but during the execution lifecycle I may decide to stop listening to ‘click’ only… and at the end of the execution I could stop listening to everything.

> TBH, that should not be an argument. This way we will accept everything if it's already coded and working.

In this case we’re talking about a pretty standard API that we “expect” to be used. We’re not inventing anything new here. That’s the way people are used to have such kind of API out there, where you pass the same arguments to listen and stop listening to events.

> And these series are IMO unnecessarily complex.

Well, they’re the very minimum needed to have those feature available.

Ofc if the above arguments don’t convince we can remove features and code (and risk to have to get then back later).

Still the idea of having a `{ off() }` object returned by `on()` sounds good and it definitely fits a new issue.

Reinmar: > Just like a button click. You can do button.on( ‘click’ ) or me.listenTo( button, ‘click’ ). Then I want to listen to other stuff from button, but during the execution lifecycle I may decide to stop listening to ‘click’ only… and at the end of the execution I could stop listening to everything.

The thing is that I would love to know that I'm removing exactly the listener __I__ added. If you do `me.stopListening( button, 'click' )` you know that you stopped listening to "click", but what if a child or parent class started listening as well? Booom, happy debugging ;P. Ofc, you can pass the callback too or even the context and priority, but then it's awful code.

That's why I find `listener.off()` much nicer. You store the listener and remove just it.

> In this case we’re talking about a pretty standard API that we “expect” to be used.

But is it API that we want people to use? I don't want therefore I'll open a new issue for that.

src/emitter.js
```diff
@@ -0,0 +1,284 @@
...
+		off: function( event, callback, ctx ) {
+			var callbacks = getCallbacksIfAny( this, event );
+
+			if ( !callbacks ) {
+				return;
+			}
+
+			for ( var i = 0; i < callbacks.length; i++ ) {
+				if ( callbacks[ i ].callback == callback ) {
+					if ( !ctx || ctx == callbacks[ i ].ctx ) {
```
gregpabian: Why separate `if`?

fredck: Just to have cleaner/readable code and avoid long if's.

We can change it if it looks so bad.

PR #13, opened 2015-02-09
t/10: Basic Model implementation

### LINKED ISSUE #10: Introduce Model (MVC)
Independently of other MVC aspects, the first thing to get developed is the Model class, providing the basics for model data and observables.

fredck:
Fixes #10.

This branch is based on t/11, which should be reviewed first.


fredck:
Ok, the branch is ready for review with support for Model and Collection, both observable, based on the event system introduced with #12.

src/utils.js
```diff
@@ -64,6 +64,83 @@ CKEDITOR.define( function() {
...
+			return function() {
+				return next++;
+			};
+		} )()
 	};
+
+	return utils;
 } );
+
+/**
```
gregpabian: What is that from?

fredck: This is something I'm proposing for documentation purposes only.

Here we're providing the documentation that should be used for all `extend()` uses in our code... the ones mixing in `extendMixin`.

For example, it is used here:
https://github.com/cksource/ckeditor5-core/blob/t/10/src/mvc/model.js#L46

I'm not totally sure that this is the right syntax for it, because we still don't have JSDuck in place here, but we'll be able to check it in the future. The important now is the general idea.

gregpabian: I see. BTW, have we already decided to stay with JSDuck? I think there was a discussion going on regarding this topic?

fredck: Much probably yes, but as you said this is another topic.

src/mvc/collection.js
```diff
@@ -0,0 +1,109 @@
...
+		/**
+		 * Adds an item into the collection.
+		 *
+		 * Note that this is an array-like collection, so the same item can be present more than once. This behavior is
+		 * for performance purposes only and is not guaranteed to be kept in the same way in the future.
+		 *
+		 * @param {Model} model The item to be added.
+		 */
+		add: function( model ) {
+			model.addParentEmitter( this );
```
gregpabian: I'm not really into this approach. Why don't we use the stuff that's already there - good ol' `listenTo()` and `stopListening()`? I think the full control over the event flow should remain on parents, not children calling their parent's `fire` method. I fill that this solution makes too many assumptions.

src/mvc/collection.js
```diff
@@ -0,0 +1,109 @@
...
+
+			this._models.push( model );
+
+			this.fire( 'add', model );
```
gregpabian: What do you think about a common "change" event triggered when something is added, removed or changed? Right now we have to listen to 3 different types of event just to reflect that something changed in the collection.

tests/mvc/collection/collection.js
```diff
@@ -0,0 +1,149 @@
...
+
+var modules = bender.amd.require( 'mvc/collection', 'mvc/model' );
+
+describe( 'add', function() {
+	it( 'should change length and enable get', function() {
+		var Model = modules[ 'mvc/model' ];
+
+		var box = getCollection();
+
+		expect( box ).to.have.length( 0 );
```
gregpabian: I've replaced all these `expect( box.length ).to.equals( 0 );` assertions with more explicit ones.

src/mvc/collection.js
fredck: I'm not a big fan of this idea. It's all about the [confusion on events bubbling I was talking about earlier](https://github.com/ckeditor/ckeditor5-design/issues/7#issuecomment-73204464).

Both `add` and `remove` represent **collection** changes. Currently, `change` represent changes in the **items in the collection** (fired by the items themselves). So while they look like similar, they're semantically different and one may have different expectations from it on a case-by-case basis.

src/mvc/collection.js
```diff
@@ -0,0 +1,109 @@
...
+		 *
+		 * Note that this is an array-like collection, so the same item can be present more than once. This behavior is
+		 * for performance purposes only and is not guaranteed to be kept in the same way in the future.
+		 *
+		 * @param {Model} model The item to be added.
+		 */
+		add: function( model ) {
+			model.addParentEmitter( this );
+
+			this._models.push( model );
```
gregpabian: A model doesn't have any reference to it's parent (in this case a collection) and I'm pretty sure we're gonna need one.

src/mvc/collection.js
fredck: I was unsure about this as well, but it brought two benefits:

 1. It makes it much easier to have such event bubbling system implemented in parent objects. It's just a matter of calling a single function, instead of listening to an event and having to fire it again (which in any case is still possible to be done).

 2. It performs better because it is executed inside the `fire()` method itself through a simple loop over the parents instead of using the events API machinery.

 3. Then, the critical part... listenTo what? We would have to introduce a new `all` event just for that, which would have an even bigger impact on the API.

As for assumptions, these functions are dedicated to work with Emmitters. That's the one and only assumption and it's by design (and part of the documentation).

Finally, this feature will be much probably used mainly inside the MVC API itself, which is a place we have full control.

We may step back and go with `listenTo()` as well. I would be fine. But let's be sure that it's worth undoing the work which is already in place.

src/mvc/collection.js
fredck: This reference is set inside `Model.addParentEmitter()`, isn't it?

src/mvc/collection.js
gregpabian: > Then, the critical part... listenTo what? We would have to introduce a new all event just for that, which would have an even bigger impact on the API.

True, that's why I'm a bit scared of bubbling all kinds of events from a child. It may eventually happen that a child will  try to fire an event with the same name the parent (collection) uses and give us a huge headache debugging it :) That's why I'd still go with Backbone's approach and just bubble child's `change` event.

fredck: Looking purely to the model side of our MVC api, I don't see a big deal because I don't expect models to fire events other than `change`.

In the other hand, we're talking about the events API, which will not be used exclusively for model purposes but for everything that needs events.

Therefore, if we decide NOT to [go with general event bubbling](https://github.com/ckeditor/ckeditor5-design/issues/7#issuecomment-73016905), then I agree that we can remove the `addParentEmitter()` method, handling bubbling in the way you proposed, exclusively inside Collection and for the `change` event only.

src/mvc/collection.js
gregpabian: It's not a reference I'm thinking of - the one you're talking about is just for event bubbling. Which brings another question - do you expect models to belong to more than one collection and/or have more than one parent?

tests/mvc/model/model.js
```diff
@@ -0,0 +1,129 @@
...
+
+			// Check params.
+			sinon.assert.calledWithExactly( spy, sinon.match.instanceOf( EventInfo ), 'color', 'blue', 'red' );
+			sinon.assert.calledWithExactly( spy, sinon.match.instanceOf( EventInfo ), 'year', 2003, 2015 );
+			sinon.assert.calledWithExactly( spy, sinon.match.instanceOf( EventInfo ), 'wheels', 4, sinon.match.typeOf( 'undefined' ) );
+			sinon.assert.calledWithExactly( spyColor, sinon.match.instanceOf( EventInfo ), 'blue', 'red' );
+			sinon.assert.calledWithExactly( spyYear, sinon.match.instanceOf( EventInfo ), 2003, 2015 );
+			sinon.assert.calledWithExactly( spyWheels, sinon.match.instanceOf( EventInfo ), 4, sinon.match.typeOf( 'undefined' ) );
+		} );
+	} );
```
gregpabian: You should add a test for the `set()` function to not fire "change" event when passing the current attribute value.

tests/mvc/model/model.js
```diff
@@ -0,0 +1,129 @@
...
+
+	describe( 'extend', function() {
+		it( 'should create new Model based classes', function() {
```
gregpabian: There are no tests for Model properties extension.

tests/utils/utils.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+/* globals describe, it, expect, bender */
+
+'use strict';
+
+var modules = bender.amd.require( 'utils' );
+
+describe( 'extendMixin', function() {
+	it( 'should extend classes', function() {
```
gregpabian: You're missing tests for the statics extension.

src/mvc/collection.js
fredck: I don't expect anything... the way it is implemented doesn't exclude this possibility. It was just a matter of including a for loop or not in the code, so it doesn't make it much more complex..

src/utils.js
Reinmar: This is not required. You can inherit docs of any method of every class.

But I would actually do this differently. I think that we need a `Class` mixin and this mixin will expose `extend` static method. Then every class will contain `@mixins Class` and voila. `Class` could also have other methods like `_def()` for easier `Object.defineProperty()`. Plus it could do the boilerplate stuff:

```
Model.extend = utils.extendMixin;

utils.extend( Model.prototype, EmitterMixin, { ... } );
```

Basically - it's time to discuss if we want the `Class` (or however you want to call it) to exist or not. I think that there are reasons for having it, but I'm not sure about the shape of it.

fredck: That's an interesting idea. I'm all for it, maybe named `BasicClass` or `Essentials` to avoid a name so close to the `class` keyword.

It assumes that classes will always have `extend()`, be `emitters`, etc... but I think it is an acceptable assumption.

src/mvc/collection.js
Reinmar: I would go without the event bubbling. I think that it's hard now to say whether we need it and how we need it (and our case isn't so typical). If, based on the cases on which we'll work, we'll see that we often have to [forward events](https://github.com/ckeditor/ckeditor5-design/issues/7#issuecomment-74501808) we can make bubbling possible in some way. We'll also see if there are name conflicts, any performance drop, etc. But if we'll go with bubbling now I feel that it will be harder to evaluate and undo it in the future.

In other words - KISS for now and let's reevaluate this with time. Such problems are best handled iteratively.

src/mvc/collection.js
Reinmar: > Currently, change represent changes in the items in the collection (fired by the items themselves)

This is why I don't like automatic bubbling. A change on item is translated to a change on collection, but semantically collection's changes are "item was removed", "item was added" and (optionally) "item was changed". This is a similar case to bubbling in views - [click in an item is not the same as click in a panel](https://github.com/ckeditor/ckeditor5-design/issues/7#issuecomment-74500719).

src/mvc/collection.js
fredck: I agree that this is not an essential feature, but it is a pretty handy tool. Btw, take in consideration that, when it comes to model, it's not only about our needs. It's a tool to enable interacting with the editor state. For example, an outsider may create a tool that logs events from commands, doing so through the `commands` collection, instead of having to attach to events of every single command.

Anyway, even without this feature it is possible to achieve what one needs, so if this is supposed to be a blocker for discussion, I'm ok to remove if for now and review it in the future.

src/mvc/collection.js
fredck: I agree.. that's why I'm for bubbling in collections only, not among different kinds of objects (panel vs item).

src/mvc/collection.js
```diff
@@ -0,0 +1,109 @@
...
+		 * Gets one item from the collection.
+		 *
+		 * @param {Number} index The index to take the item from.
+		 * @returns {Model} The requested item.
+		 */
+		get: function( index ) {
+			var model = this._models[ index ];
+
+			if ( !model ) {
+				throw 'Index not found';
```
Reinmar: We lose useful things like e.stack() if we throw a string instead of `Error` instance.

src/mvc/collection.js
Reinmar: :D But I meant that it has similar semantical issues in collections as in other objects.

fredck: I understood that... but with a reduced impact.

Collections of apples have a very clear relation with apples... so if it fires events thrown by apples, it's fine... the panel vs items case is bad because it is orange firing events from apples.

The only clear conflict would be if apples fire "add" or "remove"... that's why we where considering limiting bubbling for "change" only.

src/utils.js
adelura: @Reinmar Do you mean that you want to introduce base `MixinClass`, and all mixin like `EmitterMixin` will inherit from base one?

fredck: Actually I think the idea is not a mixin but a real "basic" class that expose features.

tests/utils/utils.js
Reinmar: And for the `constructor` property of the `proto` param.

src/utils.js
```diff
@@ -64,6 +64,83 @@ CKEDITOR.define( function() {
 		 */
 		isObject: function( obj ) {
 			return typeof obj === 'object' && !!obj;
-		}
+		},
+
+		/**
+		 * A mixin function to be used to implement the static `extend()` method in classes. It allows for easy creation
+		 * of subclasses.
```
Reinmar: Code samples needed. It's important that we understand how you want it to be used and e.g. the `constructor` part can be incorrectly understood. This can be done together with https://github.com/ckeditor/ckeditor5-core/pull/13/files#r24818601.

src/utils.js
Reinmar: Yep - I meant the "Basic/Base/Grandpa/Mom" class. It will also be a good place to document how we imagine the classes to be organised in JS.

src/mvc/collection.js
adelura: > This is why I don't like automatic bubbling.

As I proposed [here](https://gist.github.com/adelura/d72870b378fcf64e91fe#file-app-js-L4) bubbling should be disabled by default. And decision whether event is buble`able should be made when event is fired.

Even if we see conflict here. By default bubbling is disabled, but only by passing fourth extra `target` argument into `on` method then it means that listening is on child element - see [here](https://gist.github.com/adelura/d72870b378fcf64e91fe#file-app-js-L22-L25). In [this](https://gist.github.com/adelura/d72870b378fcf64e91fe#file-app-js-L17-L19) listener bubbling doesn't affect.

tests/utils/utils.js
gregpabian: +1

Reinmar:
Replying to https://github.com/ckeditor/ckeditor5-core/issues/10#issuecomment-72665479:

> @Reinmar, I understand that you're asking for a resolution around our MVC strategy before proceeding with this. It makes sense, but the reason why this issue has been opened is that we already have a resolution about one thing: we're going to go with MVC.

I did express my concerns about the order in which we approach designing and implementing the MV* stuff, because this is an absolutely critical part of the entire codebase. It must match our rather unusual needs and our constraints. Therefore I feel unsafe when we worked on some of the details without __clearly__ defining the big picture (take [this](http://futurice.com/blog/reactive-mvc-and-the-virtual-dom) for instance).

So far we know that we are going with some MV* (yes, we were talking about MVCVM, but there's no outcome of these discussions and there were significant differences in what @gregpabian and you proposed). I'm pretty sure that different MV* approaches could lead us to different Models implementations. Now we'll have the Model, so let's see how this will affect the design. After all, we can reevaluate the Model class later on.

Back to the code. There are few things which I want to understand about this model proposal which without the big picture are unclear. (I guess that it will be best if you could answer with code samples :))

1. How and where the validation will be performed? For instance - let's say that `SomeModel#foo` attribute can't be set to `false` if `SomeModel#bar` is `true`. Or other case - `User#password` must match certain criteria.
2. How the encapsulation can be achieved? Naturally, attributes are always public and all of them can be set using the `set()` method. Only properties can be private or readonly. Is that right or are we planning to have private or readonly attributes? I'm asking because of scenarios like - `SomeModel#foo` is readonly for other components, but the `SomeModel` class can change it and then the `change` events should be fired.
3. How are we going to solve scenarios like `#foo` and `#bar` attributes can only be set together? For instance - how would you implement more classic `SomeModel#setFooAndBar( foo, bar )` which sets both and calculates something else? Plus, there's a constraint that one attribute can't be changed without the other (for instance one is a range and one is a text selected by it).

With these aspects clear I'm ok with R+ing the PR.

fredck:
> I'm pretty sure that different MV* approaches could lead us to different Models implementations.

Actually, it could lead on different **features** on Model.

What we have so far is the very basics of Model. We're simply making it possible to create a Model state without many features around it. The only feature we added is making the state observable with the events API (making this minimal as well).

I see that most of your concerns are related to the **lack of features** of the proposed implementation. This is intentional exactly because of the reasons you exposed. The fact is that most of these features will be defined with time, as we need them, and most of them will be related to the UI aspects of the MV* we gonna implement.

> I guess that it will be best if you could answer with code samples

In fact, to make the approach clear, the currently possible code samples are pretty simple. It's all about being able to create the Model state and nothing else. Again, that's on purpose.

> How and where the validation will be performed? For instance - let's say that SomeModel#foo attribute can't be set to false if SomeModel#bar is true. Or other case - User#password must match certain criteria.

This feature is not included in the proposal. We can decide, at any moment, how to do it and then we'll have several options, introducing it to any of the layers of the MV*. At that point we can enhance the Model, if necessary.

> How the encapsulation can be achieved? Naturally, attributes are always public and all of them can be set using the set() method. Only properties can be private or readonly. Is that right or are we planning to have private or readonly attributes? I'm asking because of scenarios like - SomeModel#foo is readonly for other components, but the SomeModel class can change it and then the change events should be fired.

Not sure if I understood you but again those looks like additional features. We may decide to use conventions on things that should or should not be touched. Or enforce then by code, adding features to the current Model implementation. We must be sure that it is worth.

> How are we going to solve scenarios like #foo and #bar attributes can only be set together? For instance - how would you implement more classic SomeModel#setFooAndBar( foo, bar ) which sets both and calculates something else? Plus, there's a constraint that one attribute can't be changed without the other (for instance one is a range and one is a text selected by it).

Just like the above. Either this is a convention or we add features to be able to define readonly attributes. Both cases pared with a setFooAndBar() method.

----

Anyway, the point is that by unlocking the Model class we're able to proceed with the development of other features. We may decide to stop it and bring right now all the possible features that we can expect to have in the Model in the future and we have also the possibility to develop them in the future, in parallel with other development.

I'm for moving ahead, otherwise we'll have to stop everything to make research and design the whole thing first. My intention is start building an application that has no UI/View and then build the UI part on top of it, adding the necessary feature for it just then, not now. There is a lot of code out of the MV* aspect of our API and I want to unlock the development of it asap.

So, if we're not doing anything wrong with the current proposal, I would ask to not bring unnecessary blockers to it.

Ofc, if something is understood as critical for the current phase, we must talk further.

gregpabian:
Personally, I'd move on with this very basic implementation of Model and Collection, and later on, if we decide that more features are necessary (e.g. validation), we can build a richer Model class on top of what we've got right now. The only thing I'd consider adding to the classes you proposed is some kind of unique id that could help us identifying things, e.g. right now to pick an item from the collection, only an index can be used and I think this might be not enough. (I'm not insisting to have this now, but it won't hurt :))

Reinmar:
> I see that most of your concerns are related to the lack of features 

I am not asking for features. I am asking for a plan. Currently the plan is that we will somehow solve the issues that we can name now and we will somehow define with time what we really want to achieve. I am not saying we will not solve them and that one day this cannot become a cool Model as a part of cool MV* approach. I am saying that since this is your idea and your implementation I am not able to __review it__ and __continue it__ (in the same direction as you chose). You may unblock yourself by closing this PR, but you will not unblock us. This is part of a wider issue - lack of communication that others can follow. I am pretty sure that other guys feel even more lost than I do or will feel once they start coding something.

Anyway, I did the review of the code. Once the issues that we mentioned in the comments are resolved this can be R+. Grzesiek's idea about adding an id makes sense, but I'm evaluating it in my own imagined context.

fredck:
I understand that this may be a bit confusing because Model is part of MVC and MVC is all about UI. Then I come here and say that we don't care about UI for now, so in fact we don't care about MVC for now. It is true and that's why we're talking about Model only... which is in fact the one and only part of MVC that has nothing to do with UI.

Then it is also hard to think about an application without UI but just data. That's all about this PR. It's about bringing a solution to handle pure data, in a way that we know it can be consumed later once we'll design the UI but that doesn't have any assumptions about how the UI gonna consume it.

It's all about coding what has been [designed already](https://github.com/ckeditor/ckeditor5-design/wiki/MVC#the-model) and it is critical to have it now so we can code other things that complete the editor model, like configuration, commands, plugins, language, etc. Once the (very-very) basic `Model` class is ready, we can finally proceed with `CKEDITOR.create()` and the `Editor` class, unlocking a series of things to be developed (not only me).

At the same time we proceed with our MVC design, defining the UI strategy that will finally consume the model state. At that point we'll have a very clear understanding about what features we need from Model in a UI point of view and we'll enlarge it. But at the current state this is irrelevant and that's why I didn't want to focus on that.

Do you think all this should be exposed in a different way to the team?

fredck:
Ok, more work on this one will be done as soon as #12 gets merged. For now it would be too messy to work on this because of the branches dependency.

src/mvc/collection.js
fredck: Removed with edc9dad.

src/mvc/collection.js
fredck: Event bubbling removed for now with edc9dad.

src/mvc/collection.js
fredck: This parent/child logic has been removed with edc9dad.

tests/mvc/model/model.js
fredck: Test added with faf88b6.

tests/mvc/model/model.js
fredck: Tests added with c550ed0.

tests/utils/utils.js
fredck: Tests added with c550ed0 and dc701b9.

src/mvc/collection.js
fredck: Good one ;) Fixed with 02949c9.

src/mvc/collection.js
```diff
@@ -0,0 +1,106 @@
...
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/**
+ * Collections are ordered sets of models.
+ *
+ * Any event that is triggered on a model in a collection will also be triggered on the collection directly, for
```
gregpabian: I think this part is no longer true, isn't it?

src/mvc/collection.js
```diff
@@ -0,0 +1,106 @@
...
+		Object.defineProperty( this, '_models', {
+			value: []
+		} );
+
+		/**
+		 * The number of items available in the collection.
+		 *
+		 * @property length
+		 */
+		Object.defineProperty( this, 'length', {
```
gregpabian: You can define this property on the Collection.prototype as well.

src/mvc/collection.js
```diff
@@ -0,0 +1,106 @@
...
+		 *
+		 * Note that this is an array-like collection, so the same item can be present more than once. This behavior is
+		 * for performance purposes only and is not guaranteed to be kept in the same way in the future.
+		 *
+		 * @param {Model} model The item to be added.
+		 */
+		add: function( model ) {
+			this._models.push( model );
+
+			this.fire( 'add', model );
```
gregpabian: Missing documentation for the event.

src/mvc/collection.js
```diff
@@ -0,0 +1,106 @@
...
+				}
+			}
+
+			var removedModel = this._models.splice( modelOrIndex, 1 )[ 0 ];
+
+			if ( !removedModel ) {
+				throw new Error( 'Index not found' );
+			}
+
+			this.fire( 'remove', removedModel );
```
gregpabian: Missing documentation for the event.

src/mvc/model.js
```diff
@@ -0,0 +1,94 @@
...
+	Model.extend = utils.extendMixin;
+
+	utils.extend( Model.prototype, EmitterMixin, {
+		/**
+		 * Creates and sets the value of a model property of this object. This property will be part of the model state
+		 * and will be observable.
+		 *
+		 * It accepts also a single object literal containing key/value pairs with properties to be set.
+		 *
+		 * @param {String} name The property name.
```
gregpabian: Let's do another round of :fire: :fire: :fire: "property vs. attribute" :fire: :fire: :fire: fight! :trollface:

src/mvc/collection.js
```diff
@@ -0,0 +1,106 @@
...
+			this.fire( 'add', model );
+		},
+
+		/**
+		 * Gets one item from the collection.
+		 *
+		 * @param {Number} index The index to take the item from.
+		 * @returns {Model} The requested item.
+		 */
+		get: function( index ) {
```
gregpabian: Shouldn't `get()` with no arguments return the entire collection?

src/mvc/model.js
Reinmar: IIRC all of us already did the same mistake :D. Unfortunately it's intuitive (because correct :P) to use the name "property", but as we can see in the model constructor model can also have normal properties, so this name is reserved. However, if you think that it may not be that confusing if we'll call both "properties" (e.g. "normal" and "observable"), then I'm ok with that.

(Is there an emoji for extinguishing the fire? :P)

src/emitter.js
```diff
@@ -214,13 +214,17 @@ CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
 		 * @param {...*} [args] Additional arguments to be passed to the callbacks.
 		 */
 		fire: function( event, args ) {
-			var callbacks = getCallbacksIfAny( this, event );
+			var eventInfo = event instanceof EventInfo && event;
```
Reinmar: Aren't the changes in this file and `src/eventinfo.js` a leftover? If not, then they miss tests and changes in documentation.

src/mvc/model.js
gregpabian: :fire_engine: ? :D

src/mvc/model.js
```diff
@@ -0,0 +1,94 @@
...
+		 *
+		 * @property _attributes
+		 * @private
+		 */
+		Object.defineProperty( this, '_attributes', {
+			value: {}
+		} );
+
+		// Extend this instance with the additional (out of state) properties.
+		if ( properties ) {
```
gregpabian: Code coverage says this condition is not covered.

src/mvc/model.js
fredck: Nah... my mistake.

"attribute" is a basic concept of models related to the model state. They're clearly different from properties, which are said to not be part of the state being simply a js feature. Therefore I'm all for using the correct name consistently in a way to differentiate them.

It's a matter of getting used to it and then things will sound very clear for us.

I'll fix this.

src/mvc/collection.js
fredck: Well... we didn't predict this. Do we need it? Can we have this as a feature request later or you feel is an essential feature?

src/emitter.js
fredck: Hum... I thought about that, but I found this a nice feature to have, especially considering that it was already there. I mean, it's a way to fire an event using an EventInfo object (e.g. can be used to re-fire an event from other objects).

Would you prefer me to remove it? Otherwise I'm fine to have tests for it.

Reinmar: Hm... I've got no strong opinion about this. Seems that it may be useful, but I'm not 100% sure if passing `EventInfo` object from another event (emitter) will be correct. I mean - it contains a reference to "source" plus some methods. Perhaps it all depends on how we document this (and in the future perhaps other) property. But the `stop()` and `off()` methods create tricky cases to be tested. For instance if we forward one event to another (as we did in bubbling) the `stop()` method can be called in both listeners. Will it cancel both events? Which one it will cancel if it's called in the first listener which ten fires another event using its own `EventInfo` object? :D.

So... a lot to test and a lot to document. If you see a future for this event, then why not, but perhaps it would be better to wait.

fredck: Ok, you did a good job on scaring me :) I'll revert this as well.

src/utils.js
fredck: Introduced BasicClass with 9c505ff.

src/mvc/collection.js
fredck: Fixed with e75d5e7.

src/mvc/collection.js
fredck: Nice catch... fixed with f7093ea.

src/mvc/collection.js
fredck: Fixed with e5e48b6.

src/mvc/collection.js
fredck: Fixed with e5e48b6.

src/mvc/model.js
fredck: Fixed with 1ae16ef.

src/emitter.js
fredck: Fixed with 9104c27.

src/mvc/model.js
fredck: Fixed with c1c8404. 100% code coverage now.

fredck:
Ok, PR back on review.

PR #14, opened 2015-02-12
t/7: Introduced Lo-Dash as part of "utils"

### LINKED ISSUE #7: Introduce utils lib

fredck:
Fixes #7.

This code introduces `grunt lodash`, which updates the lodash code available at `src/lib/lodash` with the list of methods available in `utils-lodash.js`.


src/utils.js
```diff
@@ -12,58 +12,18 @@
...
-		 * Checks if the provided object is a JavaScript function.
-		 *
-		 * @param obj The object to be checked.
-		 * @returns {Boolean} `true` if the provided object is a JavaScript function. Otherwise `false`.
-		 */
-		isFunction: function( obj ) {
-			return typeof obj == 'function';
-		},
+	// Extend "utils" with Lo-Dash methods.
+	lodashIncludes.forEach( function( name ) {
```
gregpabian: Replaced good ol' for loop with `forEach`. It's 2015, we've got technology! :sunglasses:

gregpabian:
One general comment regarding tests - so what do we do about the tests for utils? Should we still test all the stuff that comes from Lo-Dash?

src/utils.js
fredck: I know... I know... I'm still reluctant on replacing plain basic for loops with methods calls and callbacks. It gives me this lower performance feeling :D but ofc it is fine.

gregpabian: The performance IS lower, however in this case it really doesn't matter - we're talking about few dozens of items at most ;)

fredck:
> Should we still test all the stuff that comes from Lo-Dash?

I don't think so, in fact. We should do that just for our stuff inside utils. This means that all tests for utils should be removed for now.

src/utils.js
fredck: I know... I know... but I tend to code better performance code anyway, if I can :P

Anyway forEach() is fine as well in this case, as you said.

gregpabian:
In this case we're still gonna need some tests for the code that exposes Lo Dash's methods in the "utils" module and then I could call it done.

src/utils.js
Reinmar: I feel the same way as Fred. God only knows if the performance is really lower (micro benchmarks will not test it) and what's the memory usage, but I never liked `forEach` because of this feeling :D.

gregpabian: So this is going to be a FDD* project? :P

* feeling-driven development

Reinmar:
One thing. When we were discussing ckeditor/ckeditor5-design/issues/13 I wrote: 

> We could integrate it with our tools by copying methods we want to our object, so we will be able to extend it without modifying the original object. This way we'll be able to use the same Lo-Dash library which the website uses.

How can this be achieved with the code that you wrote Fred? How can page author build CKEditor without lodash or with the same lodash which he/she uses on his/her page?

src/utils-lodash.js
```diff
@@ -0,0 +1,46 @@
...
+	var lodashInclude = [
+		/**
+		 * Assigns own enumerable properties of source object(s) to the destination object.
+		 *
+		 * From Lo-Dash: https://lodash.com/docs#assign (alias)
+		 *
+		 * @member utils
+		 * @method extend
+		 *
+		 * @param {Object} object The destination object.
```
Reinmar: I don't think we need those. The link to lodash's docs is IMO enough. Otherwise our lists of params may with time become outdated and cross-checking them is something we don't want to do :D.

fredck:
> How can page author build CKEditor without lodash or with the same lodash which he/she uses on his/her page?

This is not covered.

src/utils-lodash.js
fredck: It would be certainly easier to maintain... so, how do we want this? Like the following?

```
		/**
		 * See Lo-Dash: https://lodash.com/docs#assign (alias)
		 *
		 * @member utils
		 * @method extend
		 */
		'extend'
```

Reinmar: Yep. This or this + one line description (as before). I'm ok with both.

Reinmar:
> This is not covered

Actually, I asked because I think that it might be doable now, I'm just not sure about the details. Is it possible to configure almond while building CKE to get `lodash-ckeditor` from other location? If yes, then perhaps this is all we need. On the other hand, a builder's option to build CKEditor package with no dependencies included (no lodash, no promises tools lib, etc.) would be much more useful (cause easier to use).

gregpabian:
> Is it possible to configure almond while building CKE to get lodash-ckeditor from other location?

It's possible to configure Require.js optimizer to load some modules from a different location using maps, however there's no easy way to configure it at the moment and I've already created a ticket for that ckeditor/ckeditor5#12

fredck:
> In this case we're still gonna need some tests for the code that exposes Lo Dash's methods in the "utils" module and then I could call it done.

Hum... I'm laking on creativity for this :)

I've just pushed updates for this PR. I've still left one test for `utils.extend()` because I remember that this was critical for one part of the code. It doesn't sound like a big deal to have that test.

So, is that test enough to ensure that Lo-Dash stuff is being loaded?

After all, if utils gets fucked up everything gonna melt down because it is used everywhere. So the rest of the code should already be enough for testing it.

In any case, please feel free to point out (or code) other things we could test.

fredck:
> > In this case we're still gonna need some tests for the code that exposes Lo Dash's methods in the "utils" module and then I could call it done.
> Hum... I'm laking on creativity for this :)

Ok, I've figured out a possible test for this. Please wait while I'm working on this.

src/utils.js
adelura: > It gives me this lower performance feeling

We shouldn't optimize everything, but only stuff which we found out that significantly slow down things. If we found something working slow while performance testing, then we should care.

OT: Comming back to `forEach` function - I love this one, because it creates scope and we can reuse function which is defined somewhere else. We know profits :)

fredck:
Ok... ready for review again.

src/utils.js
fredck: > We shouldn't optimize everything

My point is that coding it that way doesn’t make any difference for me… it’s not more effort… it’s not more code. I just said that I do that automatically (and the reason why I **unconsciously** do it). It’s not that I prefer one way and reject the other ;)

gregpabian: > It’s not that I prefer one way and reject the other ;)

Yet we're back with the `for` loop...

src/utils.js
```diff
@@ -12,58 +12,13 @@
...
-			if ( !this.isObject( source ) ) {
-				return target;
-			}
+CKEDITOR.define( [ 'utils-lodash', 'lib/lodash/lodash-ckeditor' ], function( lodashIncludes, lodash ) {
+	var utils = {};
 
-			if ( arguments.length > 2 ) {
-				var args = Array.prototype.splice.call( arguments, 1 );
+	// Extend "utils" with Lo-Dash methods.
+	for ( var i = 0; i < lodashIncludes.length; i++ ) {
```
gregpabian: Guess who's back!

PR #15, opened 2015-02-23
t/9: Introduced the Editor class, CKEDITOR.create and CKEDITOR.instances.

### LINKED ISSUE #9: Introduce CKEDITOR.create
This ticket is related to the creation part described here:
https://github.com/ckeditor/ckeditor5-design/wiki/Editor-Instances

fredck:
Fixes #9.

Note that the intent of this PR is purely introducing `CKEDITOR.create` and the basic infrastructure around it. It doesn't include all other things related to the creation of editors. For example, it doesn't accept the configuration parameter because the whole configuration system is still not covered.

### DIFF
--- a/src/ckeditor.js
+++ b/src/ckeditor.js
@@ -33,7 +33,7 @@ CKEDITOR.define( [ 'editor', 'mvc/collection', 'promise' ], function( Editor, Co
 		 *         // Manipulate "editor" here.
 		 *     } );
 		 *
-		 * @param {String|Object} element A element selector or the DOM element, which will be the source for the
+		 * @param {String|Object} element An element selector or a DOM element, which will be the source for the
 		 * created instance.
 		 * @returns {Promise} A promise, which will be fulfilled with the created editor.
 		 */
--- a/src/ckeditor.js
+++ b/src/ckeditor.js
@@ -3,6 +3,8 @@
  * For licensing, see LICENSE.md.
  */
 
+/* globals document */
+
 'use strict';
 
 /**
@@ -12,8 +14,55 @@
  * @singleton
  */
 
-CKEDITOR.define( function() {
+CKEDITOR.define( [ 'editor', 'mvc/collection', 'promise' ], function( Editor, Collection, Promise ) {
 	var CKEDITOR = {
+		/**
+		 * A collection containing all editor instances created.
+		 */
+		instances: new Collection(),
+
+		/**
+		 * Creates an editor instance for the provided DOM element.
+		 *
+		 * The creation of editor instances is an asynchronous operation, therefore a promise is returned by this
+		 * method.
+		 *
+		 *     CKEDITOR.create( '#content' );
+		 *
+		 *     CKEDITOR.create( '#content' ).then( function( editor ) {
+		 *         // Manipulate "editor" here.
+		 *     } );
+		 *
+		 * @param {String|Object} element A element selector or the DOM element, which will be the source for the
+		 * created instance.
+		 * @returns {Promise} A promise, which will be fulfilled with the created editor.
+		 */
+		create: function( element ) {
+			var that = this;
+
+			return new Promise( function( resolve, reject ) {
+				// If a query selector has been passed, transform it into a real element.
+				if ( typeof element == 'string' ) {
+					element = document.querySelector( element );
+
+					if ( !element ) {
+						reject( new Error( 'Element not found' ) );
+					}
+				}
+
+				var editor = new Editor( element );
+
+				that.instances.add( editor );
+
+				// Remove the editor from `instances` when destroyed.
+				editor.once( 'destroy', function() {
+					that.instances.remove( editor );
+				} );
+
+				resolve( editor );
+			} );
+		},
+
 		/**
 		 * Gets the full URL path for the specified plugin.
 		 *
--- a/src/ckeditor.js
+++ b/src/ckeditor.js
@@ -18,6 +18,9 @@ CKEDITOR.define( [ 'editor', 'mvc/collection', 'promise' ], function( Editor, Co
 	var CKEDITOR = {
 		/**
 		 * A collection containing all editor instances created.
+		 *
+		 * @readonly
+		 * @property {Collection}
 		 */
 		instances: new Collection(),
 
@@ -27,13 +30,13 @@ CKEDITOR.define( [ 'editor', 'mvc/collection', 'promise' ], function( Editor, Co
 		 * The creation of editor instances is an asynchronous operation, therefore a promise is returned by this
 		 * method.
 		 *
-		 *     CKEDITOR.create( '#content' );
+		 *		CKEDITOR.create( '#content' );
 		 *
-		 *     CKEDITOR.create( '#content' ).then( function( editor ) {
-		 *         // Manipulate "editor" here.
-		 *     } );
+		 *		CKEDITOR.create( '#content' ).then( function( editor ) {
+		 *			// Manipulate "editor" here.
+		 *		} );
 		 *
-		 * @param {String|Object} element An element selector or a DOM element, which will be the source for the
+		 * @param {String|HTMLElement} element An element selector or a DOM element, which will be the source for the
 		 * created instance.
 		 * @returns {Promise} A promise, which will be fulfilled with the created editor.
 		 */
--- a/src/editor.js
+++ b/src/editor.js
@@ -19,7 +19,7 @@ CKEDITOR.define( [ 'ckeditor', 'mvc/model', 'utils' ], function( CKEDITOR, Model
 		 * This constructor should be rarely used. When creating new editor instance use instead the
 		 * {@link CKEDITOR#create CKEDITOR.create() method}.
 		 *
-		 * @param {Object} element The DOM element that will be the source for the created editor.
+		 * @param {HTMLElement} element The DOM element that will be the source for the created editor.
 		 * @constructor
 		 */
 		constructor: function Editor( element ) {
@@ -28,7 +28,7 @@ CKEDITOR.define( [ 'ckeditor', 'mvc/model', 'utils' ], function( CKEDITOR, Model
 			 * editor creation and is not subject to be modified.
 			 *
 			 * @readonly
-			 * @property {Object}
+			 * @property {HTMLElement}
 			 */
 			this.element = element;
 		},
--- a/src/editor.js
+++ b/src/editor.js
@@ -1,0 +1,55 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/**
+ * Represents a single editor instance.
+ *
+ * @class Editor
+ */
+
+CKEDITOR.define( [ 'ckeditor', 'mvc/model', 'utils' ], function( CKEDITOR, Model ) {
+	var Editor = Model.extend( {
+		/**
+		 * Creates a new instance of the Editor class.
+		 *
+		 * This constructor should be rarely used. When creating new editor instance use instead the
+		 * {@link CKEDITOR#create CKEDITOR.create() method}.
+		 *
+		 * @param {Object} element The DOM element that will be the source for the created editor.
+		 * @constructor
+		 */
+		constructor: function Editor( element ) {
+			/**
+			 * The original host page element upon which the editor is created. It is only supposed to be provided on
+			 * editor creation and is not subject to be modified.
+			 *
+			 * @readonly
+			 * @property {Object}
+			 */
+			this.element = element;
+		},
+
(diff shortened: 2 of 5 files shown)


src/editor.js
```diff
@@ -0,0 +1,55 @@
...
+			 *
+			 * @readonly
+			 * @property {Object}
+			 */
+			this.element = element;
+		},
+
+		/**
+		 * Destroys the editor instance, releasing all resources used by it. If the editor replaced an element, the
+		 * element will be recovered.
```
gregpabian: I'm not sure on which JSDuck version our clone is based on, but there's a [`@fires`](https://github.com/senchalabs/jsduck/wiki/@fires) tag which we should use here.

tests/ckeditor/ckeditor.js
```diff
@@ -0,0 +1,97 @@
...
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals describe, it, expect, beforeEach, document */
+
+'use strict';
+
+var modules = bender.amd.require( 'ckeditor', 'editor', 'promise' );
+
+var content = document.getElementById( content );
```
gregpabian: Erm, am I missing something here?

gregpabian: This by the way shows that all tests are green if we pass `null` to the `create()` function. But should they?

src/ckeditor.js
```diff
@@ -12,9 +14,56 @@
...
+		 * created instance.
+		 * @returns {Promise} A promise, which will be fulfilled with the created editor.
+		 */
+		create: function( element ) {
+			var that = this;
+
+			return new Promise( function( resolve, reject ) {
+				// If a query selector has been passed, transform it into a real element.
+				if ( typeof element == 'string' ) {
+					element = document.querySelector( element );
```
gregpabian: So in case we want to create instances for every `<textarea>` element on the page, we have to collect all the elements first and then call create(element) on every single item? Have you considered a different approach?

tests/ckeditor/ckeditor.js
fredck: WTF!

The point is that we don't care about the element for now, so it can be anything... but ofc, the test should be correct at least.

src/editor.js
fredck: Hum... I don't think we're using it in ckeditor-dev... curious what @Reinmar has to say about it.

tests/ckeditor/ckeditor.js
gregpabian: > The point is that we don't care about the element for now, so it can be anything...

Well we actually do care about the argument that comes to the `create` function, cause if it's a string, then we treat it as a selector and evaluate it. So I think the `create` function could "care" more.

src/ckeditor.js
fredck: I remember that the original concept included this feature but it was not included in [the docs](https://github.com/ckeditor/ckeditor5-design/wiki/Editor-Instances) and it therefore it wasn't implemented.

But in such case, how to we want to resolve the promise? Right now it passes the (one) created editor instance. **If** we decide to make this change, we have the follow options:

 1. Pass an array back in case of multiple editors and an Editor instance in case of single. Error if none.
 2. Always pass an array back, even for one editor only. Empty array if no element found.
 3. Always pass an array back. Error if no element found.

For consistency, I would go with "3", but it'll be a pain in the ass for developers... especially because those that will use `then()` will usually be creating a single editor only.

Therefore, I think we may consider not having this feature right now.

tests/ckeditor/ckeditor.js
fredck: You know... it's about parameters validation and this is something that we don't do on purpose, to make the code lighter. Otherwise we should start having validation on every method.

So, create() really doesn't care about what you pass to it... for now. At some point we'll start doing something with that element and by then things will not work if you pass something that the docs didn't tell you to pass.... like in any other method in the API.

src/editor.js
Reinmar: I didn't know about it. Either I forgot that it is possible or it was added after we adopted JSDuck. But it is very useful, because we often document this manually, so I'm of course for using it.

fredck:
Ok, minor fixes applied.

I'm opting to not go with the [discussed feature addition to create()](https://github.com/ckeditor/ckeditor5-core/pull/15#discussion_r25243678) right now because it is still unclear for me what the promise should return then.

TODO: This deserves a dedicated ticket after closing this PR.

src/editor.js
fredck: Fixed with 8aa5402.

tests/ckeditor/ckeditor.js
fredck: Fixed with 4eb3bf4.

src/ckeditor.js
mlewand: Ofc consistency is most very important here, therefore I consider only 2 and 3 as viable strategies.

> 2. Always pass an array back, even for one editor only. Empty array if no element found.

I like this option more.

* If I'm very lazy developer I'll simply put `CKEDITOR.create('.editor')` on each page. I don't care if on given subpage there's one editor or none.
* Selectors might be very generic, it's a perfectly normal situation that it they won't match any element on a page.
	* `document.querySelectorAll` does not consider it to be erorr if our selector does not match anything.
	* jQuery does not either.

That makes me think that it's fine that no editors were created with my selector. It won't do any harm, because in `then()` method I'm `for each`ing result anyway because we work on array.

Verdict: option 2

fredck: The cases that makes me sad about the array approach are these:

```
// When pointing to a specific ID... for a single element
CKEDITOR.create( '#content' ).then( function( editors ) {
    // Why not just the editor that I created :(
    var editor = editors[ 0 ];
} );

// Even worst, when passing a single element.
var el = document.getElementById( 'content' );
CKEDITOR.create( el ).then( function( editors ) {
    // Why not just the editor that I created :(
    var editor = editors[ 0 ];
} );
```

Those are the most common cases because usually developers control each editor creation instead of creating a bunch of them.

For now, I would like to wait with this features, creating just a single editor per call.

adelura: How about introduce a function called `createOne` which will always create only the one instance. And let `create` function return an array.

And then `createOne` will be promoted everywhere and `create` will be keept for power developers.

Reinmar: IMO - KISS. Let's have the `CKEDITOR.create()` do one thing - create an editor based on DOM element. We need this method anyway, because the more complex scenarios should be based on a basic method like this.

Then, we can think how to initialise many editors. For instance that could be done by:

```
CKEDITOR.find( '.editor' ).forEach( CKEDITOR.create );
```

Or if you have jQ:

```
$( '.editor' ).each( CKEDITOR.create );
```

Or if you have nothing and there's no `CKEDITOR.find`:

```
[].slice.call( ( document.querySelectorAll( '.editor' ) ) ).forEach( CKEDITOR.create );
```

So the only think I would assure is that the `create()` method has the context bound.

fredck: I agree with @Reinmar... still create() should accept a selector, because this is one of the most basic features. It should create one and only element, just like implemented right now.

Then, one can still interact through elements like you exemplified and have the problem solved. And in the future, if really needed, something like @adelura proposed is interesting... but in the opposite way, by introducing a `createAll()`.

Anyway, for now, as I commented already, let's leave this discussion as a followup ticket for the PR.

Reinmar: > I agree with @Reinmar... still create() should accept a selector, because this is one of the most basic features. 

I really don't feel it's necessary. As a general approach I would choose to make less in our code (to make it smaller and simpler) and let the end-devs do the rest. The DOM isn't anymore a dark place where you need a tank to visit your grandma.

Other than that - I don't like one more thing - we spend time on implementing a solution for "create editor on all elements which match XYZ", when the more general problems is "create editor on all chosen elements". How do I create editors on all `.editor` elements inside a specific `.container` (to which I have a reference) if the `.container` class might be used in many places.

```
CKEDITOR.createAll( '.container .editor' );
```

Isn't what I want. Because what I want is:

```
$( mySpecificContainer ).find( '.editor' ).each( CKEDITOR.create );
```

And why did I use jQuery and not `querySelectorAll`? Ha... because `querySelectorAll` sucks, because it does not match the selector starting from the context, but with the entire tree (starting from document element). There are reasons why it was implemented this way, but basically it's not what you want. Therefore libs like jQuery create temporary id on the context element to be able to build a right selector.

Anyway, instead of trying to solve 90% of cases and making CKEditor bigger and bigger, let's try to solve 100% and make it as small as possible. This is my general approach to CKE5, not only applicable here. In this specific case the `createAll()` method won't harm anyone, so you can add, but as I said "I don't feel it's necessary".

adelura: > Ha... because querySelectorAll sucks, because it does not match the selector starting from the context

Dunno if I understand correctly but `querySelectorAll` start looking from context. It's available not only for document, but for elements as well.
``` js
mySpecificContainer.querySelectorAll( '.editor' );
```
Will return array of elements which are **descendants** of `mySpecificContainer`.

Reinmar: It is available on elements, but it starts matching the selector from the document element (`<html>`). Try: `document.body.querySelector( 'html h1' );`

Reinmar: From the [spec](http://www.w3.org/TR/selectors-api/#processing-selectors):

> Selectors are evaluated against a given element in the context of the entire DOM tree in which the element is located.

Reinmar: BTW. This is an interesting story why it was implemented this way. AFAIR the spec authors had to make this move to make the chances for faster implementation higher. As you may know browsers match selectors from the end (from the last token). They first find all nodes matching the last token and then filter out the rest by using next tokens to the left. Selectors were previously used only for the stylesheets, so in the entire document, hence it made sense for performance reasons to implement it this way.

Therefore it was easier to browser vendors to implement `element.querySelector` in the way it is implemented, because it simply reused the existing algorithms. Otherwise, spec authors would risk that the new feature wouldn't be adopted and as you can imagine it's pretty critical.

fredck: > In this specific case the createAll() method won't harm anyone, so you can add, but as I said "I don't feel it's necessary".

I totally agree.. that's why I said:

> in the future, if really needed

tests/ckeditor/ckeditor.js
```diff
@@ -0,0 +1,97 @@
...
+	it( 'should remove the editor from the `instances` collection on `destroy` event', function() {
+		var CKEDITOR = modules.ckeditor;
+		var editor1, editor2;
+
+		// Create the first editor.
+		return CKEDITOR.create( content ).then( function( editor ) {
+			editor1 = editor;
+
+			// Create the second editor.
+			return CKEDITOR.create( content ).then( function( editor ) {
```
Reinmar: Creating second editor on the same element? That should be prohibited.

tests/editor/editor.js
```diff
@@ -0,0 +1,45 @@
...
+
+'use strict';
+
+var modules = bender.amd.require( 'editor' );
+
+describe( 'constructor', function() {
+	it( 'should create a new editor instance', function() {
+		var Editor = modules.editor;
+
+		var editor = new Editor( document.body );
```
Reinmar: As soon as we'll start really initialising editors on this elements this will affect the other tests. I think we should use specific elements from the beginning, to avoid changing these tests in the future.

tests/ckeditor/ckeditor.js
fredck: That's what I've discussed earlier with @gregpabian... we don't care much about what is passed as element for now. We just know that the element parameter is necessary. Later on we gonna use it for real and this test may eventually fail and have to be fixed. Therefore I'm for fixing the test to use a different element.

Additionally, at this stage we're not having validation of parameters. That's on purpose.

tests/editor/editor.js
fredck: Ok, it makes sense.

tests/ckeditor/ckeditor.js
Reinmar: I'm not saying we must now validate the param. I say that we should avoid writing tests which we'll need to rewrite when the code gets better. This test is logically wrong because we know that only one editor can be created on one element, even though we don't check this now.

gregpabian: I think this still should be Editor's concern to check on what element it is being instantiated.

Reinmar:
I pushed one commit with API docs fixes and commented on two issues in tests. The rest is fine for me.

tests/ckeditor/ckeditor.js
Reinmar: > I think this still should be the Editor's concern to check on what element it is being instantiated.

Yep. Once we introduce the check I agree - it should be done by the Editor constructor.

fredck:
Ok, tests fixed. Ready for final review.

PR #19, opened 2015-02-27
t/16: Introduced the configuration API

### LINKED ISSUE #16: Introduce the Configuration System
The overall specs for the Configuration System are available here:
https://github.com/ckeditor/ckeditor5-design/wiki/Configuration

fredck:
Fixes #16.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -63,9 +63,8 @@ CKEDITOR.define( [ 'mvc/model', 'utils' ], function( Model, utils ) {
 		 *     config.toolbar.color; // 'red'
 		 *
 		 * @param {String|Object} nameOrConfigurations The configuration name or an object from which take properties as
-		 * configuration entries.
-		 * @param {*} [value] The configuration value. Used if a name is passed to nameOrConfigurations. If undefined,
-		 * the configuration is set to `null`.
+		 * configuration entries. Configuration names are case-insensitive.
+		 * @param {*} [value=null] The configuration value. Used if a name is passed to nameOrConfigurations.
 		 */
 		set: function( name, value ) {
 			// Just pass the call to the original set() in case of an object. It'll deal with recursing through the
@@ -128,7 +127,7 @@ CKEDITOR.define( [ 'mvc/model', 'utils' ], function( Model, utils ) {
 		 *
 		 *     config.get( 'toolbar.collapsed' );
 		 *
-		 * @param {String} name The configuration name.
+		 * @param {String} name The configuration name. Configuration names are case-insensitive.
 		 * @returns {*} The configuration value or `undefined` if the configuration entry was not found.
 		 */
 		get: function( name ) {
--- a/src/config.js
+++ b/src/config.js
@@ -12,7 +12,7 @@
  * @extends Model
  */
 
-CKEDITOR.define( [ 'ckeditor', 'mvc/model', 'utils' ], function( CKEDITOR, Model, utils ) {
+CKEDITOR.define( [ 'mvc/model', 'utils' ], function( Model, utils ) {
 	var Config = Model.extend( {
 		/**
 		 * Creates an instance of the {@link Config} class.
--- a/src/config.js
+++ b/src/config.js
@@ -1,0 +1,201 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/**
+ * Handles a configuration dictionary.
+ *
+ * @class Config
+ * @extends Model
+ */
+
+CKEDITOR.define( [ 'ckeditor', 'mvc/model', 'utils' ], function( CKEDITOR, Model, utils ) {
+	var Config = Model.extend( {
+		/**
+		 * Creates an instance of the {@link Config} class.
+		 *
+		 * @param {Object} [configurations] The initial configurations to be set.
+		 * @constructor
+		 */
+		constructor: function Config( configurations ) {
+			// Call super-constructor.
+			Model.apply( this );
+
+			if ( configurations ) {
+				this.set( configurations );
+			}
+		},
+
+		/**
+		 * Set configuration values.
+		 *
+		 * It accepts both a name/value pair or an object, which properties and values will be used to set
+		 * configurations.
+		 *
+		 * It also accepts setting a "deep configuration" by using dots in the name. For example, `'resize.width'` sets
+		 * the value for the `width` configuration in the `resize` subset.
+		 *
+		 *     config.set( 'width', 500 );
+		 *     config.set( 'toolbar.collapsed', true );
+		 *
+		 *     // Equivalent to:
+		 *     config.set( {
+		 *         width: 500
+		 *         toolbar: {
+		 *             collapsed: true
+		 *         }
+		 *     } );
+		 *
+		 * Passing an object as the value will amend the configuration, not replace it.
+		 *
+		 *     config.set( 'toolbar', {
+		 *         collapsed: true,
+		 *     } );
+		 *
+		 *     config.set( 'toolbar', {
+		 *         color: 'red',
+		 *     } );
+		 *
+		 *     config.toolbar.collapsed; // true
+		 *     config.toolbar.color; // 'red'
+		 *
+		 * @param {String|Object} nameOrConfigurations The configuration name or an object from which take properties as
+		 * configuration entries.
+		 * @param {*} [value] The configuration value. Used if a name is passed to nameOrConfigurations. If undefined,
+		 * the configuration is set to `null`.
+		 */
+		set: function( name, value ) {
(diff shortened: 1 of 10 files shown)


src/config.js
```diff
@@ -0,0 +1,201 @@
...
+		 *
+		 *     config.get( 'toolbar.collapsed' );
+		 *
+		 * @param {String} name The configuration name.
+		 * @returns {*} The configuration value or `undefined` if the configuration entry was not found.
+		 */
+		get: function( name ) {
+			// The target for this configuration is, for now, this object.
+			//jscs:disable safeContextKeyword
+			var source = this;
```
adelura: Why we need this variable? And why it's sometimes called `source` and sometimes `target`? I suggest to follow common approach and call them `that` or `self` .

fredck: `that` is a trick to reference `this` inside inner-functions. That's not the case here. 

You should read the rest of the code... here `source` (on `get()`) and `target` (on `set()`) refer to the source/target object for the configurations, which at start is  `this` but then eventually moves to child properties of `this`. Therefore a more semantical name is used for the variable and unfortunately the jscs rules we have applied are too strict for this.

src/config.js
```diff
@@ -0,0 +1,201 @@
...
+'use strict';
+
+/**
+ * Handles a configuration dictionary.
+ *
+ * @class Config
+ * @extends Model
+ */
+
+CKEDITOR.define( [ 'ckeditor', 'mvc/model', 'utils' ], function( CKEDITOR, Model, utils ) {
```
adelura: I asked myself: "Why the heck we need CKEDITOR, ideally config should be independent". And yes - we don't use it :+1:

fredck: Ops, true... leftover from a previous implementation :)

It's weird that jshint doesn't catch this... maybe because `CKEDITOR` is configured as a global, still it is passed as parameter here, which overrides the global :/

src/config.js
```diff
@@ -0,0 +1,201 @@
...
+		 *     config.set( 'toolbar', {
+		 *         color: 'red',
+		 *     } );
+		 *
+		 *     config.toolbar.collapsed; // true
+		 *     config.toolbar.color; // 'red'
+		 *
+		 * @param {String|Object} nameOrConfigurations The configuration name or an object from which take properties as
+		 * configuration entries.
+		 * @param {*} [value] The configuration value. Used if a name is passed to nameOrConfigurations. If undefined,
```
mlewand: I belive we should change type markup here:

```
@param {Mixed} [value]
```

Rather than asterisk char.

tests/ckeditor/ckeditor.js
```diff
@@ -48,6 +48,14 @@ describe( 'create', function() {
 		} );
 	} );
 
+	it( 'should set configurations on the new editor', function() {
+		var CKEDITOR = modules.ckeditor;
+
+		return CKEDITOR.create( content, { test: 1 } ).then( function( editor ) {
+			expect( editor.config.test ).to.equals( 1 );
```
gregpabian: Man you really can't resist using `to.equals`, huh? My inner grammar nazi cries every time I read this :crying_cat_face:
Can we agree that every expect should produce a grammatically correct sentence?
For example: *"Expect A to equal B."*, *"Expect A to be an instance of B"* etc.

src/config.js
gregpabian: `Mixed` suggest a class name so I wouldn't go with that. Besides, if we decide to go with JSDoc3 then `*` is a valid representation of "any type". See: http://usejsdoc.org/tags-param.html

src/config.js
```diff
@@ -0,0 +1,201 @@
...
+				return;
+			}
+
+			// The target for this configuration is, for now, this object.
+			//jscs:disable safeContextKeyword
+			var target = this;
+			//jscs:enable
+
+			// The configuration name should be split into parts if it has dots. E.g: `resize.width`.
+			var parts = name.toLowerCase().split( '.' );
```
mlewand: We need to stress it across API docs that keys are automatically lowercased. I can't see it anywhere in docs ATM.

tests/config/config.js
```diff
@@ -0,0 +1,252 @@
...
+		expect( config ).to.have.property( 'language' ).to.equals( 'pl' );
+		expect( config ).to.have.property( 'resize' ).to.have.property( 'minheight' ).to.equals( 300 );
+		expect( config ).to.have.property( 'resize' ).to.have.property( 'maxheight' ).to.equals( 800 );
+		expect( config ).to.have.property( 'resize' ).to.have.property( 'icon' )
+			.to.have.property( 'path' ).to.equals( 'xyz' );
+		expect( config ).to.have.property( 'toolbar' ).to.equals( 'top' );
+	} );
+
+	it( 'should work with no parameters', function() {
+		var Config = modules.config;
```
gregpabian: Do you really have to do this in every test case that uses this module? Can't you use `before` hook to make a reference once and reuse it later?

src/config.js
gregpabian: +1

tests/ckeditor/ckeditor.js
fredck: > Man you really can't resist using to.equals, huh? My inner grammar nazi cries every time I read this

:D I always found it weird as well, but I thought we agreed on that, that's why I was using it. Certainly my mistake.

Let's come to a final agreement (please open an issue, not here) and change this in all tests after we close this PR.

gregpabian: I'm almost certain we agreed to use the correct form (`to.equal`) right after this issue appeared for the first time.

tests/config/config.js
fredck: Please have an issue for this.

src/config.js
mlewand: True, didn't catch it that in JSDoc 3 asterisk is used to indicate that. Stuck on IDE completion which was suggesting me `@param {Mixed}`.

What about def value `[value=null]` and shorter descr?

tests/ckeditor/ckeditor.js
gregpabian: Found it: https://github.com/ckeditor/ckeditor5/pull/11#discussion_r23288489 :)

src/config.js
fredck: The docs are pointing to `get()` to retrieve configurations, which is also case-insensitive so it doesn't make much difference.

Ofc it is in any case worth to at least make it notice that the config keys are case-insensitive.

Btw, this is for now documented here:
https://github.com/ckeditor/ckeditor5-design/wiki/Configuration

src/config.js
gregpabian: > What about def value [value=null] and shorter descr?

Yup, this is the right syntax, we used it before already and it should be used here as well.

src/config.js
mlewand: y, it make sense - good naming.

tests/ckeditor/ckeditor.js
gregpabian: Created an issue for that - #20.

src/config.js
```diff
@@ -0,0 +1,201 @@
...
+		 *
+		 * This method is supposed to be called by plugin developers to setup plugin's configurations. It would be
+		 * rarely used for other needs.
+		 *
+		 * @param {String|Object} nameOrConfigurations The configuration name or an object from which take properties as
+		 * configuration entries.
+		 * @param {*} [value] The configuration value. Used if a name is passed to nameOrConfigurations. If undefined,
+		 * the configuration is set to `null`.
+		 */
+		define: function( name, value ) {
```
gregpabian: I'm not really convinced to the name of this method.. Wouldn't it make more sense to name it `default` and store its values in `this.defaults`?

Reinmar: I'm out of context right now so ignore my message if it misses the point. But I remember that defining config properties was aimed at not only defining defaults, but also what configuration options are out there (possible validation).

gregpabian: OK, that makes sense.

fredck: > aimed at not only defining defaults, but also what configuration options are out there (possible validation)

Exactly... that's why it's named "define".

src/config.js
fredck: Weird... I've created a test file with this:

```js
/* globals test */

'use strict';

test.example = function( test ) {
	return 1;
};
```

... and here jshint caught the unused parameter... we'll have to spend more time on this at some point... for now, I'm cleaning the params list, ofc.

fredck:
Ok, minor fixes applied. Ready for review again.

gregpabian:
What about all those assertions we discussed earlier? Are you gonna address them now or while resolving #20?

fredck:
> What about all those assertions we discussed earlier? Are you gonna address them now or while resolving #20?

Let's do all in a roll right after closing this PR, if you don't mind.

gregpabian:
Alright, so if there is no more comments, then I'm OK with closing this one.

src/config.js
adelura: Ahh I see now, my bad.

PR #24, opened 2015-04-09
t/23: Introduced the Plugin loading API

### LINKED ISSUE #23: Introducing the Plugins API
Introduce the Plugins API according to the design docs:
https://github.com/ckeditor/ckeditor5-design/wiki/Plugins

fredck:
Fixes #23.

Note that relative updates ended up in [ckeditor5](https://github.com/ckeditor/ckeditor5), where a new test plugin has been introduced [ckeditor5-plugin-devtest](https://github.com/ckeditor/ckeditor5-plugin-devtest). Therefore `git pull` and `npm install` must be executed in `ckeditor5` to catch these updates.

### DIFF
--- a/src/editor.js
+++ b/src/editor.js
@@ -54,12 +54,12 @@ CKEDITOR.define( [ 'mvc/model', 'editorconfig', 'plugincollection' ], function( 
 		/**
 		 * Initializes the editor instance object after its creation.
 		 *
-		 * The initialization consists on the following procedures:
+		 * The initialization consists of the following procedures:
 		 *
 		 *  * Load and initialize the configured plugins.
 		 *  * TODO: Add other procedures here.
 		 *
-		 * This method should be rarely used as `CKEDITOR.create` calls it one should never use the `Editor` contrusctor
+		 * This method should be rarely used as `CKEDITOR.create` calls it one should never use the `Editor` constructor
 		 * directly.
 		 *
 		 * @returns {Promise} A promise which resolves once the initialization is completed.
--- a/src/editor.js
+++ b/src/editor.js
@@ -11,7 +11,7 @@
  * @class Editor
  */
 
-CKEDITOR.define( [ 'mvc/model', 'editorconfig' ], function( Model, EditorConfig ) {
+CKEDITOR.define( [ 'mvc/model', 'editorconfig', 'plugincollection' ], function( Model, EditorConfig, PluginCollection ) {
 	var Editor = Model.extend( {
 		/**
 		 * Creates a new instance of the Editor class.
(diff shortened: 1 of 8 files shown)


fredck:
Right after creating this PR, one thing came to my mind... the way it is coded now (using Promise.all to wait for all init() calls to resolve) may not guarantee the initialization order if a asynchronous plugin depends on a non-asynchronous one.

I'll be checking this, so this aspect should be ignored on review, but still a R+ should not take place until we'll not have this issue clarified. I'll follow up on this PR about this.

fredck:
> Right after creating this PR, one thing came to my mind... the way it is coded now (using Promise.all to wait for all init() calls to resolve) may not guarantee the initialization order if a asynchronous plugin depends on a non-asynchronous one.

In fact I was right with my doubt. I've pushed an additional commit to solve this issue. Ready for full review.

fredck:
Updated the PR description with additional information.

src/editor.js
```diff
@@ -42,6 +42,63 @@ CKEDITOR.define( [ 'mvc/model', 'editorconfig' ], function( Model, EditorConfig
...
+		 * directly.
+		 *
+		 * @returns {Promise} A promise which resolves once the initialization is completed.
+		 */
+		init: function() {
+			var that = this;
+			var config = this.config;
+
+			// Create and cache a promise that resolves when all initialization procedures get resolved.
+			this._initPromise = this._initPromise || Promise.all( [
```
gregpabian: Shouldn't you require the "promise" module instead of using global Promise?

src/plugincollection.js
```diff
@@ -0,0 +1,144 @@
...
+
+		/**
+		 * Gets a plugin from the collection.
+		 *
+		 * @param {String} name The plugin name.
+		 * @returns {Plugin} The requested plugin, if available in the collection.
+		 */
+		get: function( name ) {
+			if ( typeof name != 'string' ) {
+				return Collection.prototype.get.apply( this, arguments );
```
gregpabian: Do we really need this?

src/plugincollection.js
```diff
@@ -0,0 +1,144 @@
...
+			// The hash table used to store pointers to loaded plugins by name.
+			this._names = {};
+		},
+
+		/**
+		 * Loads a set of plugins and add them to the collection.
+		 *
+		 * @param {String} plugins A comma separated list of plugin names to get loaded.
+		 */
+		load: function( plugins ) {
```
gregpabian: Missing documentation for the return value.

tests/editor/editor.js
```diff
@@ -21,6 +21,60 @@ beforeEach( function() {
...
+		init: sinon.spy().named( 'F' )
+	} );
+} );
+
+var asyncSpy = sinon.spy().named( 'async-call-spy' );
+
+CKEDITOR.define( 'plugin!async', [ 'plugin' ], function( Plugin ) {
+	return Plugin.extend( {
+		init: sinon.spy( function() {
+			return new Promise( function( resolve ) {
```
gregpabian: You should require the "promise" module instead of using the global one.

src/plugincollection.js
gregpabian: Ok, I saw you use this case in the tests, but is there any other practical use for this?

src/editor.js
```diff
@@ -42,6 +42,63 @@ CKEDITOR.define( [ 'mvc/model', 'editorconfig' ], function( Model, EditorConfig
...
+				}
+
+				// Return the promise chain.
+				return promise;
+
+				function getInitResolveFn( index ) {
+					return function() {
+						// Resolve with the return value of init(). If it is a Promise, it'll inherit its state. For
+						// everything else it resolves immediately.
+						return Promise.resolve( that.plugins.get( index ).init() );
```
gregpabian: Don't know why but my previous comment somehow disappeared... Anyway, I think You don't have to wrap the returned value in a promise again. This should work just fine:

```js
return that.plugins.get( index ).init();
```

src/plugincollection.js
fredck: You mean, the ability to retrieve a plugin by name? Yes this is very important when one wants to interacts with the plugin instance in the editor directly, for example calling methods of that plugin.

gregpabian: No, I mean the ability to retrieve a plugin by index.

fredck: Ah, then yes, definitely. That's one of the basic features in collections so one can iterate through the items in a loop. (e.g. listing all plugins).

gregpabian: OK, makes sense.

src/plugincollection.js
fredck: Fixed with de01ab4.

src/editor.js
fredck: Definitely! Fixed with 93ac426. I've additionally removed `Promise` from the JSHint globals with cb00b6e so it'll now throw errors in such cases.

tests/editor/editor.js
fredck: Fixed with 93ac426.

src/editor.js
fredck: Good one... there is not reason to have all `.then()` calls in the chain returning promises.

Fixed with 216a50a.

src/editor.js
gregpabian: > I've additionally removed Promise from the JSHint globals

+1 I was about to suggest it :)

fredck:
Ok, all issues fixed. Awaiting for review again.

PR #27, opened 2015-06-09
t/26: Introduced the API for creators

### LINKED ISSUE #26: Introduce the first "creators"
The basics for the following [creators](https://github.com/ckeditor/ckeditor5-design/wiki/Editor-Instances#the-creator) should be introduced:

 * Inline Editor Creator (`ckeditor5-plugin-inlinecreator`)
 * Classic Editor Creator (`ckeditor5-plugin-classiccreator`)

The minimal expected features:

 * Make the minimal necessary DOM manipulation expected by the creator.
 * Load input data from the source element.
 * Initialize the root editable.
 * On destroy:
   * Revert the initial DOM manipulation.
   * Set output data back to the source element.

fredck:
Fixes #26.

A very initial version of the Classic Creator using the proposed API is being proposed separately at ckeditor/ckeditor5-plugin-classiccreator#2. Therefore a cross review is not a bad idea. For that intend, it is enough to add the following line in `package.json` at your local `ckeditor5` repo:

```
    "ckeditor5-plugin-classiccreator": "ckeditor/ckeditor5-plugin-classiccreator"
```

The above line will be added definitively once this PR is closed.

The code in this PR changed drastically in the middle of the road. It has been initially coded according to our very first design were we would have a single editable per editor instance. A working version of it is available at a3332e2. Then, we decided that supporting multiple editables per instance would be very interesting and so things have been refactored to support this. Therefore the reviewer will have a better experience by comparing this PR with master instead of checking commits individually.

### DIFF
--- a/src/editable.js
+++ b/src/editable.js
@@ -11,10 +11,18 @@
  * Represents an editable inside the editor.
  *
  * @class Editable
+ * @extends Model
  */
 
 CKEDITOR.define( [ 'model', 'config', 'editablecollection', 'promise' ], function( Model, Config, EditableCollection, Promise ) {
 	var Editable = Model.extend( {
+		/**
+		 * Creates an instance of the Editable class.
+		 *
+		 * @constructor
+		 * @param {HTMLElement|Editable} element The DOM element that will be linked to this editable. If an editable is
+		 * provided, that same editable is returned by the constructor.
+		 */
 		constructor: function Editable( element ) {
 			// We need to renew the module request because of cross-reference with Editable<->EditableCollection.
 			EditableCollection = CKEDITOR.require( 'editablecollection' );
@@ -31,19 +39,50 @@ CKEDITOR.define( [ 'model', 'config', 'editablecollection', 'promise' ], functio
 			 * The DOM element managed by this editable.
 			 *
 			 * @readonly
-			 * @property {HTMLElement}
+			 * @property {HTMLElement} element
 			 */
 			this.set( 'element', element );
 
+			/**
+			 * The DOM element that represents the editable view.
+			 *
+			 * @readonly
+			 * @property {HTMLElement} view
+			 */
 			this.set( 'view' );
 
+			/**
+			 * Configurations specific to this editable. It inherits configurations from parent editables.
+			 *
+			 * @readonly
+			 * @property {Config} config
+			 */
 			this.set( 'config', new Config() );
 
+			/**
+			 * The child editables of this editable.
+			 *
+			 * @readonly
+			 * @property {EditableCollection} editables
+			 */
 			this.set( 'editables', new EditableCollection( this ) );
 
 			return this;
 		},
 
+		/**
+		 * Initializes the editable.
+		 *
+		 * The initialization consists of the following procedures:
+		 *
+		 *  * Setup the editable view.
+		 *  * Initialize children.
+		 *  * Enable editing in the editable.
+		 *
+		 * This method should be rarely used as `editor.init` calls it during editor initialization.
+		 *
+		 * @returns {Promise} A promise that resolves once the initialization is completed.
+		 */
 		init: function() {
 			var that = this;
 
@@ -114,6 +153,11 @@ CKEDITOR.define( [ 'model', 'config', 'editablecollection', 'promise' ], functio
 	 * @member Editable
 	 */
 	Object.defineProperties( Editable.prototype, {
+		/**
+		 * The parent editable of this editable.
+		 *
+		 * @property {Boolean}
+		 */
 		parent: {
 			set: function( parent ) {
 				this.config.parent = parent.config;
--- a/src/editable.js
+++ b/src/editable.js
@@ -1,0 +1,82 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/**
+ * Represents an editable inside the editor.
+ *
+ * @class Editable
+ */
+
+CKEDITOR.define( [ 'model', 'promise' ], function( Model, Promise ) {
+	var Editable = Model.extend( {
+		constructor: function Editable( element ) {
+			// Accept and Editable instance as parameter as well, for simplification in other parts of the code.
+			if ( element instanceof Editable ) {
+				return element;
+			}
+
+			// Call the base constructor.
+			Model.apply( this );
+
+			/**
+			 * The DOM element managed by this editable.
+			 *
+			 * @readonly
+			 * @property {HTMLElement}
+			 */
+			this.element = element;
+
+			this.isEditable = true;
+
+			return this;
+		},
+
+		/**
+		 * Sets the editable data.
+		 *
+		 * @param {String} data The data to be set into the editable.
+		 * @returns {Promise} A promise that resolves as soon as the data is processed and fully loaded inside the
+		 * editable.
+		 */
+		setData: function( data ) {
+			this.element.innerHTML = data;
+
+			return Promise.resolve();
+		},
+
+		/**
+		 * Gets the current editable data.
+		 *
+		 * @returns {String} The data.
+		 */
+		getData: function() {
+			return this.element.innerHTML;
+		}
+	} );
+
+	/**
+	 * @member Editable
+	 */
+	Object.defineProperties( Editable.prototype, {
+		/**
+		 * Indicates whether or not the editable has editing enabled.
+		 *
+		 * @property {Boolean}
+		 */
+		isEditable: {
+			get: function() {
+				return this.element.contentEditable;
+			},
+
+			set: function( value ) {
+				this.element.contentEditable = value;
+			}
+		}
+	} );
+
+	return Editable;
+} );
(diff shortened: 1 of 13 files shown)


src/editor.js
```diff
@@ -44,16 +38,28 @@ CKEDITOR.define( [
...
+
+			/**
+			 * The editables attached to this editor instance.
+			 *
+			 * @readonly
+			 * @type {EditableCollection} editables
+			 */
+			this.set( 'editables', new EditableCollection() );
+
+			this._creators = {};
```
pjasiun: Shouldn't `_creators` be a CREDITOR property? The editor constructor only takes one and use it. There is no point in registering creator for every editor instance.

src/editor.js
```diff
@@ -104,6 +115,58 @@ CKEDITOR.define( [
...
+				//      be done at this stage because both fireCreator() and initEditables() will drive the data to
+				//      editable.view instead.
+				if ( typeof that._data != 'string' ) {
+					that._data = that.getData();
+				}
+			}
+
+			function fireCreator() {
+				// Take the name of the creator to use (config or any of the registered ones).
+				var creatorName = config.creator || Object.keys( that._creators )[ 0 ];
```
pjasiun: It would be better if creators would have `match` method and the first matching creator would be used. It means I can add classic creator for `textarea` element, inline for `div` and oneline for `input` and proper creator would be used if the `create` method will be called on the element.

src/editor.js
```diff
@@ -111,11 +174,68 @@ CKEDITOR.define( [
...
 		 * @fires destroy
+		 * @returns {Promise} A promise that resolves once the editor instance is fully destroyed.
 		 */
 		destroy: function() {
+			var that = this;
+
 			this.fire( 'destroy' );
 
-			delete this.element;
+			this._destroyPromise = this._destroyPromise || Promise.resolve()
```
pjasiun: There is missing test for it. I can remove `this._destroyPromise ||` and tests are still green. This is easy feature which can be easily tested and event easier skipped during future refactoring so it needs tests.

src/editable.js
```diff
@@ -0,0 +1,191 @@
...
+		 *  * Enable editing in the editable.
+		 *
+		 * This method should be rarely used as `editor.init` calls it during editor initialization.
+		 *
+		 * @returns {Promise} A promise that resolves once the initialization is completed.
+		 */
+		init: function() {
+			var that = this;
+
+			this._initPromise = this._initPromise || Promise.resolve()
```
pjasiun: There is missing test for it. I can remove `this._initPromise ||` and tests are still green.

src/editor.js
```diff
@@ -104,6 +115,58 @@ CKEDITOR.define( [
...
+					}
+
+					// Create an instance of the creator.
+					that._creator = new Creator( that );
+
+					// Finally fire the creator. It may be asynchronous, returning a promise.
+					return that._creator.create();
+				}
+
+				return 0;
```
pjasiun: The comment would be useful what this `return 0` means.

src/editablecollection.js
```diff
@@ -0,0 +1,79 @@
...
+		 */
+		add: function( editableOrElement ) {
+			var editable = editableOrElement;
+
+			if ( editableOrElement instanceof HTMLElement ) {
+				editable = new Editable( editableOrElement );
+			}
+
+			if ( this.parent ) {
+				editable.parent = this.parent;
```
pjasiun: Missing assertion for this code. I can remove it and all tests pass.

src/editor.js
```diff
@@ -61,22 +67,27 @@ CKEDITOR.define( [
...
 			// Create and cache a promise that resolves when all initialization procedures get resolved.
-			this._initPromise = this._initPromise || Promise.all( [
-				loadPlugins().then( initPlugins )
-			] );
+			this._initPromise = this._initPromise || Promise.resolve()
+				.then( loadPlugins )
+				.then( initPlugins )
+				.then( prepareData )
+				.then( fireCreator )
+				.then( initEditables )
```
pjasiun: `initEditables` should be before `fireCreator`. Editables should be ready when we call creators, creator may need to ready some properties of editable.

src/editable.js
```diff
@@ -0,0 +1,191 @@
...
+				.then( initChildren )
+				.then( enableEditing );
+
+			return this._initPromise;
+
+			function setupView() {
+				if ( !that.view ) {
+					// TODO: For now we're simply pointing to the element. Once the Document Model (dm) will be implemented,
+					// this will point to a real View object.
+					that.view = that.element;
```
pjasiun: I think that this should be done only by creators. `that.view = that.element;` makes sense only for inline editable, for any other it could be a root of problems it `textarea` became a `view`. Since it is not not the final form of code `setupView` could be removed for now and, eventually, introduces later, if needed.

pjasiun:
Also: the ticket is about both `inlinecreator` and `classiccreator`. Here is only one implemented. The ticket description should be changed or the second constructor should be added.

pjasiun:
See also comments here:
 - https://github.com/ckeditor/ckeditor5-plugin-classiccreator/pull/2
 - https://github.com/ckeditor/ckeditor5-core/issues/26
 - https://github.com/ckeditor/ckeditor5-design/issues/60
 - https://github.com/ckeditor/ckeditor5-core/issues/30

pjasiun:
And a note: `editor` on the one hand apply `Model`, can be observed, on the other it init everything, call creator which will create a view, so do typical controllers works, It means that `editor` is at the moment both: Model and Controller. We already talked about it and I agree that it is the best solution in this case, but we should be aware that we break the classic MVC separation here.

src/editor.js
fredck: Creators are strictly related to editor instances:

 * They're loaded on demand through instance plugins. 
 * In most of the cases, an editor will load just one creator which would be the default one, making `config.creator` not necessary.
 * They could be customized (overridden) in a per instance fashion.

Creators have no use at `CKEDITOR`. Additionally we would have the problem of defining who should load them if not the editor instance.

Still one could argue that we could use `CKEDITOR` as the repository for creators but there is zero benefit on this. What we have in each instance is simply (1) a list of available creators for that specific editor instance with (2) pointers to the creator class. It's a very lightweight structure that fulfils the requirement of "1" giving quick access to the creators with "2".

src/editor.js
fredck: For now, it's (1) over-engineering and (2) would easily have issues with creators that match the same kinds of elements (e.g. div on classic and inline).

Our current approach is "less". Features like this can be added in the future if really needed.

src/editor.js
fredck: Fixed with d53bffa.

src/editable.js
fredck: Fixed with ce80685.

src/editor.js
fredck: I'm still used to make functions always return anything if they ever return (line 150)... it's an old jslint check... I'll simply remove that line.

pjasiun: My question was: what should happen if there is no creator? There should be a note that doing nothing (no error) is designed behavior, because it was not clear for me.

src/editor.js
pjasiun: I think about `CKEDITOR` as a factory for `editors` (since it has a `create` method). And see no point in putting whole list of creators to each instance of `editor` when only one will be used. Also I think that user should not be forced to add the creator as a plugin every time he want to create an editor. It should be already defined when he build package what type of the creator will be used.

src/editor.js
fredck: Fixed with 37e13aa.

src/editablecollection.js
fredck: Fixed with e8c90bb.

src/editor.js
fredck: Fixed with b2b9fd4.

fredck:
> Also: the ticket is about both inlinecreator and classiccreator. Here is only one implemented. The ticket description should be changed or the second constructor should be added.

This was intentional. As soon as the shape of one creator is approved, we'll move on with the next one. It'll be up to us to do this in this PR on by opening another issue.

fredck:
I've just reverted the change in the initialization order (b2b9fd4) because editor.view, which is initialized by the creator, must be available to enable editing.

fredck:
@Reinmar come with the idea of avoiding having editor.addCreator and simply using naming conventions to detect which creator to use for each editor instance. So creators would still be plugins, with names like `ckeditor5-plugin-creator-classic` and `ckeditor5-plugin-creator-inline`.

This is the only missing part of this PR.

Reinmar:
BTW. There was one more thing that we discussed (although I'm not sure if it's related to this piece of code) – what the editable is, what the creator is, etc. For instance, the creator should not be responsible for creating an `<iframe>` element, because initialisation of that kind of editable is the role of editable's view. This way, this (non trivial) piece of code will be reusable, so it will be easy to implement different creators which still use iframed editable.

Another thing is where should be a code reading data from editor element. This job may need to be extended by a specific implementation, because e.g.:

* someone may want to load data from the `<template>` element,
* someone may want to pass data to the creator,
* someone may want to load data asynchronously (although, I would ignore this case and require preloading data by the developer)

This, I think, clearly shows that reading the data is the creator's job (so it should be a creator's method). But... I'm not sure if this is what we decided, Fred. Do you recall? :D

Reinmar:
As I've made some changes on `master` and I'm taking now this branch over, I'm closing this PR and I'll rebase the branch.

PR #28, opened 2015-06-12
t/18: Introduced the Mixin class

### LINKED ISSUE #18: Introduce the Mixin class
Following [recent talks](https://github.com/ckeditor/ckeditor5-design/issues/52), to make the use of mixins more elegant, we should introduce the Mixin class.

The following is a usage example of it:

```js
var EmitterMixin = new Mixin( {
    method1: ...,
    method2: ...
} );

function TargetClass() {}

EmitterMixin.mixin( TargetClass.prototype );
```

The constructor basically accepts the mixin properties and the class exposes a single `mixin()` method to mix such properties in a target.

fredck:
Fixes #18.

### DIFF
--- a/src/basicclass.js
+++ b/src/basicclass.js
@@ -17,7 +17,7 @@ CKEDITOR.define( [ 'emittermixin', 'utils' ], function( EmitterMixin, utils ) {
 	}
 
 	// Injects the events API.
-	utils.extend( BasicClass.prototype, EmitterMixin );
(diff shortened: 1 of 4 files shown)


pjasiun:
The PR looks outdated. Do you agree to close it, or will you update it?

fredck:
I will not work on this any further, but this PR doesn't have to be closed. Anyone can take this over from here.

Reinmar:
I proposed #147. This PR cannot be reused any more.

PR #29, opened 2015-06-12
t/21: Made fire() return a promise.

### LINKED ISSUE #21: Make .fire() return a promise
It would be interesting to have `.fire()` return a promise.

This would enable:

 * Listeners to throw errors.
 * Have asynchronous listeners (Yeah!!!)

Emitters would be able to do this (e.g.):

```
editor.fire( 'something' )
    .then( function() {
        // Do something on success.
    } )
    .catch( function( err ) ) {
        // Handle errors.
    } );
```

fredck:
Fixes #21.

`fire()`is now a synch/asynch hybrid. It is synchronous when all listeners are synchronous and asynchronous if any of them return a promise.

### DIFF
--- a/src/emittermixin.js
+++ b/src/emittermixin.js
@@ -12,7 +12,7 @@
  * @singleton
  */
 
-CKEDITOR.define( [ 'eventinfo', 'utils' ], function( EventInfo, utils ) {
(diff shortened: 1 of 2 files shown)


pjasiun:
Since the code was moved to the https://github.com/ckeditor/ckeditor5-utils this PR is outdated.

PR #39, opened 2015-10-06
Added log and ckeditorerror modules.

### LINKED ISSUE #38: Introduce a simple error logging system and error class
Implement https://github.com/ckeditor/ckeditor5-design/issues/14#issuecomment-138527386

Reinmar:
Fixes #38. Code in t/38.

### DIFF
--- a/src/ckeditorerror.js
+++ b/src/ckeditorerror.js
@@ -1,0 +1,48 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/**
+ * The CKEditor error class.
+ *
+ * All errors will be shortened during the minification process in order to reduce the code size.
+ * Therefore, all error messages should be documented in the same way as those in {@link CKEDITOR.core.log}.
+ *
+ * Read more in the {@link core.log} module.
+ *
+ * @class CKEditorError
+ * @extends Error
+ */
+
+CKEDITOR.define( function() {
+	class CKEditorError extends Error {
+		/**
+		 * Logs an error to the console.
+		 *
+		 * Read more about error logging in the {@link core.log} module.
+		 *
+		 * @constructor
+		 * @param {String} message The error message in an `error-name: Error message.` format.
+		 * During the minification process the "Error message" part will be removed to limit the code size
+		 * and a link to this error documentation will be logged to the console.
+		 * @param {Object} [data] Additional data describing the error.
+		 */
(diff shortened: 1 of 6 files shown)


Reinmar:
It's a super simple implementation of the proposed API. There will be much more fun in the builder, but that'll be done in the future.

Reinmar:
One second.... I forgot to fix already existing usage of `Error()`. I want to cover it in this PR.

Reinmar:
OK, there's still the thing with https://github.com/benderjs/benderjs/issues/245#issuecomment-138467961 (that PluginCollection turns a meaningful error into a useless one), but I think I'll handle this in a separate ticket: #40.

Other than that this PR is back on review.

src/ckeditorerror.js
```diff
@@ -0,0 +1,54 @@
...
+ * Therefore, all error messages should be documented in the same way as those in {@link CKEDITOR.core.log}.
+ *
+ * Read more in the {@link core.log} module.
+ *
+ * @class CKEditorError
+ * @extends Error
+ */
+
+CKEDITOR.define( function() {
+	class CKEditorError extends Error {
```
pjasiun: There is no class documentation. We should decide is we document both class and constructor, because they are basically the same.

src/ckeditorerror.js
```diff
@@ -0,0 +1,54 @@
...
+		/**
+		 * Creates an instance of the class.
```
pjasiun: That is pretty obvious for the constructor ;)

src/ckeditorerror.js
fredck: I think both end up in different places in the documentation so they're needed. Btw, class documentation has nothing to do with the constructor documentation. For instance, the constructor documentation will include parameters.

src/ckeditorerror.js
fredck: Well... but that's what it is in fact :)

I would make an improvement: "Creates an instance of the CKEditorError class." to make it clear.

src/ckeditorerror.js
Reinmar: +1 for Fred. Class documentation should describe how to use the class, how it connects with other classes, the idea behind it, etc. Constructor is just like any other method – it accepts stuff and does something. This needs to be documented as well.

Besides, it's really not time  to decide how to document this as we haven't even decided yet with which docs generator we'll go :D. So I'm just following what's already in the code base.

src/ckeditorerror.js
Reinmar: I'll change the description to match Fred's proposal.

PR #43, opened 2015-10-13
t/42: Plugincollection is greedy while loading plugin dependencies

### LINKED ISSUE #42: Plugincollection is greedy while loading plugin dependencies
At the moment, if `plugin!A` has a dependency `plugin!A/b`, `PluginCollection` loads `plugin!/A/b` and immediately invokes a new plugin instance out of `plugin!/A/b`, even though `plugin!/A/b` does not inherit from `Plugin`.

I discovered it loading ui components from the library like `plugin!ui-library/button`. There was a number of errors because of that.

Long story short, `PluginCollection` should not assume that plugin dependencies are also plugins.

http://requirejs.org/docs/plugins.html#names

oleq:
Fixes #42.

### DIFF
--- a/src/plugincollection.js
+++ b/src/plugincollection.js
@@ -12,7 +12,13 @@
  * @extends Collection
  */
 
-CKEDITOR.define( [ 'collection', 'promise', 'log' ], function( Collection, Promise, log ) {
+CKEDITOR.define( [
+	'collection',
+	'promise',
+	'plugin',
+	'ckeditorerror',
+	'log'
+], function( Collection, Promise, Plugin, CKEditorError, log ) {
 	class PluginCollection extends Collection {
 		/**
 		 * Creates an instance of the PluginCollection class, initializing it with a set of plugins.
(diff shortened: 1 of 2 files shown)


src/plugincollection.js
```diff
@@ -58,20 +58,27 @@ CKEDITOR.define( [ 'collection', 'promise', 'log' ], function( Collection, Promi
 					CKEDITOR.require( [ 'plugin!' + plugin ],
 						// Success callback.
 						function( LoadedPlugin ) {
-							var loadedPlugin = new LoadedPlugin( that._editor );
-							loadedPlugin.name = plugin;
-							loadedPlugin.path = CKEDITOR.getPluginPath( plugin );
-							loadedPlugin.deps = getPluginDeps( plugin );
+							var deps = getPluginDeps( plugin );
+							var isPluginDep = plugin.indexOf( '/' ) > 0;
```
fredck: Should we check the path or check `instanceof Plugin`?

Reinmar: There should be no real difference if all your plugins and paths are correct, but it's an interesting question in case of some mistakes.

* Checking by path makes sure that you can only have a plugin file under a specific path.
* Checking by `instanceof` will allow you to have more plugins inside one plugin package what perhaps makes some sense, but then we would need to make sure that they really work there (if we want it to become a feature).
* Not checking `instanceof` is also unsafe because someone could mistakenly return a wrong value in his/her plugin module.

So I'm for keeping the current check and adding `instanceof` check inside. If `LoadedPlugin` is not an instance of `Plugin` then I would throw an error.

oleq:
Added `loadedPlugin instanceof Plugin` check and an additional test to check if an error is produced.

PR #47, opened 2015-10-15
Introduces basic creator mechanism

Reinmar:
Fixes (or rather starts) #26.


oleq:
Good job. Useful API for UI development.

PR #78, opened 2015-11-09
Reimplemented the Collection class.

### LINKED ISSUE #77: Fix the collections
Followup of #58.

We decided that a single `Collection` class with functionality of both – unordered map and an ordered array will make the most sense. There will be no ambiguity which collection to choose for the specific case that you're dealing with and you won't need two collections when you need to store named items, accessible by names, but also ordered.

The changes:

1. Remove `NamedCollection`.
1. Add `Collection._modelMap`.
1. Change `Collection.add()` so when adding a model without an `id` it's automatically generated.
1. `Collection.add()` should accept all objects, not only models. That will guarantee that we'll be able to generate collections of other kinds of objects.
1. Allow configuring the name of the property which represent's model `id` (e.g. to `name`).
1. Add  `Collection.at( index )` and `Collection.get( id )`.
1. Change `Collection.remove()` to accept id (when a string), index (when a number) or a model (when an object).
1. Remove `forEach()` but add an iterator. Add `map()`.

Reinmar:
Fixes #77.


src/collection.js
```diff
@@ -20,15 +25,47 @@ CKEDITOR.define( [ 'emittermixin', 'ckeditorerror', 'utils' ], function( Emitter
 		 * Creates a new Collection instance.
 		 *
 		 * @constructor
+		 * @param {Iterale} [items] Items to be added to the collection.
+		 * @param {String} [idName='id'] The name of the property which is considered to identify an item.
 		 */
-		constructor() {
+		constructor( items, idName ) {
```
fredck: If I could add my bits to this, I dislike a lot configuration parameters. As for methods, I think we should start getting used to have one and only `options` parameter that can be extended as needed.

In the case of a class constructor though, we could have a more elegant solution, like a configuration property. I see that it has been decided to pass the initial items to the constructor though, so we need to know it beforehand. Therefore I would still propose to call it `options`.

Reinmar: So `Collection( items, options )`?

fredck: Yes, I would propose that. Actually, there are two alternatives:

1. `Collection( items, options )`
2. `Collection()` + configuration property + `add( item | items )`

Both are ok for me.

Reinmar: Actually... I'd choose a 3rd way – `Collection( options )` and chainable `add( item | items )`. That's best for single resp principle.

Reinmar: I've just realised that `add()` cannot accept an iterable collection of items, because it will be impossible to tell whether that object should be added as a single collection item or whether we should iterate over it. That could lead to nasty problems in the future if one wants to create collection of such objects. Let's wait to see how much we'll miss this method of adding items.

Reinmar:
Ready for a review again.

src/collection.js
```diff
@@ -37,106 +69,189 @@ CKEDITOR.define( [ 'emittermixin', 'ckeditorerror', 'utils' ], function( Emitter
...
-		add( model ) {
-			this._models.push( model );
+		add( item ) {
+			var itemId;
+			var idProperty = this._idProperty;
+
+			if ( ( idProperty in item ) ) {
+				itemId = item[ idProperty ];
+
+				if ( typeof itemId != 'string' ) {
```
pjasiun: Do we need such limitation? Why "symbol" can not be a key? We already have a case in the `Document` where we have a collection of root elements and the root of graveyard is kept by a symbol, so it will not conflict with any other name of the root.

Reinmar: Good idea. I exposed it to #81.

PR #79, opened 2015-11-10
Editor should throw if config.creator is not set

### LINKED ISSUE #55: Throw when more than one creator was defined, but config.creator isn't set
Currently:

* If config.creator was defined we try to use it. We throw if such creator was not defined.
* Otherwise, we take the first of the defined creators.
* If both didn't work we throw.

This has one major flaw – if someone chooses only the `creator-super` when building a package she/he expects that this is the only creator, so it will be used automatically. Unfortunately, it's not going to work this way, because `creator-super` may inherit from the `creator-classic` which will be automatically included in the package and loaded **and** initialised. So finally, it may be automatically chosen even though the developer wanted the `creator-super`. Super confusing.

Therefore, we may do this differently:

* If config.creator is defined, use it. Throw if such creator does not exist.
* If there's just one creator defined, use it.
* Otherwise, throw that config.creator wasn't defined.

So in this case if the config.plugin setting listed only one creator and this creator does not require any other one, then we can safely use it as there can't be any confusion. But in all other cases we expect `config.creator` to be defined.

Reinmar:
Fixes #55.


Reinmar:
I merged master into my branch because there were conflicts. Ready for review again.

PR #88, opened 2015-11-18
t/69: Transaction and deltas.

### LINKED ISSUE #69: Transaction and deltas
Based on the discussion in https://github.com/ckeditor/ckeditor5-design/issues/68 the `Transaction` class will be created. It will group multiple operations together as a one undo step. There will be one `Transaction` class and multiple `Deltas` which register as transactions methods.

Each `Delta` will be build from operations, for example `SplitDelta` will be an `InsertOperation` and `MoveOperation`. `Delta` may have a separate logic for OT.

The `Transaction` class will also have a `commit` method which fires events that a set of changes is done, with an optional parameters.

The `Transaction` will be the basic API for document changes. The sample usage of it will looks like this:

`document.createTransaction().split( node, position ).move( nodes ).commit( { dontRender: true } );`

pjasiun:
Fixes #69.


src/document/node.js
```diff
@@ -232,19 +232,14 @@ CKEDITOR.define( [ 'document/attribute', 'utils', 'ckeditorerror' ], ( Attribute
...
 		 */
-		_getAttrCount() {
-			let count = 0;
-
-			for ( let attr of this._attrs ) { // jshint ignore:line
-				count++;
-			}
-
-			return count;
+		getAttrIterator() {
```
Reinmar: OK... this is weird. Why iterator? Why not returning an array?

Reinmar: Actually, after a sec I'm OK with the iterator. But the method should definitely not state that. It should be `getAttrs()`. There simply needs to be one method of accessing attributes and that's what this method will do.

src/document/document.js
```diff
@@ -131,6 +142,15 @@ CKEDITOR.define( [
 		get _graveyard() {
 			return this.getRoot( graveyardSymbol );
 		}
+
+		/**
+		 * Creates a {@link document.Transaction} instance which allows to change the document.
+		 *
+		 * @returns {document.Transaction} Transaction instance.
+		 */
+		makeTransaction() {
```
Reinmar: make => create

Reinmar: Even the docs say "creates" :D.

src/document/delta/transaction-base.js
```diff
@@ -0,0 +1,139 @@
...
+			delta.transaction = this;
+			this.deltas.push( delta );
+
+			return delta;
+		}
+
+		/**
+		 * Transaction provides iterator interface which will iterate over deltas in the transaction.
+		 */
+		[ Symbol.iterator ]() {
```
Reinmar: Unnecessary since `this.delta` is public.

scofalik: I don't agree, there is already one thing you have to do each time you add delta (set `delta.transaction`) and there might be more in the future. Also the internals might change. I think it is nice to do `transaction.addDelta( delta )` than `transaction.deltas.push( delta )`.

src/document/delta/transaction-base.js
```diff
@@ -0,0 +1,139 @@
...
+				 * @error transaction-register-taken
+				 * @param {String} name
+				 */
+				throw new CKEditorError(
+					'transaction-register-taken: This transaction method is already taken.',
+					{ name: name } );
+			}
+
+			Transaction.prototype[ name ] = function() {
+				creator.apply( this, [ this.doc, this ].concat( Array.from( arguments ) ) );
```
Reinmar: I don't like that instead of assigning the creator directly to the prototype we wrap it and concatenate args with what would already be accessible as `this`. This is an unnecessary step.

src/document/delta/transaction-base.js
Reinmar: So if we consider that the internals may change, then `this.delta` should be private. In other words – you need the iterator only if you cannot access the array.

src/document/node.js
Reinmar: Note for me in the future: it's better to have an iterator if attrs are private because we can change them to map or set or whatever and all these types have the same common API – they are iterable. We could consider always returning an array, because, even as the API docs state, you will normally access the entire set. However, after starting keeping attr in map or set we would be creating an array everytime, even though you don't need it. On the other hand, `Array.from( attrsMap )` may be faster than `Array.from( node.getAttrs() )` when `getAttrs()` returns iterable, but it's too early to think about it.

src/document/delta/splitdelta.js
```diff
@@ -0,0 +1,67 @@
...
+
+		const insert = new InsertOperation( Position.createAfter( splitElement ), copy, doc.version );
+		doc.applyOperation( insert );
+		delta.addOperation( insert );
+
+		const move = new MoveOperation(
+			position,
+			Position.createFromParentAndOffset( copy, 0 ),
+			splitElement.getChildCount() - position.offset,
+			doc.version );
```
scofalik: New line.

Reinmar:
I corrected the API docs and commented on bits I didn't like.

src/document/document.js
pjasiun: Changed.

src/document/delta/transaction-base.js
pjasiun: Removed. I also removed iterator form delta, because of the same reason.

src/document/delta/transaction-base.js
pjasiun: Done. I also moved returning `this` to the registered functions, so the register simple add function to the prototype. And I added tests to check if functions are chainable.

Reinmar:
> Done. I also moved returning this to the registered functions, so the register simple add function to the prototype. And I added tests to check if functions are chainable.

So is the branch back on R? Or did something remain?

src/document/node.js
pjasiun: I renamed `getAttrIterator` to `getAttrs`.

pjasiun:
Now it is :)

PR #90, opened 2015-11-25
T/59: OT for Tree Model

### LINKED ISSUE #59: OT for operations
It should be possible to transform one operation with another. To do it:
 - `Operation.transform( anotherOperation, anotherMoreImportant )` - method should be implemented for all types of operations, where `anotherOperation` is operation to transform with and `anotherMoreImportant` is boolean to determine which operation should we use in case of conflicts,
 - `NoOperation` class should be introduced, for operations which do nothing,
 - existing operation classes should be changed, if needed, most likely move and remove will not need to have nodes,
 - `Operation.reverseOperation()` method should be changed, if needed.

After this ticket it should be possible to combine two sets of operations and undo any operation from the history, still manually working with operations, without a mechanism for handling operation collections.

scofalik:
Fixes #59.


pjasiun:
First of all I am really impressed. No only the code, but also number and quality of test docs and comments is impressive. Good job. But... ;)

src/document/transformoperation.js
```diff
@@ -0,0 +1,434 @@
+/**
```
pjasiun: I think we could move transformoperation to the operation namespace. It is not an operation, but I think that the namespace is something more than the place only for operations. It could be called transform, so it will not be confused with the operations, so: `document/transformoperation` -> `document/operation/transform`

tests/_tools/tools.js
```diff
@@ -67,4 +67,29 @@
 			return count;
 		}
 	};
+
+	bender.tools.operations = {
+		expectOperation: ( Position, Range ) => {
```
pjasiun: This could be moved to the `test/transformoperations` as an internal function.

src/document/operation/changeoperation.js
```diff
@@ -77,45 +80,15 @@ CKEDITOR.define( [ 'document/operation/operation', 'ckeditorerror' ], ( Operatio
...
-					if ( !value.node.hasAttr( oldAttr ) ) {
-						/**
-						 * The attribute which should be removed does not exists for the given node.
-						 *
-						 * @error operation-change-no-attr-to-remove
-						 * @param {document.Node} node
-						 * @param {document.Attribute} attr
-						 */
-						throw new CKEditorError(
-							'operation-change-no-attr-to-remove: The attribute which should be removed does not exists for given node.',
```
pjasiun: You removed both checks and errors (`operation-change-no-attr-to-remove` and `operation-change-attr-exists`), as we talked F2F you said they were problematic during the OT. The problem is that they are here because of reason: if the operation with the wrong `oldAttr` will be applied, then reversing that operation will now work properly. So the issue is not solved, it is hidden, and may be very hard to solve later. Also this check helped me in finding issues when I was working on the change delta multiple times, so I prefer it stays here.

src/document/operation/changeoperation.js
```diff
@@ -124,6 +97,35 @@ CKEDITOR.define( [ 'document/operation/operation', 'ckeditorerror' ], ( Operatio
 		getReversed() {
 			return new ChangeOperation( this.range, this.newAttr, this.oldAttr, this.baseVersion + 1 );
 		}
+
+		clone() {
+			return new ChangeOperation( this.range.clone(), this.oldAttr, this.newAttr, this.baseVersion );
```
pjasiun: You introduced `clone` methods to all operation classes, position class and range class. On the one hand we could use `utils.clone`, but on the other cloned object is not the instance of the class (`( utils.clone( position ) instanceof Position ) === false`) and `utils.clone` will not know how deep the object should be cloned (in this case range have to be cloned, but attributes not), so this is fine for me.

src/document/operation/changeoperation.js
```diff
@@ -124,6 +97,35 @@ CKEDITOR.define( [ 'document/operation/operation', 'ckeditorerror' ], ( Operatio
...
+
+		/**
+		 * Checks whether this operation has conflicting attributes with given {@link document.operation.ChangeOperation}.
+		 * This happens when both operations changes an attribute with the same key and they either set different
+		 * values for this attribute or one of them removes it while the other one sets it.
+		 *
+		 * @param {document.operation.ChangeOperation} otherOperation Operation to check against.
+		 * @returns {boolean} True if operations have conflicting attributes.
+		 */
+		conflictsAttributesWith( otherOperation ) {
```
pjasiun: `conflictsAttributesWith` should got to `transformoperation`. On the other hand we could have helper to compare two attributes and better handle the "null" operation. I met similar problem creating `ChangeDelta`: https://github.com/ckeditor/ckeditor5-core/blob/master/src/document/delta/changedelta.js#L144-L145.I think it will be changed when we will close https://github.com/ckeditor/ckeditor5-core/issues/63, so we could leave comment there and fix it later.

src/document/position.js
```diff
@@ -97,13 +119,13 @@ CKEDITOR.define( [ 'utils', 'ckeditorerror' ], ( utils, CKEditorError ) => {
 		}
 
 		/**
-		 * Two positions equal if paths are equal.
+		 * Two positions equal if paths are equal and roots are the same.
 		 *
 		 * @param {document.Position} otherPosition Position to compare.
 		 * @returns {Boolean} True if positions equal.
 		 */
 		isEqual( otherPosition ) {
```
pjasiun: I am not sure if if we need `Position.isEqual` anymore is one can use `Position.compareWith == Position.SAME`. Or if we assume that comparing position will be popular we could add also `Position.isBefore` and `Position.isAfter`.

src/document/position.js
```diff
@@ -113,11 +135,21 @@ CKEDITOR.define( [ 'utils', 'ckeditorerror' ], ( utils, CKEditorError ) => {
 		 *
 		 * @returns {Number[]} Path to the parent.
 		 */
-		getParentPath() {
```
pjasiun: We should decide something here: https://github.com/ckeditor/ckeditor5-design/issues/88.

src/document/position.js
```diff
@@ -171,7 +203,196 @@ CKEDITOR.define( [ 'utils', 'ckeditorerror' ], ( utils, CKEditorError ) => {
...
+		 * positions `[ 1, 1, 3 ]`, `[ 1, 1, 4 ]`, `[ 1, 1, 5 ]`. Since our position was in the second moved node,
+		 * the transformed position will be in a sub-tree of a node at `[ 1, 1, 4 ]`. Looking at original path, we
+		 * took care of `[ 2, 3 ]` part of it. Now we have to add the rest of the original path to the transformed path.
+		 * Finally, the transformed position will point to `[ 1, 1, 4, 1 ]`.
+		 *
+		 * @param {document.Position} from Beginning of the moved range.
+		 * @param {document.Position} to Position where the range is moved.
+		 * @returns {document.Position} Combined position.
+		 */
+		getCombined( from, to ) {
```
pjasiun: `getCombined` is messy: `from` is a begging of the range, but there is not end of that range. `to` is a position (not the end of that range as one could assume). We assume that the position is inside the range, so the function can be called only in specific cases. It looks more like the helper, not the end-user API. It could be moved to the `transformoperation` and changed to be an internal helper or it could be fixed: the first parameter should be a range and the function should check if he position is inside that range. In this case it could be renamed to `getTransformedByMove` and stay here.

src/document/range.js
```diff
@@ -72,6 +72,211 @@ CKEDITOR.define( [ 'document/positioniterator', 'document/position' ], ( Positio
 		}
 
 		/**
+		 * Checks whether this {document.Range range} intersects with the given {document.Range range}.
+		 *
+		 * @param {document.Range} otherRange Range to check.
+		 * @returns {boolean} True if ranges intersect.
+		 */
+		intersectsWith( otherRange ) {
```
pjasiun: We do not need `intersectsWith` when you can check exactly the same using `getCommon`.

src/document/range.js
```diff
@@ -72,6 +72,211 @@ CKEDITOR.define( [ 'document/positioniterator', 'document/position' ], ( Positio
...
+		 * 	let otherRange = new Range( new Position( [ 1 ], root ), new Position( [ 2 ], root ) );
+		 * 	let transformed = range.getCommon( otherRange ); // null - ranges have no common part
+		 *
+		 * 	otherRange = new Range( new Position( [ 3 ], root ), new Position( [ 5 ], root ) );
+		 * 	transformed = range.getCommon( otherRange ); // range from [ 3 ] to [ 4, 0, 1 ]
+		 *
+		 * @param {document.Range} otherRange Range to compare with.
+		 * @returns {document.Range} Range that is common part of given ranges.
+		 */
+		getCommon( otherRange ) {
```
pjasiun: `getCommon` should be renamed to `getIntersection`.

src/document/range.js
```diff
@@ -72,6 +72,211 @@ CKEDITOR.define( [ 'document/positioniterator', 'document/position' ], ( Positio
...
+		 * Examples:
+		 * 	let range = new Range( new Position( [ 2, 7 ], root ), new Position( [ 4, 0, 1 ], root ) );
+		 * 	let otherRange = new Range( new Position( [ 1 ], root ), new Position( [ 2 ], root ) );
+		 * 	let transformed = range.getCommon( otherRange ); // null - ranges have no common part
+		 *
+		 * 	otherRange = new Range( new Position( [ 3 ], root ), new Position( [ 5 ], root ) );
+		 * 	transformed = range.getCommon( otherRange ); // range from [ 3 ] to [ 4, 0, 1 ]
+		 *
+		 * @param {document.Range} otherRange Range to compare with.
+		 * @returns {document.Range} Range that is common part of given ranges.
```
pjasiun: It should be `{document.Range|Null}`.

src/document/transformoperation.js
```diff
@@ -0,0 +1,434 @@
...
+			/**
+			 * Returns an array containing the result of transforming
+			 * {document.operation.InsertOperation} by {document.operation.InsertOperation}.
+			 *
+			 * @param {document.operation.InsertOperation} a Operation that will be transformed.
+			 * @param {document.operation.InsertOperation} b Operation to transform by.
+			 * @param {Boolean} isStrong Flag indicating whether this operation should be treated as more important
+			 * when resolving conflicts.
+			 * @returns {Array.<document.operation.InsertOperation>} Result of the transformation.
+			 */
```
pjasiun: Because these are not the part of the API (it is an internal function) is should not use jsDoc style for the comment. It not be parsed by the documentation tool.

src/document/transformoperation.js
```diff
@@ -0,0 +1,434 @@
...
+ * to create two operations as a result. It would be impossible to create just one operation that handles
+ * modifications needed to be applied to the tree. This is why Array is returned instead of single object.
+ * All returned operations has to be applied (or further transformed) to get an effect that was intended in
+ * pre-transformed operation.
+ *
+ * @function transformOperation
+ * @param {document.operation.Operation} a Operation that will be transformed.
+ * @param {document.operation.Operation} b Operation to transform by.
+ * @param {Boolean} isStrong Flag indicating whether this operation should be treated as more important
+ * when resolving conflicts.
```
pjasiun: I think you could add one more paragraph about the `isStrong` parameter. Explain why we need this parameter. Explain why we can not assume that operation a is always stronger and switch the order of the parameters when we need. In fact it makes no sense, but I believe that this is the question one will ask when he will read this code for the first time.

src/document/transformoperation.js
```diff
@@ -0,0 +1,434 @@
...
+CKEDITOR.define( [
+	'document/operation/insertoperation',
+	'document/operation/changeoperation',
+	'document/operation/moveoperation',
+	'document/operation/nooperation',
+	'document/range',
+	'utils'
+], ( InsertOperation, ChangeOperation, MoveOperation, NoOperation, Range, utils ) => {
+
+	// When we don't want to update an operation, we create and return a clone of it.
```
pjasiun: We don't want or don't need to update an operation?

pjasiun:
Also `transformoperation` tests could have no document as a requirement. They can create root node without a document and use number as a `version`. It's not a big deal but it would be nice proof that you do not need any document to do transformations.

tests/_tools/tools.js
scofalik: I agree, previously it was used in multiple files, I forgot to bring it back from `tools.js`.

src/document/transformoperation.js
scofalik: I leave it for discussion.

src/document/operation/changeoperation.js
scofalik: Yes it was connected with some problems when I was doing ChangeOperation by ChangeOperation OT. I don't remember the case exactly but in oldAttr was wrong in that scenario and it required making more complicated OT script. Maybe this already got solved during one of a few refactorings of all those scripts. I'll check it once again.

src/document/operation/changeoperation.js
scofalik: It's better as it is.

src/document/position.js
scofalik: Well, you are right. It all boils down to code readability vs API maintenance. `if ( a.isEqual( b ) ) { }` for sure looks better than `if ( a.compareWith( b ) == Position.SAME )` and is easier to remember. Also when you try to understand part of code your brain has to process what compareWith Position SAME actually means, so I think that it is better to have function names that sounds natural and are not distracting. I like the idea of `Position.isBefore` and `Position.isAfter`. I know that on several occasions you fought for keeping API simple but the main reason for keeping it simple is easy maintenance and those methods would be very simple and will be based on just one method `compareWith`. I am all for such methods.

src/document/position.js
scofalik: This is a method that I was hestitating the most and honestly, I don't like it and I am not sure if it should be a part of API, at least in this form. Hence my super-long JSDoc for it, since it is very hard to understand what is really going on with this function.

scofalik: I think that `getTransformedByMove` would be too complicated. But of course it depends how we  will understand it in the context of `Position`. I will think on the future of this method.

src/document/range.js
scofalik: This is true. When combined with your other comment about changing `getCommon` to `getIntersection` it would also read natural:
`if ( a.getIntersection( b ) === null ) { }`.

src/document/range.js
scofalik: I agree.

src/document/position.js
scofalik: I thought a bit about it and commented on that issue. According to my comment, this should be a normal method, not a getter.

src/document/transformoperation.js
scofalik: So, basically, earlier these functions were parts of classes so I JSDoc'ed them. Also the docs were a bit more different. I didn't want to delete these comments, but you are right, now it doesn't have much sense to keep them in this state.

src/document/position.js
pjasiun: Also to do `a.compareWith( b ) == Position.SAME` you need to require `Position` and you do not need to do it to do `a.isEqual( b )`. I agree that `isEqual` is faster to process in the brain, so we could add `Position.isBefore` and `Position.isAfter`.

src/document/transformoperation.js
scofalik: Don't need to.

src/document/position.js
pjasiun: So let's move it to `transformoperation` as a private function, at least for now.

src/document/range.js
scofalik: After short investigation ( :) ) it appeared, that `intersectsWith` is used only within `Range` class and it was overcomplicated. I removed it and provided simple condition to check in it's place.

Of course I also removed tests for `intersectsWith` but before I did it, I used the mentioned check with the tests for `intersectsWith` and they were all green.

src/document/range.js
scofalik: Will fix.

src/document/transformoperation.js
scofalik: Will fix.

src/document/operation/changeoperation.js
scofalik: Are you sure there won't be any other case where the user might like to compare two ChangeOperations? Well, on the other hand, we won't be encouraging users to create operations on their own but use `Transaction` API so you might be right. I'll move it.

src/document/position.js
scofalik: This is a correct observation. BTW. I am fine with `utils.compareArrays` flags (`SAME`, `EXTENSION`, `PREFIX`) because you have to require `utils` anyway to use `compareArrays`.

scofalik: One more thing - I am not sure if `isEqual` is a better name than `isSame` but since we also have `isEqual` methods in other classes I'd just stick with this name to make it unified across classes.

src/document/position.js
scofalik: It might be useful for `SmartPosition` and `SmartRange`.

scofalik:
This is a fine point. Document was used there because it was already used in xxxxOperations tests. I will get rid of it.

I fixed most of issues. The ones that are still open are:
* moving `transformoperation.js` to `operation` directory
* removed errors in `ChangeOperation`
* `Position.getCombined`

scofalik:
I fixed the rest of issues.

Some comment on `.getCombined`. I couldn't come up with a perfect solution. I introduced `Position.getTransformedByMove` which:
* is something that a user might expect of our API
* includes some code that was used in, AFAIR, two places in `operation/transform.js` and it simplified the code a bit
* hides `.getCombined` in it.

Unfortunately, it would be impossible to write a function that covers all places where `.getCombined` was used. Mainly because sometimes I "combine" position which technicaly is not in moved range. This happens when I want to "combine" whole range and it is exact as moved range. Then, range boundaries are the same, so they are outside of moved range. I can't "hack" it with extra condition because this is sometimes expected behavior.

So it works for `Position`s and does not work for `Range`s. Well, after all this is `Position` method so it should not be expected to work for moving `Range` boundaries. That's why I decided to keep `.getCombined` as protected method (named `_getCombined`) and use it for ranges.

There is a temptation to create `getTransformedByMove` for `Range` class but it would be complicated (well, it would take most of the complex and long code from OT algorithms). I am not sure if we won't need it though - for `SmartRanges` or something.

src/document/operation/transform.js
```diff
@@ -0,0 +1,409 @@
...
+			return null;
+		}
+
+		let start = ranges[ 0 ].start;
+		let end = ranges[ 0 ].end;
+		const root = ranges[ 0 ].root;
+
+		for ( let i = 1; i < ranges.length; i++ ) {
+			if ( ranges[ i ].root != root ) {
+				continue;
```
pjasiun: There is missing coverage for this line. Please add test of remove this condition if it is not possible to get here.

pjasiun:
We need to decide something about `transformoperation` file location. Beside that and the missing test the rest of changes are fine for me.

scofalik:
I simplified `joinRanges` so it does exactly what is needed as it is private helper function. Well, it is private helper at least for now because I have a feeling I will need something like this in further developement, but right now it is kept simple.

I also fixed branch coverage in `haveConflictingAttributes` which actually (surprise, surprise) let me to discovering a bug.

scofalik:
> We need to decide something about transformoperation file location. Beside that and the missing test the rest of changes are fine for me.

I already moved it to `operation/transform.js` as you suggested.

Reinmar:
Make sure to sort methods in the code in the following order:

* public:
  * instance
  * class
* private
  * instance
  * class

src/document/position.js
```diff
@@ -5,7 +5,12 @@
 
 'use strict';
 
-CKEDITOR.define( [ 'utils', 'ckeditorerror' ], ( utils, CKEditorError ) => {
+CKEDITOR.define( [ 'document/rootelement', 'utils', 'ckeditorerror' ], ( RootElement, utils, CKEditorError ) => {
+	const SAME = 0;
+	const AFTER = 1;
+	const BEFORE = -1;
+	const DIFFERENT = -2;
```
Reinmar: Why so negative?

src/document/range.js
```diff
@@ -72,6 +72,192 @@ CKEDITOR.define( [ 'document/positioniterator', 'document/position' ], ( Positio
...
+			return position.isAfter( this.start ) && position.isBefore( this.end );
+		}
+
+		/**
+		 * Returns an array containing one or two {document.Range ranges} that are results of transforming this
+		 * {document.Range range} by inserting `howMany` nodes at `insertPosition`. Two {document.Range ranges} are
+		 * returned if the insertion was inside this {document.Range range}.
+		 *
+		 * Examples:
+		 * 	let range = new Range( new Position( [ 2, 7 ], root ), new Position( [ 4, 0, 1 ], root ) );
```
Reinmar: Indentation: spaces -> 2tabs. Or we can fix it later after we'll figure out which doc generator we want to use. But in general, since we use MD, 2tabs are the minimum to make it a code block in this flavour. We can later agree though to use the GH's alternative (triple backticks). So... ignore this is you're lazy :D.

src/document/position.js
scofalik: I did it so you can easily use `Array.prototype.sort` with `Position.compareWith`

Reinmar: OK, @scofalik says that `-1, 0, 1` can be passed to `Array#sort()`. It sounds useful, but it's a bit hacky. Especially that `DIFFERENT` should most likely result in an error being thrown, as this means that positions come from different documents (or something). Therefore, I would propose to normalise these so `compareWith()` returns `0, 1, 2, 3` and we could add a static method to `Position` which could be used to generate `sort()`'s callbacks.

src/document/operation/transform.js
```diff
@@ -0,0 +1,406 @@
...
+
+CKEDITOR.define( [
+	'document/operation/insertoperation',
+	'document/operation/changeoperation',
+	'document/operation/moveoperation',
+	'document/operation/nooperation',
+	'document/range',
+	'utils'
+], ( InsertOperation, ChangeOperation, MoveOperation, NoOperation, Range, utils ) => {
+	// When we don't want to update an operation, we create and return a clone of it.
```
Reinmar: Move these helpers below the `return` statement.

src/document/operation/changeoperation.js
Reinmar: :+1: for `subject.clone()` if there are any doubts.

Reinmar:
I commented on all pieces which I didn't understand or like and pushed a few commits with API docs or code style fixes. Once all these matters are resolved, R+ from me.

scofalik:
As I said, I made flags in `Position` negative to use them in `Array.sort`, however it appeared that after refactoring this is not used anymore. I brought back "normal" values for those flags.

The sorting was used only in `operation/transform.js`. It wasn't even really `Position`s sorting, but `Range`s sorting, so sort-callback should be written in `Range` class if anywhere. Moreover, you suggested throwing error when one tries to sort ranges with different roots. But in that case it is expected that some ranges may have different roots and we want them sorted anyway (it's enough if all ranges of same root are "sticking together").

So we can't even tell what sorting method is proper and how oftern will we use it (I expect that not often). For those reasons I wouldn't include it in the API. For now, I just changed the callback function in `operation/transform.js` so it takes care of different roots.

I also sorted ( :) ) methods in classes and I hope that this PR is finally ready for merge.

scofalik:
I've pushed one more commit:

* There was a bug with ranges sorting. It was more complicated than I thought. I fixed it.
* I added a special case when MoveOperation is transformed by MoveOperation and one of those operations range and target is contained in the other one's range. The results were okay-ish before but now are more in line with what a user would expect.
* Code coverage is now 100%.

PR #94, opened 2015-11-27
t/93: Implement data binding between the models

### LINKED ISSUE #93: Implement data binding between the models
To make binding between features, Controllers and Views easier, Model should implement binding interface, so some part (selected attributes) of the model follows the state of another model.

With such an interface, it's quite easy to "delegate" a part of the Controller (or feature) model into View model without duplicated `on( 'change:*', ... )` listeners.

Required by #92.

oleq:
Fixes #93.

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -30,6 +30,24 @@ CKEDITOR.define( [ 'emittermixin', 'ckeditorerror', 'utils' ], ( EmitterMixin, C
 			 */
 			this._attributes = {};
 
+			/**
+			 * Map containing bindings of this model to external models.
+			 * See {@link #bind}.
+			 *
+			 * @property {Map}
+			 * @private
+			 */
+			this._boundTo = new Map();
+
+			/**
+			 * Object that stores which attributes of this model are bound.
+			 * See {@link #bind}.
+			 *
+			 * @property {Object}
+			 * @private
+			 */
+			this._bound = {};
+
 			// Extend this instance with the additional (out of state) properties.
 			if ( properties ) {
 				utils.extend( this, properties );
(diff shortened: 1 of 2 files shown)


Reinmar:
I think you overdid with the number of checks in `bind().to().as()`. It's good to check some dynamic conditions which you expect to be hard to debug or which could frequently occur. But I think that this is too much:

```
			if ( !toModel ) {
				/**
				 * No model supplied.
				 *
				 * @error model-bind-to-no-model
				 */
				throw new CKEditorError( 'model-bind-to-no-model: No model supplied.' );
			}

			if ( !( toModel instanceof Model ) ) {
				/**
				 * An instance of Model is required.
				 *
				 * @error model-bind-to-wrong-model:
				 */
				throw new CKEditorError( 'model-bind-to-wrong-model: An instance of Model is required.' );
			}

			if ( !isStringArray( toAttrs ) ) {
				/**
				 * Model attributes must be strings.
				 *
				 * @error model-bind-to-wrong-attrs
				 */
				throw new CKEditorError( 'model-bind-to-wrong-attrs: Model attributes must be strings.' );
			}
```

scofalik:
To reduce the amount of throwed errors I try to combine them. In the quoted code snippet, I'd put `error model-bind-to-no-model` and `model-bind-to-wrong-model` in one error. Maybe we should start from there. Also you have typo in `model-bind-to-wrong-model` doc (colon at the end of error name).

Reinmar:
E.g. if no model specified or that arg is not a model it will quickly fail anyway. I'd remove those 3 checks as well as:


```
			if ( !callback ) {
				/**
				 * No callback function supplied.
				 *
				 * @error model-bind-as-no-callback
				 */
				throw new CKEditorError( 'model-bind-as-no-callback: No callback function supplied.' );
			} else if ( typeof callback !== 'function' ) {
				/**
				 * Callback must be a Function.
				 *
				 * @error model-bind-as-wrong-callback
				 */
				throw new CKEditorError( 'model-bind-as-wrong-callback: Callback must be a Function.' );
			}
```

And:

```

				if ( !isStringArray( unbindAttrs ) ) {
					/**
					 * Attributes must be strings.
					 *
					 * @error model-unbind-wrong-attrs
					 */
					throw new CKEditorError( 'model-unbind-wrong-attrs: Attributes must be strings.' );
				}
```

As all of them check things that could be checked statically (so are rather easy to debug). Alternatively, we can wrap it with `isDebug`.

scofalik:
I don't understand why it would be good to wrap it in `isDebug`. We would have to keep them in code and doc anyway so what is the point? These checks won't take much time to process. The only drawback is the sheer amount of code and docs. So we either remove/combine some of errors or leave them as they are.

Reinmar:
> I don't understand why it would be good to wrap it in isDebug.

Mainly code size and somewhere at the end also performance (in this specific case neglectable). And I don't think that it's a problem that those will stay in the docs, because most often developers will code their features using a dev version of CKE anyway. For the other cases it does not matter.

Anyway, I leave it up to Olek. We can do whatever we want in this specific case (even keep this as this is a very low level function that could be coded in this way), but I wanted to communicate the general approach that we should have, which is to not control things that could be controlled statically (in a typical strongly typed language :D).

src/model.js
```diff
@@ -104,6 +122,424 @@ CKEDITOR.define( [ 'emittermixin', 'ckeditorerror', 'utils' ], ( EmitterMixin, C
...
+	 * attributes. Also sets the initial state of `chain._bindModel` bound attributes.
+	 *
+	 * @private
+	 * @param {BindChain} chain The chain initialized by {@link Model#bind}.
+	 */
+	function setupBinding( chain ) {
+		// If there's already a binding between the models (`chain._bindModel` listens to
+		// `chain._current.model`), there's no need to create another `change` event listener.
+		if ( !getModelBindingsToCurrent( chain ) ) {
+			chain._bindModel.listenTo( chain._current.model, 'change', ( evt, attrName, value ) => {
```
Reinmar: I commented this line directly on a commit. For the reference: https://github.com/ckeditor/ckeditor5-core/pull/94#commitcomment-14682505

src/model.js
```diff
@@ -104,6 +122,424 @@ CKEDITOR.define( [ 'emittermixin', 'ckeditorerror', 'utils' ], ( EmitterMixin, C
...
+	/**
+	 * Check if the `model` has given `attrs`.
+	 *
+	 * @private
+	 * @param {Model} model Model to be checked.
+	 * @param {Array} arr An array of `String`.
+	 * @returns {Boolean}
+	 */
+	function hasAttributes( model, attrs ) {
+		return attrs.findIndex( a => {
```
Reinmar: Wouldn't this be `array#some`?

src/model.js
```diff
@@ -104,6 +122,424 @@ CKEDITOR.define( [ 'emittermixin', 'ckeditorerror', 'utils' ], ( EmitterMixin, C
...
+
+	/**
+	 * Check if all entries of the array are of `String` type.
+	 *
+	 * @private
+	 * @param {Array} arr An array to be checked.
+	 * @returns {Boolean}
+	 */
+	function isStringArray( arr ) {
+		return arr.findIndex( a => typeof a !== 'string' ) == -1;
```
Reinmar: The same as above.

Reinmar:
I've made a quick review and to sum it up:

1. Listening on `sourceModel#change` will spoil the performance. Can be reported as a followup though, as we have a functional method already.
2. `array#indexOf() == -1` can be rewritten as `!array#some()` which IMO is more readable.
3. Too many assertions in the code. No strong opinion in this case (so it's up to you whether to remove them, leave or keep only in the debug mode), but in general we should not validate so basic things.

Reinmar:
One more reason for R-: ckeditor5-core/src/model.js doesn't have 100% code coverage.

oleq:
## Performance issues

Yes. I also realised that rather than using events, we should implement it the opposite way (for `A.bind( 'x' ).to( B )`): `x` attribute setter in `B` should make changes to `A.x` rather than `A` listen to the changes in `B` and react when `B.x` has changed. But it was too late when I noticed it.

It's not gonna be a huge change to implement though as I understand how to do it and what to re–write On the other hand I'm not quite sure the negative performance impact will be visible. As long as it's not DOM – it's fast. I created a followup in #101.

## Errors

My idea was (is!) that low–level API should be fail–safe in terms of DX. It's **not** about us but 3rd–party developers who will use `Model#bind` in many places and deal with occasional errors. While the concept of model binding is quite clear, the implementation definitely isn't and debugging it might become quite a nightmare for inexperienced developers. I think that for the sake of high DX level, we can create and maintain more errors in the crucial places. But as @scofalik pointed out, I reduced the number of errors by merging them.

## `Array#indexOf`

I used `every` instead of `some` and shortened the code even more, making it more intuitive at the same time.

## Test code coverage

Well... I increased it :P

Reinmar:
> My idea was (is!) that low–level API should be fail–safe in terms of DX

I generally agree that the low-level API can be safer for the sake of DX. But it's a question of how far we want to go. IMO, checking things which will fail quite clearly just one line below will help some people in understanding what's going on, but there always will be some for whom this won't be enough verbosity. It's like perfecting some task – at the beginning you have a significant improvement by making just small changes. Later, big changes bring small improvements, because you're just that close to the perfection.

I'm also comparing our approach with other projects which either don't do any validation (and fail quietly or with std JS exceptions) or do a little of it. Let's head for a bit more than "a little", but not try to perfect it.

PR #106, opened 2015-12-04
T/89 Changes emitter for tree model document

pjasiun:
Fixes https://github.com/ckeditor/ckeditor5-core/issues/89.


src/document/operation/changeoperation.js
```diff
@@ -130,6 +134,12 @@ CKEDITOR.define( [
 					value.node.setAttr( newAttr );
 				}
 			}
+
+			return {
+				range: this.range,
+				oldAttr: oldAttr ? oldAttr : undefined,
```
scofalik: Is this really needed? Also I don't think that this should be `undefined` while the operation expected argument is `null` when it is "empty". We should stick with one of them and I propose `null` as for me it passes more data. Then, we should simply return `oldAttr` and `newAttr`. Remember to change docs in `document.Document`.

src/document/operation/insertoperation.js
```diff
@@ -58,6 +64,10 @@ CKEDITOR.define( [
 
 		_execute() {
 			this.position.parent.insertChildren( this.position.offset, this.nodeList );
+
+			return {
+				range: new Range( this.position, Position.createFromPositionAndShift( this.position, this.nodeList.length ) )
```
scofalik: In my "OT" pull request I introduced `Range.createFromPositionAndOffset` which does exactly that.

src/document/operation/moveoperation.js
```diff
@@ -121,6 +128,11 @@ CKEDITOR.define( [
 			const removedNodes = sourceElement.removeChildren( sourceOffset, this.howMany );
 
 			targetElement.insertChildren( targetOffset, removedNodes );
+
+			return {
+				sourcePosition: this.sourcePosition,
+				range: new Range( this.targetPosition, Position.createFromPositionAndShift( this.targetPosition, this.howMany ) )
+			};
```
scofalik: `Range.createFromPositionAndOffset`

src/document/position.js
```diff
@@ -221,6 +221,25 @@ CKEDITOR.define( [ 'document/rootelement', 'utils', 'ckeditorerror' ], ( RootEle
...
+		 *		 const pos = new Position( [ 1, 2, 8 ], root );
+		 *
+		 *		 Position.createFromPositionAndShift( pos, -3 ) // [ 1, 2, 5 ]
+		 *		 Position.createFromPositionAndShift( pos, 5 )  // [ 1, 2, 13 ]
+		 *
+		 * @param {document.Position} referencePosition Reference position.
+		 * @param {Number} shift Offset shift.
+		 * @returns {document.Position}
+		 */
+		static createFromPositionAndShift( referencePosition, shift ) {
```
scofalik: We may keep it if we want to, but if it's only use is creating "flat" ranges I'd consider removing it.

tests/document/document-change-event.js
```diff
@@ -0,0 +1,186 @@
+/**
```
scofalik: Create seperate directory `tests/document/document` and put `document.js` and `change.js` (or `event-change.js` or `change-event.js`) there?

scofalik:
I've pushed small commit with minor fixes (mostly typos). Please consider comments I've made.

src/document/operation/changeoperation.js
pjasiun: Agreed. This is a leftover after the previous idea of having separate events for add, removing and changing attribute.

src/document/operation/insertoperation.js
pjasiun: Changed. Additional I created: https://github.com/ckeditor/ckeditor5-core/issues/107.

src/document/operation/moveoperation.js
pjasiun: Done.

src/document/position.js
pjasiun: Removed.

tests/document/document-change-event.js
pjasiun: A was also thinking about it. Done.

scofalik:
I've added one commit with removed dependency and merged the branch.

Issue #100, opened 2015-12-01
Range iteration tools


pjasiun:
There are multiple ways one could want to iterate over a range:
 - standard way - get all opening, closing tags and characters,
 - get nodes - return all nodes, but only once, so the same element will not be returned twice for opening and closing tags,
 - get flat - get only top level nodes, for example if one moved something he could be interested in the nodes which changed parent, so the to level nodes, 
 - merge characters - return text per group of sibling characters with the same attributes, in many cases you want to get text and do not want to handle every character separately. 

Note that one may want to get flat & merge character or get nodes & merge characters, but characters on different levels should not be merged.

This could be handled by passing additional parameters to `positionIteration` (which will be renamed to TreeWalker https://github.com/ckeditor/ckeditor5-core/issues/85), but we should avoid making `PositionIteration` too complex. If it became complex code at this stage of development will became unreasonable complex code soon it the future. Alternatively these could be range methods or simple tools.

scofalik:
`Range.getPositions()` was requested -- iterate through all positions of range. It is kind-of a "standard way" described above, but it is impossible to do in `for..of` loop, because `position` is a property of the `PositionIterator`.

scofalik:
`getAllNodes`, `getPositions`, `getTopLevelNodes` and `getMinimalFlatRanges` have been introduced as `Range` methods. Some of them may take `mergeCharacters` flag as a parameter and then return all characters with the same attributes as one `TextNode`. See PR #139.

pjasiun:
With `TreeWalker` options it becomes a very powerful tools. But some changes need to be still added to close this tickets.

First of all `Range.getTopLevelNodes` is very unoptimised. It iterates over all children, and children of these children to simply get to the next sibling. It would be much simpler if `ignoreChildren` would be another option in the `TreeWalker`. The `TreeWalker` would have an additional `if`, but thanks to it not only the code will work much faster, but also we will be able to use any combination of options we need. Now `Range.getTopLevelNodes` calls `getMinimalFlatRanges` and iterate over all of them. User have no simple way to check if one flat range ends and another starts or execute some code every time the flat range changes. This method is also the most useful for flat ranges and it such cases there is no need to call `getMinimalFlatRanges` at all. I think the method should be removed and the option to the `TreeWalker` should be added, so the code would looks like this:
```
const flatRanges = this.getMinimalFlatRanges();
for ( let range of flatRanges ) {
[...]

Issue #65, opened 2015-10-28
Refactor PositionIterator


pjasiun:
`PositionIterator` keeps only the `Position` so it calculates a node and offset every time the `next()` method is called and recalculates a new position based on the new node and offset.

This could be improved by caching node and offset, but it will not work if the document will change during iteration.

On the other hand, if document changes, the position makes no sense, so we could assume that document is immutable during iterating.

Note that then, iterator should work differently for `SmartPosition` (see #64). There changing document during iteration makes sense.

pjasiun:
Or there could be two iterators: `Static` and `Dynamic`, because caching may improve also iterating `SmartPosition` as long as document does not change during iteration.

scofalik:
In my implementation of simple Tree Model for OT purposes, my `Node` class had properties `.prevNode` and `.nextNode` pointing to, well, previous and next siblings of that node in the structure. Those were updated whenever a node has been added or removed from the tree (this is fairly easy operation done in a static amount of time). If you had these, you could cache the last visited node and use it's `nextNode` and `prevNode` properties. Whenever `nextNode` or `prevNode` is `null`, you know that you reached the last sibling and you have to "move up". I think this approach is less affected by changes on the document during iteration. You could easily remove some nodes during iteration and still have a valid iterator state. Of course if you move the last visited node *somewhere* strange things may happen :).

By the way, having `nextSibling` and `previousSibling` as real, set properties that are not re-evaulated each time they are called will bring some optimization anyway, so I propose to change them. I am sure you don't need code snippet for that, but if you want to have a quick look how it could look, you may check out: https://github.com/ckeditor/ckeditor5-design/blob/op-trans/ot_ranges/ot.js#L44-L63

pjasiun:
Now we can not store `nextSibling` and `previousSibling` in each node, because it cost too much memory. Now we store only `character` and `parent` per each `Character` node. With `nextSibling` and `previousSibling` it would be more-less twice that much. 

But I hope we will introduce text nodes compression (#62) soon. Then it would be a very nice solution.

Reinmar:
I'm not sure how and when iterators will be used as we don't have any use cases implemented yet. However, in CKE4 we always had the same issue that our ranges or walker may easily get desynced. We got used to that and learned to deal with the situation, so I would say that we can live with such iterator in CKE5 too.

Second thing – natively you have two iterators as well – `TreeWalker` and `NodeIterator`:

[...]

Issue #62, opened 2015-10-28
Node list compression


pjasiun:
Because every character is now a separate class, tree model can consume a lot of memory. The situation is event worse because model will keep the history of changes including all historical nodes and operations with lists of nodes to change.

The solution is to introduce compression inside `NodeList` class which is used for both operations and child list. The idea is to keep existing interface with characters as children, but to implement it internally in the way that all siblings characters with the same attributes will be stored as as a single `Text` (string with attributes). 

Then characters may be a proxy witch act like a single character, but have reference to the whole text and modify it if characters attribute changes. The character class does not need to be lightweight anymore, because there will not be many references to these classes kept in the code.

If user call `getChild` at the middle of the string the new character class will be created with the reference to that string. Disadvantage would be that it will not be possible to simple compare references to two characters, but is does not seems to be a common operation. On the other hand, it will be possible to get `Position` before/after both character and compare them.

This mechanism needs to be smart. The most common operations should work fast, without creating a special API for them. For example consider changing attribute on the set of following characters (like applying bold on the range). The change attribute for the character should check if the new list of attributes is the same as the list of attributes of the previous text and move that character to the previous text. Special method to apply style on the range would work faster, but having separate API to work on whole text nodes is something we want to avoid, at least at the begging.

Thanks to this compression we will not only save memory because of having less character classes, but also reduce number of attributes and maybe be able to change the attribute data structure, which is getting strange and slow because of decisions to save memory.

scofalik:
Here is the new behavior of `NodeList`:

We have `Text` class, which is only for representing a string with attached attributes to it. It is mostly used to append characters with set attributes to `Element` instances. `Text` is not a `Node` class.

We got rid of `Character` class and introduced similiar `TextNode` class. `TextNode` is extending `Node` class. It is used as a representation of some text in tree model. This class is a bit weird but is a best consensus between user expectations, code efficiency, API complication and amount of time spent coding it.

`NodeList` does not contain singular characters like before. Now all consecutive characters with same attributes are kept as one instance of `Text` class. So internal `NodeList` array is keeping instances of `Element` and `Text` inside it. Still, outside, characters are visible as singular beings. So if your `NodeList` contain a `P` element, `foo` string and `DIV` element if you iterate over it (without merging characters) you get `P` element, `f` text node, `o` text node, `o` text node, `DIV` element. Similary, when you use `nodeList.get( 3 )` you will get `o` text node.

[...]

PR #122, opened 2015-12-21
T/63 AttributeList API + Range iterators.

### LINKED ISSUE #63: Data structure for attributes
To save memory, attributes are instanced of `Attribute` class which can be reused. We introduced special method to force `Attribute` reuse (`Attribute.register`). Then, not to split this attributes into separate key and value, node contains a `Set` of attributes.

The problem is that then attributes are not stored by key all attribute operations (`hasAttr`, `getAttr`, `setAttr`, `removeAttr`) need to scan whole set, what is very ineffective. During the text transformation from the Tree Model to DOM every pair of siblings needs to be compared to ensure that attributes do not change.

When the node compression will be done (#62) there will be less attributes stored and the memory usage may not such a be a problem anymore. Then we could store attributes in a map, instead of set, where the attribute key will be a map key and whole attribute or the attribute value will be a map value. 

Also current way of compering attributes will need to be changed. Now the `_hash` attribute is kept which is build from the key and value. If they will be stored and compared separately we could build `_hash` only from value. Or we could consider storing values stringified.

Finally `Attribute.register` may not makes sens after these changes, becoming an unnecessary complication.

scofalik:
Fixes #63, #100

### DIFF
--- a/src/treemodel/attributelist.js
+++ b/src/treemodel/attributelist.js
@@ -8,111 +8,148 @@
 import Attribute from './attribute.js';
 
 /**
- * List of attributes. Used to manage a set of attributes added to and removed from an object containing
- * AttributeList.
+ * List of attributes.
  *
  * @class treeModel.AttributeList
  */
-export default class AttributeList {
+export default class AttributeList extends Map {
 	/**
-	 * Creates a list of attributes.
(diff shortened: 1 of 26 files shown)


scofalik:
During short F2F discussion we decided that this PR should also contain #62 and probably #65.

scofalik:
Will create new PR for this.

PR #124, opened 2015-12-30
t/123: Model binding fails with as() when using custom attribute names.

### LINKED ISSUE #123: Model binding fails with as() when using custom attribute names
The following binding 
```js
A.bind( 'x' )
	.to( B, 'y' )
	.to( C, 'z' )
	.as( ( y, z ) => y + z );
```
results in `A.x === undefined`.

oleq:
Fixes #123. It turned out that it is OK that `boundAttrs === undefined` when `chain._callback` is set (called `.to( ... ).as( _callback )`).


oleq:
This pull request is included in #134 and will no longer be required after #133 is closed.

PR #137, opened 2016-01-12
t/135: Unable to move items with auto–generated ids between Collections.

### LINKED ISSUE #135: Unable to move items with auto–generated ids between Collections
```js
const colA = new Collection();
const colB = new Collection();
const mA = new Model();
const mB = new Model();

colA.add( mA );
colB.add( mB );
colB.add( colA.remove( mA ) );
```
ends up with `Error: collection-add-item-already-exists`. This is because `colA` sets `mA.id` to 1 and `colB` sets `mB.id` to 1 so, when moving `mA` from `colA` to `colB`, there's a collision between ids.

Because of this issue, moving `Controllers` between `ControllerCollections` of different parent `Controllers` is pretty impossible.

----

Solution 1: Remove `id` when calling `Collection#remove`. 

Solution 2: Make auto–generated ids globally unique (at the moment ids are unique within an instance of the `Collection` only).

----

Solution 1 means that an item cannot reside in multiple `Collections`, which is IMHO wrong. There could be many kinds of `Collections`, which collect items by different (i.e. logical) criteria, so it must be possible that an item is shared between different `Collections`.

So I'm for Solution 2.

oleq:
Fixes #135.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -8,6 +8,7 @@
 import EmitterMixin from './emittermixin.js';
 import CKEditorError from './ckeditorerror.js';
 import utilsObject from './lib/lodash/object.js';
+import utils from './utils.js';
 
 /**
  * Collections are ordered sets of objects. Items in the collection can be retrieved by their indexes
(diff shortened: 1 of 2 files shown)


Reinmar:
I pushed two more commits. The last one fixes a test which now fails, due to a bug in the new implementation.

PR #145, opened 2016-01-18
Merge batch, batch-base and register.

### LINKED ISSUE #129: Merge batch, batch-base and register
These 3 files separation was an awful circular dependencies workaround. Now, with ES6 modules we can merge these 3 files back together.

Also `register` can be exported as a separate function, so one can use:

`import register form batch.js`

If `register` is everything it need.

scofalik:
Fixes #129


tests/treemodel/node.js
```diff
@@ -22,8 +22,6 @@ describe( 'Node', () => {
 	let attrFooBar;
 
 	before( () => {
-		charB = new Character( 'b' );
```
pjasiun: It should not be here.

src/treemodel/document.js
```diff
@@ -229,4 +229,4 @@ export default class Document {
 	 */
 }
 
-objectUtils.extend( Document.prototype, EmitterMixin );
```
pjasiun: It should not be here.

pjasiun:
Rebase changes on master and revert changes in these two files. Beside these everything looks much better.

PR #147, opened 2016-01-19
T/18b: Introduce utils.mix()

### LINKED ISSUE #18: Introduce the Mixin class

Reinmar:
Alternative version to #28.

Fixes #18.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -7,7 +7,6 @@
 
 import EmitterMixin from './emittermixin.js';
 import CKEditorError from './ckeditorerror.js';
-import utilsObject from './lib/lodash/object.js';
 import utils from './utils.js';
 
 /**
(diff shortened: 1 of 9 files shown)


Reinmar:
See https://github.com/ckeditor/ckeditor5-core/issues/18#issuecomment-172804286.

Reinmar:
See https://github.com/ckeditor/ckeditor5-core/issues/18#issuecomment-172868364

PR #161, opened 2016-01-25
T/63 b Attributes API

scofalik:
Another attempt at #63.

There is one fragment of code I'd like to discuss. See `TextFragment.setAttribute` and `TextFragment.removeAttribute`.

So, basically, `TextFragment` should be seen as a collection of `CharacterProxy` instances. `CharacterProxy` is readonly so `TextFragment` naturally also should be. `TextFragment` only represents some nodes in tree model and because it is a "proxy" to them, we should not be able to make changes on it that are reflected in the tree model.

For some simplicity reasons, `setAttribute` and `removeAttribute` has been introduced. They are used by `AttributeOperation#_execute`. These are `protected` methods, What I don't like about them is that they change whole text node, not only the fragment of the text node represented by `TextFragment`. I highlighted it in a comment. So when a user irresponsibly use this method, he/she may change not only nodes in `TextFragment` but also others.

I could change this behavior. Why I haven't done it yet?
* This will probably be used only in `AttributeOperation`.
* This is a protected method, so should be used only in core internals.
* This could be only solved by spliting text node in `NodeList`, applying changes and then checking if text node should be merged. This would have to be done for every call to `setAttribute` and `removeAttribute`. Compared to current implementation, splitting and merging is called only once for attribute operation execution. Also, the code in `TextFragment` would get more complicated. Also we would have to properly handle that `TextFragment` if we end up splitting it's text node or merging it's text node.

So, to sum up, I would have to write less clean and slower code. I was also thinking about throwing error if we want to `setAttribute` or `removeAttribute` on a `TextFragment` that does not contain whole text node but it would be silly to have such a "basic" method that works only on some instances. Also I would have to refer `NodeListText` in the error description and we want to keep that class as hidden as possible, so we won't confuse users any further.


pjasiun:
Cannot `setAttribute` and `removeAttribute` be the methods of the `NodeList`? I understand it would causes changes in the attribute operation algorithm, but it should not be a big nor complex change. I prefer avoiding quirky API, even the protected one. One date someone will work on this piece of code and we we should reduce the number of WTFs to the minimum.

Also these changes are not what #63 is about. The ticket is about removing `Attributre` class, at least as a parameter of the `setAttribute`, switching to the simpler API using key and value, and this PR is the follow up of https://github.com/ckeditor/ckeditor5-core/issues/62.

scofalik:
In F2F discussion we decided that we can keep `setAttribute` and `removeAttribute` for `Element` because it may come in handy. Consider following scenario: we create an `Element` in one part of code and in totally different part we want to decide whether that `Element` should have any attributes. We can't do it if only way to set attributes is via constructor.

After spending more than half of a day thinking about introducing `setAttribute` and `removeAttribute` to `NodeList`, I came to conclusion that it is not that easy.

By doing so, we are defeating the primary purpose of this refactoring. We were supposed to simplify `AttributeOperation` by removing "hacky way" of setting attribute for `TextFragment`. You were against a simple line of code. After removing it, you propose a complicated loop that will retrive parts of `NodeList`s that are in the range and then use method of those `NodeList`. We are doing it wrong. We can't propose introducing complicated algorithms to "clean code" :).

I spent a lot of time thinking on balancing code complication, code efficiency and clean API. Right now I feel we are starting to waste time. I'd do one of those:

1) If we are keeping `setAttribute` and `removeAttribute` for `Element`, I'd keep it for `TextFragment`. I can tweak those methods for `TextFragment` so they do automatic `NodeList` splitting. It will add extra code but at least we are introducing additonal feature and we don't have to make those methods protected.

2) If we want to keep `TextFragment` a readonly representation I suggest adding different handling 
for `TextFragment` in `AttributeOperation`:
```javascript
// ...
if ( node instanceof TextFragment ) {
node.first._nodeListText._attrs.set( ... ); // ( a )
}
else {
node.setAttribute( ... );
}
```

3) Like in 2), but line `( a )` might be simplified to
```javascript
node._attrs.set( ... )
```
By introducing getter in `TextFragment`
```javascript
get _attrs() {
return this.firsts._nodeListText._attrs;
}
```

scofalik:
BTW. I pushed some code to t/63-d that you might like to compare to t/63-c. You will see how much more complicated code we would have to introduce. This code is buggy but I have no intention of making it right because I think it is pointless.

scofalik:
We can do
```javascript
if ( node instanceof TextFragment ) {
node.first.parent._children.setAttribute( node.first.getIndex(), node.text.length, ... );
}
else {
node.setAttribute( ... );
}
```

pjasiun:
:+1: for the last option.

src/treemodel/document.js
```diff
@@ -215,10 +215,11 @@ export default class Document {
 	 * {@link #graveyard graveyard root}.
 	 * @param {treeModel.Position} [changeInfo.sourcePosition] Change source position. Exists for 'remove', 'reinsert' and 'move'.
 	 * Note that for 'reinsert' the source position will be in the {@link #graveyard graveyard root}.
-	 * @param {treeModel.Attribute} [changeInfo.oldAttr] Only for 'attr' type. If the type is 'attr' and `oldAttr`
-	 * is `undefined` it means that new attribute was inserted. Otherwise it contains changed or removed attribute.
-	 * @param {treeModel.Attribute} [changeInfo.newAttr] Only for 'attr' type. If the type is 'attr' and `newAttr`
-	 * is `undefined` it means that attribute was removed. Otherwise it contains changed or inserted attribute.
+	 * @param {*} [changeInfo.key] Only for 'attr' type. Key of changed / inserted / removed attribute.
```
pjasiun: Isn't key a string?

src/treemodel/element.js
```diff
@@ -117,4 +117,44 @@ export default class Element extends Node {
...
+	 * Removes all attributes from the element and sets given attributes.
+	 *
+	 * @param {Iterable.<*>} attrs Iterable object containing attributes to be set. See {@link treeModel.Node#getAttributes}.
+	 */
+	setAttributesTo( attrs ) {
+		if ( langUtils.isPlainObject( attrs ) ) {
+			this._attrs = utils.objectToMap( attrs );
+		} else {
+			this._attrs = new Map( attrs );
+		}
```
pjasiun: These 4 lines repeat 4 times (5 including treeView). Maybe it should be a tool?

pjasiun:
I fixed both issues. Rest of code looks much better, simpler. Good job!

PR #185, opened 2016-02-05
T/72 Deltas Transformations

### LINKED ISSUE #72: OT for deltas
It should be possible to transform delta with another delta. By default it should transform all operations inside one delta with all operations inside another delta, but, for some special cases, special behavior should be provided, to bring result which user expect most.

Note that, in these special cases, two deltas should have special behavior to get the same result on both sides. This is why delta can not have special behavior with operations, because operations can not have special behavior with deltas. 

Deltas and operations are two different abstraction levels and we should not mix them.

scofalik:
This PR contains changed to position move algorithms, shell for deltas transformations and a few examples of special cases for transformations with tests.

After this gets merged we can close #72 as I created two follow-up issues (#183 #184) with what is left to be done.

### DIFF
--- a/src/treemodel/delta/mergedelta.js
+++ b/src/treemodel/delta/mergedelta.js
@@ -6,6 +6,7 @@
 'use strict';
 
 import Delta from './delta.js';
+import SplitDelta from './splitdelta.js';
 import { register } from '../batch-base.js';
 import Position from '../position.js';
 import Element from '../element.js';
@@ -19,7 +20,35 @@ import CKEditorError from '../../ckeditorerror.js';
  *
  * @class treeModel.delta.MergeDelta
  */
-export default class MergeDelta extends Delta {}
+export default class MergeDelta extends Delta {
+	get moveOperation() {
+		return this.operations[ 0 ] || null;
+	}
+
+	get removeOperation() {
+		return this.operations[ 1 ] || null;
+	}
+
+	get position() {
+		return this.removeOperation ? this.removeOperation.sourcePosition : null;
+	}
+
+	getReversed() {
+		if ( this.moveOperation ) {
+			let delta = new SplitDelta();
+
+			for ( let op of this.operations ) {
+				delta.addOperation( op.getReversed() );
+			}
+
+			delta.operations.reverse();
+
+			return delta;
+		}
+
+		return null;
+	}
+}
 
 /**
  * Merges two siblings at the given position.
--- a/src/treemodel/delta/transform.js
+++ b/src/treemodel/delta/transform.js
@@ -1,0 +1,170 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+import OT from '../operation/transform.js';
+
+import AttributeDelta from './attributedelta.js';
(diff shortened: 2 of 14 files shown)


src/treemodel/delta/transform.js
```diff
@@ -0,0 +1,170 @@
...
+import MergeDelta from './mergedelta.js';
+import MoveDelta from './movedelta.js';
+import RemoveDelta from './removedelta.js';
+import SplitDelta from './splitdelta.js';
+import Delta from './delta.js';
+
+import arrayUtils from '../../lib/lodash/array.js';
+import utils from '../../utils.js';
+
+const baseDeltas = [ AttributeDelta, InsertDelta, MoveDelta, RemoveDelta ];
```
pjasiun: base deltas?

scofalik: This is for auto resolving some of the conflicts. `MoveOperation` from `MoveDelta` and `RemoveDelta` should always be more important than `MoveOperation`s from other deltas.  I might just hardcode it because it will probably won't have more use than this.

pjasiun:
We decided not to put this code into the master branch, yet. I renamed the branch name to something more meaningful: https://github.com/ckeditor/ckeditor5-core/tree/ot-for-deltas Lets close this PR for now, we will talk about these changes later.

src/treemodel/delta/transform.js
pjasiun: I am not sure about the hardcoded list of deltas here, if the whole mechanism was supposed to be scalable.

pjasiun: If it only used to get priority should not it be called `priorities`? Or would not it be better if deltas have priorities?

src/treemodel/delta/mergedelta.js
```diff
@@ -19,7 +20,35 @@ import CKEditorError from '../../ckeditorerror.js';
  *
  * @class treeModel.delta.MergeDelta
  */
-export default class MergeDelta extends Delta {}
+export default class MergeDelta extends Delta {
+	get moveOperation() {
+		return this.operations[ 0 ] || null;
+	}
+
+	get removeOperation() {
```
pjasiun: I would be cool if deltas encapsulate informations about operations and return only meta informations about the delta (taken from operations). For example instead of: `delta.howManyMoved` instead of `delta.moveOperation.howMany`.

src/treemodel/position.js
```diff
@@ -287,20 +288,22 @@ export default class Position {
 	 * @param {Boolean} insertBefore Flag indicating whether moved nodes are pasted before or after `insertPosition`.
 	 * This is important only when `targetPosition` and this position are same. If that is the case and the flag is
 	 * set to `true`, this position will get transformed by range insertion. If the flag is set to `false`, it won't.
+	 * @param {Boolean} [sticky] Flag indicating whether this position "sticks" to range, that is if it should be moved
+	 * with the moved range if it is equal to one of range's boundaries.
 	 * @returns {treeModel.Position} Transformed position.
 	 */
-	getTransformedByMove( sourcePosition, targetPosition, howMany, insertBefore ) {
+	getTransformedByMove( sourcePosition, targetPosition, howMany, insertBefore, sticky ) {
```
pjasiun: As we were talking, we should consider moving `sticky` parameter to the places where this function is called. Function with 5 parameters looks like something no one will be able use properly.

PR #191, opened 2016-02-08
t/187: Weak instanceof Node check in DOMEmitterMixin

### LINKED ISSUE #187: Weak instanceof Node check in DOMEmitterMixin
Checks like [this](https://github.com/ckeditor/ckeditor5-core/blob/956ae659806d86abdcd4f39af2efe38888d67888/src/ui/domemittermixin.js#L178) fail when Node belongs to a different window (i.e. in an iframe). Then `window1.Node != window2.Node`.

oleq:
Fixes #187.

### DIFF
--- a/src/ui/domemittermixin.js
+++ b/src/ui/domemittermixin.js
@@ -175,7 +175,7 @@ const DOMEmitterMixin = extend( {}, EmitterMixin, {
 
 		// Check if emitter is an instance of DOM Node. If so, replace the argument with
 		// corresponding ProxyEmitter (or create one if not existing).
-		if ( emitter instanceof Node ) {
+		if ( isDomNode( emitter ) ) {
 			args[ 0 ] = this._getProxyEmitter( emitter ) || new ProxyEmitter( emitter );
 		}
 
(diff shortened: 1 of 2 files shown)


src/ui/domemittermixin.js
```diff
@@ -254,6 +254,15 @@ function getNodeUID( node ) {
 	return node[ 'data-ck-expando' ] || ( node[ 'data-ck-expando' ] = utils.uid() );
 }
 
+// Checks if given node is native DOM Node.
+//
+// @private
+// @param {Node} node
+// @return {Boolean} True when native DOM Node.
+function isDomNode( node ) {
+	return node && node.nodeType && node.nodeName;
```
Reinmar: That's too weak. If we'll implement DOM-like class structure, then it's enough that we had these two (rather popular) properties and it blows up unless `addEventListener` was also defined. So I would also check here the presence of this method.

BTW. Are you sure that all kinds of DOM nodes have the `nodeName` defined? I see they most likely have (https://developer.mozilla.org/en/docs/Web/API/Node/nodeName) but I've never used this property in the wild.

oleq: I think it unlikely to happen but OK, I'll add `addEventListener` check. 

> BTW. Are you sure that all kinds of DOM nodes have the nodeName defined?

[Yes](https://www.w3.org/TR/DOM-Level-2-Core/core.html#ID-1841493061).

Reinmar: > Yes.

I didn't ask about the theory :P. I asked about the practice. But let's assume that we're safe here.

PR #194, opened 2016-02-09
t/188: Allow easy binding syntax for existing elements.

### LINKED ISSUE #188: Allow easy binding syntax for existing elements
The following View class syntax is easy to use and extend:

```js
this.template = {
	tag: 'iframe',
	attrs: {
		class: [
			'ck-framededitable',
		],
		width: this.bindToAttribute( 'width' ),
		height: this.bindToAttribute( 'height' )
	},
	on: {
		load: 'loaded',
		input: 'input'
	}
};
```

but there's not easy way to re–use it in case of an existing element, i.e. when it's passed directly to View instance (like inline instance's editable). In such case, all bindings (`bindToAttribute` and `on:`) must be created manually, which is pretty annoying. 

A method like `View.setElementBindings( element, def )` would do the trick. It could be called as a standalone method but still used internally in View to create new elements from `this.template`.

**Note:** In the future it is expected that for better performance View instances will clone existing DOM element (once created from `this.template`, cached) instead of invoking it from scratch. So such method will be needed to set bindings and listeners of a clone anyway.

oleq:
Resolves #188.

### DIFF
--- a/src/ui/view.js
+++ b/src/ui/view.js
@@ -97,7 +97,7 @@ export default class View {
 		}
 
 		// Prepare pre–defined listeners.
-		this._prepareElementListeners( this.template );
+		this._prepareTemplateListeners( this.template );
 
 		this._template = new Template( this.template );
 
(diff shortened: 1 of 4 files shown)


oleq:
I'm still not quite sure about the solution. On the one hand, it will do fine for cloned DOM elements (with children) because they have their attributes and `textContent` already set, so `View#applyTemplateToElement` activates model bindings (`attrs`) and listeners (`on`) **only**. On the other hand it is useless if one wants to actually set all attributes (also not bound), because current implementation ignores them.

So perhaps, if current implementation is OK, `View#applyTemplateToElement` should be split into `View#applyAttributeBindings` and `View#applyDOMListeners`, which handle `attrs: {}` and `on: {}` sections separately and have a meaning. Otherwise `View#applyTemplateToElement` suggests that it will handle all attributes, `textContent`, etc. WDYT?

oleq:
Also take a look on #195.

src/ui/view.js
```diff
@@ -377,7 +405,7 @@ export default class View {
 
 		// Repeat recursively for the children.
 		if ( def.children ) {
-			def.children.map( this._prepareElementListeners, this );
+			def.children.map( this._prepareTemplateListeners, this );
```
Reinmar: Use `forEach()`.

oleq:
The conclusion is that `applyTemplateToElement` will not create new Nodes but will re–set all attributes, listeners and `textContent`.

oleq:
New implementation in #198.

PR #198, opened 2016-02-11
t/188a: Allow easy binding syntax for existing elements

### LINKED ISSUE #188: Allow easy binding syntax for existing elements

oleq:
Closes #188.

### DIFF
--- a/src/ui/view.js
+++ b/src/ui/view.js
@@ -424,7 +424,7 @@ export default class View {
 
 		// Repeat recursively for the children.
 		if ( def.children ) {
-			def.children.map( this._createTemplateListenerAttachers, this );
+			def.children.forEach( this._createTemplateListenerAttachers, this );
 		}
 	}
 }
--- a/src/ui/view.js
+++ b/src/ui/view.js
@@ -268,9 +268,10 @@ export default class View {
 
 	/**
 	 * Destroys the view instance. The process includes:
-	 *  1. Removal of child views from {@link #regions}.
-	 *  2. Destruction of the {@link #regions}.
-	 *  3. Removal of {#link #_el} from DOM.
+	 *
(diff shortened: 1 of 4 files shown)


src/ui/view.js
```diff
@@ -336,19 +369,26 @@ export default class View {
...
-		let on = def.on;
+	_createTemplateListenerAttachers( def ) {
+		const on = def.on;
 
-		if ( on ) {
+		// Don't create attachers if they're already here or in the context of the same (this) View instance.
+		if ( on && ( !on._listenerAttachers || on._listenerView != this ) ) {
 			let domEvtName, evtNameOrCallback;
 
+			Object.defineProperty( on, '_listenerAttachers', {
```
Reinmar: Shouldn't you use some symbol here?

Reinmar:
I reported #201 for the issue I commented on, because it's a deeper problem in fact. Other than that, +1.

PR #213, opened 2016-02-17
Marking test valid only for browser environment.

### LINKED ISSUE #212: Mark tests that are valid only for browser environment.
After completing issue [#84](https://github.com/ckeditor/ckeditor5/issues/84) from `ckeditor5` repository we need to mark tests that are not valid for Node.js environment ( for example using browser DOM API ). This can be achieved by adding `browser-only` tag to `bender-tags:` comment in test file.

szymonkups:
Adding proper tags to fix #212.


Reinmar:
This PR got 100% outdated after the repo split.

PR #225, opened 2016-03-01
T/224

pjasiun:
Fixes https://github.com/ckeditor/ckeditor5-core/issues/224 with some additional clean-up.

### DIFF
--- a/src/treemodel/operation/attributeoperation.js
+++ b/src/treemodel/operation/attributeoperation.js
@@ -96,7 +96,7 @@ export default class AttributeOperation extends Operation {
 				 * The attribute which should be removed does not exists for the given node.
 				 *
 				 * @error operation-attribute-no-attr-to-remove
-				 * @param {core.treeModel.Node|core.treeModel.TextFragment} item
+				 * @param {core.treeModel.Item} item
 				 * @param {String} key
 				 * @param {*} value
 				 */
--- a/src/treemodel/operation/attributeoperation.js
+++ b/src/treemodel/operation/attributeoperation.js
@@ -90,19 +90,19 @@ export default class AttributeOperation extends Operation {
 	}
 
 	_execute() {
-		for ( let item of this.range.getAllNodes( true ) ) {
+		for ( let item of this.range.getItems( true ) ) {
 			if ( this.oldValue !== null && item.getAttribute( this.key ) !== this.oldValue ) {
 				/**
 				 * The attribute which should be removed does not exists for the given node.
 				 *
 				 * @error operation-attribute-no-attr-to-remove
-				 * @param {core.treeModel.Node} node
+				 * @param {core.treeModel.Node|core.treeModel.TextFragment} item
 				 * @param {String} key
 				 * @param {*} value
 				 */
 				throw new CKEditorError(
 					'operation-attribute-no-attr-to-remove: The attribute which should be removed does not exists for given node.',
-					{ node: item, key: this.key, value: this.oldValue }
+					{ item: item, key: this.key, value: this.oldValue }
 				);
 			}
 
--- a/src/treemodel/operation/attributeoperation.js
+++ b/src/treemodel/operation/attributeoperation.js
@@ -90,7 +90,7 @@ export default class AttributeOperation extends Operation {
 	}
 
 	_execute() {
-		for ( let item of this.range.getItems( true ) ) {
+		for ( let item of this.range.getItems() ) {
 			if ( this.oldValue !== null && item.getAttribute( this.key ) !== this.oldValue ) {
 				/**
 				 * The attribute which should be removed does not exists for the given node.
--- a/src/treemodel/position.js
+++ b/src/treemodel/position.js
@@ -411,6 +411,8 @@ export default class Position {
 	/**
 	 * Creates a new position after given node.
 	 *
+	 * @see core.treeModel.TreeWalkerValue
+	 *
 	 * @param {core.treeModel.Node} node Node the position should be directly after.
 	 * @returns {core.treeModel.Position}
 	 */
@@ -430,6 +432,8 @@ export default class Position {
 
 	/**
 	 * Creates a new position before the given node.
+	 *
+	 * @see core.treeModel.TreeWalkerValue
 	 *
 	 * @param {core.treeModel.node} node Node the position should be directly before.
 	 * @returns {core.treeModel.Position}
--- a/src/treemodel/position.js
+++ b/src/treemodel/position.js
@@ -301,7 +301,7 @@ export default class Position {
 	/**
 	 * Checks whether this position is after given position.
 	 *
-	 * **Note:** see {treeModel.Position#isBefore}.
+	 * **Note:** see {@link core.treeModel.Position#isBefore}.
 	 *
 	 * @param {core.treeModel.Position} otherPosition Position to compare with.
 	 * @returns {Boolean} True if this position is after given position.
@@ -411,7 +411,7 @@ export default class Position {
 	/**
 	 * Creates a new position after given node.
 	 *
-	 * @see core.treeModel.TreeWalkerValue
+	 * @see {@link core.treeModel.TreeWalkerValue}
 	 *
 	 * @param {core.treeModel.Node} node Node the position should be directly after.
 	 * @returns {core.treeModel.Position}
@@ -433,7 +433,7 @@ export default class Position {
 	/**
 	 * Creates a new position before the given node.
 	 *
-	 * @see core.treeModel.TreeWalkerValue
+	 * @see {@link core.treeModel.TreeWalkerValue}
 	 *
 	 * @param {core.treeModel.node} node Node the position should be directly before.
 	 * @returns {core.treeModel.Position}
--- a/src/treemodel/range.js
+++ b/src/treemodel/range.js
@@ -253,7 +253,7 @@ export default class Range {
 
 	/**
 	 * Returns an iterator that iterates over all {@link core.treeModel.Item items} that are in this range and returns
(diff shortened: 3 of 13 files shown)


src/treemodel/operation/attributeoperation.js
```diff
@@ -90,19 +90,19 @@ export default class AttributeOperation extends Operation {
...
 	_execute() {
-		for ( let item of this.range.getAllNodes( true ) ) {
+		for ( let item of this.range.getItems( true ) ) {
 			if ( this.oldValue !== null && item.getAttribute( this.key ) !== this.oldValue ) {
 				/**
 				 * The attribute which should be removed does not exists for the given node.
 				 *
 				 * @error operation-attribute-no-attr-to-remove
-				 * @param {core.treeModel.Node} node
+				 * @param {core.treeModel.Node|core.treeModel.TextFragment} item
```
scofalik: This can be `core.treeModel.Item` now

src/treemodel/position.js
```diff
@@ -411,6 +411,8 @@ export default class Position {
 	/**
 	 * Creates a new position after given node.
 	 *
+	 * @see core.treeModel.TreeWalkerValue
```
scofalik: Are you sure this is the correct way to use @see tag? I am not sure either and have seen various way to use it.

Reinmar: Run `gulp docs` and you'll know :D.

src/treemodel/range.js
```diff
@@ -261,32 +252,51 @@ export default class Range {
 	}
 
 	/**
-	 * Returns an iterator that iterates over all {@link core.treeModel.Node nodes} that are in this range and returns them.
+	 * Returns an iterator that iterates over all {@link core.treeModel.Item items} that are in this range and returns
+	 * them together together with additional information like length or {@link core.treeModel.Position positions},
```
scofalik: "together together"

src/treemodel/range.js
```diff
@@ -261,32 +252,51 @@ export default class Range {
...
+	 *
+	 * **Note:** this method will not return a parent node of start position. This is in contrary to
+	 * {@link core.treeModel.TreeWalker} which will return that node with `'ELEMENT_END'` type. This method also
+	 * returns each {@link core.treeModel.Element} once, while simply used {@link core.treeModel.TreeWalker} might
+	 * return it twice: for `'ELEMENT_START'` and `'ELEMENT_END'`.
 	 *
-	 * **Note:** this method will not return a parent node of start position. This is in contrary to {@link core.treeModel.TreeWalker}
-	 * which will return that node with `'ELEMENT_END'` type. This method also returns each {@link core.treeModel.Element} once,
-	 * while simply used {@link core.treeModel.TreeWalker} might return it twice: for `'ELEMENT_START'` and `'ELEMENT_END'`.
+	 * **Node:*** because this method does not return {@link core.treeModel.TreeWalkerValue values} with the type of
```
scofalik: Node :)

src/treemodel/range.js
```diff
@@ -261,32 +252,51 @@ export default class Range {
...
-	 * (`false`) objects. Defaults to `false`.
-	 * @returns {Iterable.<core.treeModel.Node|treeModel.TextFragment>}
+	 * should be returned as one {@link core.treeModel.TextFragment} (`true`) or one by one as multiple
+	 * {@link core.treeModel.CharacterProxy} (`false`) objects. Defaults to `false`.
+	 * @returns {Iterable.<core.treeModel.TreeWalkerValue>}
 	 */
-	*getAllNodes( mergeCharacters ) {
-		let it = new TreeWalker( { boundaries: this, mergeCharacters: mergeCharacters } );
-		let step;
+	*getValues( mergeCharacters ) {
```
scofalik: I am not sure if I like the name of the method, despite it returning an object which class has "Value" in name.

It feels strange to ask about "range values". What values range has? Range might contain nodes or items, but not values.

Reinmar: +1

src/treemodel/treewalker.js
```diff
@@ -238,13 +250,23 @@ function formatReturnValue( type, item ) {
...
 
 /**
  * Object returned by {@link core.treeModel.TreeWalker} when traversing tree model.
  *
- * @typedef {Object} core.treeModel.TreeWalkerItem
- * @property {treeModel.TreeWalkerItemType} type
- * @property {treeModel.Node|treeModel.TextFragment} item Value between old and new position of {@link core.treeModel.TreeWalker}.
+ * @typedef {Object} core.treeModel.TreeWalkerValue
+ * @property {treeModel.TreeWalkerValueType} type
+ * @property {core.treeModel.Item} item Item between previous and next positions of {@link core.treeModel.TreeWalker}.
```
scofalik: I don't like "between previous and next positions of TreeWalker". Old and new seems more appropriate and doesn't care whether we go `.next()` or `.previous()`. Alternative description might be `Item encountered during TreeWalker iteration` or smth like this.

src/treemodel/treewalker.js
```diff
@@ -238,13 +250,23 @@ function formatReturnValue( type, item ) {
...
+ * @property {treeModel.Position} previousPosition Previous position of the iterator. For `'ELEMENT_END'` it is the last
```
scofalik: Please update to `core.treeModel.Position`

src/treemodel/treewalker.js
```diff
@@ -238,13 +250,23 @@ function formatReturnValue( type, item ) {
...
+ * position inside the element. For all other types it is the position before the item. Note that it is more
+ * efficient to use this position then calculate the position before the node using
+ * {@link core.treeModel.Position.createBefore}. It is also more efficient to get the position after node by shifting
+ * `previousPosition` by `length`, using {@link core.treeModel.Position#getShiftedBy}, then calculate the position using
+ * {@link core.treeModel.Position.createAfter}.
+ * @property {treeModel.Position} nextPosition Next position of the iterator. For `'ELEMENT_START'` it is the first
```
scofalik: Please update to `core.treeModel.Position`

src/treemodel/range.js
scofalik: After reading tests I am not sure whether someone would even like to use `getValues` if they have `getItems` which essentialy does very similar things, but is easier to operate with.

pjasiun: So could you suggest a better name?

scofalik: Maybe this should not have a name but just be a default range iterator?

pjasiun: The whole point of this change is to return `position` and `length` together with the `item`. I have no better idea to call the object will contains such a group of data.

pjasiun: Maybe `getData` and `TreeWalkerData`?

scofalik: AFAIR we do not use range iterator all that often because it was cumbersome.

pjasiun: I removed default iterator, because we do not know if one what to `mergeCharacters` or not and if one want to return `ELEMENT_END` or not. I agree it was useless with the form it was. For me we could make this method default iterator with the `mergeCharacters`, so it will skip `ELEMENT_END` and merge characters. It is the most convenient for me, but I do not know if it is what people expect.

scofalik: The most common usage will be either this or `getItems`. If someone want to "go deep", they always have the possibility to use `TreeWalker` and set boundaries to given range or use `getPositions`. Cases where you just want to see what items are in range are the most common.

src/treemodel/range.js
```diff
@@ -261,32 +252,51 @@ export default class Range {
...
+	/**
+	 * Use {@link core.treeModel.Range#getValues}, but return only {@link core.treeModel.Item items} from
+	 * {@link core.treeModel.TreeWalkerValue}.
+	 *
+	 * @see {core.treeModel.Range#getValues}
+	 * @param {Boolean} [mergeCharacters] See {@link core.treeModel.Range#getValues}.
+	 * @returns {Iterable.<core.treeModel.Item>}
+	 */
+	*getItems( mergeCharacters ) {
+		for ( let value of this.getValues( mergeCharacters ) ) {
```
pjasiun: In the F2F discussion we decided that `mergeCharacters` will be usually true so should be changed to `singleCharacters`, which is false by default.

src/treemodel/operation/attributeoperation.js
pjasiun: Fixed.

src/treemodel/position.js
pjasiun: Fixed.

src/treemodel/range.js
pjasiun: Fixed.

src/treemodel/range.js
pjasiun: Fixed.

src/treemodel/treewalker.js
pjasiun: Fixed.

src/treemodel/treewalker.js
pjasiun: Fixed, here and is some other places.

src/treemodel/treewalker.js
pjasiun: Fixed.

pjasiun:
I have changed default `TreeWalker` option to merge characters, since this is the way we usually use this class. The change is pretty big, because it means that we prefer `TextFragment` over `CharacterProxy` right now, but it also looks like a change in the good direction. I also removed the `Range.getValues` method and added a iterable interface to the `Range` which now returns merged values without `ELEMENT_END`. It looks like a most useful structure.

scofalik:
I've pushed some minor fixes.

PR #229, opened 2016-03-04
Deltas transformations

### LINKED ISSUE #72: OT for deltas

scofalik:
This pull request contains changes introducing transformations for deltas. Fixes #72.

There are some things that might be done but I'd propose to add them as separate tickets. If you feel that they should be included in this PR, let me know, but I'd like to merge this with master but not spend more time on it right now.

These are:
1. Tests and documentation for tests utils tools.
2. Way to automate delta creation (some kind of "default" constructor). Right now delta has to be build from scratch everytime because delta constructor does not create any operations. After creating delta instance (i.e. `let wrapDelta = new WrapDelta()`) developer has to add all operations by hand. This won't be used by outside developers but deltas are created in a few places in code (i.e. tests, methods registered to `Batch`, delta transformations) and it is a bit painful that one has to create all the operations by hand. Something like `let wrapDelta = new WrapDelta( wrapRange, wrapElement );` would be great and is doable.
3. Some methods in `Range` and `Position` got complicated by introducing `sticky` flag. It would be a nice idea to try and move the "sticky"-related stuff directly to OT algorithms. I don't know if it is doable and how much refactor will it take, but right now `core.treeModel.Position#getTransformedByMove` takes 5 parameters which is not ideal.
4. Right now we have (or should have) 100% code coverage but it does not mean that all transformation cases have been covered. It would be good to write more tests that do not test special case scenarios but "generic" / "default" transformations where deltas conflicts should be resolved on operational transformation level. I actually found three annoying bugs when writing tests for such "default" transformations and there might be more to be discovered.


src/treemodel/delta/attributedelta.js
```diff
@@ -19,7 +19,58 @@ import Element from '../element.js';
  * @memberOf core.treeModel.delta
  * @extends core.treeModel.delta.Delta
  */
-export default class AttributeDelta extends Delta {}
+export default class AttributeDelta extends Delta {
+	/**
+	 * @see core.treeModel.delta.Delta#_reverseDeltaClass
```
Reinmar: You can use `@inheritdoc`.

src/treemodel/delta/transform.js
```diff
@@ -0,0 +1,490 @@
...
+ */
+export const priorities = new Map();
+priorities.set( MergeDelta, 0 );
+priorities.set( UnwrapDelta, 1 );
+priorities.set( WrapDelta, 1 );
+priorities.set( SplitDelta, 1 );
+priorities.set( AttributeDelta, 2 );
+priorities.set( InsertDelta, 2 );
+priorities.set( MoveDelta, 2 );
+priorities.set( RemoveDelta, 3 );
```
pjasiun: Maybe there priorities should be: 10, 20, 30, in case someone want to put his delta between existing.

src/treemodel/delta/transform.js
```diff
@@ -0,0 +1,490 @@
+/**
```
pjasiun: I would split this file into 2: one which define the transformation API and one which adds all special cases. Also priorities might be moved to deltas as their properties.

src/treemodel/delta/transform.js
Reinmar: No one said that only integers are accepted :P.

src/treemodel/position.js
```diff
@@ -279,20 +280,22 @@ export default class Position {
 	 * @param {Boolean} insertBefore Flag indicating whether moved nodes are pasted before or after `insertPosition`.
 	 * This is important only when `targetPosition` and this position are same. If that is the case and the flag is
 	 * set to `true`, this position will get transformed by range insertion. If the flag is set to `false`, it won't.
+	 * @param {Boolean} [sticky] Flag indicating whether this position "sticks" to range, that is if it should be moved
+	 * with the moved range if it is equal to one of range's boundaries.
 	 * @returns {core.treeModel.Position} Transformed position.
 	 */
-	getTransformedByMove( sourcePosition, targetPosition, howMany, insertBefore ) {
+	getTransformedByMove( sourcePosition, targetPosition, howMany, insertBefore, sticky ) {
```
pjasiun: I am not sure if these methods should not be protected. No one will be able to use them properly, anyway, especially with new parameters. What end users should use is `LivePosition`. It would be useful to put here `@see` reference to the live position.

tests/treemodel/delta/transform/mergedelta.js
```diff
@@ -0,0 +1,194 @@
...
+
+	beforeEach( () => {
+		doc = getFilledDocument();
+		root = doc.getRoot( 'root' );
+		gy = doc.graveyard;
+		baseVersion = doc.version;
+	} );
+
+	describe( 'MergeDelta by', () => {
+		let mergeDelta, mergePosition;
```
pjasiun: I miss merge by merge test for 2 merges in one paragraph.

pjasiun:
> 1. Tests and documentation for tests utils tools.
> 2. Way to automate delta creation (some kind of "default" constructor). Right now delta has to be build from scratch everytime because delta constructor does not create any operations. After creating delta instance (i.e. let wrapDelta = new WrapDelta()) developer has to add all operations by hand. This won't be used by outside developers but deltas are created in a few places in code (i.e. tests, methods registered to Batch, delta transformations) and it is a bit painful that one has to create all the operations by hand. Something like let wrapDelta = new WrapDelta( wrapRange, wrapElement ); would be great and is doable.
> 3. Some methods in Range and Position got complicated by introducing sticky flag. It would be a nice idea to try and move the "sticky"-related stuff directly to OT algorithms. I don't know if it is doable and how much refactor will it take, but right now core.treeModel.Position#getTransformedByMove takes 5 parameters which is not ideal.
> 4. Right now we have (or should have) 100% code coverage but it does not mean that all transformation cases have been covered. It would be good to write more tests that do not test special case scenarios but "generic" / "default" transformations where deltas conflicts should be resolved on operational transformation level. I actually found three annoying bugs when writing tests for such "default" transformations and there might be more to be discovered.

I agree with all 4 points, and that they should be moved to separate tickets.

Also I afraid that we tested mostly single Delta X single Delta transformation. What I am not sure about is that the whole concept of operations and deltas is that the result of default transformation of deltas may not behave like the original delta anymore, what may means that special conversions stop working. It means that all cases should be handled then as special cases or none of them. In the worst case the whole idea of having two lavers (operations and deltas) will not have sense anymore and we will have to rewrite deltas as addition operations with transformations for all cases.

It would be cool to have a manual test to create two sets of deltas and check if the result of merging is fine, but it would be much easier to do it based on the editor with working selection, so may be a future task.

scofalik:
I created #236 #237 #238 #239 as follow-ups.

scofalik:
Additonally, this PR closes #184 and #183

scofalik:
I pushed fixes and additional test, as was requested.

scofalik:
I rebased the branch with master.

pjasiun:
I pushed additional commit, which should be reviewed, but not to play with this PR anymore I have merged it.

@scofalik please review these changes, and good job :)

scofalik:
+1 for the changes, thanks for managing it.

Issue #186, opened 2016-02-05
Verify destruction chain


Reinmar:
Once we have some stable implementation of first basic editors we should verify that each component has the `destroy()` method and that each `Emitter` or `Observable` calls `stopListening()`.

Reinmar:
One thing which is definitely missing is `Plugin.destroy()` method – it should call `stopListening()` and editor should destroy all plugins before destroying itself.

Reinmar:
And the second thing is that the editor should destroy all plugins.

Reinmar:
Other things which needs to be destroyed are commands – `stopListening()` should be called on them. That could be done by the feature, but since we add commands to the `editor.commands` it will be much easier if the editor does it. We may also think about implementing commands repo, so it's its obligation to detach commands.

Reinmar:
Another case: #262.

Reinmar:
Moved to https://github.com/ckeditor/ckeditor5/issues/114

PR #234, opened 2016-03-07
Features base: Commands and Schema

scofalik:
This PR contains changes that introduces `Command` abstract class and first command: `AttributeCommand`. It also brings in first working version of `Schema`. `Feature` class is also introduced, but it is very very simple right now.

Please take a look at `editor.js`. I've put there some integration stuff which was needed to implement and test commands. You may have better ideas where to put this stuff.

The one thing I am not sure about is storing selection attributes. We agreed that this will be done through prefixing but where exactly put this code is a tough choice. Keep in mind that this has to be done through `Batch` API because those changes has to be acknowledged by undo and OT. The idea I implemented is that attribute storing is responsibility of a command.

The other ideas were to make it a more automatic -- i.e. whenever selection changes and was collapsed in empty node, store attributes and create a batch. We don't even need `Schema` for this as we would store only attributes from selection and save them prefixed. But this would be creating a batch "behind the scenes" and I am not sure if this won't create some problems. If you think this is a better idea, let me know and I will change it.


src/command/attributecommand.js
```diff
@@ -0,0 +1,201 @@
...
+	 * * if selection is collapsed in empty node, the command applies attribute to the parent node of selection (note
+	 * that selection inherits all attributes from a node if it is in empty node).
+	 *
+	 * If the command is disabled (`isEnabled == false`) when it is executed, nothing will happen.
+	 *
+	 * @private
+	 * @param {Boolean} [forceValue] If set it will force command behavior. If `true`, command will apply attribute,
+	 * otherwise command will remove attribute. If not set, command will look for it's current value to decide what it should do.
+	 */
+	_execute( forceValue ) {
```
Reinmar: It was meant to be public `execute()` and protected `_doExecute()`.

src/emittermixin.js
```diff
@@ -225,6 +225,7 @@ const EmitterMixin = {
 	 * @param {String} event The name of the event.
 	 * @param {...*} [args] Additional arguments to be passed to the callbacks.
 	 * @method core.EmitterMixin#fire
+	 * @returns {*} Value returned by the last callback processed in the event loop.
```
Reinmar: Let's not do this change in this PR. Please use the existing emitter API. You can implement the `checkEnabled` event in this way:

```js
const data = { isEnabled: true };
fire( 'checkEnabled', data );

if ( data.isEnabled ) ....
```

src/feature.js
```diff
@@ -0,0 +1,36 @@
...
+ * The main responsibilities for Feature are:
+ * * setting required dependencies (see {@link core.Plugin#requires},
+ * * configuring, instantiating and registering commands to editor,
+ * * registering converters to editor (if the feature operates on Tree Model),
+ * * setting and registering UI components (if the feature uses it).
+ *
+ * @memberOf core
+ */
+export default class Feature extends Plugin {
+	/**
```
Reinmar: For now the class can be empty as neither the constructor nor `init()` needs to be defined.

src/command/attributecommand.js
```diff
@@ -0,0 +1,201 @@
...
+		 * @member {String} core.command.AttributeCommand#attributeKey
+		 */
+		this.attributeKey = attributeKey;
+	}
+
+	/**
+	 * Flag indicating whether command is active. For collapsed selection it means that typed characters will have
+	 * the attribute set. For range selection it means that all nodes inside have the attribute applied.
+	 *
+	 * @returns {Boolean}
```
Reinmar: Shouldn't it be documented as a property?

src/command/attributecommand.js
```diff
@@ -0,0 +1,201 @@
...
+ * attribute on a text or element with value `true`. AttributeCommand uses {@link core.treeModel.Document#selection} to
+ * decide which nodes (if any) should be changed, and applies or removes attributes from them.
+ * See {@link core.treeView.Converter#execute} for more.
+ *
+ * The command checks {@link core.treeModel.Document#schema} to decide if it should be enabled.
+ * See {@link core.treeView.Converter#checkSchema} for more.
+ *
+ * @memberOf core.command
+ */
+export default class AttributeCommand extends Command {
```
Reinmar: I wonder which pieces of this command could be useful outside it. Right now we assume that to operate on attributes you need this command, but what if your functionality needs to perform similar action but without registering a command?

I see that either we can expose some of this command's functionality into some `AttrbiuteApplier` or we can tell – you want a piece of this command, then create it (what can be done without registering).

src/command/command.js
```diff
@@ -0,0 +1,135 @@
...
+			} );
+		}
+	}
+
+	/**
+	 * Checks if a command should be enabled according to its own rules. Mostly it will check schema to see if the command
+	 * is allowed to be executed in given position. This method can be defined in child class (but is not obligatory).
+	 * If it is defined, it will be added as a callback to `refreshState` event.
+	 *
+	 * @method core.command.Command#checkEnabled
```
Reinmar: Sounds like a protected method, rather than a public one. The public interface is made of a `refreshState()` method and `isEnabled` property.

src/command/attributecommand.js
Reinmar: It may be a bit too early to understand what's the best way to expose this logic, so we can postpone this decision and keep the current implementation.

src/treemodel/selection.js
```diff
@@ -5,11 +5,15 @@
...
 'use strict';
 
+import Position from './position.js';
+import Range from './range.js';
 import LiveRange from './liverange.js';
 import EmitterMixin from '../emittermixin.js';
 import CKEditorError from '../ckeditorerror.js';
 import utils from '../utils.js';
 
+const storePrefix = 'selection_store:';
```
Reinmar: Do we have a naming convention for these? I would say that it should be `selectionStore:`. Or even better – `selection:`.

src/treemodel/selection.js
```diff
@@ -83,11 +86,26 @@ export default class Selection {
 	}
 
 	/**
+	 * Flag indicating whether the selection has any range in it.
+	 *
+	 * @readonly
+	 * @type {Boolean}
+	 */
+	get hasAnyRange() {
```
Reinmar: Maybe `isEmpty()`?

src/treemodel/selection.js
```diff
@@ -215,51 +290,69 @@ export default class Selection {
...
+	 * if given range is intersecting with any range that is already stored in this selection.
 	 *
-	 * @param {String} key Key of attribute to remove.
-	 * @returns {Boolean} `true` if the attribute was set on the selection, `false` otherwise.
+	 * @private
+	 * @param {core.treeModel.Range} range Range to add.
 	 */
-	removeAttribute( key ) {
-		return this._attrs.delete( key );
+	_pushRange( range ) {
```
Reinmar: This method is odd. It's very confusing what is it supposed to do.

src/treemodel/selection.js
```diff
@@ -135,8 +156,44 @@ export default class Selection {
 	}
 
 	/**
+	 * Returns the first range in the selection. First range is the one which {@link core.treeModel.Range#start start} position
```
Reinmar: Does it mean that ranges aren't sorted upon adding them? It sounds like it could be better to:

* sort them when adding,
* require that they are added in the right order.

Or is it caused by the fact that live ranges are used, so they may change their order after being added to the selection?

src/treemodel/selection.js
```diff
@@ -83,11 +86,26 @@ export default class Selection {
...
+
+	/**
 	 * Returns whether the selection is collapsed. Selection is collapsed when all it's ranges are collapsed.
+	 * If selection has no ranges, returns null instead.
 	 *
 	 * @type {Boolean}
 	 */
 	get isCollapsed() {
+		if ( !this.hasAnyRange ) {
+			return null;
```
Reinmar: It should not be null according to the docs.

I would return `true` if the selection is empty. This is what the native selection does in Firefox and it was my first choice as well.

src/command/attributecommand.js
```diff
@@ -0,0 +1,201 @@
...
+		const value = ( forceValue === undefined ) ? !this.value : forceValue;
+
+		if ( selection.isCollapsed ) {
+			let selectionParent = selection.getFirstPosition().parent;
+
+			if ( selectionParent.getChildCount() === 0 ) {
+				// If selection is collapsed and in empty node, operate on stored selection attributes.
+				const storeKey = Selection.getStoreAttributeKey( this.attributeKey );
+
+				document.enqueueChanges( () => {
```
pjasiun: My initial idea was to have `_doExecute` already inside the `enqueueChanges` block, but then we may have an issue if one command wants to execute another command. Or `changesDone` event will be fired even if there were no real changes in the document (on the other hand `changesDone` event should not harm).

Reinmar: I wouldn't go this way. If we require using `enqueueChanges()` then let's make it clear that you should call it before you do something. Otherwise we risk creating an unclear situation in which the developers sometimes will have to call it and sometimes it will be called automatically.

PS. I'd even consider throwing if you're trying to make some changes when not inside `enqueueChanges()`.

Reinmar:
I wanted to merge master to t/features, but it's a mess. @scofalik, please resolve these conflicts (without rebasing the branch so we won't lose the existing comments).

scofalik:
Why did you want to merge `master` to `t/features`?

src/treemodel/selection.js
scofalik: I think that `.hasAnyRange` is more precise. If I've seen `isEmpty` I'd think to myself whether this mean "has no ranges" or "has no nodes / collapsed". I've changed it to `hasAnyRanges` (added "s").

src/treemodel/selection.js
scofalik: 1. Some time ago we discussed whether ranges should be sorted upon adding. If yes, we would have to change `focus` and `anchor` implementation. Live ranges are also an issue. Most of time there will be one range in selection anyway so I don't think that sorting on-the-fly will be that difficult. What we could do is sort upon adding and sort ranges whenever it's live ranges changes but I suspect that this might happen more often or as often as using `.getFirstRange()`. The other idea is to introduce some kind `._dirty` flag that will be checked whenveer `.getFirstRange()` is fired and if selection is not dirty, there won't be any sorting. But this doesn't seem good to me either.

2. We cannot require that ranges are added in the right order. Multiple ranges are for multiple selection and multiple selection might be done in any order.

src/treemodel/selection.js
scofalik: We decided that `Selection` will always have at least one range.

src/treemodel/selection.js
scofalik: We decided that `Selection` will always have at least one range.

src/treemodel/selection.js
Reinmar: It seems that the current approach is the simplest, so I'm ok with it.

scofalik:
> It was meant to be public execute() and protected _doExecute().

Fixed.

> Let's not do this change in this PR. Please use the existing emitter API.

Reverted changes and implemented the mechanism in suggested way.

> For now the class can be empty as neither the constructor nor init() needs to be defined.

Made `Feature` an empty class.

> Shouldn't it be documented as a property?

Changed this and other places where documentation for getters was wrong.

> Sounds like a protected method, rather than a public one. The public interface is made of a refreshState() method and isEnabled property

Honestly it sounds like a private method. I've changed it to private.

> Maybe isEmpty()?

Now `Selection` is never empty.

> Does it mean that ranges aren't sorted upon adding them?

Left as it was.

> [isCollapsed] should not be null according to the docs.

The issue is solved by making `Selection never empty. It will return `false` or `true`.

src/treemodel/document.js
```diff
@@ -128,38 +148,48 @@ export default class Document {
...
 
 		return root;
 	}
 
 	/**
-	 * Enqueue a callback with document changes. Any changes to be done on document (mostly using {@link core.treeModel.Document#batch} should
-	 * be placed in the queued callback. If no other plugin is changing document at the moment, the callback will be
+	 * Removes all events listeners set by document instance.
+	 */
+	detach() {
```
Reinmar: `s/detach/destroy`. You don't attach & detach the document to anything. Plus, we use `destroy()` in other places in exactly this meaning.

src/command/command.js
```diff
@@ -0,0 +1,135 @@
...
+		 */
+		this.editor = editor;
+
+		/**
+		 * Flag indicating whether a command is enabled or disabled.
+		 * A disabled command should do nothing upon it's execution.
+		 *
+		 * @member {Boolean} core.command.Command#isEnabled
+		 */
+		this.isEnabled = true;
```
Reinmar: `Command#isEnabled` and `Command#value` must be observable. Otherwise, there will be no easy way to create e.g. a button's model based on an existing command.

So:

1. `Command` must implement `utils.Observable`.
2. `Command#value` must not be a getter, but a property which is automatically updated once the value is changed. In case of the attribute command that can be implemented as a selection's observer. On selection changes, check the value of the command.

src/command/command.js
```diff
@@ -0,0 +1,135 @@
...
+			} );
+		}
+	}
+
+	/**
+	 * Checks if a command should be enabled according to its own rules. Mostly it will check schema to see if the command
+	 * is allowed to be executed in given position. This method can be defined in child class (but is not obligatory).
+	 * If it is defined, it will be added as a callback to `refreshState` event.
+	 *
+	 * @private
```
Reinmar: All the things which you documented as `@private` should be `@protected`, otherwise specific commands cannot use them.

src/command/command.js
```diff
@@ -0,0 +1,135 @@
...
+	 * @method core.command.Command#_checkEnabled
+	 * @returns {Boolean} `true` if command should be enabled according to {@link core.treeModel.Document#schema}. `false` otherwise.
+	 */
+
+	/**
+	 * Fires `refreshState` event and checks it's resolve value to decide whether command should be enabled or not.
+	 * Other parts of code might listen to `refreshState` event on this command and add their callbacks. This
+	 * way the responsibility of deciding whether a command should be enabled is shared.
+	 *
+	 * @fires {@link core.command.Command.refreshState refreshState}
```
Reinmar: Should be `#refreshState`.

src/command/command.js
```diff
@@ -0,0 +1,135 @@
...
+		const data = { isEnabled: true };
+		this.fire( 'refreshState', data );
+
+		this.isEnabled = data.isEnabled;
+	}
+
+	/**
+	 * Executes the command if it is enabled.
+	 *
+	 * @protected
```
Reinmar: I wanted to say that it should be public (because the name does not start with an underscore), but as I think about it more I see that it could stay protected. It's only used by the `editor.execute()` method, so the same package. In this case you will need to add the underscore.

src/treemodel/selection.js
```diff
@@ -20,7 +28,7 @@ export default class Selection {
 	/**
 	 * Creates an empty selection.
```
Reinmar: Missing `@param`.

src/treemodel/selection.js
```diff
@@ -131,12 +141,44 @@ export default class Selection {
...
+	 * @returns {core.treeModel.Position}
+	 */
+	getFirstPosition() {
+		return Position.createFromPosition( this.getFirstRange().start );
 	}
 
 	/**
 	 * Removes all ranges that were added to the selection. Fires update event.
 	 *
+	 * @fires {@link core.treeModel.Selection.update update}
```
Reinmar: `#update`

src/treemodel/selection.js
```diff
@@ -212,54 +274,230 @@ export default class Selection {
...
+
+					batch.setAttr( storeKey, attr[ 1 ], selectionParent );
+				}
+			} );
 		}
 	}
 
-	this._ranges.push( LiveRange.createFromRange( range ) );
+	/**
+	 * Updates this selection attributes basing on it's position in the Tree Model.
```
Reinmar: "based on its position"

src/treemodel/selection.js
```diff
@@ -212,54 +274,230 @@ export default class Selection {
...
+	 *
+	 * @private
```
Reinmar: This method cannot be private, because the document won't be able to call it.

src/treemodel/document.js
```diff
@@ -128,38 +148,48 @@ export default class Document {
...
+	/**
+	 * Enqueue a callback with document changes. Any changes to be done on document (mostly using {@link core.treeModel.Document#batch}
+	 * should be placed in the queued callback. If no other plugin is changing document at the moment, the callback will be
 	 * called immediately. Otherwise it will wait for all previously queued changes to finish happening. This way
 	 * queued callback will not interrupt other callbacks.
 	 *
-	 * When all queued changes are done {@link core.treeModel.Document#changesDone} event is fired.
+	 * When all queued changes are done {@link core.treeModel.Document.changesDone} event is fired.
 	 *
+	 * @fires {@link core.treeModel.Document.changesDone}
```
Reinmar: `#changesDone`

src/treemodel/document.js
```diff
@@ -128,38 +148,48 @@ export default class Document {
 	/**
 	 * Creates a new top-level root.
 	 *
+	 * @param {String|Symbol} id Unique root id.
 	 * @param {String|Symbol} name Unique root name.
```
pjasiun: It does not make sense.

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+'use strict';
+
+import clone from '../lib/lodash/clone.js';
+import CKEditorError from '../ckeditorerror.js';
+
+/**
+ * SchemaItem is a singular registry item in {@link core.treeModel.Schema} that groups and holds allow/disallow rules for
+ * one entity. This class is used internally in {@link core.treeModel.Schema} and should not be used outside it.
+ *
+ * @see core.treeModel.Schema
```
Reinmar: Maybe let's make it `@protected`?

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+		this._addPath( '_disallowed', path, attribute );
+	}
+
+	/**
+	 * Adds path to the SchemaItem instance.
+	 *
+	 * @param {String} member Name of the array member into which the path will be added. Possible values are `_allowed` or `_disallowed`.
+	 * @param {Array.<String>|String} path Path to be added. String with item names separated by spaces may be passed.
+	 * @param {String} [attribute] If set, this path will be used only for entities that have an attribute with this key.
+	 * @private
```
Reinmar: `@private` goes before all other tags.

Reinmar: Rule of thumb: document stuff in the order you would naturally read it. There's a private method which accepts the following params and returns foo.

src/treemodel/document.js
```diff
@@ -68,12 +74,26 @@ export default class Document {
...
 		 * @readonly
-		 * @member {core.treeModel.Selection} core.treeModel.Document#selection
+		 * @protected
+		 * @member {Map} core.treeModel.Document#roots
 		 */
-		this.selection = new Selection();
+		this._roots = new Map();
+
+		// Add events that will update selection attributes.
+		this.selection.on( 'update', () => {
```
Reinmar: I'm not sure on what events the selection should update its attributes itself. This should not conflict with the situation when the feature sets the selection and its attributes (e.g. bold command when executed will add attributes to the selection and that will, if I understand the flow correctly, trigger `#changesDone`).

src/treemodel/selection.js
```diff
@@ -99,17 +106,20 @@ export default class Selection {
...
 	/**
 	 * Adds a range to the selection. Added range is copied and converted to {@link core.treeModel.LiveRange}. This means
-	 * that passed range is not saved in the Selection instance and you can safely operate on it. Accepts a flag
-	 * describing in which way the selection is made - passed range might be selected from {@link core.treeModel.Range#start}
-	 * to {@link core.treeModel.Range#end} or from {@link core.treeModel.Range#start} to {@link core.treeModel.Range#end}. The flag
-	 * is used to set {@link #anchor} and {@link #focus} properties.
+	 * that passed range is not saved in the Selection instance and you can safely operate on it.
 	 *
+	 * Accepts a flag describing in which way the selection is made - passed range might be selected from
+	 * {@link core.treeModel.Range#start} to {@link core.treeModel.Range#end} or from {@link core.treeModel.Range#start}
```
Reinmar: This sentence makes no sense. I guess the latter selection should be from `#end` to `#start`.

src/treemodel/selection.js
```diff
@@ -131,12 +141,44 @@ export default class Selection {
...
+
+		for ( let i = 0; i < this._ranges.length; i++ ) {
+			let range = this._ranges[ i ];
+
+			if ( !first || range.start.isBefore( first.start ) ) {
+				first = range;
+			}
+		}
+
+		return first && Range.createFromRange( first ) || this._getDefaultRange();
```
Reinmar: Please use `?:` :P. If anytime in the future `Range.createFromRange()` would return a falsy value there would be a very nasty bug here. I've really seen those.

src/treemodel/selection.js
```diff
@@ -212,54 +274,230 @@ export default class Selection {
...
+	 * the parent node is empty.
+	 *
+	 * @private
+	 * @param {String} key Key of attribute to set.
+	 * @param {*} value Attribute value.
 	 */
-	clearAttributes() {
-		this._attrs.clear();
+	_storeAttribute( key, value ) {
+		const selectionParent = this.getFirstPosition().parent;
```
Reinmar: How do you know that first position's parent is a block?

Reinmar: Actually... it does not need to be a block, right? It can be some inline widget's container or something. Is it all taken into consideration now?

src/treemodel/selection.js
```diff
@@ -212,54 +274,230 @@ export default class Selection {
...
+
+		this._ranges.push( LiveRange.createFromRange( range ) );
+	}
+
+	/**
+	 * Iterates through all attributes stored in current selection's parent.
 	 *
+	 * @returns {Iterable.<*>}
+	 */
+	*_getStoredAttributes() {
```
Reinmar: It's been a bit confusing for me that in one place you have `set/get/removeAttribute()` and next to it you have `_storeAttribute()` and the rest. We need to find a more distinguishable name for the "hoist attributes to the block" functionality.

Some ideas:

* `hoistAttribute()`, `removeHoistedAttribute()`,
* `setBlockAttribute()`, `removeBlockAttribute()`,
* `preserveAttribute()`

src/treemodel/selection.js
```diff
@@ -212,54 +274,230 @@ export default class Selection {
...
+	 * Stores given attribute key and value in current selection's parent node if the selection is collapsed and
+	 * the parent node is empty.
+	 *
+	 * @private
+	 * @param {String} key Key of attribute to set.
+	 * @param {*} value Attribute value.
 	 */
-	clearAttributes() {
-		this._attrs.clear();
+	_storeAttribute( key, value ) {
```
Reinmar: Since we store those attributes in order to preserve selection attributes, doesn't it mean that the value could be taken from the selection's attributes? And that perhaps all attribute keys could be taken from it?

src/command/attributecommand.js
Reinmar: I'm looking at this class again and I feel that it will be crucial to make it (or related commands/classes) flexible and extensible. For instance, this command will need to share some logic with command which is able to set attribute with a specified value (e.g. the link command would need such a command). Then, we have base link feature extensions, like for instance link styles. I don't know yet how those extensions will be defined, either as separate commands or some logic pluggable into existing commands. Tough stuff to predict, so let's just keep this in our minds.

src/command/command.js
```diff
@@ -0,0 +1,135 @@
...
+		this.refreshState();
+	}
+
+	/**
+	 * Executes command.
+	 * This is an abstract method that should be overwritten in child classes.
+	 *
+	 * @private
+	 */
+	_doExecute() {
```
Reinmar: I don't like the new line after `{` :P.

src/editor.js
```diff
@@ -42,7 +43,7 @@ export default class Editor {
 		/**
 		 * Holds all configurations specific to this editor instance.
 		 *
-		 * This instance of the {@link Config} class is customized so its {@link Config#get} method will retrieve
+		 * This instance of the {@link core.Config} class is customized so its {@link core.Config#get} method will retrieve
```
Reinmar: `s/core/utils`

src/editor.js
```diff
@@ -60,6 +61,20 @@ export default class Editor {
 		this.plugins = new PluginCollection( this );
 
 		/**
+		 * Tree Model document managed by this editor.
+		 *
+		 * @member {core.treeModel.Document} core.Editor#document
```
Reinmar: + `@readonly`

src/editor.js
```diff
@@ -60,6 +61,20 @@ export default class Editor {
...
+		 */
+		this.document = new Document();
+
+		/**
+		 * Commands registered to the editor.
+		 *
+		 * @member {Map} core.Editor#commands
```
Reinmar: + `@readonly`

src/treemodel/document.js
```diff
@@ -128,38 +148,48 @@ export default class Document {
...
-	 * be placed in the queued callback. If no other plugin is changing document at the moment, the callback will be
+	 * Removes all events listeners set by document instance.
+	 */
+	detach() {
+		this.selection.detach();
+		this.stopListening();
+	}
+
+	/**
+	 * Enqueue a callback with document changes. Any changes to be done on document (mostly using {@link core.treeModel.Document#batch}
```
Reinmar: KISS: "Enqueues document changes."

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+	 * Creates SchemaItem instance.
+	 *
+	 * @param {core.treeModel.Schema} schema Schema instance that owns this item.
+	 */
+	constructor( schema ) {
+		/**
+		 * Schema instance that owns this item.
+		 *
+		 * @member {core.treeModel.Schema} core.treeModel.SchemaItem#_schema
+		 * @private
```
Reinmar: `@private @member`

src/treemodel/selection.js
```diff
@@ -212,54 +274,230 @@ export default class Selection {
...
+	}
 }
 
 utils.mix( Selection, EmitterMixin );
+
+/**
+ * Fired whenever selection ranges are changed through {@link core.treeModel.Selection Selection API}. Not fired when
+ * {@link core.treeModel.LiveRange live ranges} inserted in selection change because of Tree Model changes.
+ *
+ * @event core.treeModel.Selection.update
```
Reinmar: `Selection#update`

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+		this._disallowed = [];
+	}
+
+	/**
+	 * Allows entity, represented by this item, to be in given path.
+	 *
+	 * @param {Array.<String>|String} path Path in which entity is allowed. String with item names separated by spaces may be passed.
+	 * @param {String} [attribute] If set, this path will be used only for entities that have an attribute with this key.
+	 */
+	addAllowed( path, attribute ) {
```
Reinmar: Couldn't we simply name these `allow()` and `disallow()`?

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+	 * @param {Array.<String>|String} path Path to be added. String with item names separated by spaces may be passed.
+	 * @param {String} [attribute] If set, this path will be used only for entities that have an attribute with this key.
+	 * @private
+	 */
+	_addPath( member, path, attribute ) {
+		if ( typeof path === 'string' ) {
+			path = path.split( ' ' );
+		}
+
+		path = path.slice();
```
Reinmar: Why cloning it if we've just created it? And why cloning anyway (comment missing perhaps)?

Reinmar: I see we also clone it in `_getPaths()`. I sense problems in the future here – it should be clarified when the stuff needs to be cloned and secured that it is if there's some special risk.

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+	 * @param {String} type Paths' type. Possible values are `ALLOW` or `DISALLOW`.
+	 * @param {Array.<String>} checkPath Path to check.
+	 * @param {String} [attribute] If set, only paths registered for given attribute will be returned.
+	 * @returns {Boolean} `true` if item has any registered matching path, `false` otherwise.
+	 * @private
+	 */
+	_hasMatchingPath( type, checkPath, attribute ) {
+		const itemPaths = this._getPaths( type, attribute );
+
+		checkPath = checkPath.slice();
```
Reinmar: AFAICS, it's unnecessary.

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+ * is in correct state.
+ *
+ * Schema consist of {@link core.treeModel.SchemaItem schema items}, each describing different entity. Entity can be
+ * either a Tree Model element or an abstract group for similar elements. Entities are represented by names. Names
+ * of special/abstract entities should be prefixed by `$` sign.
+ *
+ * Each entity in Schema may have a set of allow/disallow rules. Every rule describes in which entities given entity
+ * can or cannot be.
+ *
+ * Entities can extend other entities. This mechanism allows for grouping entities under abstract names. Whenever a rule
```
Reinmar: This paragraph made me think that perhaps extensions could be implemented JS-way.  If one item extends another item, make it a child class of the parent item. The theoretical `getAllowed()` method would be implemented like this:

```
getAllowed() {
    const allowedByParent = super.getAllowed();

    return allowedByParent.concat( this._allowed );
}
```

So the logic of extending one item by another item would be implemented totally inside `SchemaItem`. Perhaps this would clarify the implementation.

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+
+		// Due to circular references we need to remove parent reference.
+		json._schema = '[treeModel.Schema]';
+
+		return json;
+	}
+}
+
+/**
+ * Schema is a run-time created and modified description of which entity in Tree Model is allowed to be inside another
```
Reinmar: This class needs a more high-level introduction to be understandable. Something like:

Schema is a definition of the structure of the document. It allows to define which tree model items (element, text, etc.) can or be nested within which and which attributes can be applied to them. It's created during the run-time of the application, typically by features. Also, the features can query the schema to learn what's the allowed structure and act accordingly.

For instance, if a feature wants to define that an attribute `bold` is allowed on the text it needs to register this rule like this:

```
editor.document.schema.allow( '$text', 'bold' );
```

Note: items prefixed with `$` are special group of items. `$text` means text nodes and text-like content. `$block` means block elements.

When registering an item it's possible to tell that this item should inherit from some other existing item. E.g. `p` can inherit from `$block`, so when an attribute `center` is allowed on the `$block` it will automatically be also allowed on the `p` element.

etc. etc. Try to write this from the user perspective. Don't document the implementation, but the feature.

src/treemodel/schema.js
```diff
@@ -0,0 +1,417 @@
...
+		/**
+		 * Description of what entities are a base for given entity.
+		 *
+		 * @member {Map} core.treeModel.Schema#_extensionChains
+		 * @private
+		 */
+		this._extensionChains = new Map();
+
+		// Register some default abstract entities.
+		this.registerItem( '$inline', null );
```
Reinmar: Why requiring passing `null`?

Reinmar:
OK... I've made a very general review of the schema and by reading the implementation I can only say that it's unclear what's going on there :D. However, the API seems to be fine and with some improvements to the docs (some high-level intro and more examples) it should be understandable.

As for its completeness I've got no opinion at this point. Let's see it in practice. If there will be many WTFs or missing (and hard to implement) features we'll need to review what we have. But that would be a waste of time to do this now.

Finally, I think that some more OO approach could be used to implement relations between schema items. I think that it could help developers understand all this and perhaps would be more flexible in the future. But that sounds like a topic for a new issue.

src/command/command.js
scofalik: Changed to `_execute`.

src/treemodel/selection.js
scofalik: Added `@param`

src/treemodel/schema.js
scofalik: Added `@protected`.

src/treemodel/selection.js
scofalik: From Tree Model point of view I belive that it does not make any difference. If selection position can be put into something it means that almost certainly characters can be put there too (unless schema forbids). So I think it is fine to save selection attributes in that container and not make any additional logic for this.

In Tree Model we have `Element`s ("containers") and characters and things that are inline in TreeView/DOM are attributes on characters so unless someone creates converter that will convert `Element` to inline stuff, you will be fine. But even then I think it is correct to save attributes in selections first parent.

src/treemodel/selection.js
scofalik: 1) Well, we could get attribute value for selection just basing on it's key, but it is an additional call and we have that value anyway.

2) We could have just one method which would remove all stored attributes and set all current attributes. But that would mean additional reads of keys / values and additional batches / operations. I don't think it's worth it.

src/treemodel/document.js
scofalik: I don't know what part of the comment I should change. I think it is clear and simple now :).

src/treemodel/schema.js
scofalik: This was one of the approaches I've been thinking of after implementing current solution. It is true that it might turn out in more understandable `Schema` internals, but it would be a bit cumbersome for end-users. For each new element you introduce, you will have to create a new class + some would have to be already introduced (the default `$inline` and `$block` items). OTOH, most features will already have multiple files in their packages (feature class, command classes, ui integration, etc.) so one more probably would not make a harm.

So I think that current solution is easy as all you need to do is to write one line of code somewhere inside feature class. It is in the same vein as tree model `Element`s where they are distinguished only by their name.

For sure I will look forward to enhance `Schema`. There are a lot of ideas and I already described some here https://github.com/ckeditor/ckeditor5-design/issues/129#issuecomment-188137346

A wild thought: we could do the extension thing also with Tree Model `Element`s... But right now they are nice and simple. Maybe (extended) `Element` could be coupled with `Schema` by `SchemaItem` (or even without it, but I don't want to put everything in one place). This is just a wild thought though, nothing to discuss now, really.

scofalik:
OK, I pushed suggested changes and replied in few places where I wanted to discuss some things. Thanks for through and careful review.

src/treemodel/schema.js
scofalik: I will investigate this.

scofalik: Okay. So the rule I go by is that whenever a class method takes an array as a parameter and stores it in class member we should consider cloning the array instead saving the reference to the passed array. Of course there are some cases where you would want to store the reference, but in most cases its best to avoid the situation where class member can be edited without accessing a class at all. It opens up possibilities for mistakes.
```
let myArr = [ 1, 2, 3 ];
let myObj = newMyObj( myArr );
// ... a lot of code
myArr.push( 4 ); // do you really want this to silently impact myObj?
```

Of course when it comes to adding paths to `SchemaItem`s those arrays will mostly be anonymous, but I thought it will be better to be safe than sorry.

scofalik: As far as `_getPaths` is concerned -- we use that function in `_hasMatchingPath` which shifts `itemPath` array. If we won't copy the array in `_getPaths` we will be shifting the original array that is stored as class member. We need a copy of that array because we use it for temporary processing.

src/treemodel/document.js
Reinmar: It's details, but it should be "Enqueues". And the fact that it's a callback is a detail which isn't needed in the very first sentence of this method documentation.

Reinmar:
> This was one of the approaches I've been thinking of after implementing current solution. It is true that it might turn out in more understandable Schema internals, but it would be a bit cumbersome for end-users. For each new element you introduce, you will have to create a new class + some would have to be already introduced (the default $inline and $block items). OTOH, most features will already have multiple files in their packages (feature class, command classes, ui integration, etc.) so one more probably would not make a harm.

I think that we could keep the current API. The extension would happen under the hood. One class would be enough – you would just configure its instances differently for each item.

scofalik:
Oh, so you don't really mean JS-y extensions. This is another approach I've been thinking of and this is one of the enhancements I want to do in next `Schema` iteration.

src/treemodel/schema.js
Reinmar: In general I agree, you accept or return an array, then copy it if it's also stored inside. However, we must think about performance and memory. This code will be executed very frequently, so we may need to rethink this algorithm.

Reinmar:
> Oh, so you don't really mean JS-y extensions. This is another approach I've been thinking of and this is one of the enhancements I want to do in next Schema iteration.

It could be a mix of both. You would have a simple method for creating items like you have today, but you could also create them by inheriting from something. Or something like that :D.

Reinmar:
I've commented earlier that `document.detach()` should be renamed to `document.destroy()` and I even wanted to comment that `liverange.detach()` makes sense :D I see you changed all to `destroy()` in 514f50ad. I'll revert those changes in LiveRange, LivePosition and perhaps in Selection as well.

Reinmar:
Followup which hasn't been answered here: https://github.com/ckeditor/ckeditor5-core/issues/259

Reinmar:
And another one: #260.

Reinmar:
OK, it's enough improvements for now. That's a very good start for our first features. There are definitely things which we'll need to improve in the future like `editor#commands` (see #186), schema and the attr command, but right now it would be to much guessing.

So, mergin'!

Issue #164, opened 2016-01-27
TreeView API for converters should be introduced


pjasiun:
Is should be simple to wrap some text into an inline style which will be automatically merged with siblings:

```
<b>foo</b>bar<i><b>bom</b></i>
->apply <b> on bar
<b>foobar<i>bom</i></b>
```

It may be done by the additional class: `treeView.Writter`.

pjasiun:
During the meeting we decided that TreeView elements will have a property which will say:

 - Is it block (<p>), inline (<b>) or inline-block  (<img>) element,
 - what is the priority (for inline),
 - can two elements can be merged (if they have different priority they can not).

pjasiun:
Also we decided that switching order of elements will be a future task, not handled by this ticket, so for the example in the fist post it will be:

`<b>foobar</b><i><b>bom</b></i>`

pjasiun:
We decided that block will be called `container`, inline will be called `attribute` and we do not need inline-block.

pjasiun:
Closed by #245.

PR #245, opened 2016-03-09
TreeView writer

szymonkups:
This PR introduces TreeView Writer class mentioned in #164.


src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+'use strict';
+
+import Position from './position.js';
+import Element from './element.js';
+import Text from './text.js';
+import Range from './range.js';
+import CKEditorError from '../ckeditorerror.js';
+
+/**
+ * Tree model Writer class.
```
pjasiun: Tree View, not tree model.

src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+ * @memberOf core.treeView
+ */
+ export default class Writer {
+	constructor() {
+		/**
+		 * Priorities map. Maps node to priority.
+		 * Nodes with priorities are considered as attributes.
+		 *
+		 * @member {WeakMap} core.treeView.Writer#_priorities
+         * @protected
```
pjasiun: Something is wrong with the indentation.

src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+         */
+		this._priorities = new WeakMap();
+	}
+
+	/**
+	 * Returns `true` if provided node is a `container` node.
+	 * Node is considered as `container` when it is a {@link core.treeView.Element Element} instance and has no priority
```
pjasiun: It would be cool to explain that the container is. The fact it has no priority is not a point. The point is method will break attribute, but not break containers, nor enter them.

src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+	 * @returns {Boolean} `True` if provided node is an attribute.
+	 */
+	isAttribute( node ) {
+		const isElement = node instanceof Element;
+
+		return isElement && this._priorities.has( node );
+	}
+
+	/**
+	 * Sets node priority. When priority is defined, node is considered as `attribute`
```
pjasiun: No dot.

src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+	 *
+	 * @param {core.treeView.Position} position Position used as a start point to locate parent container.
+	 * @returns {core.treeView.Element|null} Parent container element or `null` if container is not found.
+	 */
+	getParentContainer( position ) {
+		let parent = position.parent;
+
+		while ( !this.isContainer( parent ) ) {
+			if ( !parent ) {
+				return null;
```
pjasiun: Maybe `undefined` should be returned instead of `null` as `getPriority` (`WeakMap.get`) do.

src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+			parent = parent.parent;
+		}
+
+		return parent;
+	}
+
+	/**
+	 * Breaks attribute nodes at provided position. It breaks `attribute` nodes inside `container` node.
+	 *
+	 *		<p>{foo}<b><u>{bar}|</u></b></p> -> <p>{foo}<b><u>{bar}</u></b>|</p>
```
pjasiun: It may be unclear what is a container and what is an attribute in this example. Even if it is pretty obvious for us, it would be cool to note.

src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+	 *		<p>{foo}<b><u>|{bar}</u></b></p> -> <p>{foo}|<b><u>{bar}</u></b></p>
+	 *		<p>{foo}<b><u>{b|ar}</u></b></p> -> <p>{foo}<b><u>{b}</u></b>|<b><u>{ar}</u></b></p>
+	 *
+	 * @see {@link core.treeView.Writer#isContainer}
```
pjasiun: You do not need @link in @see tag. Change to `@see core.treeView.Writer#isContainer`.

src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+		}
+	}
+
+	/**
+	 * Uses {@link core.treeView.Writer#breakAttributes breakAttribute} method to break attributes on
+	 * {@link core.treeView.Range#start start} and {@link core.treeView.Range#end end} positions of
+	 * provided {@link core.treeView.Range Range}.
+	 *
+	 * Throws {@link core.CKEditorError CKEditorError} `treeview-writer-invalid-range-container` when
+	 * {@link core.treeView.Range#start start} and {@link core.treeView.Range#end end} positions are not placed inside
```
pjasiun: and -> or?

src/treeview/treeview.js
```diff
@@ -47,6 +48,8 @@ export default class TreeView {
 		 */
 		this.observers = new Set();
 
+		this.writer = new Writer();
```
pjasiun: Docs needed.

src/treeview/writer.js
```diff
@@ -0,0 +1,623 @@
...
+			throw new CKEditorError( 'treeview-writer-invalid-range-container' );
+		}
+
+		// If range is collapsed - nothing to remove.
+		if ( range.isCollapsed ) {
+			return [];
+		}
+
+		// Break attributes at range start and end.
+		const breakRange = this.breakRange( range );
```
pjasiun: Why not destructuring? :D
`const { start: breakStart, end: breakEnd } = this.breakRange( range );`

src/treeview/element.js
```diff
@@ -196,17 +222,62 @@ export default class Element extends Node {
...
 		for ( let i = index; i < index + number; i++ ) {
 			this._children[ i ].parent = null;
 		}
 
 		return this._children.splice( index, number );
 	}
+
+	/**
+	 * Checks if this element is similar to other element.
+	 * Both elements should have the same name and attributes to be considered as similar.
```
pjasiun: You can add that these element can have different children to make it clear.

src/treeview/node.js
```diff
@@ -129,6 +136,20 @@ export default class Node {
...
 
 	/**
+	 * Clones this node.
+	 *
+	 * @method treeView.Node#clone
+	 * @returns {treeView.Node} Clone of this node.
+	 */
+
+	/**
+	 * Checks if provided node is the same as this node.
```
pjasiun: They are not the same. Elements can have different children. You could note it here to clear what does to it mean that nodes are similar.

src/treeview/node.js
```diff
@@ -129,6 +136,20 @@ export default class Node {
...
+	 *
+	 * @method treeView.Node#isSimilar
+	 * @returns {Boolean} True if nodes are the same.
```
pjasiun: Again, not "the same".

pjasiun:
Test file is too huge it should be split into per-method files.

pjasiun:
And we could export `create` tool as a generic tool to create `treeView` structure. It may be useful for others in the future.

pjasiun:
Beside these details I do not see any big issues with the code. I am really impressed how it looks: very complex methods are split into multiple smaller and simpler, so there is no big, monolithic methods, everything well documented, described and tested. Good job!

pjasiun:
One more think: you do not need to declare `globals beforeEach, describe, it` they are declared globally for the tests folder: https://github.com/ckeditor/ckeditor5-core/blob/master/tests/.jshintrc

src/utils.js
```diff
@@ -176,6 +176,24 @@ const utils = {
...
 
 	/**
+	 * Returns the number of items return by the iterator.
+	 *
+	 *		utils.count( [ 1, 2, 3, 4, 5 ] ); // 5;
+	 *
+	 * @param {Iterable.<*>} iterator Any iterator.
+	 * @returns {Number} Number of items returned by that iterator.
+	 */
+	count( iterator ) {
```
pjasiun: Since we moved utils to the separate package (https://github.com/ckeditor/ckeditor5-core/pull/251), we should move this change there too.

PR #277, opened 2016-03-17
Implemented FocusObserver

### LINKED ISSUE #273: Implement focus observer
Actually `focus` and `blur`.

Reinmar:
Fixes #273.

### DIFF
--- a/src/treeview/observer/focusobserver.js
+++ b/src/treeview/observer/focusobserver.js
@@ -21,7 +21,15 @@ export default class FocusObserver extends DomEventObserver {
 	}
 
 	onDomEvent( domEvt ) {
-		this.fire( domEvt.type, domEvt.target );
+		const treeView = this.treeView;
+
+		this.fire( domEvt.type, {
+			get viewTarget() {
+				return treeView.domConverter.getCorrespondingViewElement( domEvt.target );
+			},
+
+			domTarget: domEvt.target
+		} );
 	}
 }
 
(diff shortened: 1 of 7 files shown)


src/treeview/observer/focusobserver.js
```diff
@@ -0,0 +1,40 @@
...
+	onDomEvent( domEvt ) {
+		this.fire( domEvt.type, domEvt.target );
+	}
+}
+
+/**
+ * Fired when one of the editables gets focus.
+ *
+ * @event core.treeView.TreeView#focus
+ * @param {HTMLElement} The event target (editable element that was focused).
```
pjasiun: ViewTree should build abstraction over the DOM and use only View elements in events.

Reinmar:
> ViewTree should build abstraction over the DOM and use only View elements in events.

I fixed that, so now focus and blur are fired with tree view target. However, I also decided that since we have the DOM target anyway, there's no reason to not expose it, because that will make the event much more useful for the whole UI and other possible pieces of code which may need something straight from the DOM.

It looks though as if we need some wrapper for that event info, because the code I wrote will repeat. It could work like this:

```js
this.fire( domEvt.type, new EventInfo( domEvt, { additionalProps: 1 } ) );
```

Then, EventInfo would provide `viewTarget`, `domTarget`, ref to the native event, and all other additional props.

pjasiun:
Maybe `fire` method in the `DomEventObserver` could be extended so it add event info automatically. Anyway I renamed `viewTarget` to `target` since `View` object should be default and do not need any prefixing here (otherwise we will need to prefix everything).

Please merge master and resolve conflicts and I we can merge this PR.

PR #287, opened 2016-03-18
T/250 History and other improvements in core for undo

scofalik:
This PR introduces improvements done to core repository that are connected with undo feature:
* `core.treeModel.History` class
* `core.treeModel.Document#batch` event
* multiple fixes in deltas, operations and OT.

There are two doubts I have about the code:
* Maybe `core.treeModel.Document#version` should be moved to `History` class? OTOH, right now `History` doesn't know anything about `Document`. But it feels natural to have that property there. I even had to implement `_nextHistoryPoint`...
* Placement of "new batch event". There are many possibilities to handle this problem:
 * leave it in `Batch.addDelta`
 * move it to `Batch` constructor
 * move it to `Document#batch`
 * move it to `Document#applyOperation` but then we need a registry for batches that we already sent an event for
 * move it to `History` but then we need a registry as well
 * do not create an event but make batches history in `History` (we will need an array + a registry)

### DIFF
--- a/src/treemodel/document.js
+++ b/src/treemodel/document.js
@@ -319,12 +319,6 @@ export default class Document {
 	 *
 	 * @event core.treeModel.Document#changesDone
 	 */
-
-	/**
-	 * Fired when a first delta is added to a batch. Can be treated as "new batch created" event.
-	 *
-	 * @event core.treeModel.Document#batch
-	 */
 }
 
 utils.mix( Document, EmitterMixin );
--- a/src/treemodel/document.js
+++ b/src/treemodel/document.js
@@ -11,6 +11,7 @@ import transformations from './delta/basic-transformations.js'; // jshint ignore
 
 import RootElement from './rootelement.js';
 import Batch from './batch.js';
+import History from './history.js';
 import Selection from './selection.js';
 import EmitterMixin from '../../utils/emittermixin.js';
 import CKEditorError from '../../utils/ckeditorerror.js';
@@ -93,6 +94,14 @@ export default class Document {
 
 		// Graveyard tree root. Document always have a graveyard root, which stores removed nodes.
 		this.createRoot( graveyardSymbol );
+
+		/**
+		 * Document's history.
+		 *
+		 * @readonly
+		 * @member {core.treeModel.History} core.treeModel.Document#history
+		 */
+		this.history = new History();
 	}
 
 	/**
@@ -130,6 +139,8 @@ export default class Document {
 		let changes = operation._execute();
 
 		this.version++;
+
+		this.history.addOperation( operation );
 
 		const batch = operation.delta && operation.delta.batch;
 		this.fire( 'change', operation.type, changes, batch );
--- a/src/treemodel/document.js
+++ b/src/treemodel/document.js
@@ -119,8 +119,7 @@ export default class Document {
 	 * {@link core.treeModel.operation.Operation operations}. To create operations in the simple way use the
 	 * {@link core.treeModel.Batch} API available via {@link core.treeModel.Document#batch} method.
 	 *
-	 * This method calls {@link core.treeModel.Document#change} event.
-	 *
+	 * @fires @link core.treeModel.Document#change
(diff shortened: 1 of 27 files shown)


src/treemodel/document.js
```diff
@@ -309,6 +319,12 @@ export default class Document {
 	 *
 	 * @event core.treeModel.Document#changesDone
 	 */
+
+	/**
+	 * Fired when a first delta is added to a batch. Can be treated as "new batch created" event.
+	 *
+	 * @event core.treeModel.Document#batch
```
pjasiun: "Can be treated as new batch created event." Can be called `batchCreated` too ;)

src/treemodel/history.js
```diff
@@ -0,0 +1,126 @@
...
+
+	/**
+	 * Transforms out-dated delta by all deltas that were added to the history since the given delta's base version. In other
+	 * words, it makes the delta up-to-date with the history. The transformed delta(s) is (are) ready to be applied
+	 * to the {@link core.treeModel.Document document}.
+	 *
+	 * @param {core.treeModel.delta.Delta} delta Delta to update.
+	 * @returns {Array.<core.treeModel.delta.Delta>} Result of transformation which is an array containing one or more deltas.
+	 */
+	updateDelta( delta ) {
```
pjasiun: `transform` is better then `update`. Update suggest that you change value of the delta stored in the history.

pjasiun:
`_nextHistoryPoint` is fine for me. `version` is a version of the document, not a version of history, so I would keep it in the document.

src/treemodel/history.js
```diff
@@ -0,0 +1,126 @@
...
+
+	/**
+	 * Transforms given delta by another given delta.
+	 *
+	 * @private
+	 * @param {core.treeModel.delta.Delta} toTransform Delta to be transformed.
+	 * @param {core.treeModel.delta.Delta} transformBy Delta to transform by.
+	 */
+	static _transform( toTransform, transformBy ) {
+		return transform( toTransform, transformBy, false );
```
pjasiun: What?!... I mean... Why?

scofalik: For testing purposes. We have discussed it on the chat and a bit live with @reinmar.

"I want to test and I want to test whether `transform` was called and with what arguments. I see 3 solutions:
1. Test the effect of the whole loop. This sucks because it is mostly testing what `transform` does and I already have tests for this.
2. Make `transform` static member of that class, which will allow spying it. This is weird because w change the class only because of tests requirements. OTOH it would let us bring in different `transform` method but that won't happen, like, ever.
3. Move the inner `for` to other method and test whether that method got called.  This kinda solves the issue but "the other method" would probably be private and not tested."

We chose to go with solution 2.

pjasiun:
About the `batch` event: I do not like how it works now. It is fired not when the batch is created, not when batch is applied, but when the first delta is added. I perfectly understand why, but it could be a big WTF for a user who would try to use it. Especially because the first delta can be added asynchronously.

Also, every plugin which use this event need to handle manually the state which this event implies.

We could keep the batches history. Then every plugin may change on change if their batch is still the last one or whatever they need. I am not sure if it should be the same history object, or we could have separate deltas history and batches history if they are separated.

src/treemodel/history.js
pjasiun: Ach, ok then. But please add a comment and change to `@protected`.

pjasiun:
And I am thinking if this approach could be used by the collaboration editing plugin later. Deltas which come from the server could be transformed by the `updateDelta` method before applying, what is nice. But local changes also need to be transform against remote changes, and I wonder if we could reuse some of this code.

But most probably I, just think too much about it now.

src/treemodel/history.js
scofalik: I don't know if it is right.
1) The comment will show up in documentation and I don't know whether it is right to document such things.
2) Why `@protected`? This is a private method, but it being private does not prohibit us from testing it, right?

Issue #256, opened 2016-03-10
Extend View Element class.


pjasiun:
View element class is missing multiple methods. The most important are to get classes and styles as collections, because now they are returned only as regular attributes.

We could also check DOM Element API and CKEditor 4 element API to ensure that there is not important methods missing.

Reinmar:
When talking about class list the implementation can follow the native API. But when talking about styles we need to figure out whether we shouldn't implement some style normalization/matching logic. For instance, when a feature tries to convert `font-family` and the incoming data has `font: 12px 'Arial'` style, it would be great if some method was able to extract `font-family` from the existing style. We missed this feature in CKEditor 4 and there were some issue because of that. However, it's not crucial – it's an enhancement that we can predict today, but implement later.

szymonkups:
Starting with classes. Each access to classes should be synchronized with `class` attribute. Should we always operate directly on this attribute, eg. create `hasClass` / `removeClass` methods that parse `class` attribute each time it is called. Maybe we can follow native API and create `ClassList` that will internally store the class list? It will update `class` attribute when it is needed and update itself when `class` attribute is accessed directly?

pjasiun:
I think that `Element` should have private/protected properties: `attributes`, `styles` i `classes` and do not keep styles and classes in the attributes collection. Then, the `getAttribute` and `setAttribute` methods should be smarter and handle `style` and `class` attribute in different way then other attributes.

PR #289, opened 2016-03-21
Classes and styles handling in tree view Element.

szymonkups:
Added classes and styles handling to `treeView.Element` class as mentioned in #256.

### DIFF
--- a/src/treeview/element.js
+++ b/src/treeview/element.js
@@ -119,9 +119,10 @@ export default class Element extends Node {
 
 		const cloned = new Element( this.name, this._attrs, childrenClone );
 
-		// Classes are cloned separately - it solution is faster than adding them back to attributes and
+		// Classes and styles are cloned separately - it solution is faster than adding them back to attributes and
 		// parse once again in constructor.
 		cloned._classes = new Set( this._classes );
+		cloned._styles = new Map( this._styles );
 
 		return cloned;
 	}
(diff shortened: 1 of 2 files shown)


src/treeview/element.js
```diff
@@ -254,27 +358,188 @@ export default class Element extends Node {
...
+			}
 		}
 
 		return true;
 	}
+
+	/**
+	 * Adds style to the element.
+	 *
+	 * @example
```
pjasiun: We do not use `@example`. We use indentation for samples.

src/treeview/element.js
```diff
@@ -254,27 +358,188 @@ export default class Element extends Node {
...
+	 * element.setStyle( 'color', 'red' );
+	 * element.setStyle( {
+	 * 		color: 'red',
+	 * 		position: 'fixed'
+	 * } );
+	 *
+	 * @param {String|Object} property Property name or object with property - value pairs.
```
pjasiun: It is usually `key - value` or `name - value`, so I would rename property to `key` or `name`.

src/treeview/element.js
```diff
@@ -254,27 +358,188 @@ export default class Element extends Node {
...
+	 * Returns true if style property is present.
+	 * If more then one style property is provided - returns true only when all properties are present.
+	 *
+	 * @example
+	 * element.hasStyle( 'color' ); // Returns true if 'border-top' style is present.
+	 * element.hasStyle( 'color', 'border-top' ); // Returns true if 'color' and 'border-top' styles are both present.
+	 *
+	 * @param {...String} property
+	 */
+	hasStyle( ...property ) {
```
pjasiun: Shouldn't it be plural (properties)? Or better `names` or `keys`.

PR #294, opened 2016-03-24
ViewConsumable

szymonkups:
`ViewConsumable` class described in #275.

### DIFF
--- a/src/treecontroller/viewconsumable.js
+++ b/src/treecontroller/viewconsumable.js
@@ -69,15 +69,9 @@ class ViewElementConsumables {
 		if ( elementOnly ) {
 			if ( this._canConsumeElement === false ) {
 				this._canConsumeElement = true;
-
-				return false;
 			}
 
-			if ( this._canConsumeElement === true ) {
-				return false;
-			}
-
-			return null;
+			return;
 		}
 
 		for ( let type in consumables ) {
@@ -140,10 +134,14 @@ class ViewElementConsumables {
 		const consumables = this._consumables[ type ];
 
 		for ( let name of items ) {
+			if ( type === 'attribute' && ( name === 'class' || name === 'style' ) ) {
+				// TODO: comment error
+				throw new CKEditorError( 'viewconsumable-invalid-attribute: Classes and styles should be handled separately.' );
+			}
 			const value = consumables.get( name );
 
 			if ( value === false ) {
-				consumables.get( name ).set( true );
+				consumables.set( name, true );
 			}
 		}
 	}
--- a/src/treecontroller/viewconsumable.js
+++ b/src/treecontroller/viewconsumable.js
@@ -7,6 +7,7 @@
 
 import isArray from '../../utils/lib/lodash/isArray.js';
 import CKEditorError from '../../utils/ckeditorerror.js';
+import ViewElement from '../treeview/element.js';
 import ViewText from '../treeview/text.js';
 import ViewDocumentFragment from '../treeview/documentfragment.js';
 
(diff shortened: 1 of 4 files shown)


src/treecontroller/viewconsumable.js
```diff
@@ -0,0 +1,486 @@
...
+	 *		viewConsumable.add( p, { style: 'color' } ); // This is properly handled style.
+	 *
+	 * @param {engine.treeView.Element} element
+	 * @param {Object} consumables
+	 * @param {Boolean} consumables.name If set to true element's name will be included.
+	 * @param {String|Array} consumables.attribute Attribute name or array of attribute names.
+	 * @param {String|Array} consumables.class Class name or array of class names.
+	 * @param {String|Array} consumables.style Style name or array of style names.
+	 */
+	add( element, consumables ) {
```
pjasiun: For all these methods `element` may be not only the element, but also a text node or document fragment (root). For all there there is not `consumables` details, but it should be possible to consume text node or document fragment.

src/treecontroller/viewconsumable.js
```diff
@@ -0,0 +1,486 @@
...
+			consumables.class.push( className );
+		}
+
+		const styles = element.getStyleNames();
+
+		for ( let style of styles ) {
+			consumables.style.push( style );
+		}
+
+		return consumables;
```
pjasiun: It could also iterate over children and add them and all their attributes, styles and classes to the consumable list.

src/treecontroller/viewconsumable.js
```diff
@@ -0,0 +1,486 @@
...
+	}
```
pjasiun: Additional static method would be useful which takes an element or document fragment and build instance of a  `ViewConsumable` class with all elemenets in the tree.

src/treecontroller/viewconsumable.js
```diff
@@ -0,0 +1,486 @@
...
+	 * @param {Object} consumables Object describing which parts of the element should be tested.
+	 * @param {Boolean} consumables.name If set to `true` element's name will be tested.
+	 * @param {String|Array} consumables.attribute Attribute name or array of attribute names to test.
+	 * @param {String|Array} consumables.class Class name or array of class names to test.
+	 * @param {String|Array} consumables.style Style name or array of style names to test.
+	 * @returns {Boolean|null} Returns `true` when all tested items can be consumed, `null` when even one of the items
+	 * were never marked for consumption and `false` when even one of the items were already consumed.
+	 */
+	test( consumables ) {
+		// Check if name can be consumed
```
pjasiun: Full stop. ;)

src/treecontroller/viewconsumable.js
```diff
@@ -0,0 +1,486 @@
...
+	 *		viewConsumable.consume( p, { attribute: 'name' } ); // Consumes element's attribute.
+	 *		viewConsumable.consume( p, { class: 'foobar' } ); // Consumes element's class.
+	 *		viewConsumable.consume( p, { style: 'color' } ); // Consumes element's style.
+	 *		viewConsumable.consume( p, { attribute: 'name', style: 'color' } ); // Consumes attribute and style.
+	 *		viewConsumable.consume( p, { class: [ 'baz', 'bar' ] } ); // Multiple consumables can be consumed.
+	 *
+	 * Throws {@link utils.CKEditorError CKEditorError} `viewconsumable-invalid-attribute` when `class` or `style`
+	 * attribute is provided - it should be handled separately by providing actual style/class.
+	 *
+	 *		viewConsumable.consume( p, { attribute: 'style' } ); // This call will throw an exception.
```
pjasiun: It should be possible to consume or test `style` or `class` attribute. It would consume all styles or class, return false if there it any of styles/classes which is not consumable.

PR #295, opened 2016-03-24
T/284 DocumentFragment, TextFragment.setAttribute

### LINKED ISSUE #284: Set attribute on detached nodes list
It should be possible to set attribute on detached nodes list:
```js
for( let item of Range.createFromElement( nodeList ) ) {
	item.setAttribute( 'bold', true );		
}
```

There are 2 missing parts in this code:
 - it is not possible to create a range over the node list nor use tree walker with the node list,
 - it is not possible to `setAttribute` on the item if the item is `TextFragment`.

We may consider creating class `DocumentFragment` which wrap `NodeList` and add iteration to it, but I am not sure if it is needed.

We may consider renaming `createFromElement` or creating new method, since `NodeList` is not an element.

scofalik:
Introduces `DocumentFragment` class and attributes manipulation API for `TextFragment.setAttribute` which should fix #284.

### DIFF
--- a/src/treemodel/delta/insertdelta.js
+++ b/src/treemodel/delta/insertdelta.js
@@ -67,7 +67,6 @@ export default class InsertDelta extends Delta {
  * @method engine.treeModel.Batch#insert
  * @param {engine.treeModel.Position} position Position of insertion.
  * @param {engine.treeModel.NodeSet} nodes The list of nodes to be inserted.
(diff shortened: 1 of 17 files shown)


pjasiun:
More-less follow-up: https://github.com/ckeditor/ckeditor5-engine/issues/303.

PR #302, opened 2016-03-29
T/300 Schema

### LINKED ISSUE #297: Schema improvements
Because of converters, schema need some improvements:

 - it should be possible to check element with multiple attributes at once (not each attribute separately),
 - it should be possible to pass array of ancestor elements, instead of names (so schema should map this array to its name, so converters do not need to do it).

scofalik:
Fixes #297 , #300 .

### DIFF
--- a/src/treemodel/schema.js
+++ b/src/treemodel/schema.js
@@ -6,6 +6,7 @@
 'use strict';
 
 import clone from '../../utils/lib/lodash/clone.js';
+import isArray from '../../utils/lib/lodash/isArray.js';
 import CKEditorError from '../../utils/ckeditorerror.js';
 
 /**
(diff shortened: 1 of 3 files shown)


src/treemodel/schema.js
```diff
@@ -241,11 +267,21 @@ export default class Schema {
 	 * @param {engine.treeModel.SchemaQuery} query Disallowed query.
 	 */
 	disallow( query ) {
-		this._getItem( query.name ).disallow( query.inside, query.attribute );
+		this._getItem( query.name ).disallow( query.inside, query.attributes );
 	}
 
 	/**
-	 * Checks whether entity with given name (and optionally, with given attribute) is allowed at given position.
+	 * Makes a requirement in schema that entity represented by given item has to have given set of attributes.
```
pjasiun: Example needed, to show not only how, but also why.

src/treemodel/schema.js
```diff
@@ -264,7 +300,28 @@ export default class Schema {
...
+	/**
+	 * Checks whether entity with given name (and optionally, with given attribute(s)) is allowed in given chain of
+	 * parent {@link engine.treeModel.Element elements}.
+	 *
+	 * @param {Array.<engine.treeModel.Element>} elements Elements that are parents of queried entity.
+	 * @param {String} name Entity name to check.
+	 * @param {Array.<String>|String} [attributes] If set, schema will check for entity with given attribute(s).
+	 * @returns {Boolean} `true` if entity is allowed, `false` otherwise
+	 */
+	checkInElements( elements, name, attributes ) {
```
pjasiun: The method name is missleading. It looks like the schema is checked for many elements. I do not think this method at all is needed, one can use `checkQuery`.

src/treemodel/schema.js
```diff
@@ -264,7 +300,28 @@ export default class Schema {
...
+		if ( !this.hasItem( name ) ) {
+			return false;
+		}
```
pjasiun: Should not `checkQuery` check this?

src/treemodel/schema.js
```diff
@@ -253,10 +289,10 @@ export default class Schema {
 	 *
```
pjasiun: More examples would be useful, for example with multiple attributes.

src/treemodel/schema.js
```diff
@@ -274,7 +331,7 @@ export default class Schema {
 	 *		// Check whether bold text is allowed in header element.
 	 *		let query = {
 	 *			name: '$text',
-	 *			attribute: 'bold',
+	 *			attributes: 'bold',
 	 *			inside: 'header'
 	 *		};
 	 *		if ( schema.checkQuery( query ) ) { ... }
```
pjasiun: I feel that the API is now overcomplicated. There is `checkAtPosition`, `checkInElements ` and `checkQuery` methods. My idea is to rename `checkQuery` to simple `check` and make it a little bit smarter so it check if `this.hasItem( name )`  and what is the type of `inside`, accepting array of elements, array of strings and position. Thanks to such change user would not need to think which method he should use.

tests/treemodel/schema/schema.js
```diff
@@ -206,6 +250,52 @@ describe( 'checkQuery', () => {
...
+		// Because of inherting from $inline, image can have bold
+		expect( schema.checkQuery( { name: 'img', inside: '$block', attributes: [ 'alt', 'src', 'bold' ] } ) ).to.be.true;
+		// But it can't have only bold without alt or/and src.
+		expect( schema.checkQuery( { name: 'img', inside: '$block', attributes: [ 'alt', 'bold' ] } ) ).to.be.false;
+		expect( schema.checkQuery( { name: 'img', inside: '$block', attributes: [ 'src', 'bold' ] } ) ).to.be.false;
+		expect( schema.checkQuery( { name: 'img', inside: '$block', attributes: [ 'bold' ] } ) ).to.be.false;
+
+		// Even if image has src and alt, it can't have attributes that weren't allowed
+		expect( schema.checkQuery( { name: 'img', inside: '$block', attributes: [ 'alt', 'src', 'attr' ] } ) ).to.be.false;
 	} );
```
pjasiun: There should be a test to check it `requireAttributes` was called on the parent item it does not affect a child item (not allow nor disallow element with required attributes).

src/treemodel/schema.js
scofalik: Added.

Issue #306, opened 2016-04-04
Tests: Create a convenient utils to test features


Reinmar:
In order to test a feature we'll need a tools to:

- [x] 1. Compare data and selection in the model with an expected value,
- [x] 2. Set data and selection in the model to a specified value (needed to test algorithms operating on the model and converters to the view),
- ~~3. Test the history (undo/redo) if needed (?),~~
- [x] 4. Compare data and selection in the view with an expected value,
- [x] 5. Set data and selection in the view to a specified value (needed to test converters to the model),
- ~~6. Way to browse contents of the model and view (needed for manual tests),~~
- ~~7. Compare DOM to an HTML string (https://github.com/ckeditor/ckeditor5-utils/issues/19) (needed for example for testing view rendering).~~
- [x] 8. Extend the tools created in 1. so they can print any position which is passed (useful for debugging).
- [x] 9. Publish these utils in `console.dump()` so you don't need to import anything. Such module, implementing this API should be loaded by default in tests and on demand in other envs. This method could be smart to accept any `( model element || model doc frag || view element || view doc frag ) && ( position || selection || range )`.
- [ ] 10. Use `[` and `]` in model test utils and pass additional selection parameters in the separate object.
- [ ] 11. Use view testing tools with converters for model testing.
- [x] 12. Make `setData()` of the model utils work in the same way as you would expect loading data to editor (so data should be rendered) – #455.

Reinmar:
Main question is about the format of the expected values.

* In the view we may use something similar like in CKEditor 4, so a well formatted HTML string.
* It would be good to find a nice string pattern for the model, but that may be much harder due to complexities and format of the model.

Reinmar:
OK, this is a quick proposal for a format that we'll use to set and read data and selection from the model. WDYT?

Full:

```html
<p><bold value=true><selection />foo</bold>bar</p>
<p><bold value=true>foo</bold><selection />bar</p> (when the selection doesn't have the bold attr)
<p><bold value=true>fo</bold><selection /><bold value=true>o</bold>bar</p>
<p><bold value=true><selection backwards>foo</selection></bold>bar</p>
<p><link value=http://foo><linkTitle value=something><selection>fo</selection>o</linkTitle></link>bar</p>
```

Ignore attribute values:

```
<p><bold>foo</bold>bar</p>
<p><link><linkTitle><selection>fo</selection>o</linkTitle></link>bar</p>
```

[...]

PR #323, opened 2016-04-06
Implemented test utils getData() and setData()

Reinmar:
Part of #306.

### DIFF
--- a/tests/_utils-tests/model.js
+++ b/tests/_utils-tests/model.js
@@ -1,0 +1,192 @@
+/**
+ * @license Copyright (c) 2003-20'INSERT'6, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+import { getData } from '/tests/engine/_utils/model.js';
+import Document from '/ckeditor5/engine/treemodel/document.js';
+import Element from '/ckeditor5/engine/treemodel/element.js';
+import Text from '/ckeditor5/engine/treemodel/text.js';
+import Range from '/ckeditor5/engine/treemodel/range.js';
+
+describe( 'model test utils', () => {
+	let document, root, selection;
+
+	beforeEach( () => {
+		document = new Document();
+		root = document.createRoot( 'main', '$root' );
+		selection = document.selection;
+
+		selection.removeAllRanges();
+	} );
+
+	describe( 'getData', () => {
+		it( 'writes elements and texts', () => {
+			root.appendChildren( [
+				new Element( 'a', null, [ 'atext' ] ),
+				new Element( 'b', null, [
+					new Element( 'c1' ),
+					'ctext',
+					new Element( 'c2' )
+				] ),
+				new Element( 'd' )
+			] );
+
+			expect( getData( document, 'main' ) ).to.equal(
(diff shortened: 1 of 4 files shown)


tests/_utils-tests/model.js
```diff
@@ -0,0 +1,192 @@
...
+		root = document.createRoot( 'main', '$root' );
+		selection = document.selection;
+
+		selection.removeAllRanges();
+	} );
+
+	describe( 'getData', () => {
+		it( 'writes elements and texts', () => {
+			root.appendChildren( [
+				new Element( 'a', null, [ 'atext' ] ),
```
scofalik: FYI: 'atext' does not have to be in an array., `new Element( 'a', null, 'atext' )` would be fine as well.

tests/_utils-tests/model.js
```diff
@@ -0,0 +1,192 @@
...
+				);
+			} );
+
+			it( 'writes selection collapsed at the text boundary', () => {
+				selection.collapse( elA, 'AFTER' );
+
+				expect( getData( document, 'main', options ) ).to.equal(
+					'<a></a><selection />foo<$text bold=true>bar</$text><b></b>'
+				);
+			} );
```
scofalik: Same test twice. Was this test supposed to test something else or just too much copypasta?

tests/_utils-tests/model.js
Reinmar: Roger that!

tests/_utils-tests/model.js
```diff
@@ -0,0 +1,359 @@
...
+
+			it( 'writes backward selection', () => {
+				selection.addRange(
+					Range.createFromParentsAndOffsets( elA, 0, elB, 0 ),
+					true
+				);
+
+				expect( getData( document, 'main', options ) ).to.equal(
+					'<a><selection backward></a>foo<$text bold=true>bar</$text><b></selection></b>'
```
scofalik: From what I've seen, attributes are ordered alphabetically. What about `backward` keyword when selection has attributes?

tests/_utils-tests/model.js
Reinmar: I changed the second test to check right text boundary.

tests/_utils-tests/model.js
```diff
@@ -0,0 +1,359 @@
...
+			} ).to.throw();
+		} );
+
+		it( 'throws when missing closing tag', () => {
+			expect( () => {
+				setData( document, 'main', '<a><b></b>' );
+			} ).to.throw();
+		} );
+
+		it( 'throws when missing closing tag for text', () => {
```
scofalik: This test is more like "missing opening tag for text". How it should be?

tests/_utils-tests/model.js
Reinmar: I treat backward as a special attribute. It's always first.

Reinmar: PS. The reason why I don't think that documenting how this function work in detail makes no sense is of course that it will change, but also that you use it for tests. If you more or less understand how it works, then it's just a matter of tuning up TCs that you wrote so they expect the correct format. It's jut that the format should not change.

PR #328, opened 2016-04-11
treeView.Selection class

szymonkups:
View selection representation as described in #308.

### DIFF
--- a/src/treeview/node.js
+++ b/src/treeview/node.js
@@ -106,6 +106,25 @@ export default class Node {
 	}
 
 	/**
+	 * Returns ancestors array of node. Node itself is not included in the array.
+	 *
+	 * @param {Boolean} [parentFirst=false] When set to `true`, array will be sorted from node's parent to root element,
+	 * otherwise root element will be the first item in the array.
+	 * @returns {Array} Array with ancestors. Empty array is returned when node has no ancestors.
+	 */
+	getAncestors( parentFirst ) {
+		const ancestors = [];
+		let parent = this.parent;
+
+		while ( parent !== null ) {
+			ancestors[ parentFirst ? 'push' : 'unshift' ]( parent );
+			parent = parent.parent;
+		}
+
+		return ancestors;
+	}
+
+	/**
 	 * Sets the {@link engine.treeView.TreeView} of the node. Note that not all of nodes need to have {@link engine.treeView.TreeView}
 	 * assigned, see {@link engine.treeView.Node#getTreeView}.
 	 *
--- a/src/treeview/node.js
+++ b/src/treeview/node.js
@@ -106,15 +106,16 @@ export default class Node {
 	}
 
 	/**
-	 * Returns ancestors array of node. Node itself is not included in the array.
+	 * Returns ancestors array of this node.
 	 *
+	 * @param {Boolean} [includeNode=false] When set to `true` this node will be also included in parent's array.
 	 * @param {Boolean} [parentFirst=false] When set to `true`, array will be sorted from node's parent to root element,
 	 * otherwise root element will be the first item in the array.
-	 * @returns {Array} Array with ancestors. Empty array is returned when node has no ancestors.
+	 * @returns {Array} Array with ancestors.
 	 */
-	getAncestors( parentFirst ) {
+	getAncestors( includeNode, parentFirst ) {
 		const ancestors = [];
-		let parent = this.parent;
+		let parent = includeNode ? this : this.parent;
 
 		while ( parent !== null ) {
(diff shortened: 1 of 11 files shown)


src/treeview/selection.js
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	/**
+	 * Returns an array of ranges added to the selection. The method returns a copy of internal array, so
+	 * it will not change when ranges get added or removed from selection.
+	 *
+	 * @returns {Array.<engine.treeView.Range>}
+	 */
+	getRanges() {
+		return this._ranges.slice();
```
pjasiun: It could be a generator: `return this. _ranges[ Symbol.iterator ]();` See, for instance: https://github.com/ckeditor/ckeditor5-engine/blob/master/src/treemodel/node.js#L176-L178

src/treeview/selection.js
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	/**
+	 * Returns copy of a range at specified index, or `null` if there is no range under that index.
+	 *
+	 * @param {Number} index Index of range in selection.
+	 * @returns {engine.treeView.Range|null}
+	 */
+	getRangeAt( index ) {
```
pjasiun: `getRangeAt` suggest it takes position as a parameter. Maybe simple `getRange`?

src/treeview/selection.js
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	/**
+	 * Removes range at given index.
+	 *
+	 * @fires engine.treeView.Selection#change:range
+	 * @param {Number} index
+	 * @returns {engine.treeView.Range|null} Returns removed range or `null` if there is no range under given index.
+	 */
+	removeRangeAt( index ) {
```
pjasiun: As above.

src/treeview/selection.js
```diff
@@ -0,0 +1,330 @@
...
+	 * Throws {@link utils.CKEditorError CKEditorError} `view-selection-range-intersects` if added range intersects
+	 * with ranges already stored in Selection instance.
+	 *
+	 * @fires engine.treeView.Selection#change:range
+	 * @param {engine.treeView.Range} range
+	 */
+	addRange( range, isBackward ) {
+		this._pushRange( range );
+		this._lastRangeBackward = !!isBackward;
+		this.fire( 'change:range' );
```
pjasiun: We do not change nothing else then `range` so it could be a simple `change` event.

src/treeview/node.js
```diff
@@ -106,6 +106,26 @@ export default class Node {
...
 
 	/**
+	 * Returns ancestors array of this node.
+	 *
+	 * @param {Boolean} [includeNode=false] When set to `true` this node will be also included in parent's array.
+	 * @param {Boolean} [parentFirst=false] When set to `true`, array will be sorted from node's parent to root element,
+	 * otherwise root element will be the first item in the array.
+	 * @returns {Array} Array with ancestors.
+	 */
+	getAncestors( includeNode, parentFirst ) {
```
pjasiun: Maybe `options` objects would be better? `getAncestors( { parentFirst: true } )` seems to be more readable them `getAncestors( false, true )`. And more scaleable.

src/treeview/selection.js
Reinmar: But model.Selection has `getRanges` returning an array.

src/treeview/selection.js
Reinmar: Is this method needed at all? model.Selection does not have it. Plus, why do you clone range in this case, but not in `getRanges()`? If we need to clone (hopefully we don't need to), then we should always clone or never clone.

src/treeview/selection.js
szymonkups: We decided that it should return iterator with clones and also model method should do the same.

src/treeview/selection.js
szymonkups: For now we will drop `getRangeAt` and `removeRangeAt` methods. If they will be needed, should be implemented both in model and view.

src/treeview/selection.js
Reinmar: Why? (I mean it's quite evident, but still explanation is needed for future reference; especially when you'll be planning changes in the model).

szymonkups: I will add documentation to this class explaining why it's behaving like this and inserts and returns copies of ranges. If we allow to return ranges that are already inside the selection we will need a mechanism to track changes made on these ranges (to fire selection `change` event for example).

Reinmar: In the API docs it will be enough to say that this method returns copies of ranges. It was more a question to us, because we've been changing our opinion couple of times on this.

Issue #330, opened 2016-04-12
Introduce AttributeElement and ContainerElement


pjasiun:
Introduce `AttributeElement` and `ContainerElement`.

PR #334, opened 2016-04-12
t/330: Introduce AttributeElement and ContainerElement.

pjasiun:
Fixes: #330.

### DIFF
--- a/src/treeview/element.js
+++ b/src/treeview/element.js
@@ -12,6 +12,14 @@ import isPlainObject from '../../utils/lib/lodash/isPlainObject.js';
 /**
  * Tree view element.
  *
+ * Editing engine does not define fixed HTML DTD. This is why the type of the {@link engine.treeView.Element} need to
+ * be defined by the feature developer. Creating an element you should use {@link engine.treeView.ContainerElement}
+ * class or {@link engine.treeView.AttributeElement}.
+ *
+ * Note that for view elements which are not created from model, like elements from mutations, paste or
+ * {@link engine.treeController.DataController#set data.set} it is not possible to define the type of the element, so
+ * these will be instances of the {@link engine.treeView.Element}.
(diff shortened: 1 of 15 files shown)


src/treeview/element.js
```diff
@@ -110,7 +118,8 @@ export default class Element extends Node {
 			}
 		}
 
-		const cloned = new Element( this.name, this._attrs, childrenClone );
+		// ContainerElement and AttributeElement should be also cloned properly.
+		const cloned = new this.constructor( this.name, this._attrs, childrenClone );
```
szymonkups: We should also clone `AttributeElement`'s priority.

src/treeview/writer.js
```diff
@@ -414,10 +342,19 @@ import DocumentFragment from './documentfragment.js';
 	 * and {@link engine.treeView.Range#end} positions are not placed inside same parent container.
 	 *
 	 * @param {engine.treeView.Range} range Range to wrap.
-	 * @param {engine.treeView.Element} attribute Attribute element to use as wrapper.
+	 * @param {engine.treeView.AttributeElement} attribute Attribute element to use as wrapper.
 	 * @param {Number} priority Priority to set.
 	 */
 	wrap( range, attribute, priority ) {
```
szymonkups: `AttributeElement` should store priority so there is no need now to pass it as third parameter.

szymonkups: `AttributeElement` store priority itself so there is no need now to pass it as third parameter.

src/treeview/element.js
pjasiun: Fixed.

pjasiun:
Pushed commits with all mentioned changes. Also I added `isSimmilar` method to the `AttributeElement` which additionally check if priorities are the same.

PR #346, opened 2016-04-14
wrapPosition in treeView.Writer

szymonkups:
PR adds `wrapPosition` method to `treeView.Writer` class, as described in #312.

### DIFF
--- a/src/treeview/position.js
+++ b/src/treeview/position.js
@@ -6,6 +6,7 @@
 'use strict';
 
 import utils from '../../utils/utils.js';
+import Text from './text.js';
 
 /**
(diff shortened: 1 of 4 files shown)


src/treeview/position.js
```diff
@@ -16,7 +17,7 @@ export default class Position {
 	/**
 	 * Creates a position.
 	 *
-	 * @param {engine.treeView.Element} parent Position parent element.
+	 * @param {engine.treeView.Node} parent Position parent node.
```
pjasiun: Isn't parent always an element? It has to have children, hasn't it?

Issue #296, opened 2016-03-25
Converter Builders


pjasiun:
## View To Model:

It is not that simple to build an flexible helper functions for converters with readable API, they take too many parameters. But it should be simple to build converters.

This is why I suggest chainable API:

```js
BuildViewConverterUsing( editor.data.toModel, editor.editing.toModel )
	.fromElement( 'strong' )
	.fromElement( 'b' )
	.fromElement( { element: 'span', style: { 'font-weight': fontWeightBold } } ).consuming( { style: 'font-weight' } )
	.toAttribute( { bold: true } );

BuildViewConverterUsing( editor.data.toModel, editor.editing.toModel )
	.fromElement( 'img' ).consuming( { name: true, class: 'src' } )
	.toElement( ( viewElement ) => new ModelElement( 'img', { 'src': viewElement.getAttribute( 'src' ) } ) );

BuildViewConverterUsing( editor.data.toModel, editor.editing.toModel )
	.fromElement( 'p' )
	.toElement( ( viewElement ) => new ModelElement( 'p' ) );

BuildViewConverterUsing( editor.data.toModel, editor.editing.toModel )
	.fromAttribute( { class: 'important' } )
	.toAttribute( { important: true } );

BuildViewConverterUsing( editor.data.toModel, editor.editing.toModel )
	.fromElement( 'a' ).consuming( { name: true, attribute: 'href' } )
	.toAttribute( ( viewElement ) => { return { 'link': viewElement.getAttribute( 'href' ) }; } );
```

Note that:
 - `BuildViewConverterUsing()` may take many `ViewConversionDispatcher`s, so it may handle many pipelines (editing and data) at the same time, or user can build separate converters for each pipeline.
 - `fromElement` take `Matcher` (https://github.com/ckeditor/ckeditor5-engine/issues/274) instance or  string or object which will be used to build `Matcher`.
 - it is possible to add multiple elements listeners using `fromElement`,
 - `consuming` defines the list of elements to consume from the last added element, or, if not defined, the consuming list is created based on the match,
 - the API is scalable, is should be possible to add more methods in the future if needed (i.e. `.withoutChildren()`).

## Model To View:

Converters from model to view could provide similar API:

```js
BuildModelConverterUsing( editor.data.toView, editor.editing.toView )
	.fromElement( 'p' )
	.toElement( ( modelElement ) =>  new ViewElement( 'p' ) );

[...]

Issue #180, opened 2016-02-04
TreeView to TreeModel conversion controller


pjasiun:
- The converter will be able to convert whole document or part of it, including the document fragment (multiple siblings without a parent, https://github.com/ckeditor/ckeditor5-core/issues/173).

- It will be configurable, by callbacks fired for every converted item (elements, texts).

- Callbacks can be assigned to the specific elements, because of the performance.

- It will be possible to register generic callback, called every time.

- Callbacks will have priorities.

- User will be able to set the position for the next converter callback (mess the document, and said where the conversion cursor should be put), by default the cursor will be moved to the next position.

- Each step will have a list of items to consume (classes, attributes, styles, name, text), and methods:
   - `consume` - will remove item from that list,
   - `isConsumed` - check if the item has been consumed.

- The list of items to consume will shrink, not grow, because we would end up with the whole list next to each item at the end of the conversion. We could initialise the list when it is used for the first time on each item.

- There will be tool to create basic converter (e.g. attribute converter).

pjasiun:
Note that TreeView to TreeModel conversion will be called on:
 - initialisation,
 - paste/drop,
 - setData.

In all cases it is data insertion. In contrast to model -> view conversion it does not need to be able to handle move or deletion. If one select text and pressed "delete" proper controller will map selected range and delete content on the model. No data conversion is needed.

pjasiun:
One of the problems TreeView to TreeModel controller need to handle is incorrect data. It need to do some clean-up based on the registered controllers. This controller work is similar to the ACF in the CKEditor 4.

For example if the `<h1>` element is converting:
 - header converter should handle it, if exists,
 - if headed converter does not exist, the paragraph controller should handle it and transform it into a paragraph,
 - if the paragraph converted is not able to handle it, maybe a `<br>` need to be inserted,
 - at the end of the chain, there may be an universal controller, equivalent of the `allowedContent: true` option, with will transform any HTML tag into a model element.

pjasiun:
During the F2F discussion with @scofalik we realised a problem with consuming mechanism.

Consider such example.

Lets have such view to convert into a model:
`<p><b><em>foo</em>bar</b></p>`

The cursor is at the begging and convert a `<p>`:
```
|<p><b><em>foo</em>bar</b></p> -> p
```

The cursor move and do nothing, because it can not convert `<b>` without text. That is fine.
```
<p>|<b><em>foo</em>bar</b></p> -> p
```
```
[...]

PR #347, opened 2016-04-15
T/179 Converters API

### LINKED ISSUE #179: TreeModel to TreeView conversion controller
- Converter will listen on the change event on the Tree Model and modify Tree View using Writer class (https://github.com/ckeditor/ckeditor5-core/issues/164).

- It will be configurable, by callbacks fired for every converted item (element, text, attribute).

- Whole text with the same attribute will be a single converted item ("f" ( bold ), "o" ( bold ), "o" ( bold ), "b", "a", "r" will fire two text conversion events: one for "foo" and one for "bar").

- Callbacks can be assigned to the specific item types, because of the performance.

- It will be possible to register generic callback, called every time.

- Callbacks will have priorities.

- User will be able to set the position for the next converter callback (mess the document, and said where the conversion cursor should be put), by default the cursor will be moved to the next position.

- Each step will have a list of items to consume (text, element, attributes), and methods:
   - `consume` - will remove item from that list,
   - `isConsumed` - check if the item has been consumed.

- The list of items to consume will shrink, not grow, because we would end up with the whole list next to each item at the end of the conversion. We could initialise the list when it is used for the first time on each item.

- Generic attribute may be consumed by the specific converter, for example image converter may consume bold attribute on the image, so we will have `` instead of `<b></b>`

- There will be tool to create basic converter (e.g. attribute converter).

scofalik:
This PR brings Converters API in `treeController` namespace. Fixes #179 #180 . There will be a number of follow-up issues but the core for converters is already ready to be used.

Follow-ups:
#296 (This is the main follow-up, this issue should be the next step of taking care of converters.)
#266 (It appeared that it is important to have this feature. We can try to workaround it but it makes some converters complicated and some maybe even impossible.)
#267 (As above.)

(Refactor stuff. We sooner we will do it, the less we will have to re-write):
#348
#349
#350
#351
#353
#355
#356


pjasiun:
You should not change `DataController` and `EditingController` there is a separate ticket for it.

pjasiun:
We should ensure that View to Model converters flattens structure. I mean that if only text is supported then `<b>foo<i>bar<u>bom</u></i></b>` will return [ 'foo', 'bar', 'bom' ], not [ 'foo', [ 'bar', [ 'bom' ] ] ].

scofalik:
I've added a list of follow-ups to the first post.

scofalik:
I've added a test you requested and tried to clean up mess with editing/data controllers.

pjasiun:
I like the code, docs, tests and examples. We could now discuss docs, for instance if the `ModelConsumable` should have an example of a `modelConversionDispatcher` listener. But I think that part of this documentation need to be moved to the architecture docs anyway and we will be able to make docs better when we will have architecture docs. We could also spend some time improving advance tests, but their purpose was to prove it is possible what they did.

I think it is time to close it. Good job!

PR #358, opened 2016-04-18
Improvements to treeView.writer.wrap().

szymonkups:
This PR contains improvements added to `treeView.writer.wrap()` method. Now it handles `AttributeElement` wrapping by merging attributes/styles/classes whenever possible, as described in #252.

### DIFF
--- a/src/treeview/writer.js
+++ b/src/treeview/writer.js
@@ -309,7 +309,7 @@ import DocumentFragment from './documentfragment.js';
 	}
 
 	/**
-	 * Wraps elements within range with provided attribute element.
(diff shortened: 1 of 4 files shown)


pjasiun:
I merged master into this branch and get red tests. Please check all tests, especially:
http://tests.ckeditor.dev:1030/build/amd/tests/engine/treecontroller/advanced-converters

szymonkups:
It looks like I need to improve `treeView.writer.unwrap()` method too. It should be able to unwrap part of a single `AttributeElement`, for example:

```
<a href="foo" title="bar"></a> 
unwrap <a href="foo"></a>
<a title="bar"></a>
```
Also should take special care about styles and classes, same as during wrapping.

Issue #344, opened 2016-04-13
Implement TreeWalker options.direction


Reinmar:
It's currently impossible to iterate backwards which is a super common task (most algorithms work forwards and backwards – e.g. delete and backspace). I need to first scan the whole document and then pick up the last item.

This will become a performance issue pretty soon as simple backspace needs to scan the document to the left.

Reinmar:
It'd be also good to close https://github.com/ckeditor/ckeditor5-engine/issues/345 to clarify which positions are returned (what may be unclear when we iterate backward).

pjasiun:
We had `previous();` method in the `TreeWalker` not so long ago: https://github.com/ckeditor/ckeditor5-engine/pull/241/files But then maintaining that code, adding support to all options, when no one wast using it was a waist of time. No you can restore (and updated) that code if such logic is needed.

oskarwrobel:
I have a question about detail.

What will be better?
`option.direction = 'backward|forward'; // default forward`
or
`option.backward = true|false; // default false`;

pjasiun:
I like `direction` option. With a small change: "backward" and "forward", as flag should be all upper case.

PR #362, opened 2016-04-18
Introduced model composer.

### LINKED ISSUE #341: Introduce model composer
We need a handful of functions for most common operations on the document. We decided to group them under the name "composer". It's not perfect, but we could not find a better one. The view writer will be renamed to view composer as well.

Couple example ones:

* `deleteContents`,
* `weakInsertContents` and/or `insertContents`,
* `getContents`,
* `modifySelection` (expand, shrink),

The main idea about these functions is that they need to be extensible. They will be used by backspace/delete/typing/enter/etc. handlers which cannot know about all the features that can exist. Hence, some basic implementations can exist, but many features (like lists, tables, widgets) will want to tune up these algorithms.

We will also have view events such as "delete", "input", "enter", "copy", etc. Some features will want to listen to those too (e.g. to enter), but some actions combine multiple operations (e.g. typing does `deleteContents` and then `insertContents`) . If one of the features needs to override the delete action and if only the view events were the extension point, then it would also need to override input and enter events to tune how the delete action works there. Furthermore, there may be other algorithms about which that feature might not know and which might be deleting or inserting contents some.

The way how composer extensions will work is that events will be fired by `document.composer` methods. They will have default actions in form of a functions that can also be imported directly (so their logic can be reused without firing next events).

Reinmar:
Fixes #341.

You can test it on https://github.com/ckeditor/ckeditor5/tree/delete-feature (together with https://github.com/ckeditor/ckeditor5-delete/issues/1).

### DIFF
--- a/src/treemodel/composer/modifyselection.js
+++ b/src/treemodel/composer/modifyselection.js
@@ -13,7 +13,7 @@ import Range from '../range.js';
  * Modifies the selection. Currently the supported modifications are:
  *
  * * Extending. The selection focus is moved in the specified `options.direction` with a step specified in `options.unit`
- * (defaults to `'character'`, other values are not not yet supported).
+ * (defaults to `'CHARACTER'`, other values are not not yet supported).
  * Note: if you extend a forward selection in a backward direction you will in fact shrink it.
  *
  * @method engine.treeModel.composer.modifySelection
--- a/src/treemodel/composer/modifyselection.js
+++ b/src/treemodel/composer/modifyselection.js
@@ -21,11 +21,7 @@ import Range from '../range.js';
  * @param {Object} [options]
  * @param {'FORWARD'|'BACKWARD'} [options.direction='FORWARD'] The direction in which the selection should be modified.
  */
-export default function modifySelection( selection, options ) {
-	if ( !options ) {
-		options = {};
-	}
-
+export default function modifySelection( selection, options = {} ) {
 	const isForward = options.direction != 'BACKWARD';
 
 	const focus = selection.focus;
--- a/src/treemodel/composer/modifyselection.js
+++ b/src/treemodel/composer/modifyselection.js
@@ -1,0 +1,88 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+import Position from '../position.js';
+import TreeWalker from '../treewalker.js';
+import Range from '../range.js';
+
+/**
+ * Modifies the selection. Currently the supported modifications are:
+ *
+ * * Extending. The selection focus is moved in the specified `options.direction` with a step specified in `options.unit`
+ * (defaults to `'character'`, other values are not not yet supported).
+ * Note: if you extend a forward selection in a backward direction you will in fact shrink it.
+ *
+ * @method engine.treeModel.composer.modifySelection
+ * @param {engine.treeModel.Selection} The selection to modify.
+ * @param {Object} [options]
+ * @param {'FORWARD'|'BACKWARD'} [options.direction='FORWARD'] The direction in which the selection should be modified.
+ */
+export default function modifySelection( selection, options ) {
+	if ( !options ) {
+		options = {};
+	}
+
+	const isForward = options.direction != 'BACKWARD';
+
+	const focus = selection.focus;
+	const walker = new TreeWalker( {
+		boundaries: getSearchRange( focus, isForward ),
+		singleCharacters: true
+	} );
+
+	const items = Array.from( walker );
+	let next = items[ isForward ? 'shift' : 'pop' ]();
+
+	// 1. Nothing to do here.
+	if ( !next ) {
+		return;
+	}
+
+	// 2. Consume next character.
+	if ( next.type == 'CHARACTER' ) {
+		selection.setFocus( next[ isForward ? 'nextPosition' : 'previousPosition' ] );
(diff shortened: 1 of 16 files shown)


Reinmar:
I need to fix two things yet:

1. One incorrect test in http://tests.ckeditor.dev:1030/build/amd/tests/engine/treecontroller/advanced-converters which fails since I added validation in `Position.createFromParentAndOffset()`.
2. One minor doc thing.

Reinmar:
OK, ready for review!

pjasiun:
Follow ups:
https://github.com/ckeditor/ckeditor5-engine/issues/363 Merge model document.composer with document.
https://github.com/ckeditor/ckeditor5-engine/issues/364 Rename or merge view.writer.

Reinmar:
And also: #365.

Reinmar:
And also: #366.

Reinmar:
I see that composer events' docs are missing. Working on it.

Reinmar:
Done.

src/treemodel/selection.js
```diff
@@ -229,44 +229,49 @@ export default class Selection {
...
-	 * * parent element and `'END'` (sets selection at the end of that element),
-	 * * node and `'BEFORE'` or `'AFTER'` (sets selection before or after the given node).
-	 *
-	 * @fires {@link engine.treeModel.Selection#change:range change:range}
+	 * @fires engine.treeModel.Selection#change:range
 	 * @param {engine.treeModel.Node|engine.treeModel.Position} nodeOrPosition
 	 * @param {Number|'END'|'BEFORE'|'AFTER'} [offset=0] Offset or one of the flags. Used only when
 	 * first parameter is a node.
 	 */
 	collapse( nodeOrPosition, offset ) {
```
scofalik: I propose `collapseAt` unless both arguments are optional.

src/treemodel/selection.js
```diff
@@ -229,44 +229,49 @@ export default class Selection {
...
+	 * Sets {@link engine.treeModel.Selection#focus} in the specified location.
+	 *
+	 * The location can be specified in the same form as {@link engine.treeModel.Position.createAt} parameters.
+	 *
+	 * @fires engine.treeModel.Selection#change:range
+	 * @param {engine.treeModel.Node|engine.treeModel.Position} nodeOrPosition
+	 * @param {Number|'END'|'BEFORE'|'AFTER'} [offset=0] Offset or one of the flags. Used only when
+	 * first parameter is a node.
+	 */
+	setFocus( nodeOrPosition, offset ) {
```
scofalik: The result of this function may be a range that intersects with other ranges currently in seleciton. AFAIR at one point we said that this should not happen. This behavior may have a negative influence on some scripts (i.e. when a scripts takes all ranges from selection and processes them).

src/treemodel/selection.js
Reinmar: There's [collapse()](https://developer.mozilla.org/en-US/docs/Web/API/Selection/collapse) method in the DOM. Besides, that method wasn't added in this PR.

src/treemodel/selection.js
Reinmar: If I read this correctly, since the addRange() method is used the `_pushRange()` method will validate that ranges do not intersect. Isn't it enough?

src/treemodel/composer/modifyselection.js
```diff
@@ -0,0 +1,88 @@
...
+
+	const isForward = options.direction != 'BACKWARD';
+
+	const focus = selection.focus;
+	const walker = new TreeWalker( {
+		boundaries: getSearchRange( focus, isForward ),
+		singleCharacters: true
+	} );
+
+	const items = Array.from( walker );
```
scofalik: I am not sure if this is efficient, but I suppose that `modifySelection` won't be fired that often so I guess it's okay.

Reinmar: That's due to missing https://github.com/ckeditor/ckeditor5-engine/issues/344.

src/treemodel/composer/modifyselection.js
```diff
@@ -0,0 +1,88 @@
...
+ * * Extending. The selection focus is moved in the specified `options.direction` with a step specified in `options.unit`
+ * (defaults to `'CHARACTER'`, other values are not not yet supported).
+ * Note: if you extend a forward selection in a backward direction you will in fact shrink it.
+ *
+ * @method engine.treeModel.composer.modifySelection
+ * @param {engine.treeModel.Selection} The selection to modify.
+ * @param {Object} [options]
+ * @param {'FORWARD'|'BACKWARD'} [options.direction='FORWARD'] The direction in which the selection should be modified.
+ */
+export default function modifySelection( selection, options ) {
```
Reinmar: Options should default to `{}`

src/treemodel/position.js
```diff
@@ -463,11 +504,21 @@ export default class Position {
...
 	 * @returns {engine.treeModel.Position}
 	 */
 	static createFromParentAndOffset( parent, offset ) {
+		if ( !( parent instanceof Element || parent instanceof DocumentFragment ) ) {
+			/**
+			 * Position must be anchored in an element.
+			 *
+			 * @error position-in-text
+			 */
+			throw new CKEditorError( 'position-in-text: Position cannot be in located inside a text node.' );
```
Reinmar: I updated the comment, but not this message ;/ Will fix it later.

src/treemodel/selection.js
scofalik: If this is fine with you, it is ok.

Issue #360, opened 2016-04-18
Tests: Create utility functions for testing view tree. 


szymonkups:
This is a part of #306.
We need to create utility functions that will allow to compare view data and selection to expected value, and set view data and selection to given value.

szymonkups:
I'm almost finished with the tools but have problem defining final API that will suit all our needs, but still be easy to use.
#### getData() ####
##### getData( node [, options ] ) #####
Returns HTML-like string representing `TreeView` begining from `node`.
```javascript
getData( element );  // <a>foobar</a>
```
##### getData( node, selection [, options ] )#####
Returns HTML-like string representing `TreeView` begining from `node` and also includes ranges from provided `Selection` instance object.
```javascript
getData( element, selection ) // <a>{foobar}</a>
```
Options object can include two boolean flags: `showType` and `showPriority`. 
When `showType` will be set to `true`, `AttributeElements` and `ContainerElements` will be printed in a form of `<attribute:a>` and `<container:div>`. 
When `showPriority` will be set to `true`, `AttributeElements` priority will be printed: `<a:12>`.

#### setData() ####
##### setData( data ) #####
Parses provided data and returns nodes with all parsed elements.
```javascript
setData( 'foobar'); // Returns Text node.
setData( '<a>foobar</a>' ); // Returns Element with text node inside.
setData( '<container:div></container:div>' ); // Returns ContainerElement.
setData( '<a:12></a:12>' ); // Returns AttributeElement that has priority set to `12`.
setData( '<a></a><b></b>' ); // Returns DocumentFragment that contains two nodes.
```
Selection ranges can be included. For example:
```javascript
setData( '<a>{foobar}</a>', selection );
```
All ranges inside selection will be replaced by ranges that are set inside provided `data`.
Order of the ranges inside selection can be included. For example:
```javascript
setData( '<a>{foo}{bar}</a>' ), selection, '21' );
```
All ranges inside selection will be replaces by ranges in desired order (first range placed as second, second as first). 

**Questions**

1. Should we include some other ways of getting the view data? For exampe: ```getData( treeView, rootName, selection [, options] )``` - will return HTML-like string that include nodes  from `TreeView` instance starting from given `rootName`  and include provided selection ranges.
1. Should `setData` also include a parameter that will allow to set selection order, something like `lastRangeBackward`?
1. Should we include other ways of setting view data?
[...]

PR #381, opened 2016-04-27
Tree View test utilties.

szymonkups:
This PR contains view test utility functions, described as first level API in #360. Tools introduce two functions `parse` and `stringify` that can be used to compare view nodes.
I've also changed `treeView.writer` tests to use these new functions and removed old utilities used there.


src/treeview/documentfragment.js
```diff
@@ -123,10 +125,23 @@ export default class DocumentFragment {
...
+
 		for ( let i = index; i < index + howMany; i++ ) {
 			this._children[ i ].parent = null;
 		}
 
 		return this._children.splice( index, howMany );
 	}
+
+	/**
+	 * @param {engine.treeView.ChangeType} type Type of the change.
```
Reinmar: Should be documented as private.

src/treeview/documentfragment.js
```diff
@@ -123,10 +125,23 @@ export default class DocumentFragment {
...
+	 * @param {engine.treeView.Node} node Changed node.
+	 * @fires engine.treeView.Node#change
+	 */
+	_fireChange( type, node ) {
+		this.fire( 'change', type, node );
```
Reinmar: Why not `change:type`, @pjasiun?

pjasiun: Because events namespaces were introduced later.

tests/_utils-tests/view.js
```diff
@@ -0,0 +1,423 @@
...
+		it( 'should parse DocumentFragment as root', () => {
+			const text1 = new Text( 'foobar' );
+			const text2 = new Text( 'bazqux' );
+			const b1 = new Element( 'b', null, text1 );
+			const b2 = new Element( 'b', null, text2 );
+			const fragment = new DocumentFragment( [ b1, b2 ] );
+			expect( stringify( fragment, null ) ).to.equal( '<b>foobar</b><b>bazqux</b>' );
+		} );
+
+		it( 'should not write ranges outside elements', () => {
```
Reinmar: Split such tests into multiple ones. Single test should check single case.

tests/_utils-tests/view.js
```diff
@@ -0,0 +1,423 @@
...
+			const text = new Text( 'foobar' );
+			const b = new Element( 'b', null, text );
+			const p = new Element( 'p', null, b );
+			const range1 = Range.createFromParentsAndOffsets( p, 0, p, 5 );
+			const range2 = Range.createFromParentsAndOffsets( p, -1, p, 1 );
+			const range3 = Range.createFromParentsAndOffsets( text, 0, text, 7 );
+			const range4 = Range.createFromParentsAndOffsets( text, -1, text, 2 );
+			const range5 = Range.createFromParentsAndOffsets( text, 6, text, 8 );
+			const selection = new Selection();
+			selection.addRange( range1 );
```
Reinmar: I'm missing vertical spacing in this code. It's made of logical parts - preparation, case1, case2, case3, etc. Blank lines would make it less overwhelming :D.

tests/_utils/view.js
```diff
@@ -0,0 +1,836 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/* jshint latedef:false */
```
Reinmar: That's to suppress the late class definition warning? I always thought that you can't define a class after it's used (in code). That didn't make sense to me, so I would remove that rule if it's true.

szymonkups: Yes - this is used to suppress the late class definition JSHint warning. Class is used in code that is located above definition of the class but instance is not created until execution of `parse` / `stringify` methods. I wanted to have my helper/private classes to be located later in the file. What is better then to avoid JSHint errors: to move classes to the beginning of the file or put them in separate files?

Reinmar: No, no – we should remove (or add) that option to JSHint config. It's totally natural that you want the code of tests to be located at the front and AFAIS class definition works like a function definition (worth confirming with the spec or some article). So the same way we define functions at the end of the scope we should define the classes as well.

Reinmar: E.g. I already caught myself three times on reading https://github.com/ckeditor/ckeditor5-engine/blob/master/src/treemodel/schema.js#L23 as `Schema` because I was expecting `Schema` to be defined first.

tests/_utils/view.js
```diff
@@ -0,0 +1,836 @@
...
+ * will be included in output data.
+ * If range position is placed inside element node, it will be represented with `[` and `]`:
+ *
+ *		const text = new Text( 'foobar' );
+ *		const b = new Element( 'b', null, text );
+ *		const p = new Element( 'p', null, b );
+ *		const selection = new Selection();
+ *		selection.addRange( Range.createFromParentsAndOffsets( p, 0, p, 1 ) );
+ *
+ *		stringify( p, selection ); // <p>[<b>foobar</b>]</p>
```
Reinmar: Is it possible to pass range to that method directly, without creating a selection? Could I pass position as well?

szymonkups: Yes it is possible. Range and position support was added later on so this example is outdated and need to be changed or other examples showing that the same can be achieved more easily would be nice.

src/treeview/documentfragment.js
Reinmar: Followup – https://github.com/ckeditor/ckeditor5-engine/issues/383

tests/_utils/view.js
Reinmar: @szymonkups Could you fix this in a separate ticket on ckeditor/ckeditor5 and other repositories?

tests/_utils/view.js
```diff
@@ -0,0 +1,836 @@
...
+ *
+ * Multiple nodes provided on root level will be converted to {@link engine.treeView.DocumentFragment DocumentFragment}:
+ *
+ *		parse( '<b>foo</b><i>bar</i>' ); // Returns DocumentFragment with two child elements.
+ *
+ * Method can parse multiple {@link engine.treeView.Range ranges} provided in string data and return
+ * {@link engine.treeView.Selection Selection} instance containing these ranges. Ranges placed inside
+ * {@link engine.treeView.Text Text} nodes should be marked using `{` and `}` brackets:
+ *
+ * 		const { text, selection } = parse( 'f{ooba}r' );
```
Reinmar: I see here a different indentation. There should be two tabs (no spaces) in all of those code snippets.

tests/_utils/view.js
```diff
@@ -0,0 +1,836 @@
...
+ *
+ * Ranges placed outside text nodes should be marked using `[` and `]` brackets:
+ *
+ *		const { root, selection } = parse( '<p>[<b>foobar</b>]</p>' );
+ *
+ * Sometimes there is a need for defining order of ranges inside created selection. This can be achieved by providing
+ * ranges order array as additional parameter:
+ *
+ *		const { root, selection } = parse( '{fo}ob{ar}{ba}z', { order: [ 2, 3, 1 ] } );
```
Reinmar: As for `order` it would be easier to pass it as `'1,2,3'`. That'd be much easier to read.

tests/_utils/view.js
```diff
@@ -0,0 +1,836 @@
...
+			const children = [ ...node.getChildren() ];
+
+			for ( let child of children ) {
+				this._getPositions( child );
+			}
+		}
+
+		if ( node instanceof ViewText ) {
+			const regexp = new RegExp(
+				`[ ${TEXT_RANGE_START_TOKEN}${TEXT_RANGE_END_TOKEN}\\${ELEMENT_RANGE_END_TOKEN}\\${ELEMENT_RANGE_START_TOKEN} ]`,
```
Reinmar: For readability we put spaces inside values in template strings: `'foo ${ value }'`.

tests/_utils/view.js
```diff
@@ -0,0 +1,836 @@
...
+	}
+}
+
+/**
+ * Private helper class used to convert given HTML-like string to view tree.
+ *
+ * @private
+ */
+class ViewParser {
+
```
Reinmar: Blank line :P.

tests/_utils/view.js
Reinmar: On the other hand it'll be super rarely used, so let's leave it as it is now.

tests/treeview/writer/breakrange.js
```diff
@@ -34,198 +42,52 @@ describe( 'Writer', () => {
...
-				instanceOf: ContainerElement,
-				name: 'p',
-				rangeStart: 1,
-				rangeEnd: 1,
-				children: [
-					{ instanceOf: Text, data: 'foo' },
-					{ instanceOf: Text, data: 'bar' }
-				]
-			} );
+			test(
```
Reinmar: OK, I've just want to say that this is amazing how these tests shortened :D.

Reinmar:
To sum up:

1. We also need two functions `getData( treeView, rootName, options )` and `setData( treeView, rootName, options )` which will work directly on the tree view, with that view's selection.
2. We need `parse()` and `stringify()` in the model. They should have similar capabilities to view's functions. `getData()` and `setData()` are defined there already.
3. Model  `getData()`'s `options.selection` should be renamed to `options.withSelection` to conform to the naming in the view.
4. All those methods should be published in `bender.view|model` namespace. This needs to be done in ckeditor/ckeditor5 repo.

tests/_utils/view.js
szymonkups: Sure.

szymonkups:
I've created two separate follow-up issues: 

- to unify test tools https://github.com/ckeditor/ckeditor5-engine/issues/385,
- to publish them inside `bender` namespace  https://github.com/ckeditor/ckeditor5/issues/144.

PR #382, opened 2016-04-27
Added nodes validation to engine.treeView.Writer.insert() method.

szymonkups:
As described in #377 writer should check types of inserted elements. `Writer.wrap()` and `Writer.unwrap()` methods are already checking if element used for wrapping/unwrapping is an instance of `AttributeElement`.


src/treeview/writer.js
```diff
@@ -219,11 +220,27 @@ import DocumentFragment from './documentfragment.js';
...
+	 *
 	 * @param {engine.treeView.Position} position Insertion position.
-	 * @param {engine.treeView.Node|Iterable.<engine.treeView.Node>} nodes Node or nodes to insert.
+	 * @param {engine.treeView.Text|engine.treeView.AttributeElement|engine.treeView.ContainerElement
+	 * |Iterable.<engine.treeView.Text|engine.treeView.AttributeElement|engine.treeView.ContainerElement>} nodes Node or
+	 * nodes to insert.
 	 * @returns {engine.treeView.Range} Range around inserted nodes.
 	 */
 	insert( position, nodes ) {
+		if ( !utils.isIterable( nodes ) ) {
```
Reinmar: You can remove the `!` by switching the order of blocks.

src/treeview/writer.js
```diff
@@ -833,9 +856,38 @@ function unwrapAttributeElement( wrapper, toUnwrap ) {
...
+//
+// Throws {@link utils.CKEditorError CKEditorError} `treeview-writer-insert-invalid-node` when nodes to insert
+// contains instances that are not {@link engine.treeView.Text Texts},
+// {@link engine.treeView.AttributeElement AttributeElements} or
+// {@link engine.treeView.ContainerElement ContainerElements}.
+//
+// @private
+// @param Iterable.<engine.treeView.Text|engine.treeView.AttributeElement|engine.treeView.ContainerElement> nodes
+function validateNodesToInsert( nodes ) {
+	for ( let node of nodes ) {
```
Reinmar: If that's a generator, then after iterating it once it will be empty.

Reinmar: If we accept an iterable and need to check all instances anyway, then doing `Array.from()` at the beginning makes sense.

szymonkups: That's true - I am converting iterable to array in `insert` method, so maybe it should be documented simply as `array` here.

src/treeview/writer.js
```diff
@@ -833,9 +856,38 @@ function unwrapAttributeElement( wrapper, toUnwrap ) {
...
+		if ( !( node instanceof Text || node instanceof AttributeElement || node instanceof ContainerElement ) ) {
```
Reinmar: Don't we have a single class which is a parent of these classes?

src/treeview/writer.js
Reinmar: Right. So the doc is wrong. But if you change the doc then the `validateNodesToInsert( node.getChildren() )` call below will be incorrect :D.

src/treeview/writer.js
szymonkups: Yes. But here, this checking is exactly for preventing usage of `Element` instance (which is not valid here).

src/treeview/writer.js
szymonkups: That's indeed inconsistent, looks like converting to array at the beginning is  the way to go.

Reinmar: The reason why I mentioned this issue is that the `getChildren()` call would be fine – we could keep that and everything would work. It's a thing that I don't like about iterables vs arrays... in one place you can't pass the object, because some other function would consume it once and for all, in other place it's just fine to make that simple call, without special conversion to an array or pass the array that you already have. And we could say that it's a theoretical issue, but as we've seen it's not. When writing a function like `validateNodesToInsert` you don't know how it will be used and in case of the iterables it's the user who needs to think about it. It's the opposite of the usual problem with arrays, which you need to remember to not touch inside that function... which is easier to do.

I guess I'll start an issue in the design repo tomorrow. I feel that this will be a recurring issue. Although perhaps the only thing which we need to say is that if a function accepts an iterable it should (or can) consume it and you should remember about this.

src/treeview/writer.js
```diff
@@ -833,9 +856,38 @@ function unwrapAttributeElement( wrapper, toUnwrap ) {
...
+			/**
+			 * Inserted nodes should be instance of AttributeElement, ContainerElement or Text.
```
Reinmar: Make them links to those classes.

Reinmar:
AFAICS there's only that issue with iterable and minor thing in the error doc.

Issue #389, opened 2016-04-29
Fixes required by undo.


scofalik:
Issue for fixes required by undo feature.

scofalik:
Will fix #340

scofalik:
Will fix #307

Reinmar:
Fixed by #392.

PR #392, opened 2016-05-04
T/389 ModelRange#getTransformedByMove and other

scofalik:
This PR combines changes from #389 which are required by undo manager. The biggest change is introducing `engine.treeModel.Range#getTransformedByMove` method.


scofalik:
Follow up #393

src/treemodel/document.js
```diff
@@ -307,16 +311,15 @@ export default class Document {
 	 * * 'removeRootAttribute' when attribute for root is removed,
 	 * * 'changeRootAttribute' when attribute for root changes.
 	 *
-	 * Change event is fired after the change is done. This means that any ranges or positions passed in
```
Reinmar: Why did you remove this line?

Reinmar:
Important reminder – please remember to rebase your branch onto the latest master before putting it on review.

src/treemodel/history.js
```diff
@@ -107,13 +100,33 @@ export default class History {
...
-			index++;
 		}
 
 		return transformed;
 	}
 
 	/**
+	 * Returns all deltas from history, starting from given history point (if passed).
+	 *
+	 * @param {Number} from History point.
```
Reinmar: Missing `@returns`. Can't it be a generator, btw?

src/treemodel/document.js
scofalik: Because it was not true - `data.range` is indeed after the change, but `data.position` is before the change. I added additional note in params comments.

scofalik:
> Important reminder – please remember to rebase your branch onto the latest master before putting it on review.

Is there really a difference? I always rebase with master (`git rebase master`, I might forgot once) and there were never any problems with my PRs (I mean they were but because of different stuff). And when I tried the other method I got some problems. I have experience using my workflow so if there isn't big difference I'd like to kindly ask if I can keep doing as I was always doing?

src/treemodel/history.js
scofalik: Done

Reinmar:
@scofalik: I was asking exactly for rebasing :D. We only need to merge once an issue is already on master.

Anyway, it's not a big issue – it just may delay or confuse the reviewer. We had some issues because of that in the past. Especially, when more things depend on each other, being really up to date with your changes is important to ensure that tests pass.

scofalik:
Sure, sure. I understand the point of being up-to-date with master :), I just thought you are talking about different command.

Issue #163, opened 2016-01-27
Quirks and boguses


pjasiun:
Tree view should handle quirks and boguses. All of them should be transparent for the user, he should not see any ZWS nor bogus br in the tree view. They should be handled by the `treeView.DomConverter` and appear only in the dirty DOM. Some configuration options may be added to the `treeView.DomConverter` to defined how boguses should be handled.

pjasiun:
Problems with ZWS in the CKEditor 4:
http://dev.ckeditor.com/ticket/13593
http://dev.ckeditor.com/ticket/13816

pjasiun:
Some examples of boguses and ZWSes to handle.
```
<p>|<br bogus></p>
<p>foo<b>ZWS^</b>bar</p>
<p><b>foo</b>ZWS^<b>bar</b></p>
<p>ZWS^<b>foo</b></p>
<ul><li>^<br bogus><ul><li>x</li></ul></li></ul>
foo<br>bar<br bogus>
<hr /><br bogus><hr />
<p><a>text<br bogus></a></p> // ?
```

pjasiun:
ZWSes should be also removed on `keydown` or `selectionchange` events by the `SelectionObeserver`.

PR #397, opened 2016-05-11
t/163: Selection handling

pjasiun:
Fixes: #163, #311 and #310.


src/treeview/domconverter.js
```diff
@@ -215,12 +304,128 @@ export default class DomConverter {
...
+				return null;
+			}
+
+			if ( startsWithFiller( domParent ) ) {
+				offset -= INLINE_FILLER_LENGTH;
+				offset = offset < 0 ? 0 : offset;
+			}
+
+			return new ViewPosition( viewParent, offset );
+		} // domParent instanceof HTMLElement
```
Reinmar: The comment should go between in a new line `}` and `else {`.

Reinmar:
I'd use `data-cke-filler` attribute, instead of `data-filler` to follow CKE4's naming scheme for internals.

Reinmar:
```
 * `<br> filler creator. This is a function which creates `<br data-filler="true">` element, but should be understood as
 * configuration option more than used directly.
```

Not sure what you've meant by the last sentence.

Reinmar:
`getDataWithoutFiller()` lacks documentation.

(I'm writing comments here because GH isn't able to handle such a big PR and it hangs my browser pretty badly.)

Reinmar:
Rename the `view#selectionchange` event to `view#selectionChange`.

Reinmar:
`TreeView#addObserver()` returns `undefined` if an observer already exists. I think that it should always return an instance, so a feature doesn't have to care whether it's the first to add some observer or not.

Reinmar:
I guess you know that this is ugly:

```js
		// Assign key observer which move cursor from the end of the inline filler to the begging of it when
		// the left arrow is pressed, so the filler does not break navigation.
		this.on( 'keydown', jumpOverInlineFiller );
```

First of all, TreeView doesn't even know now that the event is going to be fired. Second, it should not care about stuff like this. Couldn't render or some other class more related to the DOM do that?

Reinmar:
> I guess you know that this is ugly

Another idea could be to move the code (function) to the `filler.js` and hide it in `injectQuirks( document )` function. That function would do whatever it needs to inject those quirks and would be called by the document (tree view).

Reinmar:
OK, I've seen all the code which has changed in `src/`. I'm going to check manual tests tomorrow, but so far this PR is nearly perfect. Taken its size and complexity it's really great. Pleasure to read.

Reinmar:
TC:

1. Open http://tests.ckeditor.dev:1030/build/amd/tests/engine/treeview/manual/immutable
2. Press left arrow.
3. Type a letter.

That latter should be removed from the DOM, but isn't.

Reinmar:
Not sure whether we want to address this now, but obviously all tests requiring native selectionchange event won't pass on FF. E.g. http://tests.ckeditor.dev:1030/build/amd/tests/engine/treeview/manual/immutable. I think that there's no point in adding a comment to all manual tests now, because we'll have to work on polyfill ASAP, definitely in 0.1.0.

pjasiun:
> TC:
> 
> Open http://tests.ckeditor.dev:1030/build/amd/tests/engine/treeview/manual/immutable
> Press left arrow.
> Type a letter.
> That latter should be removed from the DOM, but isn't.

Fixed.

pjasiun:
> Not sure whether we want to address this now, but obviously all tests requiring native selectionchange event won't pass on FF. E.g. http://tests.ckeditor.dev:1030/build/amd/tests/engine/treeview/manual/immutable. I think that there's no point in adding a comment to all manual tests now, because we'll have to work on polyfill ASAP, definitely in 0.1.0. Do we have a ticket for that?

I created a follow-up: https://github.com/ckeditor/ckeditor5-engine/issues/407.

pjasiun:
> I'd use `data-cke-filler` attribute, instead of `data-filler` to follow CKE4's naming scheme for internals.

For:
```
const fillerBr = domDocument.createElement( 'br' );
fillerBr.dataset[ 'cke-filler' ] = true;
```
I get:
`x-index:109 DOMException: Failed to set the 'cke-filler' property on 'DOMStringMap': 'cke-filler' is not a valid property name.(…)`

Soo... `cke_filler`?

pjasiun:
Ha! camelCase is automatically change to dashes. How cool is it?!
```
let filler = document.createElement( 'br' );
filler.dataset.ckeFiller = true;
filler.outerHTML // <br data-cke-filler="true">
```

Reinmar:
1. Open http://tests.ckeditor.dev:1030/build/amd/tests/engine/treeview/manual/x-index
2. Execute the TC.
3. Click outside the editable element.

```
x-index:108 Uncaught TypeError: Cannot read property 'parent' of nulltreeView.on @ x-index:108fire @ emittermixin.js:280_handleSelectionChange @ selectionobserver.js:124domDocument.addEventListener @ selectionobserver.js:94```

**Comment by pjasiun** [comment]
> Open http://tests.ckeditor.dev:1030/build/amd/tests/engine/treeview/manual/x-index
> Execute the TC.
> Click outside the editable element.

It's a part of https://github.com/ckeditor/ckeditor5-engine/issues/405. `selectionChange` should not be fired when the selection is not inside the editor what is the reason the issue.

**Comment by Reinmar** [comment]
Could you verify that these tests fail in FF due to missing native event?

http://tests.ckeditor.dev:1030/build/amd/tests/engine/treeview/observer/selectionobserver

**Comment by pjasiun** [comment]
On my FF with `dom.select_events.enabled` flag these tests are green.

File: src/treeview/filler.js [hunk 2]
```diff
@@ -142,3 +144,33 @@ export function isBlockFiller( domNode, blockFiller ) {
...
 }
+
+/**
+ * Assign key observer which move cursor from the end of the inline filler to the begging of it when
+ * the left arrow is pressed, so the filler does not break navigation.
+ *
+ * @param {engine.treeView.TreeView} treeView TreeView instance we should inject quirks handling on.
+ */
+export function injectQuirksHandling( treeView ) {
+	treeView.on( 'keydown', jumpOverInlineFiller );
```
Reinmar: For the future reference – it's fine to have this listener attached through `on()` as this code will be used by the tree view itself. Destruction of the tree view will ensure that no events are fired on it.

Reinmar:
!!!!!!!!!!

Issue #385, opened 2016-04-28
Unify model and view test tools.


szymonkups:
As described here: https://github.com/ckeditor/ckeditor5-engine/pull/381#issuecomment-215351177, model and view tools needs some unification.

> 1. We also need two functions getData( treeView, rootName, options ) and setData( treeView, rootName, options ) which will work directly on the tree view, with that view's selection.
2. We need parse() and stringify() in the model. They should have similar capabilities to view's functions. getData() and setData() are defined there already.
3. Model getData()'s options.selection should be renamed to options.withSelection to conform to the naming in the view.

pjasiun:
`setData` needs also... `data` parameter :)

And since we agreed to have a default root name (https://github.com/ckeditor/ckeditor5-engine/issues/390), the name might merged into options:

`setData( treeView, data, options )`

pjasiun:
Because I used such methods in multiple test files I introduced simple version of `getData` and `setData` in: https://github.com/ckeditor/ckeditor5-engine/commit/7cbd17bea7d65245cf80c1fd12efcc17f7768440#diff-b47ebe54b5cb8d8272af93829f8d7632R848. Feel free to use or discard this code.

PR #411, opened 2016-05-13
Unification of model and view test utils.

szymonkups:
Unified test tools for model and view as described in #385.

### DIFF
--- a/tests/_utils-tests/model.js
+++ b/tests/_utils-tests/model.js
@@ -33,7 +33,7 @@ describe( 'model test utils', () => {
 			const stringifySpy = sandbox.spy( getData, '_stringify' );
 			root.appendChildren( new Element( 'b', null, [ 'btext' ] ) );
 
-			expect( getData( document ) ).to.equal( '<b>btext</b>' );
+			expect( getData( document, { withoutSelection: true } ) ).to.equal( '<b>btext</b>' );
 			sinon.assert.calledOnce( stringifySpy );
 			sinon.assert.calledWithExactly( stringifySpy, root );
 		} );
@@ -57,7 +57,7 @@ describe( 'model test utils', () => {
 
 			setData( document, data, options );
 
-			expect( getData( document ) ).to.equal( data );
+			expect( getData( document, { withoutSelection: true } ) ).to.equal( data );
 			sinon.assert.calledOnce( parseSpy );
 			const args = parseSpy.firstCall.args;
 			expect( args[ 0 ] ).to.equal( data );
@@ -66,7 +66,7 @@ describe( 'model test utils', () => {
 		it( 'should use parse method with selection', () => {
 			const parseSpy = sandbox.spy( setData, '_parse' );
 			const options = {};
-			const data = '[<b>btext</b>]';
+			const data = '<selection><b>btext</b></selection>';
 
 			setData( document, data, options );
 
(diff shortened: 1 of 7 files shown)


tests/_utils/model.js
```diff
@@ -9,60 +9,143 @@ import TreeWalker from '/ckeditor5/engine/treemodel/treewalker.js';
...
  * @param {Object} [options]
- * @param {Boolean} [options.selection] Whether to write the selection.
+ * @param {Boolean} [options.withSelection] Whether to write the selection.
+ * @param {Boolean} [options.rootName='main'] Name of the root from which data should be stringified. If not provided
+ * default `main` name will be used.
  * @returns {String} The stringified data.
  */
-export function getData( document, rootName, options ) {
+export function getData( document, options = {} ) {
+	const withSelection = !!options.withSelection;
```
pjasiun: I prefer make `withSelection` true by default. We could rename it to `withoutSelection`, since it is more natural to have options to be false by default.

Reinmar: +1 for `withoutSelection`. Makes sense to have this enabled by default.

tests/_utils/view.js
```diff
@@ -409,26 +470,37 @@ class ViewParser {
...
-	parse( data ) {
+	parse( data, rootElement = null ) {
 		const htmlProcessor = new HtmlDataProcessor();
 
 		// Convert HTML string to DOM.
 		const domRoot = htmlProcessor.toDom( data );
 
+		// If root element is provided - clear it.
+		// if ( rootElement ) {
+		// 	rootElement.removeChildren( rootElement.getChildCount() );
```
pjasiun: Should this code be commented out?

szymonkups: It shouldn't :) In model, removing items in root element is happening in `parse` method, so the same should happen here in the view. Now it is happening in `setData`. I totally forgotten about this. Thanks.

tests/_utils-tests/model.js
```diff
@@ -5,24 +5,94 @@
...
+	} );
+
 	describe( 'getData', () => {
+		it( 'should use stringify method', () => {
+			const stringifySpy = sandbox.spy( getData, '_stringify' );
+			root.appendChildren( new Element( 'b', null, [ 'btext' ] ) );
+
+			expect( getData( document ) ).to.equal( '<b>btext</b>' );
+			sinon.assert.calledOnce( stringifySpy );
+			sinon.assert.calledWithExactly( stringifySpy, root );
```
pjasiun: I do not like checking internals this way. I mean: would it be the bug if `getData` would not use `stringify` one day?

szymonkups: It's a dirty way to check it - I admit it. But this way I know that all tests that pass for `stringify` method are also ok for `getData` so there is no need for more tests for `getData`.

pjasiun: That is true, too. And, in fact, if one day `getData` would not use `stringify` we would have to write test for it anyway. So this check is fine for me.

tests/_utils/view.js
```diff
@@ -24,7 +24,62 @@ const TEXT_RANGE_START_TOKEN = '{';
...
+ * @param {String} data HTML-like string to write into TreeView.
+ * @param {Object} options
+ * @param {String} [rootName] Root name where parsed data will be stored. If not provided, default `main` name will be
+ * used.
+ */
+export function setData( treeView, data, options = {} ) {
+	const rootName = options.rootName || 'main';
+	const root = treeView.getRoot( rootName );
+	root.removeChildren( 0, root.getChildCount() );
+	const result = setData._parse( data, { rootElement: root } );
```
pjasiun: You can use destructuring:
`const { view, selection } = setData._parse( data, { rootElement: root } );`

tests/treemodel/composer/composer.js
```diff
@@ -57,7 +57,7 @@ describe( 'Composer', () => {
 				}
 			} );
 
-			expect( getData( document, 'main', { selection: true } ) )
+			expect( getData( document, { withSelection: true } ) )
```
pjasiun: It also should be `withoutSelection`.

PR #414, opened 2016-05-13
Fixed: MoveOperation thrown errors when it should not (fixes #413).

### LINKED ISSUE #413: Model: MoveOperation throws when it should not.
`MoveOperation#execute` should be changed to not throw when you move i.e. two nodes from `graveyard` `[ 0 ]` to `root` `[ 1, 1 ]`. Right now, `MoveOperation` does not count that these are different roots and thinks that this is wrong operation (because you try to move node `1` into itself).

```javascript
} else {
            if ( compareArrays( this.sourcePosition.getParentPath(), this.targetPosition.getParentPath() ) == 'PREFIX' ) {
```

to

```javascript
} else if ( this.sourcePosition.root == this.targetPosition.root ) {
            if ( compareArrays( this.sourcePosition.getParentPath(), this.targetPosition.getParentPath() ) == 'PREFIX' ) {
```

scofalik:
Fixes #413

### DIFF
--- a/src/treemodel/operation/moveoperation.js
+++ b/src/treemodel/operation/moveoperation.js
@@ -139,7 +139,7 @@ export default class MoveOperation extends Operation {
 			throw new CKEditorError(
 				'operation-move-range-into-itself: Trying to move a range of nodes to the inside of that range.'
 			);
-		} else {
+		} else if ( this.sourcePosition.root == this.targetPosition.root ) {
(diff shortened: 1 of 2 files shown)


Reinmar:
Why do we need such change? Test should verify it.

scofalik:
I've added a test. And guess what. I even checked if it fails without the fix.

PR #418, opened 2016-05-16
T/396 TreeModel serialization / deseralization.

jodator:
This PR introduces converting TreeModel deltas to JSON and from JSON objects defined in #396.


src/treemodel/delta/delta.js
```diff
@@ -118,6 +131,31 @@ export default class Delta {
...
+	 * Custom toJSON method to make deltas serializable.
+	 *
+	 * @returns {Object} Clone of this delta with added class name.
+	 */
+	toJSON() {
+		let json = clone( this );
+
+		json.__class = this.constructor.className;
+
+		json.batch = null;
```
Reinmar: `null` will end up in the JSON... shouldn't we delete this prop instead?

src/treemodel/delta/delta.js
```diff
@@ -118,6 +131,31 @@ export default class Delta {
...
 
 	/**
+	 * Custom toJSON method to make deltas serializable.
+	 *
+	 * @returns {Object} Clone of this delta with added class name.
+	 */
+	toJSON() {
+		let json = clone( this );
+
+		json.__class = this.constructor.className;
```
Reinmar: Can't we call it `__className` for consistency?

src/treemodel/delta/delta.js
```diff
@@ -118,6 +131,31 @@ export default class Delta {
...
+		json.batch = null;
+
+		return json;
+	}
+
+	/**
+	 * Delta class name. Used by {@link engine.treeModel.delta.Delta#toJSON} method for serialization and
+	 * {@link engine.treeModel.delta.Delta.fromJSON} during deserialization.
+	 *
+	 * @type {String}
```
Reinmar: Lacks `@readonly`.

src/treemodel/delta/delta.js
jodator: This can be changed, but should be changed in other places like operation where we have:
```js
json.parent = this.parent ? this.parent.name : null;
```
So this is consistent with others (setting null). So above should be changed as:

```js
if( this.parent ) {
    json.parent = this.parent.name;
} else {
    delete json.parent;
}
```

src/treemodel/delta/delta.js
jodator: I've used `__class` to be shorter but this is not so critical so can be `__className` also.

src/treemodel/delta/delta.js
```diff
@@ -128,4 +166,64 @@ export default class Delta {
 	static get _priority() {
 		return 0;
 	}
+
+	/**
+	 * Creates InsertDelta from deserialized object, ie. from parsed JSON string
```
Reinmar: s/ie./i.e./

Reinmar: s/$/.$/

src/treemodel/delta/delta.js
Reinmar: May lack `@static` (not sure whether JSDoc is able to read automatically that property is static).

src/treemodel/delta/delta.js
```diff
@@ -128,4 +166,64 @@ export default class Delta {
...
+operations[ MoveOperation.className ] = MoveOperation;
+operations[ NoOperation.className ] = NoOperation;
+operations[ Operation.className ] = Operation;
+operations[ ReinsertOperation.className ] = ReinsertOperation;
+operations[ RemoveOperation.className ] = RemoveOperation;
+operations[ RootAttributeOperation.className ] = RootAttributeOperation;
+
+const deserializers = new Map();
+
+export function registerDeserializer( className, constructor ) {
```
Reinmar: API docs are missing.

src/treemodel/delta/delta.js
```diff
@@ -128,4 +166,64 @@ export default class Delta {
...
+	if ( deserializers.has( className ) ) {
```
Reinmar: Is it really needed? Deserializers will not be frequently added, so there's a very little chance for a bug.

src/treemodel/delta/insertdelta.js
```diff
@@ -8,6 +8,7 @@
 import Delta from './delta.js';
 import RemoveDelta from './removedelta.js';
 import { register } from '../batch.js';
+import { registerDeserializer } from './delta.js';
```
Reinmar: You can merge that with the first import.

src/treemodel/delta/mergedelta.js
```diff
@@ -47,6 +48,10 @@ export default class MergeDelta extends Delta {
 	get _reverseDeltaClass() {
 		return SplitDelta;
 	}
+
+	static get className() {
```
Reinmar: As proposed recently, let's document all inherited props and methods with `/** @inheritDoc */`.

src/treemodel/delta/delta.js
jodator: Not sure which part.
Every delta needs deserializer (even custom made by other devs). Also this part mirrors behavior from `batch#register`.

jodator: The operations part is to not use if-else block.

Reinmar: I meant throwing error and validating anything here. It's not going to be an API which is used by hundreds thousand developers. There's very little chance that this API is actually used by anyone except us in (and few times only). Hence, validation, while being correct, makes no practical sense.

src/treemodel/node.js
```diff
@@ -188,6 +188,9 @@ export default class Node {
 		// Due to circular references we need to remove parent reference.
 		json.parent = this.parent ? this.parent.name : null;
 
+		// Serialize attributes as Map object is represented as "{}" when parsing to JSON.
+		json._attrs = [ ...json._attrs ];
```
Reinmar: First of all, `_attrs` is an invalid name, because we should not use such shorthand for a property name (there's a ticket to fix this across the API). Second, it should not be prefixed with underscore, because it's not private in that JSON (cause it's not an API, just data).

tl;dr – rename it to `attributes`.

src/treemodel/nodelist.js
```diff
@@ -50,14 +51,18 @@ class NodeListText extends Text {
...
+	 * Custom toJSON method to solve child-parent circular dependencies when serializing NodeListText to JSON string.
 	 *
 	 * @returns {Object} Clone of this object with the parent property replaced with its name.
 	 */
 	toJSON() {
 		const json = clone( this );
 
-		json.parent = json.parent ? this.parent.name : null;
+		// Due to circular references we need to remove parent reference.
+		json.parent = this.parent ? this.parent.name : null;
```
Reinmar: Do we need parent at all? I guess @scofalik has been using it for debugging. But now we have different tools, so it's unnecessary here.

src/treemodel/nodelist.js
```diff
@@ -293,7 +298,7 @@ export default class NodeList {
 
 		this._indexMap.splice( index, number );
 
-		for ( let i = index; i < this._indexMap.length ; i++ ) {
+		for ( let i = index; i < this._indexMap.length; i++ ) {
```
Reinmar: This change is not part of this ticket. Please report a separate ticket for it.






trollolololo ;P

src/treemodel/operation/operation.js
```diff
@@ -81,11 +81,36 @@ export default class Operation {
...
+	/**
+	 * Name of the operation class used for serialization.
+	 *
+	 * @type {String}
+	 */
+	static get className() {
+		return 'engine.treeModel.operation.Operation';
+	}
+
+	/*jshint unused: true*/
```
Reinmar: ?

src/treemodel/operation/operation.js
```diff
@@ -81,11 +81,36 @@ export default class Operation {
...
 	 */
 	toJSON() {
-		const json = clone( this );
+		const json = clone( this, true );
+
+		json.__class = this.constructor.className;
 
 		// Due to circular references we need to remove parent reference.
-		json.delta = this.delta ? '[engine.treeModel.Delta]' : null;
+		json.delta = this.delta ? `[${this.delta.constructor.className}]` : null;
```
Reinmar: Just like in https://github.com/ckeditor/ckeditor5-engine/pull/418/files#r63502600

src/treemodel/position.js
```diff
@@ -564,6 +564,34 @@ export default class Position {
 	}
 
 	/**
+	 * Creates Element object from deserilized object, ie. from parsed JSON string.
+	 *
+	 * @param {Object} json Deserialized JSON object.
+	 * @param {engine.treeModel.Document} doc Document on which this operation will be applied.
```
Reinmar: Should be `document`.

src/treemodel/delta/delta.js
jodator: I understand this, but how this compare to: https://github.com/ckeditor/ckeditor5-engine/blob/master/src/treemodel/batch.js#L118-L132 (which is in my understanding has the same frequency of usage as you described).

src/treemodel/nodelist.js
jodator: If this is not needed anymore I could drop this across the ticket as in delta case.

src/treemodel/rootelement.js
```diff
@@ -19,16 +19,35 @@ export default class RootElement extends Element {
 	 *
 	 * @param {engine.treeModel.Document} doc {@link engine.treeModel.Document} that is an owner of the root.
 	 * @param {String} name Node name.
+	 * @param {String} rootName Root name inside parent {@link engine.treeModel.Document}.
 	 */
-	constructor( doc, name ) {
+	constructor( doc, name, rootName ) {
```
Reinmar: @pjasiun, what do you think about making this constructor protected so only the document is able to create it (then we can be sure that names are synchronised).

tests/treemodel/node.js
```diff
@@ -178,4 +179,58 @@ describe( 'Node', () => {
 			} );
 		} );
 	} );
+
+	describe( 'toJSON', () => {
+		it( 'should serialize empty node', () => {
+			let node = new Element( 'one' );
```
Reinmar: Why do you test serialising element in node's tests?

tests/treemodel/node.js
```diff
@@ -178,4 +179,58 @@ describe( 'Node', () => {
...
+						{ _attrs: [], _children: { _indexMap: [], _nodes: [] }, name: 'one', parent: null },
+						{
+							_attrs: [], _children: {
+							_indexMap: [ 0, 0, 1, 2 ], _nodes: [
+								{ _attrs: [], text: 'ba', parent: 'two' },
+								{ _attrs: [], _children: { _indexMap: [], _nodes: [] }, name: 'img', parent: 'two' },
+								{ _attrs: [], text: 'r', parent: 'two' }
+							]
+						}, name: 'two', parent: null
+						},
```
Reinmar: This object isn't correctly formatted. Two `}` at the same level of indentation mean that something is wrong here.

tests/treemodel/_utils/utils.js
```diff
@@ -34,6 +34,10 @@ const utils = {
 		}
 
 		return txt;
+	},
+
+	jsonParseStringify( object ) {
```
Reinmar: Lacks docs.

src/treemodel/rootelement.js
```diff
@@ -19,16 +19,35 @@ export default class RootElement extends Element {
...
+		this.rootName = rootName;
+	}
+
+	/**
+	 * Custom toJSON method to solve child-parent circular dependencies.
+	 *
+	 * @method engine.treeModel.RootElement#toJSON
+	 * @returns {String} Name of this root inside {@link engine.treeModel.Document} that is an owner of this root.
+	 */
+	toJSON() {
```
Reinmar: Implementation of this method is confusing. Why is the name of the root enough in `toJSON()` to represent the root? Why there are no children or attributes of it (in fact – attributes are really missing)?

If some of these things are not necessary, this should be documented why. The fact that this implementation solve child-parent circ deps is a secondary aspect, and I'm not even sure whether it's true (because the same would need to be done for all elements, while in those cases it's handled by deleting `child.parent`).

Reinmar:
I left you couple of comments here and there. Some things need to be cleaned up or adjusted a bit. Some may need clarifications. I think that the general topic which isn't defined anywhere is what needs to be stringified to represent the document in JSON. Perhaps you could write somewhere (in some most-outer stringifier or parser) few words about that.

jodator:
Well there is no most-outer stringifier besides `JSON.parse( [ /* deltas */ ] );`.

Most-outer parser would be `Delta.fromJSON()`

src/treemodel/delta/delta.js
Reinmar: IMO that `register()` makes unnecessary validation as well. I'd be ok with removing it.

Reinmar:
> Most-outer parser would be Delta.fromJSON()

Maybe it'd make sense to introduce `Serializer` class? It'd contain the whole documentation plus two `fromJSON` and `toJSON` methods which would basically execute related methods on passed items. I know that so far it's very little responsibility, but for some reason it looks like a cleaner solution anyway. WDYT?

jodator:
Yep it might be as you described.

src/treemodel/delta/delta.js
```diff
@@ -128,4 +166,64 @@ export default class Delta {
...
+				{ name: json.__className }
+			);
+		}
+
+		let Constructor = deserializers.get( json.__className );
+
+		let delta = new Constructor();
+
+		if ( json.operations.length ) {
+			json.operations.forEach( ( operation ) => delta.addOperation( operations[ operation.__className ].fromJSON( operation, doc ) ) );
```
pjasiun: Use for-of instead of `forEach`. We try to avid `forEach` and use it only when it can not be simple replaced with for-of.

src/treemodel/delta/delta.js
jodator: `@static` is not needed.
selection_020

src/treemodel/delta/delta.js
jodator: If so then ok.

src/treemodel/delta/delta.js
```diff
@@ -128,4 +167,64 @@ export default class Delta {
...
+}
+
+const operations = {};
+operations[ AttributeOperation.className ] = AttributeOperation;
+operations[ InsertOperation.className ] = InsertOperation;
+operations[ MoveOperation.className ] = MoveOperation;
+operations[ NoOperation.className ] = NoOperation;
+operations[ Operation.className ] = Operation;
+operations[ ReinsertOperation.className ] = ReinsertOperation;
+operations[ RemoveOperation.className ] = RemoveOperation;
```
pjasiun: This is definitely not `Deltas` responsibility to handle all operations instanced. If there is no better place for it you should create `OperationFactory`.

pjasiun: Or `OperationDeserializer`.

src/treemodel/node.js
jodator: It is not a part of this ticket I think. While I can change this to:

```js
json._attributes = [ ...json._attrs ];
```
I also need to remove:
```js
delete json._attrs;
```

tl;dr - this name mirrors object's properties.

src/treemodel/operation/operation.js
jodator: Base class doesn't need document - but all operations need document parameter.

Without `jshint unused: true` linter will throw errors.

This is to make methods consistent in documentation so all (base and child classes) heve the same signature.

src/treemodel/node.js
pjasiun: I think it is the bigger question: should serialised object keep private properties? I am not sure about the answers. It is simpler to use private properties, and this is what we do, for instance in `clone` or `isEqual` methods. On the other hand, serialised object, send over the network seems to be a public API. If it will use private properties then send objects will change every time private properties change, what looks like a bad smell.

src/treemodel/operation/operation.js
pjasiun: It is not possible that linter throw error on the public method.

src/treemodel/rootelement.js
pjasiun: I do not find it necessarily. If someone creates root in the document using `createRoot`, this method will create `RootElement` with the same name. There is no other way to create root in the `Document`, so there is not way to desynchronise these names. Also If someone want to create a root element, for instance controller, temporarily, for some reason, then, why should we prevent it?

src/treemodel/rootelement.js
```diff
@@ -19,16 +19,35 @@ export default class RootElement extends Element {
...
+		return typeof this.rootName === 'symbol' ? '$$graveyard' : this.rootName;
```
pjasiun: Why you assume that every symbol is a `graveyard`? You could replace it with `this === this.document.graveyard`. But  I think that we should simplified it in general and use `$graveyard` string instead of using symbols. Also one `$` should be enough.

src/treemodel/nodelist.js
pjasiun: This code exists only because of objects thrown in errors, which can not have circular dependencies. The `parent` property can be removed.

tests/treemodel/node.js
```diff
@@ -178,4 +179,58 @@ describe( 'Node', () => {
...
+				name: 'one',
+				parent: null
+			} );
+		} );
+
+		it( 'should serialize node with children', () => {
+			expect( treeModelTestUtils.jsonParseStringify( root ) ).to.deep.equal( {
+				_attrs: [],
+				_children: {
+					_indexMap: [ 0, 1, 2 ],
```
pjasiun: This is again question about private properties in serialised object. `_indexMap` is the private and internal mechanism to improve performance. If we change it none tests should fail. Now we would need to change these tests every time we change private class logics.

tests/treemodel/_utils/utils.js
pjasiun: Also it should return two separate exports instead of the `utils` object.

src/treemodel/delta/delta.js
pjasiun: But we where thinking a lot about `OperationFactory`, which might be useful also to resolve circular dependencies between operations, so I think it might be useful for more then deserialisation`.

src/treemodel/delta/delta.js
pjasiun: Similarly this function should be moved to the `DeltaFactory`.

pjasiun:
To sum up, I think there are two most important changes:

 - making serialised object "public", so they should not keep private properties, anything what is not data. 
 - introducing `DeltaFactory` and `OperationFactory`.

src/treemodel/node.js
jodator: That's some point but we have this in more places:
```js
let node = new Element( 'one' );

expect( treeModelTestUtils.jsonParseStringify( node ) ).to.deep.equal( {
	_attrs: [],
	_children: {
		_indexMap: [],
		_nodes: []
	},
	name: 'one'
} );
```

So for consistency I've left `_attrs` but the question about all private properties remains. It would be nice of you (@reinmar, @pjasiun) to give me your definitive view on this: change them or not?

pjasiun: Change. :)

src/treemodel/operation/operation.js
jodator: The linter throws error on unused parameter `document`

pjasiun: So can't you remove this parameter here since it is unused?

jodator:
All reported issues should be fixed now.

src/treemodel/delta/attributedelta.js
```diff
@@ -69,6 +70,11 @@ export default class AttributeDelta extends Delta {
 		return AttributeDelta;
 	}
 
+	/** @inheritDoc */
```
pjasiun: For consistency it should be:
```
/**
 * @inheritDoc
 */
```

src/treemodel/delta/attributedelta.js
```diff
@@ -82,7 +88,12 @@ export default class AttributeDelta extends Delta {
  * @memberOf engine.treeModel.delta
  * @extends engine.treeModel.delta.Delta
  */
-export class RootAttributeDelta extends Delta {}
+export class RootAttributeDelta extends Delta {
+	/** @inheritDoc */
```
pjasiun: Same here.

src/treemodel/delta/deltafactory.js
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		return delta;
+	}
+}
+
+const deserializers = new Map();
+
+export function registerDeserializer( className, constructor ) {
+	deserializers.set( className, constructor );
```
pjasiun: Can we move this function to the `DeltaFactory` as a static method? Also it could take only one parameter `constructor`, class name can be taken from it. And I think that it could be generic deltas register, so `registerDeserializer` could be changed to simple `register`. I can imagine more usage then only deserialisation.

tl;dr: change:
```
registerDeserializer( Delta.className, Delta );
```
To:
```
DeltaFactory.register( Delta );
```

tests/treemodel/_utils/utils.js
```diff
@@ -7,34 +7,39 @@
...
+ * Returns tree structure as a simplified string. Elements are uppercase and characters are lowercase.
+ * Start and end of an element is marked the same way, by the element's name (in uppercase).
+ *
+ *		let element = new Element( 'div', [], [ 'abc', new Element( 'p', [], 'foo' ), 'xyz' ] );
+ *		treemodelUtils.getNodesAndText( element ); // abcPfooPxyz
+ *
+ * @param {engine.treeModel.Range} range Range to stringify.
+ * @returns {String} String representing element inner structure.
+ */
+export default function getNodesAndText( range ) {
```
pjasiun: Why this export is default? Is it more important for some reasons?

tests/treemodel/element.js
```diff
@@ -233,10 +234,102 @@ describe( 'Element', () => {
...
+
+			let serialized = jsonParseStringify( el );
+
+			let deserialized = Element.fromJSON( serialized );
+
+			expect( deserialized.parent ).to.be.null;
+			expect( deserialized.name ).to.equal( 'el' );
+			expect( deserialized._children.length ).to.equal( 1 );
+
+			expect( deserialized._children._indexMap ).to.deep.equal( [ 0 ] );
```
pjasiun: Still, tests can not use (write nor read) private properties.

tests/treemodel/nodelist.js
```diff
@@ -358,13 +359,125 @@ describe( 'NodeList', () => {
...
+			let nodeList = new NodeList( text );
+
+			let serialized = jsonParseStringify( nodeList );
+
+			let deserialized = NodeList.fromJSON( serialized );
+
+			expect( deserialized._indexMap ).to.deep.equal( nodeList._indexMap );
+			expect( deserialized._nodes.length ).to.equal( nodeList._nodes.length );
+			expect( deserialized._nodes[ 0 ].text ).to.equal( nodeList._nodes[ 0 ].text );
+			expect( [ ...deserialized._nodes[ 0 ]._attrs ] ).to.deep.equal( [ ...nodeList._nodes[ 0 ]._attrs ] );
```
pjasiun: Same here. Tests can not use private properties. Use class API instead.

pjasiun:
Much better :) This time only polishing (and private properties in tests) ;)

tests/treemodel/element.js
jodator: But other test are using this :/ That's thy it is here. Just to note this, I'll change this in my changes.

src/treemodel/delta/deltafactory.js
jodator: I think that is not possible with transpiled code (I'll check this again to be sure). There were some problems with that.

tests/treemodel/element.js
pjasiun: Then please create a follow-up ticket. The Guilty will fix it ;)

PR #421, opened 2016-05-16
Renamed TreeView.TreeView to TreeView.Document.

### LINKED ISSUE #253: Rename TreeView.TreeView to the TreeView.Document
When I was thinking about the name for the class which keep single root, with DOM element, view element and all its observers I realised that `TreeViewRoot` is a poor name, because it is too similar to the `ViewRoot`.

In fact this class contains more than the `View` element, it has also DOM elements, observers, writer and renderer.

Also, since there are multiple roots it is also definitely not a tree.

Because we have `Position`, `Range`, `Element`, `Node`, etc. in both model and view it would be nice to have also `Document` in both, because these classes are in fact corresponding.

oskarwrobel:
Fixes #253 

Related to:
- https://github.com/ckeditor/ckeditor5-delete/pull/5
- https://github.com/ckeditor/ckeditor5-enter/pull/9


src/treecontroller/editingcontroller.js
```diff
@@ -5,12 +5,12 @@
 
 'use strict';
 
-import TreeView from '../treeview/treeview.js';
+import TreeDocument from '../treeview/document.js';
```
pjasiun: Should be `ViewDocument`.

pjasiun:
`tests/treeview/treeview/` folder should be renamed to `tests/treeview/document/`.

pjasiun:
Now it looks much better :+1:

PR #425, opened 2016-05-18
Added direction option to TreeWalker.

### LINKED ISSUE #344: Implement TreeWalker options.direction

oskarwrobel:
Fixes #344


src/treemodel/treewalker.js
```diff
@@ -107,6 +135,14 @@ export default class TreeWalker {
...
 		 */
 		this._visitedParent = this.position.parent;
+
+		/**
+		 * Enum type that specifies relation between two positions.
+		 *
+		 * Possible values: `'AFTER'`, `'BEFORE'`, `'SAME'`.
+		 *
+		 * @typedef {String} PositionRelation
+		 */
```
oskarwrobel: ups.

src/treemodel/treewalker.js
```diff
@@ -44,6 +44,18 @@ export default class TreeWalker {
 			throw new CKEditorError( 'tree-walker-no-start-position: Neither boundaries nor starting position have been defined.' );
 		}
 
+		if ( options && options.direction && [ 'FORWARD', 'BACKWARD' ].indexOf( options.direction ) < 0 ) {
```
Reinmar: 1. Use default param for options: `constructor( options = {} )`. That will allow you to remove first condition fragment.
2. There's no point in checking `options.direction` existence and later its value. Check both: `options.direction == 'FORWARD' || options.direction == 'BACKWARD'`.
3. There would be no performance loss, but in general creating an array just to check something as simple as two values is a bad smell.

src/treemodel/treewalker.js
```diff
@@ -44,6 +44,18 @@ export default class TreeWalker {
...
+			/**
+			 * Unknown direction.
```
Reinmar: The API doc for an error is supposed to be more verbose than the in-code message. But if the message in the code is enough, there's no point to add an API doc for an error, so you can remove this.

src/treemodel/treewalker.js
```diff
@@ -117,13 +153,37 @@ export default class TreeWalker {
...
+			 */
+			throw new CKEditorError(
+				'tree-walker-unknown-direction: Only `BACKWARD` and `FORWARD` direction allowed.',
+				{ direction: this.direction }
+			);
+		}
+	}
+
+	/**
+	 * Makes a step forward in tree model. Moves the {@link #position} to the next position and returns the encountered value.
```
Reinmar: s/tree //. We're slowly removing the "tree" adjective from the code base.

src/treemodel/treewalker.js
```diff
@@ -179,16 +239,89 @@ export default class TreeWalker {
...
 				return formatReturnValue( 'ELEMENT_END', parent, previousPosition, position );
 			}
 		}
 	}
+
+	/**
+	 * Makes a step backward in tree model. Moves the {@link #position} to the previous position and returns the encountered value.
+	 *
+	 * @private
+	 * @returns {Boolean} return.done True if iterator is done.
```
Reinmar: Doubled `@returns`.

src/treemodel/treewalker.js
```diff
@@ -230,3 +363,11 @@ function formatReturnValue( type, item, previousPosition, nextPosition, length )
  * @property {Number} [length] Length of the item. For `'ELEMENT_START'` and `'CHARACTER'` it is 1. For `'TEXT'` it is
  * the length of the text. For `'ELEMENT_END'` it is undefined.
  */
+
+/**
+ * Tree walking directions.
+ *
+ * Possible values: `'FORWARD'`, `'BACKWARD'`.
+ *
+ * @typedef {String} core.treeModel.TreeWalkerDirection
```
Reinmar: Instead of `String` you can specify the real values: `{'FORWARD'|'BACKWARD'}`.

tests/treemodel/treewalker.js
```diff
@@ -16,11 +16,9 @@ import Range from '/ckeditor5/engine/treemodel/range.js';
 import CKEditorError from '/ckeditor5/utils/ckeditorerror.js';
 
 describe( 'TreeWalker', () => {
-	let expectedSingle, expected, expectedIgnoreEnd, expectedSingleIgnoreEnd;
-	let expectedShallow, expectedCroppedStart, expectedCroppedEnd;
-
+	let expected;
```
pjasiun: I do not like this variable here. Each tests block use it and if one block forget to change the value of the `expected` it will use expected values from the previous block. Each block should define `expected` so if one forget about it, it will be undefined.

src/treemodel/treewalker.js
```diff
@@ -75,6 +95,14 @@ export default class TreeWalker {
 			Position.createFromPosition( options.boundaries.start );
 
 		/**
+		 * Walking direction. Defaults `FORWARD`.
+		 *
+		 * @member engine.treeModel.TreeWalker#direction
+		 * @type {String} core.treeModel.TreeWalkerDirection
+		 */
+		this.direction = !options.direction ? 'FORWARD' : options.direction;
```
Reinmar: `options.direction || 'FORWARD'`

pjasiun:
Missing documentation for the new option.

pjasiun:
But I do like how tests look now 👍

Reinmar:
One general note. This ticket was reported because I missed the option in this code: https://github.com/ckeditor/ckeditor5-engine/blob/master/src/treemodel/composer/modifyselection.js#L33-L34. I think that it'd be good to improve that function as a part of this ticket, just to confirm that everything works fine.

src/treemodel/treewalker.js
pjasiun: In fact we could use destucturing here:
```
constructor( { 
	boundaries,
	startPosition,
	singleCharacters=false,
	shallow=false,
	ignoreElementEnd=false,
	direction == 'FORWARD'
 } = {} )
```
Then use variables directly in code (`direction` instead of `options.direction` ).

pjasiun:
I also miss docs updates for `previousPosition` and `nextPosition` when you iterate backward, because not everything about these options is true in such case.

Reinmar:
BTW. as far as I remember, we've been considering renaming `previous|nextPosition` to `positionBefore|After`. Before would always be on the left and after on the right. That would be consistent with `Position.nodeBefore|After` and that's what I've been expecting to see when using the API for the first time.

pjasiun:
> BTW. as far as I remember, we've been considering renaming previous|nextPosition to positionBefore|After. Before would always be on the left and after on the right. That would be consistent with Position.nodeBefore|After and that's what I've been expecting to see when using the API for the first time.

Yep, I told Oskar about this. I am still against such change because `positionBefore|After` and `previous|nextPosition` return different values for `ELEMENT_START` and `ELEMENT_END`. In fact `previous|nextPosition` are not before and after in such cases. What you are talking about is true only for `TEXT`. Anyway, this is a different ticket, I do not think we should merge them here.

src/treemodel/composer/modifyselection.js
```diff
@@ -27,20 +27,23 @@ export default function modifySelection( selection, options = {} ) {
...
+	let next = walker.next();
 
 	// 1. Nothing to do here.
-	if ( !next ) {
+	if ( next.done ) {
 		return;
 	}
 
+	// Replace TreeWalker step wrapper by clean step value.
+	next = next.value;
```
pjasiun: WAT? Why not `let value = next.value;`?

src/treemodel/treewalker.js
```diff
@@ -23,6 +23,7 @@ export default class TreeWalker {
 	 * @param {Object} options Object with configuration.
 	 * @param {engine.treeModel.Range} [options.boundaries] Range to define boundaries of the iterator.
 	 * @param {engine.treeModel.Position} [options.startPosition] Starting position.
+	 * @param {'FORWARD'|'BACKWARD'} [options.direction] Walking direction.
```
pjasiun: Default value is not documented.

src/treemodel/treewalker.js
```diff
@@ -34,8 +35,17 @@ export default class TreeWalker {
...
-		if ( !options || ( !options.boundaries && !options.startPosition ) ) {
+	constructor(
+		{
+			boundaries,
+			startPosition,
+			direction = 'FORWARD',
+			singleCharacters = false,
+			shallow = false,
+			ignoreElementEnd = false,
+		} = {}
```
pjasiun: Isn't it cool? Strange, creepy, but cool ;)

src/treemodel/treewalker.js
```diff
@@ -70,25 +95,35 @@ export default class TreeWalker {
...
+			this.position = Position.createFromPosition( boundaries[ direction == 'BACKWARD' ? 'end' : 'start' ] );
+		}
+
+		/**
+		 * Walking direction. Defaults `FORWARD`.
+		 *
+		 * @member engine.treeModel.TreeWalker#direction
+		 * @type {'BACKWARD'|'FORWARD'} core.treeModel.TreeWalkerDirection
+		 */
+		this.direction = direction || 'FORWARD';
```
pjasiun: Since we use default parameter, you do not need ` || 'FORWARD';` here.

src/treemodel/treewalker.js
```diff
@@ -54,7 +71,15 @@ export default class TreeWalker {
 		 *
 		 * @member {engine.treeModel.Range} engine.treeModel.TreeWalker#boundaries
 		 */
-		this.boundaries = options.boundaries || null;
+		this.boundaries = boundaries || null;
```
pjasiun: We can move this `|| null` to the default parameter value now.

src/treemodel/treewalker.js
oskarwrobel: Right :) I didn't noticed it.

src/treemodel/composer/modifyselection.js
oskarwrobel: It was today morning (before coffee) idea :) https://github.com/ckeditor/ckeditor5-engine/pull/425/commits/d7c1f7d1cebadb56b8d8826f376563ed27c38924

pjasiun: I just realised that our code is, in fact, static typed. And, even more spuriously realised, that it is what I like in it. `next` and `value` are different types and overwriting one with another feel like a bad small, because of this static typing.

src/treemodel/treewalker.js
```diff
@@ -214,19 +341,32 @@ function formatReturnValue( type, item, previousPosition, nextPosition, length )
...
- * position inside the element. For all other types it is the position before the item. Note that it is more
- * efficient to use this position then calculate the position before the node using
- * {@link engine.treeModel.Position.createBefore}. It is also more efficient to get the position after node by shifting
- * `previousPosition` by `length`, using {@link engine.treeModel.Position#getShiftedBy}, then calculate the position using
- * {@link engine.treeModel.Position.createAfter}.
- * @property {engine.treeModel.Position} nextPosition Next position of the iterator. For `'ELEMENT_START'` it is the first
- * position inside the element. For all other types it is the position after the item.
+ * @property {engine.treeModel.Position} previousPosition Previous position of the iterator.
+ * 		Forward iteration:
+ * 			For `'ELEMENT_END'` it is the last position inside the element. For all other types it is the position
```
pjasiun: We use tab indent in docs only for code samples. Use bullet list here insted (see: https://github.com/ckeditor/ckeditor5-engine/blob/master/src/treecontroller/model-converter-builder.js#L169-L171)

pjasiun:
For me it's fine now. @Reinmar: could you check changes in `/composer/modifyselection.js`?

Reinmar:
Will do.

@oskarwrobel: could you merge master to this branch, because there are conflicts after @pjasiun's changes?

oskarwrobel:
@Reinmar done.

Reinmar:
There's one thing that doesn't make sense – checking direction in `next()`, because that's already done in  the constructor. I'll fix this.

oskarwrobel:
@Reinmar I don't know every possible use case of this class. I added this assertion in case of someone changes direction after initialization.
```
let walker = new TreeWalker();
walker.direction = 'invalid';
walker.next();
```
If such a situation is unexpected I can mark `TreeWalker#direction` property as `@readonly` in docs.

oskarwrobel:
> If such a situation is unexpected I can mark TreeWalker#direction property as @readonly in docs.

@Reinmar I see that you already did it here https://github.com/ckeditor/ckeditor5-engine/pull/425/commits/2c70a1c1eebff2e013163928227bb425302f917b :) Thanks.

Issue #288, opened 2016-03-18
Implement Editing controller


Reinmar:
Strongly connected to https://github.com/ckeditor/ckeditor5/issues/116.

Reinmar:
In #301 I introduced base classes, but the real implementation is yet to be defined.

pjasiun:
I pushed data and editing controllers prototypes to the https://github.com/ckeditor/ckeditor5-engine/tree/t/288 branch (https://github.com/ckeditor/ckeditor5-engine/commit/43948aba7b330bedcb5b1fcfe2cb927929150a70).

pjasiun:
Should solve also https://github.com/ckeditor/ckeditor5-engine/issues/292 "Define a list of observers which are always registered by the EditingController".

pjasiun:
Also related: https://github.com/ckeditor/ckeditor5-engine/issues/369 View.Document as an entry point for the view editing state

pjasiun:
Should also fix: https://github.com/ckeditor/ckeditor5-engine/issues/402 Siblings should be merged when inline filler is removed

pjasiun:
Let's split it into two tickets.

PR #442, opened 2016-05-25
t/288: Implement Editing controller

pjasiun:
Fixes: #288.

### DIFF
--- a/src/conversion/view-selection-to-model-converters.js
+++ b/src/conversion/view-selection-to-model-converters.js
@@ -23,7 +23,7 @@
  *		view.document.on( 'selectionChange', convertSelectionChange( model, mapper ) );
  *
  * @function engine.conversion.viewSelectionToModel.convertSelectionChange
- * @param {engine.model.Document} model Model document on which selection should be updated.
+ * @param {engine.model.Document} model Document model on which selection should be updated.
  * @param {engine.conversion.Mapper} mapper Conversion mapper.
  * @returns {Function} {@link engine.view.Document#selectionChange} callback function.
  */
--- a/src/conversion/view-selection-to-model-converters.js
+++ b/src/conversion/view-selection-to-model-converters.js
@@ -1,0 +1,43 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/**
+ * Contains {@link engine.view.Selection view selection} to {@link engine.model.Selection model selection} conversion
+ * helper.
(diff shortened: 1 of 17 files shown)


src/conversion/view-selection-to-model-converters.js
```diff
@@ -0,0 +1,43 @@
...
+ *		view.document.on( 'selectionChange', convertSelectionChange( model, mapper ) );
+ *
+ * @function engine.conversion.viewSelectionToModel.convertSelectionChange
+ * @param {engine.model.Document} model Model document on which selection should be updated.
+ * @param {engine.conversion.Mapper} mapper Conversion mapper.
+ * @returns {Function} {@link engine.view.Document#selectionChange} callback function.
+ */
+export function convertSelectionChange( model, mapper ) {
+	return ( evt, data ) => {
+		model.enqueueChanges( () => {
```
Reinmar: Does selection change need to enenequeueue changes?

src/editingcontroller.js
```diff
@@ -6,12 +6,157 @@
...
+ *
+ * @memberOf engine
+ */
 export default class EditingController {
-	constructor( modelDocument ) {
-		this.model = modelDocument;
+	/**
+	 * Creates editing controller instance.
+	 *
+	 * @param {engine.model.Document} model Model document.
```
Reinmar: IMO we should call it "Document model". I know we can justify why it's model's document, but document model is a known term and reflect the reality pretty well too.

src/editingcontroller.js
```diff
@@ -6,12 +6,157 @@
...
+	 *
+	 *		editing.createRoot( 'body' );
+	 *		editing.view.attachDomRoot( iframe.contentDocument.body );
+	 *
+	 * @param {Element|String} domRoot DOM root element or the name of view root element if the DOM element will be
+	 * attached later.
+	 * @param {String} [name='main'] Root name.
+	 * @returns {engine.view.ContainerElement} View root element.
+	 */
+	createRoot( domRoot, name = 'main' ) {
```
Reinmar: `createViewRoot()`?

Reinmar:
I left two comments in the code. Other than that, it's ready to be merged.

src/editingcontroller.js
Reinmar: We agreed that it's obvious that editing controller can only modify the view stuff.

src/conversion/view-selection-to-model-converters.js
pjasiun: Yep. Because changed selection have to be rendered, rendering run on `changesDone`, and without `enqueueChanges` you would not have a `changesDone` event.

PR #447, opened 2016-05-27
t/369: View.Document as an entry point for the view editing state

### LINKED ISSUE #369: View.Document as an entry point for the view editing state
Note: `view.Document` is renamed `TreeView.TreeView`: https://github.com/ckeditor/ckeditor5-engine/issues/253

Now we have `Editable` class in the core repo: https://github.com/ckeditor/ckeditor5/blob/master/src/editable.js
which keeps the information if editable `isFocused` and `isEditable`. I already met the problem that I need these informations in the `Renderer`. To avoid code duplication we should keep all informations about the view in the `View.Document` and its properties.

We need to: introduce `EditableElement` which will be child of the `view.Element` and takes the responsibility of current `Editable`. One of the reasons why we needed `Editable` is because it might be used before DOM element is created, in the iframe case. The "View" abstraction match perfectly in this case, because the `EditableElement` might be created before the DOM element and be bind later.

Then, `view.Document` can keep collections of root editables and all editables, which include also nested editabled.

pjasiun:
Fixes #369.

### DIFF
--- a/src/view/observer/focusobserver.js
+++ b/src/view/observer/focusobserver.js
@@ -22,8 +22,20 @@ export default class FocusObserver extends DomEventObserver {
 		this.domEventType = [ 'focus', 'blur' ];
 	}
 
-	onDomEvent( domEvt ) {
-		this.fire( domEvt.type, domEvt );
+	onDomEvent( domEvent ) {
+		const target = this.document.domConverter.getCorrespondingViewElement( domEvent.target );
+
+		for ( let root of this.document.roots.values() ) {
+			if ( target === root ) {
+				if ( domEvent.type == 'focus' ) {
+					root.isFocused = true;
+				} else {
+					root.isFocused = false;
+				}
+			}
+		}
+
+		this.fire( domEvent.type, domEvent );
 	}
 }
 
--- a/src/view/observer/focusobserver.js
+++ b/src/view/observer/focusobserver.js
@@ -9,6 +9,8 @@ import DomEventObserver from './domeventobserver.js';
 
 /**
  * {@link engine.view.Document#focus Focus} and {@link engine.view.Document#blur blur} events observer.
+ * Focus observer handle also {@link engine.view.RootEditableElement#isFocused isFocused} property of the
+ * {@link engine.view.RootEditableElement root elements}.
  *
  * Note that this observer is attached by the {@link engine.EditingController} and is available by default.
  *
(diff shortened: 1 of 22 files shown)


src/view/observer/focusobserver.js
```diff
@@ -22,8 +24,21 @@ export default class FocusObserver extends DomEventObserver {
 		this.domEventType = [ 'focus', 'blur' ];
 	}
 
-	onDomEvent( domEvt ) {
-		this.fire( domEvt.type, domEvt );
+	onDomEvent( domEvent ) {
+		const target = this.document.domConverter.getCorrespondingViewElement( domEvent.target );
+
+		// Update `isFocus` property of root elements.
+		for ( let root of this.document.roots.values() ) {
```
Reinmar: Shouldn't this behavior be implemented as a listener to this event?

pjasiun: Well... it could. But then we would have to take case of unbinding this event and care more about them. Also it is fine for observer to extend view. For instance `SelectionObserver` flush mutations. `MutationObserver` mark elements to be rerenders. The idea of observers was more like the strategy pattern, adding some logic to the view layer, then only firing events.

src/view/rooteditableelement.js
```diff
@@ -0,0 +1,65 @@
...
+		 * Whether the editable is focused.
+		 *
+		 * This property is updated by the {@link engine.view.obsever.FocusObserver}.
+		 * If the {@link engine.view.obsever.FocusObserver} is disabled this property will not change.
+		 *
+		 * @readonly
+		 * @observable
+		 * @member {Boolean} engine.view.RootEditableElement#isFocused
+		 */
+		this.set( 'isFocused', false );
```
Reinmar: Reported a followup for this https://github.com/ckeditor/ckeditor5-engine/issues/448.

src/view/observer/focusobserver.js
Reinmar: I'm not saying that the code would need to be moved to a different class. I'd just add it as a listener in the constructor of `FocusObserver`. That will give us the separation of concerns – that bit fires the event, that uses it.

src/view/rooteditableelement.js
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	/**
+	 * Gets {@link engine.view.Document} reference.
+	 *
+	 * @returns {engine.view.Document|null} View Document of the node or null.
+	 */
+	getDocument() {
```
Reinmar: Clearly, this should be a getter, but it cannot be according to our current rules, because in the `Node` class this implementation is too complex.

I think that we made a serious mistake defining those rules in that form, so I reopened https://github.com/ckeditor/ckeditor5-design/issues/88#issuecomment-222121223

PR #476, opened 2016-06-08
Added parameter validation for getData().

### LINKED ISSUE #456: View utils: getData() returns an empty string when a wrong param passed
I executed `getData()` with model by mistake and I got an empty string back. I'd expect an error of any kind.

oskarwrobel:
Fixes #456

### DIFF
--- a/tests/_utils-tests/model.js
+++ b/tests/_utils-tests/model.js
@@ -51,7 +51,7 @@ describe( 'model test utils', () => {
 		it( 'should throw an error when passing invalid document', () => {
 			expect( () => {
 				getData( { invalid: 'document' } );
-			} ).to.throw( TypeError, /Document needs to be an instance of engine.model.Document/ );
+			} ).to.throw( TypeError, /Document needs to be an instance of engine\.model\.Document\./ );
 		} );
 	} );
 
@@ -84,8 +84,8 @@ describe( 'model test utils', () => {
 
 		it( 'should throw an error when passing invalid document', () => {
 			expect( () => {
-				getData( { invalid: 'document' } );
-			} ).to.throw( TypeError, /Document needs to be an instance of engine.model.Document/ );
+				setData( { invalid: 'document' } );
+			} ).to.throw( TypeError, /Document needs to be an instance of engine\.model\.Document\./ );
 		} );
 	} );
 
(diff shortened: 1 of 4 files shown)


tests/_utils-tests/model.js
```diff
@@ -12,6 +12,7 @@ import Element from '/ckeditor5/engine/model/element.js';
 import Text from '/ckeditor5/engine/model/text.js';
 import Range from '/ckeditor5/engine/model/range.js';
 import Position from '/ckeditor5/engine/model/position.js';
+import CKEditorError from '/ckeditor5/utils/ckeditorerror.js';
```
szymonkups: We should use regular `Error` instances to throw inside testing tools (look at other `throws` inside this file). `CKEditorError` is used to throw errors inside the project.

Reinmar: I remember that talked about this in the past. Do you recall whether there was some clear reason for not using `CKEditorError`? I think that it might be something that it simply doesn't make sense to use `CKEditorError` because it was meant to be minified in a special way.

szymonkups: >  it was meant to be minified in a special way.

I think this was the main reason. Using `CKEditorError` inside tests will probably do no harm as long as these errors will not be confused with errors from source code (and not included in minification). I wanted to make this fix to be consistent with the rest of the file.

tests/_utils-tests/model.js
```diff
@@ -75,6 +81,12 @@ describe( 'model test utils', () => {
 			const args = parseSpy.firstCall.args;
 			expect( args[ 0 ] ).to.equal( data );
 		} );
+
+		it( 'should throw an error when passing invalid document', () => {
+			expect( () => {
+				getData( { invalid: 'document' } );
```
oskarwrobel: My bad. Should be `setData`

tests/_utils-tests/view.js
```diff
@@ -101,6 +107,12 @@ describe( 'view test utils', () => {
 				expect( args[ 1 ] ).to.be.an( 'object' );
 				expect( args[ 1 ].rootElement ).to.equal( viewDocument.getRoot() );
 			} );
+
+			it( 'should throw an error when passing invalid document', () => {
+				expect( () => {
+					getData( { invalid: 'document' } );
```
oskarwrobel: The same. Should be `setData`.

PR #477, opened 2016-06-09
Changed parameters order of model.document.createRoot() method.

### LINKED ISSUE #390: Default root name
We need a default name of the root, at least for the tests consistent. @Reinmar suggested `main` what is fine for me. I think we could go further and set it as a default parameter of the `(model|view).document.(get|create)Root`. Then a single root editor creator would not need to care about the root name.

oskarwrobel:
Fixes #390 

Related:
https://github.com/ckeditor/ckeditor5-typing/pull/11
https://github.com/ckeditor/ckeditor5/pull/200

### DIFF
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -181,8 +181,8 @@ export default class Document {
 	/**
 	 * Creates a new top-level root.
 	 *
-	 * @param {String} [rootName='main'] Unique root name.
 	 * @param {String} [elementName='$root'] Element name. Defaults to `'$root'` which also have
+	 * @param {String} [rootName='main'] Unique root name.
 	 * some basic schema defined (`$block`s are allowed inside the `$root`). Make sure to define a proper
 	 * schema if you use a different name.
 	 * @returns {engine.model.RootElement} Created root.
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -103,7 +103,7 @@ export default class Document {
 		} );
 
 		// Graveyard tree root. Document always have a graveyard root, which stores removed nodes.
-		this.createRoot( graveyardName );
+		this.createRoot( '$root', graveyardName );
 
 		/**
 		 * Document's history.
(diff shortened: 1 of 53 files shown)


tests/conversion/advanced-converters.js
```diff
@@ -45,7 +45,7 @@ let modelDoc, modelRoot, viewRoot, mapper, writer, modelDispatcher, viewDispatch
 
 beforeEach( () => {
 	modelDoc = new ModelDocument();
-	modelRoot = modelDoc.createRoot( 'root' );
+	modelRoot = modelDoc.createRoot( '$root', 'root' );
```
pjasiun: The "root" name was introduced before we had the default, `main` root name. Now it should be removed.

tests/datacontroller.js
```diff
@@ -23,8 +23,8 @@ describe( 'DataController', () => {
 
 	beforeEach( () => {
 		modelDocument = new ModelDocument();
-		modelDocument.createRoot( 'main' );
-		modelDocument.createRoot( 'title' );
+		modelDocument.createRoot( '$root', 'main' );
```
szymonkups: Can be changed to `modelDocument.createRoot()`.

tests/conversion/advanced-converters.js
szymonkups: In tests that use `doc.createRoot( '$root', 'root' );` we should use simpler version `doc.createRoot()` and eventually fix places where `root` name is used to use `main`. Is that correct @pjasiun?

src/model/document.js
```diff
@@ -181,13 +181,13 @@ export default class Document {
 	/**
 	 * Creates a new top-level root.
 	 *
-	 * @param {String} [rootName='main'] Unique root name.
 	 * @param {String} [elementName='$root'] Element name. Defaults to `'$root'` which also have
+	 * @param {String} [rootName='main'] Unique root name.
 	 * some basic schema defined (`$block`s are allowed inside the `$root`). Make sure to define a proper
```
szymonkups: This text should also be moved.

tests/conversion/advanced-converters.js
pjasiun: Yep. Exactly.

PR #487, opened 2016-06-14
t/400: Resolve infinite loop because of selection change

pjasiun:
Fixes: #400.

### DIFF
--- a/src/view/observer/selectionobserver.js
+++ b/src/view/observer/selectionobserver.js
@@ -138,7 +138,7 @@ export default class SelectionObserver extends Observer {
 		}
 
 		// Ensure we are no in the infinite loop (#400).
-		if ( this._isInfiteLoop( newViewSelection ) ) {
+		if ( this._isInfiniteLoop( newViewSelection ) ) {
 			/**
 			 * Selection infinite loop. Most probably you try to put the selection in the position which is not allowed
 			 * by the browser and browser fix it automatically what causes selectionChange event, re-rendering wrong
@@ -168,7 +168,7 @@ export default class SelectionObserver extends Observer {
 	 * @param {engine.view.Selection} newSelection DOM selection converted to view.
 	 * @returns {Boolean} True is the same selection repeat more then 10 times.
 	 */
-	_isInfiteLoop( newSelection ) {
+	_isInfiniteLoop( newSelection ) {
 		// If the position is the same a the last one or the last but one we increment the counter.
 		if ( this._lastSelection && this._lastButOneSelection &&
 			( newSelection.isEqual( this._lastSelection ) || newSelection.isEqual( this._lastButOneSelection ) ) ) {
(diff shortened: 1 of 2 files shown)


tests/view/observer/selectionobserver.js
```diff
@@ -124,6 +130,71 @@ describe( 'SelectionObserver', () => {
...
+			done();
+
+			// Prevent "done() called multiple times" error.
+			done = () => {};
+		} );
+
+		changeDomSelection();
+	} );
+
+	it( 'should not be thread as an infinite loop if the position is different', ( done ) => {
```
scofalik: "be treated"

tests/view/observer/selectionobserver.js
```diff
@@ -124,6 +130,71 @@ describe( 'SelectionObserver', () => {
...
+				setTimeout( () => changeCollapsedDomSelection( counter ) );
+			} else {
+				done();
+			}
+		} );
+
+		changeCollapsedDomSelection( counter );
+	} );
+
+	it( 'should not be thread as an infinite loop if it is less then 3 times', ( done ) => {
```
scofalik: "be treated"

scofalik:
The solution looks a bit ugly but probably could not be better, especially if we want to have it soon. Still better than setting timeouts. For me it's okay, but I'll also wait for @reinmar to look at it as it is an important fix.

tests/view/observer/selectionobserver.js
Reinmar: And you don't need `()`.

src/view/observer/selectionobserver.js
```diff
@@ -8,6 +8,8 @@
 import Observer from './observer.js';
 import MutationObserver from './mutationobserver.js';
 
+import log from '../../../utils/log.js';
```
Reinmar: I believe this is a first usage of this function :D.

tests/view/observer/selectionobserver.js
scofalik: Haven't we agreed to use `()` always, for better readability?

Reinmar: Which reminds me that we haven't clarified this in https://github.com/ckeditor/ckeditor5-design/wiki/Code-Style. Let's pretend we haven't seen this.

scofalik:
I'd also consider two things:
- getting rid of `_lastButOneSelection` and `_lastSelection` and just base on counter
- putting `_lastButOneSelection` and `_lastSelection` and probably `isInfiniteLoop` out of the class as they give really bad smell :P

pjasiun:
> getting rid of _lastButOneSelection and _lastSelection and just base on counter

The problem is how to reset counter. I do not see any good event we could listen on to reset it.

> putting _lastButOneSelection and _lastSelection and probably isInfiniteLoop out of the class as they give really bad smell :P

Well, such hacks are always bad smells. All of these properties are private and used only in one function. They can not be local and I do not feel that making them global would make them smell better.

Reinmar:
I added more comments and fixed some existing ones.

This is a hack of course, but it's quite elegant. It's enclosed in one mechanism, fully private and the solution is easy to grasp. So I'm ok with it.

PR #492, opened 2016-06-16
View TreeWalker 

### LINKED ISSUE #255: Introduce TreeView.TreeWalker
We need a tree walker which will be able to iterate over the view elements. It could be a port of the `TreeModel.TreeWalker`.

oskarwrobel:
TreeWakler for view.
Fixes #255 .

### DIFF
--- a/src/view/position.js
+++ b/src/view/position.js
@@ -194,6 +194,11 @@ export default class Position {
 	 * @returns {engine.view.Position}
 	 */
 	static createAfter( node ) {
+		// {@link engine.view.TextProxy} is not a instance of {@link engine.view.Node} so we need do handle it in specific way.
+		if ( node instanceof TextProxy ) {
+			return new Position( node._textNode, node._index + 1 );
+		}
+
 		if ( !node.parent ) {
 			/**
 			 * You can not make a position after a root.
@@ -202,11 +207,6 @@ export default class Position {
 			 * @param {engine.view.Node} root
 			 */
 			throw new CKEditorError( 'position-after-root: You can not make position after root.', { root: node } );
-		}
-
-		// {@link engine.view.TextProxy} is not a instance of {@link engine.view.Node} so we need do handle it in specific way.
-		if ( node instanceof TextProxy ) {
-			return new Position( node._textNode, node._index + 1 );
 		}
 
 		return new Position( node.parent, node.getIndex() + 1 );
@@ -219,6 +219,11 @@ export default class Position {
 	 * @returns {engine.view.Position}
 	 */
 	static createBefore( node ) {
+		// {@link engine.view.TextProxy} is not a instance of {@link engine.view.Node} so we need do handle it in specific way.
+		if ( node instanceof TextProxy ) {
+			return new Position( node._textNode, node._index );
+		}
+
 		if ( !node.parent ) {
 			/**
 			 * You cannot make a position before a root.
@@ -227,11 +232,6 @@ export default class Position {
 			 * @param {engine.view.Node} root
 			 */
 			throw new CKEditorError( 'position-before-root: You can not make position before root.', { root: node } );
-		}
-
-		// {@link engine.view.TextProxy} is not a instance of {@link engine.view.Node} so we need do handle it in specific way.
-		if ( node instanceof TextProxy ) {
-			return new Position( node._textNode, node._index );
 		}
 
 		return new Position( node.parent, node.getIndex() );
--- a/src/view/position.js
+++ b/src/view/position.js
@@ -190,7 +190,7 @@ export default class Position {
 	/**
 	 * Creates a new position after the given node.
 	 *
-	 * @param {engine.view.Node|engine.view.TextProxy} node Node after which the position should be located.
+	 * @param {engine.view.Node|engine.view.TextProxy} node Node or text proxy after which the position should be located.
 	 * @returns {engine.view.Position}
 	 */
 	static createAfter( node ) {
@@ -215,7 +215,7 @@ export default class Position {
 	/**
 	 * Creates a new position before the given node.
 	 *
-	 * @param {engine.view.Node|engine.view.TextProxy} node Node before which the position should be located.
+	 * @param {engine.view.Node|engine.view.TextProxy} node Node or text proxy before which the position should be located.
 	 * @returns {engine.view.Position}
 	 */
 	static createBefore( node ) {
--- a/src/view/position.js
+++ b/src/view/position.js
@@ -196,7 +196,7 @@ export default class Position {
 	static createAfter( node ) {
 		// {@link engine.view.TextProxy} is not a instance of {@link engine.view.Node} so we need do handle it in specific way.
 		if ( node instanceof TextProxy ) {
-			return new Position( node._textNode, node._index + 1 );
+			return new Position( node._textNode, node._index + node._data.length );
 		}
 
 		if ( !node.parent ) {
--- a/src/view/position.js
+++ b/src/view/position.js
@@ -196,7 +196,7 @@ export default class Position {
 	static createAfter( node ) {
 		// {@link engine.view.TextProxy} is not a instance of {@link engine.view.Node} so we need do handle it in specific way.
 		if ( node instanceof TextProxy ) {
-			return new Position( node._textNode, node._index + node._data.length );
+			return new Position( node.textNode, node.index + node.data.length );
 		}
 
 		if ( !node.parent ) {
@@ -221,7 +221,7 @@ export default class Position {
 	static createBefore( node ) {
 		// {@link engine.view.TextProxy} is not a instance of {@link engine.view.Node} so we need do handle it in specific way.
 		if ( node instanceof TextProxy ) {
-			return new Position( node._textNode, node._index );
+			return new Position( node.textNode, node.index );
 		}
 
 		if ( !node.parent ) {
--- a/src/view/position.js
+++ b/src/view/position.js
@@ -206,7 +206,7 @@ export default class Position {
 
 		// {@link engine.view.TextProxy} is not a instance of {@link engine.view.Node} so we need do handle it in specific way.
 		if ( node instanceof TextProxy ) {
-			return new Position( node._textNodeParent, node._index + 1 );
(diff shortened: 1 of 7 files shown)


src/view/textproxy.js
```diff
@@ -0,0 +1,115 @@
...
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+import Node from './node.js';
+
+/**
+ * Tree view text proxy.
```
pjasiun: Some additional explanation would be useful. There should be explanation why TextProxy exists. Note that is can be broken if the text node changes. Note that it is not a node, it should not be inserted as a child.

src/view/position.js
```diff
@@ -189,7 +190,7 @@ export default class Position {
 	/**
 	 * Creates a new position after the given node.
 	 *
-	 * @param {engine.view.Node} node Node after which the position should be located.
+	 * @param {engine.view.Node|engine.view.TextProxy} node Node after which the position should be located.
```
pjasiun: "Node or Text proxy after which..." to be precise.

src/view/textproxy.js
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @returns {engine.view.Document|null} View Document of the node or null.
+	 */
+	getDocument() {
+		// Parent might be Node, null or DocumentFragment.
+		if ( this.parent instanceof Node ) {
+			return this.parent.getDocument();
+		} else {
+			return null;
+		}
```
pjasiun: Can't use simply call `return this._textNode.getDocument();`?

src/view/textproxy.js
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * Gets the top parent for the node. If node has no parent it is the root itself.
+	 *
+	 * @returns {engine.view.Node}
+	 */
+	getRoot() {
```
pjasiun: Same here.

src/view/textproxy.js
```diff
@@ -0,0 +1,115 @@
...
+
+		while ( root.parent ) {
+			root = root.parent;
+		}
+
+		return root;
+	}
+
+	/**
+	 * Returns ancestors array of this node.
```
pjasiun: It is not "this node". "This" is not a node.

src/view/textproxy.js
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @param {Object} options Options object.
+	 * @param {Boolean} [options.includeNode=false] When set to `true` this node will be also included in parent's array.
+	 * @param {Boolean} [options.parentFirst=false] When set to `true`, array will be sorted from node's parent to root element,
+	 * otherwise root element will be the first item in the array.
+	 * @returns {Array} Array with ancestors.
+	 */
+	getAncestors( options = { includeNode: false, parentFirst: false } ) {
```
pjasiun: Same here.

src/view/textproxy.js
oskarwrobel: Right.

src/view/treewalker.js
```diff
@@ -0,0 +1,392 @@
...
+		if ( parent === this._boundaryEndParent && position.offset == this.boundaries.end.offset ) {
+			return { done: true };
+		}
+
+		// Get node just after current position.
+		let node;
+
+		// Text {@link engine.view.Text} element is a specific parent because it contains string instead of child nodes.
+		if ( parent instanceof Text ) {
+			node = parent._data[ position.offset ];
```
pjasiun: There is a bug. If you start in the middle of the text node, but without the singleCharacters option, the node is a letter anyway. And then, when the node is string, you return single characters till the end of the string. Test does not couch it, because in the test, the position starts one letter before the end of the text node, so it is fine there.

tests/view/treewalker.js
```diff
@@ -0,0 +1,628 @@
...
+					{ type: 'ELEMENT_END', item: bold },
+					{ type: 'TEXT', text: 'r' },
+					{ type: 'ELEMENT_START', item: img2 },
+					{ type: 'ELEMENT_END', item: img2 }
+				];
+
+				range = Range.createFromParentsAndOffsets( textBa, 1, paragraph, 3 );
+			} );
+
+			it( 'should return part of the text', () => {
```
pjasiun: This tests does not catch the bug that single letters are returned when you start in the middle of the text node.

tests/view/treewalker.js
```diff
@@ -0,0 +1,628 @@
...
+		//     |  |- A
+		//     |
+		//     |- R
+		//     |
+		//     |- img2
+		//     |
+		//     |- X
+
+		textBa = new Text( 'ba' );
+		bold = new Element( 'b', [], [ textBa ] );
```
pjasiun: `bold = new Element( 'b', null, [ textBa ] );`

pjasiun: I used to use `[]` in some tests but then we agreed that `null` is more readable.

pjasiun: Also, it should be `AttributeElement`.

tests/view/treewalker.js
```diff
@@ -0,0 +1,628 @@
...
+		charR = new Text( 'r' );
+		img2 = new Element( 'img2' );
```
pjasiun: You can use `ContainerElement`, seems you used `ContainerElement` for `p`.

tests/view/treewalker.js
oskarwrobel: It was, then I thought this element doesn't really have attributes so I can change it to Element https://github.com/ckeditor/ckeditor5-engine/pull/492/commits/3163e4f4b6fb5fd666997007b5d6a68684413e64 :)

pjasiun:
Also position does not match when you iterate with `singleCharacters` flag.

See:
```js
let iterator = new TreeWalker( { startPosition: rootBeginning, singleCharacters: true } );
for ( let value of iterator ) {
	console.log( value.previousPosition, value.item.constructor.name, value.nextPosition );
}
```
`previousPosition` of the one value does not match `nextPosition` from the previous value in case of text node boundaries.

I think that when `singleCharacters` option is `true` these positions should be inside text nodes, not next to them.

src/view/treewalker.js
```diff
@@ -0,0 +1,438 @@
...
+
+				textLength = endOffset - position.offset;
+			}
+
+			const textProxy = new TextProxy( parent, position.offset, textLength );
+
+			position.offset += textLength;
+
+			// Position at the end of Text is always out of Text node not inside.
+			if ( position.offset == parent._data.length ) {
```
oskarwrobel: I think I should move this condition here: https://github.com/ckeditor/ckeditor5-engine/pull/492/files#diff-d0dfec7faddd4ddd3ee23c8318b99147R384

src/view/textproxy.js
```diff
@@ -0,0 +1,109 @@
...
+
+		/**
+		 * The text content.
+		 *
+		 * @protected
+		 * @readonly
+		 * @member {String} engine.view.TextProxy#_data
+		 */
+
+		this._data = textNode._data.substring(
```
pjasiun: Is there a way to get data? This property might be public IMHO.

src/view/textproxy.js
```diff
@@ -0,0 +1,109 @@
...
+		this.parent = textNode.parent;
+
+		/**
+		 * Reference to the {@link engine.view.Text} element which TextProxy is a substring.
+		 *
+		 * @protected
+		 * @readonly
+		 * @member {engine.view.Text} engine.view.TextProxy#_textNode
+		 */
+		this._textNode = textNode;
```
pjasiun: There should be a way to get a text node from the text proxy. This method should be public.

src/view/textproxy.js
```diff
@@ -0,0 +1,109 @@
...
+
+		/**
+		 * Index of the substring in the `textParent`.
+		 *
+		 * @protected
+		 * @readonly
+		 * @member {Number} engine.view.TextProxy#_index
+		 */
+		this._index = startOffset;
```
pjasiun: And this is also an useful information for the user.

src/view/textproxy.js
```diff
@@ -0,0 +1,109 @@
...
+ * You should never create an instance of this class by your own. Instead, use string literals or {@link engine.model.Text}.
+ *
+ * @memberOf engine.view
+ */
+export default class TextProxy {
+	/**
+	 * Creates a tree view text proxy.
+	 *
+	 * @param {engine.view.Text} textNode Text node which text proxy is a substring.
+	 * @param {Number} startOffset Offset from beginning of {#_textNodeParent} and first character of {#_data}.
```
pjasiun: Protected properties should not be mentioned in docs.

src/view/treewalker.js
```diff
@@ -0,0 +1,448 @@
...
+
+		// We reached the walker boundary.
+		if ( parent === this._boundaryEndParent && position.offset == this.boundaries.end.offset ) {
+			return { done: true };
+		}
+
+		// Get node just after current position.
+		let node;
+
+		// Text {@link engine.view.Text} element is a specific parent because it contains string instead of child nodes.
```
pjasiun: You mixed jsDoc comments and the code comments.

src/view/treewalker.js
```diff
@@ -0,0 +1,448 @@
...
+ * * Forward iteration: For `'ELEMENT_END'` it is the last position inside the element. For all other types it is the
+ * position before the item. Note that it is more efficient to use this position then calculate the position before
+ * the node using {@link engine.view.Position.createBefore}.
+ * * Backward iteration: For `'ELEMENT_START'` it is the first position inside the element. For all other types it is
+ * the position after item.
+ * @property {engine.view.Position} nextPosition Next position of the iterator.
+ * * Forward iteration: For `'ELEMENT_START'` it is the first position inside the element. For all other types it is
+ * the position after the item.
+ * * Backward iteration: For `'ELEMENT_END'` it is last position inside element. For all other types it is the position
+ * before the item.
```
pjasiun: I miss the information that the position is always moved from the inside of the text node to its parent if the position is at the beginning or the end of the text node.

PR #496, opened 2016-06-20
Changes to selection and focus in view.Document.

### LINKED ISSUE #491: Selection & focus, second approach
**Scene:**
`view`

**Dramatis Personae:** 
`view.document`,
`view.selection`,
`FocusObserver`.

## Scene 1: The current situation.
Document has a `focusedEditable` property which is set by the `FocusObserver`. If `focusedEditable` is `null` it means that editor is blurred.

What should `focus()` method do? It should take the editable from view selection and set it as a `focusedEditable`. 

## Scene 2: The problem.
This realised an architectural issue. The selection editable is in fact the `focusedEditable`, it should be the same as `focusedEditable` if the `focusedEditable` is not null. What is selection change and `focusedEditable` does not change? Which one is the real `focusedEditable`? Which should be used?

## Scene 3: The conclusion.
It looks that we have redundant information. The selection has the real information which editable is focused. The only additional information `focusedEditable` bring is if the editor is focused or not. In fact, the selected editable is an useful information, even when the editor is blurred.

## Scene 4: The fix.
To fix this situation we should replace `focusedEditable` which two properties:
 - `selectedEditable` - which can be a getter, and always return editable which contains selection, even when editor is blurred.
 - `isFocused` - boolean property which is handled by the `FocusObserver`.

szymonkups:
Fixes #491.
Itroduces `selectedEditable` and `isFocused` on `engine.view.Document`.


src/view/document.js
```diff
@@ -74,16 +74,27 @@ export default class Document {
...
+
+		/**
+		 * {@link engine.view.EditableElement EditableElement} which is containing selection. It can be `null` if
+		 * there is no selection found.
+		 *
+		 * @readonly
+		 * @observable
+		 * @member {engine.view.EditableElement|null} engine.view.Document#selectedEditable
+		 */
+		this.set( 'selectedEditable', null );
```
pjasiun: Why `null` by default? This means that the `selectedEditable` will be `null` as long selection does not change. It should be `this.selection.getEditableElement()` by default.

src/view/selection.js
```diff
@@ -314,6 +315,29 @@ export default class Selection {
 	}
 
 	/**
+	 * Returns {@link engine.view.EditableElement EditableElement} instance that contains this selection.
+	 *
+	 * @returns {engine.view.EditableElement|null} Returns closest EditableElement or null if none is found.
+	 */
+	getEditableElement() {
+		if ( this.rangeCount ) {
+			let editable = this.getFirstPosition().parent;
```
pjasiun: We may introduce `Position.getEditableElement()` and use it here: `this.getFirstPosition().getEditableElement()`.

pjasiun:
I created a follow-up: https://github.com/ckeditor/ckeditor5-engine/issues/502.

PR #497, opened 2016-06-20
Downgraded destructing in default parameters.

oskarwrobel:
Fixes #https://github.com/ckeditor/ckeditor5/issues/190


src/model/treewalker.js
```diff
@@ -55,10 +45,10 @@ export default class TreeWalker {
 			throw new CKEditorError( 'tree-walker-no-start-position: Neither boundaries nor starting position have been defined.' );
 		}
 
-		if ( direction != 'FORWARD' && direction != 'BACKWARD' ) {
+		if ( options.direction && options.direction != 'FORWARD' && options.direction != 'BACKWARD' ) {
```
Reinmar: Where's the default option for direction coded?

Issue #443, opened 2016-05-25
Batch type


pjasiun:
We need the way to initialise data in the editor. The problem is that on the one hand we need a `change` event, so the data will be converted to the view, but we do not want to have an undo step for the initial data.

pjasiun:
More I think about it more I feel that it might be useful to mark batches which should be ignored in the undo stack. 

It would work for both initialisation and collaborative editing.

For now, we agreed that deltas from other users can be applied without any `batch`. Such batch-less deltas would be ignored then by the undo feature. But what if we want to show group of changes introduced by certain user, for instance in suggestion mode? My point is, that these changes, introduced by other, which we want to ignore in the undo stack, may need to be grouped too, batches works for such grouping very well.

Also note that creating deltas, without a batch is not an easy task, right now. So for initialisation it would not be the simplest solution.

We can not also not put these changes to the history, because then the document version will be different then the number of operations in the history, what can be a reason of issues, especially when you think about collaboration and sending this initial data for other users.

pjasiun:
Also note that re-initialization is a different case. When a use call `setData` and we decide it should reset the history we can do it. We can reset the model by clearing the history, removing all roots and reseting base version. Then we can create a root and set data the way we do it at the first place. Note that this way history will be empty but also match document version.

pjasiun:
In https://github.com/ckeditor/ckeditor5-typing/issues/7 we realised that changes caused by undo should break typing batch, but changes causes by collaboration should not. This is why be can not put all of them as deltas without a batch. We need to group these changes in batches and created batch types.

For now we need special batches for:
 - collaboration,
 - undo/redo,
 - initialization.

scofalik:
When we made this change we can consider whether it is a good practice to use `ModelDocument#applyOperation` without using a `batch`. I.e. in undo there is a code:
```javascript
for ( let delta of updatedDeltas ) {
	for ( let operation of delta.operations ) {
		this.editor.document.applyOperation( operation );
	}
}
```
This could be change to "transparent" `batch`.

I am writing this because in some places we already made kind-of-an-assumption that we check if applied `Operation` has a `batch`. If not, we treat it as "transparent" operation, that should not be considered by, i.e. undo mechanisms or typing mechanisms. Now we would be able to just check batch type instead of checking if batch is equal to `null`.

Reinmar:
[...]

PR #499, opened 2016-06-22
T/undo 2

### LINKED ISSUE #410: Fixes for OT to correct undo.
This is a followup ticket for https://github.com/ckeditor/ckeditor5-undo/pull/1 .

In this PR, there are some tests for undo feature that do not pass, most probably because of bugs in OT. OT should be corrected to fix those tests.

scofalik:
This PR includes changes in the engine that came up when fixing and re-writting Undo feature. It also "touches" #443 (added batch type to batch constructor) and finally closes #410.

I don't think that #443 is done though, because we haven't come up with closed list of batch types and I also have hope that we could get rid off `redo` batch type and maybe even `undo` batch type. https://github.com/ckeditor/ckeditor5-undo/issues/13

Biggest changes introduced by this PR is:
* `CompressedHistory` that stores deltas applied to the document but provides API to remove those deltas or updated deltas that are stored in history.
* `RemoveOperation` now creates "holder" elements in graveyard root - this was needed for `CompressedHistory`, so we can really remove deltas from history - earlier removing deltas from history messed up OT when graveyard was involved.
* OT changes for new `RemoveOperation` behavior


Reinmar:
I pushed a small commit with API doc fixes. The code looks clean an elegant and since it works for undo it's hard to say that something's wrong.

src/model/compressedhistory.js
```diff
@@ -0,0 +1,150 @@
...
+ * Modifying the original deltas history results in a compressed version of history and makes some scripts faster and easier to implement.
+ *
+ * **Note:** deltas kept in `CompressedHistory` should be used only to transform deltas. Do not use `CompressedHistory` to get original
+ * delta (especially basing on its {@link engine.model.delta.Delta#baseVersion baseVersion}). Do not trust base versions of deltas
+ * returned by `CompressedHistory`. After transforming your delta by deltas from `CompressedHistory`, fix its base version accordingly.
+ *
+ * @see engine.model.History
+ * @memberOf engine.model
+ */
+export default class CompressedHistory extends History {
```
pjasiun: `CompressedHistory` has nothing to do with compression. Also I do not see any reason to keep `CompressedHistory` and `History` separated.

src/model/compressedhistory.js
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	/**
+	 * Returns base versions of deltas which has been marked as reversed, in given base versions range.
+	 *
+	 * @param {Number} [from=0] Start of base versions range to check.
+	 * @param {Number} [to=Number.POSITIVE_INFINITY] End of base versions range to check.
+	 * @returns {Iterator.<Number>} Base versions of deltas marked as reversed.
+	 */
+	*getInactiveBaseVersions( from = 0, to = Number.POSITIVE_INFINITY ) {
```
pjasiun: Something not bangla here.

src/model/document.js
```diff
@@ -159,7 +162,10 @@ export default class Document {
 
 		this.version++;
 
-		this.history.addOperation( operation );
+		if ( operation.delta ) {
+			// Right now I can't imagine operations without deltas, but let's be safe.
```
pjasiun: You can hide an error this way. Silent exit is never a good solution. If you want to be save throw error.

src/model/operation/removeoperation.js
```diff
@@ -25,17 +26,63 @@ export default class RemoveOperation extends MoveOperation {
...
+
+	/**
+	 * Flag informing whether this operation should insert "holder" element (`true`) or should remove nodes
+	 * into existing "holder" element (`false`). It is `true` for each `RemoveOperation` that is the first `RemoveOperation`
+	 * in it's delta which points to given holder element.
+	 *
+	 * @protected
+	 * @type {Boolean}
+	 */
+	get _insertHolderElement() {
```
pjasiun: `_insertHolderElement` is a getter?

pjasiun: Maybe `_needsHolderElement`?

src/model/operation/removeoperation.js
```diff
@@ -25,17 +26,63 @@ export default class RemoveOperation extends MoveOperation {
...
+		if ( this.delta ) {
+			for ( let operation of this.delta.operations ) {
+				if ( operation instanceof RemoveOperation ) {
+					if ( operation == this ) {
+						return true;
+					} else if ( operation._holderElementOffset == this._holderElementOffset ) {
```
pjasiun: `operation._holderElementOffset == this._holderElementOffset` this check is unclear. What you check? Why?

src/model/history.js
```diff
@@ -36,109 +35,176 @@ export default class History {
...
 
-		// Fix base versions.
-		let baseVersion = transformed[ 0 ].operations[ 0 ].baseVersion;
-
-		for ( let i = 0; i < transformed.length; i++ ) {
-			transformed[ i ].baseVersion = baseVersion;
-			baseVersion += transformed[ i ].operations.length;
-		}
+		return deltas.length === 0 ? null : deltas;
+	}
```
pjasiun: Will the fact that now there might be multiple deltas with the same `baseVersion`, not break collaboration?

scofalik: It should be fine. AFAICS there are three things that should be considered:
1) Transforming incoming deltas.
2) Sending changes to other clients.
3) Repeating history when new client joins.

1) There are several cases here to consider. As long as you have correct state in compressed history and correctly "pre-transform" incoming operation it will be fine. I will write proper explanation at the bottom of this post. Anyway, the operational/delta transformations got a bit complicated and their base versions are fixed on multiple levels. I actually think it could be a good idea to get rid of fixing (incrementing) base version in OT algorithms. A feature that transforms deltas should be responsible for setting proper base version. 

2) Deltas are created and applied by `Batch` API. Nothing changed there. Each created delta will have a new, next number. OT plugin will listen for them and propagate them to server (BTW. I thought about it yesterday and I don't think `change` event is enough for this -- we might consider introducing `delta` event and/or change `engine.model.Document#applyOperation` to `#applyDelta`). This means that compressed history having multiple deltas with same base version are enclosed for each client.

3) When new client joins, it's editor contents have to be initialized. But you don't have to repeat history to do so. All you need is to make a "snapshot" of current model state and pass it as one big initial `InsertDelta`. Or you just send the "snapshot" itself and OT plugin on client side will convert it to initial `InsertDelta` (or just use `setData` method). All you need to do afterwards is to set `engine.model.Document#baseVersion` to correct value.


Here are some cases of what might happen when you get incoming delta while your history got compressed. Let's assume given history stack. `----` means that there was delta there but it got removed.
```
   A
   B
---C1---
---C2---
   Da without C
   Db without C
   E without C
---C1'---
---C2'---
   F
```
What happened here is that batch `C` got reverted and removed. Delta `D` and `E` are updated, but `D` got updated into two deltas (keep in mind this is a very rare scenario and I am not sure it could even happen :)). Delta `F` was added after undo happened.

Let's assume there is incoming, external delta `X`. Remember that the remote client will not have any updated/removed deltas, just the originals. Let's crack down following scenarios. It could base on:
1) delta `A`, `B`, or any delta before `C1`.
2) delta `C1`
3) delta `C2`
4) delta `D`, `E`, or any delta between `C2` and `C1'`
5) delta `C2'`
6) delta `F`.

Keep in mind, that the delta can't base on `Da` or `Db` because these are updated deltas and we don't send them to the server/other clients.

*In fact, as I think of it now, it could also base on `C1'`, which I haven't anticipated, because `C1'` and `C2'` are applied one after another, but they don't have to happen right after another on other client machine. Maybe we should/will have to introduce some control mechanisms, that one delta can be send/applied only together with other deltas?*

Here are solutions:
1) There are no changed deltas above, so both client's histories were same when delta `X` got created. We can simply transform delta `X` by history. It's the same case as when undoing deltas.

2) Delta `X` bases on delta `C1` that got reversed and removed from history. Deltas below remember the state "without delta `C1`". We have to convert delta `X` to the same state. To do so, we have to convert it by reversed `C1` - this way it will "forget" about `C1`. After delta `X` is transformed to `X without C` it can be normally transformed by compressed history state. Keep in mind that converting by reversed deltas is not an efficiency cost -- because we will not transform `X` later by `C1'` etc. And we would have to do so if we keep original history. Also notice that it's what we do normally while undoing deltas.

3) and 4) are actually the same scenarios as 2). We have to transform by all previously reversed deltas to get `X` to the proper state.

5) and 6) delta `X` is already in a state after `C` happened and after `C` got reversed, so there is no need to do any special things.

*If `X` bases on `C1'` it would need to be converted by `C2'` and all following deltas. So this would be another case, but similar to 2), 3) and 4). Anyway, for all those cases to work, we need to keep a registry of original/removed deltas, because we will be in need of transforming by them.*

scofalik:
I fixed all mentioned problems. I also think that after recent discussion and changes done in one of commits we may also close #443 
For now available batch types will be `default` and `transparent` - for batches that should be ignored by other features (initial batch, OT batches).

src/model/history.js
pjasiun: Thanks for the analysis. The remaining question is what should happen if one user undo. That reversed delta propagate, but how it should be applied. If it will be applied as a regular data we will have similar situation we had with undo v.1. It does not necessarily need to be an issue, since you can not undo other users changes. Alternatively we could compress deltas there too, but then we need to move the mechanism from the undo feature to the engine. Anyway, there are solution, what is enough for now.

pjasiun:
Please drop a comment there with decisions we agreed.

scofalik:
Done.

src/model/history.js
scofalik: Yes, deltas send to the remote client are applied as they are. Efficiency might be a problem because those deltas are not removed on remote client. But there should not be any problem with correctness. I mean, after all, remote client has no idea whether the delta was generated by user "by hand", by undo feature or by other feature. The problems with correct results are only connected to undo mechanism.

scofalik:
After I removed https://github.com/ckeditor/ckeditor5-engine/pull/499/files#diff-1ad6f65de8a59ca2d68c3601ab2c2aa5R164 a lot of errors popped up in tests because in early days we just applied operations without deltas. It's ~53 tests, with multiple operations in some of them. Let's move this issue to a follow-up: #501.

PR #504, opened 2016-06-27
Added view.Document#focus()

### LINKED ISSUE #488: Introduce view.Document#focus()
We need a method which features will be able to use to tell the engine to focus the editor.

The method should check in which editable the selection is now and focus that specific editable if it's not yet focused.

Related issue: https://github.com/ckeditor/ckeditor5-headings/issues/7

szymonkups:
Fixes #488.

### DIFF
--- a/src/view/document.js
+++ b/src/view/document.js
@@ -256,6 +256,17 @@ export default class Document {
 	}
 
 	/**
+	 * Focuses document. It will focus {@link engine.view.EditableElement EditableElement} that is currently having
+	 * selection inside.
+	 *
+	 * @fires engine.view.Document#focus
+	 */
+	focus() {
+		this.fire( 'focus' );
+		this.render();
+	}
+
+	/**
 	 * Disables all added observers.
 	 */
 	disableObservers() {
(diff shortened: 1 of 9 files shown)


src/view/document.js
```diff
@@ -256,6 +256,17 @@ export default class Document {
 	}
 
 	/**
+	 * Focuses document. It will focus {@link engine.view.EditableElement EditableElement} that is currently having
+	 * selection inside.
+	 *
+	 * @fires engine.view.Document#focus
+	 */
+	focus() {
+		this.fire( 'focus' );
```
pjasiun: This `focus` event has no attributes which `focus` event fired from the `FocusObserver` has. This is why why I called `focus()` in the focus manual  test I get an error from this line: https://github.com/ckeditor/ckeditor5-engine/blob/master/tests/view/manual/focusobserver.js#L18

pjasiun: We may simplify the `focus` event to make it possible to fire event with the same properties. On the other hand them we will lose cohesion between various view event based on the native DOM event. But, again, it would useful feature to be able to fire these view events manually, so the whole view events API should have native DOM events as optionally and provide a helper to fire a view event.

pjasiun: But, if it is too tricky to provide same interface, we could try to set `isFocus` and call `render`, so let renderer to provoke native `focus` event.

pjasiun: Also I am not fire if `focus` event should be fired if there already was focus. You could try to check if `isFocused` is false.

PR #505, opened 2016-06-27
T/329 Split Selection to Selection and DocumentSelection

### LINKED ISSUE #329: Split Selection into (Static)Selection and LiveSelection
The request comes from the question "what are we going to pass to methods like `deleteContents()`?". In CKEditor 4 we usually had a command that automatically works on the real selection and helper methods which accept ranges. The command of course call one of the helper methods, so they exist mostly for code reusability reasons. Other algorithms can use those in order to build some more complex behaviours.

In CKEditor 5 the 4's scheme with util methods working on ranges does not fully make sense, because:

1. Simple array of ranges doesn't specify a direction. So it would need to be accompanied with another attribute.
1. Selection defines also its attributes, so the array of ranges and direction would need also this.
1. A simple set of ranges do not have selection's method from which some may be useful.

In other words, it's clearly visible that algorithms need to work on selection. Theoretically we could now do this:

```js
const sel = new Selection( doc );

sel.addRange( someRange );

deleteUtils.delete( sel );
```

However, the current implementation of the `Selection` operates on live ranges, what mean that the selection needs to be detached once not needed. Otherwise, the editor will quickly become sluggish due to huge number of attached live positions.

Therefore, let's split the selection into two classes. Basic, static selection that can be used as a feed for utils, and live selection for the document's main selection (and those who know what they are doing).

The split should be pretty simple to do as live ranges are created in just one place. 

PS. For now we can freely live without the static selection. Simply, in the tests (which, for a time being, will be the only code that uses the feature utils) we can use live selection and detach it when needed.

scofalik:
This PR fixes #329 

Notable changes:
* Introduced `DocumentSelection` which has same features as "old" `Selection`.
* `Selection` is now static, meaning that it does not create `LiveRange`s and does not update when document changes.
* `Selection` has no attributes API.
* `Selection` has no "default range".
* `DocumentSelection` has `getSnapshot` method that returns `Selection` instance with same ranges and direction as that `DocumentSelection`.

New `Selection` class is meant to be used when you need to cached it or in methods that may/have to modify selection.

### DIFF
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -161,9 +161,7 @@ export default class Document {
 
 		this.version++;
 
-		if ( operation.delta ) {
-			this.history.addDelta( operation.delta );
-		}
+		this.history.addDelta( operation.delta );
 
 		const batch = operation.delta && operation.delta.batch;
 
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -12,7 +12,7 @@ import transformations from './delta/basic-transformations.js'; // jshint ignore
 import RootElement from './rootelement.js';
 import Batch from './batch.js';
 import History from './history.js';
-import DocumentSelection from './documentselection.js';
+import LiveSelection from './liveselection.js';
 import EmitterMixin from '../../utils/emittermixin.js';
 import CKEditorError from '../../utils/ckeditorerror.js';
 import mix from '../../utils/mix.js';
@@ -55,9 +55,9 @@ export default class Document {
 		 * Selection done on this document.
 		 *
 		 * @readonly
-		 * @member {engine.model.DocumentSelection} engine.model.Document#selection
+		 * @member {engine.model.LiveSelection} engine.model.Document#selection
 		 */
-		this.selection = new DocumentSelection( this );
+		this.selection = new LiveSelection( this );
 
 		/**
 		 * Schema for this document.
@@ -287,7 +287,7 @@ export default class Document {
 		const json = clone( this );
 
 		// Due to circular references we need to remove parent reference.
-		json.selection = '[engine.model.DocumentSelection]';
+		json.selection = '[engine.model.LiveSelection]';
 
 		return json;
 	}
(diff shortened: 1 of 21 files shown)


pjasiun:
The only important difference between `Selection` and `DocumentSelection` is that `DocumentSelection` use `LiveRange` instead of regular ranges. It means that `DocumentSelection` should be called `LiveSelection` for consistency.

scofalik:
Done

src/model/document.js
```diff
@@ -162,7 +162,6 @@ export default class Document {
 		this.version++;
 
 		if ( operation.delta ) {
-			// Right now I can't imagine operations without deltas, but let's be safe.
 			this.history.addDelta( operation.delta );
```
pjasiun: Can you merge master?

scofalik:
Done.

src/model/liveselection.js
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	/**
+	 * Creates an instance of {@link engine.model.Selection} that has same ranges and direction as this selection. Since
+	 * this will be instance of `Selection` instead of `LiveSelection`, it will not be automatically updated or rendered,
+	 * so it can be used in algorithms using and modifying selection.
+	 *
+	 * @returns {engine.model.Selection} Selection instance which ranges and direction is equal to this selection.
+	 */
+	getSnapshot() {
```
pjasiun: We have `Position.createFromPosition`, `Range.createFromRange` so why not `Selection.createFromSelection`?

pjasiun: In fact I do not like these method, I prefer `.clone( isDynamic )` but it is more important to have a consistent API now, we can discuss it in the separate issue.

scofalik: In fact, why not Zoidberg?

src/model/selection.js
```diff
@@ -60,7 +34,7 @@ export default class Selection {
 		 * Stores all ranges that are selected.
 		 *
 		 * @private
```
pjasiun: It should be protected, not private, if `LiveSelection` use it.

src/model/liveselection.js
pjasiun: I created a follow-up: https://github.com/ckeditor/ckeditor5-engine/issues/507

scofalik:
Done.

PR #516, opened 2016-07-04
Fix for focus toggling in editable

### LINKED ISSUE #512: Editable loses and regains focus when typing
1. Open http://tests.ckeditor.dev:1030/build/amd/tests/editor-classic/manual/classic
2. Init the editor.
3. Start typing.

See that on each letter isFocused is set to false and then back to true.

szymonkups:
Fixes #512.


src/view/document.js
```diff
@@ -250,11 +247,7 @@ export default class Document {
 	 * before rendering and re-enabled after that.
```
pjasiun: It would be worth to mention that this method fire `render` event.

Issue #480, opened 2016-06-13
Text nodes simplification


pjasiun:
The original concept was to have every character as a separate node. It did no work.

Then we had an idea to merge them together and make them act like a separate characters. This is why we have `CharacterProxy` which is a `Node`, `TextProxy` which is not a `Node`, `Text` to insert a text with attributes and private `NodeListText` to keep compressed text. This is why I have no idea how to copy `NodeList` it the efficient way or I need to think event time which API I should use in each case.

No more lies!

This lie, does not make much sense, since we decided that even `TreeWalker` return whole texts by default.

First we should apply some change, easy to agree:
 - `Text` is a `Node`,
 - `Text` and `NodeListText` should be merge into a single class,
 - `CharacterProxy` and `TextProxy` should be merge into a single class (`TextProxy` or `TextFragment`),
 - `NodeList` has no hidden internal structures, is contains `Text`s.

Then we will see what this change will bring. It may change the way `Positions` work and change `TreeWalker` API, but I think that we should keep the change as simple as possible at first and then introduce new features to it.

Also, note that having only `Text` and `TextFragment` (for what `TreeWalker` returns when you start iterating inside a text or ask for single characters) we will have the same classes in model and the view.

Related: https://github.com/ckeditor/ckeditor5-engine/issues/350, https://github.com/ckeditor/ckeditor5-engine/issues/351, https://github.com/ckeditor/ckeditor5-engine/issues/353, https://github.com/ckeditor/ckeditor5-engine/issues/357, https://github.com/ckeditor/ckeditor5-engine/issues/321, https://github.com/ckeditor/ckeditor5-engine/issues/428.

pjasiun:
https://github.com/ckeditor/ckeditor5-engine/issues/255 view `TreeWalker` introduce architecture similar to what we want to have in the model. 

An additional change we could apply here is merging `TEXT` and `CHARACTER` types into one (`text`).

Also if `Text` will be a node `NodeList` should return `Text` as child, but we could have additional `getItemAt` method, which provide the "old" API, for instance for OT.

scofalik:
In F2F discussion we come to conclusion that it is worthy to consider removing `NodeList` and creating `ModelWriter`.

Why so?

We already have this approach in view. In fact, in view and in model we have same problem: we have a basic structure, in which texts are nodes but we need some easy method to operate on those texts when we have a position inside a text node and we want to do something there, i.e. change attributes of part of text or insert something inside text node.

[...]

PR #519, opened 2016-07-08
Converted engine.view.Writer class to utilities functions.

### LINKED ISSUE #515: ViewWriter could be a namespace object with tools-methods.
Right now `engine.view.Writer` is a class, but it has no properties and only usage of `this` is to refer to it's other methods. It could easily be an object containing tools-methods instead of being a class.

szymonkups:
Resolves #515


src/view/writer.js
```diff
@@ -15,582 +15,597 @@ import DocumentFragment from './documentfragment.js';
...
 
-		// Position's parent is container, so no attributes to break.
-		if ( isContainerOrFragment( positionParent ) ) {
-			return Position.createFromPosition( position );
-		}
+/**
+ * Returns first parent container of specified {@link engine.view.Position Position}.
+ * Position's parent node is checked as first, then next parents are checked.
+ *
+ * Note that {@link engine.view.DocumentFragment DocumentFragment} is thread like a container.
```
scofalik: /s/thread/treated

src/view/writer.js
```diff
@@ -15,582 +15,597 @@ import DocumentFragment from './documentfragment.js';
...
-	_breakAttributes( position, forceSplitText = false ) {
-		const positionOffset = position.offset;
-		const positionParent = position.parent;
-
-		// There are no attributes to break and text nodes breaking is not forced.
-		if ( !forceSplitText && positionParent instanceof Text && isContainerOrFragment( positionParent.parent ) ) {
-			return Position.createFromPosition( position );
-		}
+export default {
+	getParentContainer,
```
scofalik: This seems more like a helper function. I am not sure if there will be any usage of this in features or core. I'd either not export it or just export it as separate function, outside of "writer" object (the function does not change anything in view).

src/view/writer.js
scofalik: And also later in comments.

src/view/writer.js
```diff
@@ -15,582 +15,597 @@ import DocumentFragment from './documentfragment.js';
...
+ * Throws {@link utils.CKEditorError CKEditorError} `view-writer-invalid-range-container` when
+ * {@link engine.view.Range#start start} and {@link engine.view.Range#end end} positions are not placed inside
+ * same parent container.
+ *
+ * Note that {@link engine.view.DocumentFragment DocumentFragment} is thread like a container.
+ *
+ * @see engine.view.writer.breakAttributes
+ * @function engine.view.writer.breakRange
+ * @param {engine.view.Range} range Range which `start` and `end` positions will be used to break attributes.
+ * @returns {engine.view.Range} New range with located at break positions.
```
scofalik: /s/with/with boundaries

src/view/writer.js
```diff
@@ -15,582 +15,597 @@ import DocumentFragment from './documentfragment.js';
...
+	if ( positionParent instanceof Text ) {
+		return position;
+	}
 
-			// Insert cloned node to position's parent node.
-			positionParent.parent.insertChildren( offsetAfter, clonedNode );
+	// When inside empty attribute - remove it.
+	if ( positionParent instanceof AttributeElement && positionParent.getChildCount() === 0 ) {
+		const parent = positionParent.parent;
+		const offset = positionParent.getIndex();
```
scofalik: I'd name that variable `index` :)

src/view/writer.js
```diff
@@ -15,582 +15,597 @@ import DocumentFragment from './documentfragment.js';
...
+ *		<p><b>[]</b></p> -> <p>[]</p>
+ *		<p><b>foo</b><i>[]</i><b>bar</b></p> -> <p><b>foo{}bar</b></p>
+ *
+ * @see engine.view.AttributeElement
+ * @see engine.view.ContainerElement
+ * @function engine.view.writer.mergeAttributes
+ * @param {engine.view.Position} position Merge position.
+ * @returns {engine.view.Position} Position after merge.
+ */
+export function mergeAttributes( position ) {
```
scofalik: I'd consider renaming this function. I never liked name `AttributeElement` in the first place, because attribute is something else than element and `AttributeElement` can have elements. This causes confusion.

I'd call this method either `mergeAttributeElements`, `mergeAtPosition` or, simply, `merge` (my vote goes here and we could rename `breakAttributes` to `break`). `merge` and `break` would be consistent with `insert`, `move`, etc. Also we do not pass any attributes to the function.

src/view/writer.js
scofalik: In similar way, `breakRange`, `breakAttributes` and `mergeAttributes` looks more like helper methods for `viewWriter`. Similar functions will be in `modelWriter` (after #480) and are now in `NodeList` - also private.

When creating features, developers should not care about things like breaking and merging nodes. They should just use `insert`, `wrap`, `move`, etc. Unfortunately I can see that `breakAttributes` is actually used in `model-selection-to-view-converters.js` which bothers me.

Except of selection conversion I can't imagine why someone would like to just break attributes, normally you would like to break attributes because you need to `insert` something or `remove` something, etc.

Since there is no quick way around this, we have to keep `breakAttributes` and (for consistency) `breakRange`, `mergeAttributes` exposed.

src/view/writer.js
```diff
@@ -15,582 +15,597 @@ import DocumentFragment from './documentfragment.js';
...
 
-		// If start position was merged - move end position back.
-		if ( !start.isEqual( newRange.start ) ) {
-			newRange.end.offset--;
-		}
-		const end = this.mergeAttributes( newRange.end );
+// Function used by both public breakRange (without splitting text nodes) and by other methods (with
+// splitting text nodes).
+//
+// @private
```
scofalik: Those tags are unnecessary here.

PR #521, opened 2016-07-12
Changes in model test utilities.

### LINKED ISSUE #441: Model parse( '<paragraph></paragraph>' ) returns RootElement
`model.parse( '<paragraph></paragraph>' );` returns `model.RootElement` with a `<paragraph>` element as a child.

Expected results are `<paragraph>` element, or `DocumentFragment` for multiple elements. There is no way `model.parse` should return `RootElement`.

I spend too much time debugging code because of this bug.

szymonkups:
Fixes #441.
To achieve this I moved basic attribute manipulation methods from `LiveSelection` to `Selection`. It was required in order to use `Selection` class when parsing model to `DocumentFragment`.

### DIFF
--- a/tests/model/selection.js
+++ b/tests/model/selection.js
@@ -249,19 +249,29 @@ describe( 'Selection', () => {
 	} );
 
 	describe( 'focus', () => {
-		it( 'should return first end start', () => {
-			selection.addRange( range );
-
-			expect( selection.focus.isEqual( range.end ) ).to.be.true;
+		let r3;
+		beforeEach( () => {
+			const r1 = Range.createFromParentsAndOffsets( root, 2, root, 4 );
+			const r2 = Range.createFromParentsAndOffsets( root, 4, root, 6 );
+			r3 = Range.createFromParentsAndOffsets( root, 1, root, 2 );
+			selection.addRange( r1 );
+			selection.addRange( r2 );
 		} );
 
-		it( 'should return first range start when backward', () => {
-			selection.addRange( range, true );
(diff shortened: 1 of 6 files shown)


tests/model/selection.js
```diff
@@ -248,6 +248,24 @@ describe( 'Selection', () => {
 		} );
 	} );
 
+	describe( 'focus', () => {
+		it( 'should return first end start', () => {
```
scofalik: This test name is weird.

tests/model/selection.js
```diff
@@ -248,6 +248,24 @@ describe( 'Selection', () => {
...
+			selection.addRange( range );
+
+			expect( selection.focus.isEqual( range.end ) ).to.be.true;
+		} );
+
+		it( 'should return first range start when backward', () => {
```
scofalik: `focus` does not return positions from first range. `focus` is connected with last added range (not "first" or "last" range in position relation meaning). No matter if selection is backward or not, `focus` and `anchor` are connected with last range (so is `isBackward` property).

Those tests should have three ranges added, which are "mixed" in position relation meaning (i.e. range 2-4, range 4-6 and range 1-2). In such test `focus` should be `1` for normal direction and `2` for backward direction.

scofalik:
`function stringify` in `_utils\model.js` probably can get some better refactoring. `document` property is not needed, since we don't need it anymore to handle selection stuff. It also seems that we don't need `DocumentFragment` wrapping. AFAIR, right now, `Position` can be created with any `Element` as a root. My only doubt is what will/should happen if passed `node` is `Text`.

PR #522, opened 2016-07-12
Added validation for Range.

### LINKED ISSUE #471: Selection.setRanges/addRange() should throw when not a range passed
I spent half an hour on this... My code was correct, but it turned out that I removed import statement at some point, so I've been using native `Range` class which works. It only breaks deep in our code, but with a non-obvious error.

oskarwrobel:
Fixes #471

### DIFF
--- a/src/model/selection.js
+++ b/src/model/selection.js
@@ -213,7 +213,7 @@ export default class Selection {
 	 */
 	addRange( range, isBackward ) {
 		if ( !( range instanceof Range ) ) {
-			throw new CKEditorError( 'model-selection-invalid-range: Invalid Range.' );
+			throw new CKEditorError( 'selection-invalid-range: Invalid Range.' );
 		}
 
 		this._pushRange( range );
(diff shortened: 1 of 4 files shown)


src/model/selection.js
```diff
@@ -126,6 +126,10 @@ export default class Selection {
 	 * or backward - from end to start (`true`). Defaults to `false`.
 	 */
 	addRange( range, isBackward ) {
+		if ( !( range instanceof Range ) ) {
+			throw new CKEditorError( 'model-selection-invalid-range: Invalid Range.' );
```
scofalik: `model-selection-invalid-range` is inconsistent with other error "codes" that we have in this file, which does not have `model-` prefix.

I just noticed that there are multiple error messages in `view` namespace that have `view-` prefix, but some do not. I see that we have consistency problem regarding this, so I will make follow-up issue, but for now, please make it consistent in one file.

scofalik: #527

oskarwrobel: I unified error codes in both files (removed `view|model` prefix).

PR #524, opened 2016-07-13
Renamed converter builders to lowerCase.

oskarwrobel:
Fixes: #467 

Related to:
- https://github.com/ckeditor/ckeditor5-basic-styles/pull/10
- https://github.com/ckeditor/ckeditor5-typing/pull/22
- https://github.com/ckeditor/ckeditor5-paragraph/pull/5
- https://github.com/ckeditor/ckeditor5-headings/pull/13

### DIFF
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -20,7 +20,7 @@ import ViewContainerElement from '../view/containerelement.js';
  * Provides chainable, high-level API to easily build basic model-to-view converters that are appended to given
  * dispatchers. In many cases, this is the API that should be used to specify how abstract model elements and
  * attributes should be represented in the view (and then later in DOM). Instances of this class are created by
- * {@link engine.conversion.BuildModelConverterFor}.
+ * {@link engine.conversion.buildModelConverter}.
  *
  * If you need more complex converters, see {@link engine.conversion.ModelConversionDispatcher},
  * {@link engine.conversion.modelToView}, {@link engine.conversion.ModelConsumable}, {@link engine.conversion.Mapper}.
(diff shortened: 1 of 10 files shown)


scofalik:
You haven't touched the file name, what was suggested in the issue. However, I don't like file name `buildmodelconverterfor.js`. I think that we could do some minor refactor here:
* name the function `buildModelConverter`
* in `ModelConverterBuilder`, remove initializing the class with dispatchers (same for `buildModelConverter`)
* add `for()` method that will take dispatchers as a parameter
* change how the converter builders are used (this should be easy search+replace)

So now it will look like this `buildModelConverter.for( dispatchers ).fromXXX( ... ).toXXX( ... );`. In my opinion, it looks better and we will have a better looking function/file name (`buildmodelconverter.js`).

Same applies to view converters builder.

Reinmar:
Doh... I've just spent 15 minutes reviewing the changes :D. I haven't noticed it's R-ed already. Anyway, I think that I agree with changes proposed by @scofalik.

Reinmar:
PS. This PR could have a branch in ckeditor5 with configuration of the branches. It took me some time to switch all that.

oskarwrobel:
@scofalik If I understood the idea correctly `buildModelConverter` will be a function, so it will work like this:
```javascript
buildModelConverter().for( dispatchers ).fromXXX( ... ).toXXX( ... );
```
instead of
```javascript
buildModelConverter.for( dispatchers ).fromXXX( ... ).toXXX( ... );
```

pjasiun:
But then we will still have different file name and function name don't we?

What do you think about skipping "build", so we will have:
```
modelConverter.for( dispatchers ).fromXXX( ... ).toXXX( ... );
```
In the `ModelConverter` file?

scofalik:
As far as second version looks and reads better, I am not sure if there is an elegant solution that would enable it. Anyway, both are fine with me.

oskarwrobel:
Maybe something like this:
```javascript
export default class ModelConverter {
    ...
    static for ( ...dispatchers ) {
        return new this( dispatchers );
    }
    ...
}
```

and then

```javascript
import ModelConverter from 'path/to/modelconverter.js';

ModelConverter.for()...
```

pjasiun:
> @pjasiun: It's fine and it was one of ideas we discussed earlier with @oskarwrobel . But it does not denote action, so the whole "sentence" has less sense. There is a feeling that it returns something that you have to use afterwards.

I get your point and it makes sens for me.

> Maybe...

Now all proposed options are fine for me, as long the file name is the same as the function/class name.

src/conversion/buildmodelconverter.js
```diff
@@ -87,6 +84,18 @@ class ModelConverterBuilder {
 	}
 
 	/**
+	 * Set one or more dispatchers which the built converter will be attached to.
+	 *
+	 * @chainable
+	 * @param {...engine.conversion.ModelConversionDispatcher} dispatchers One or more dispatchers.
+	 */
```
oskarwrobel: Missing `@returns {engine.conversion.ModelConverterBuilder}`

src/conversion/buildviewconverter.js
```diff
@@ -103,9 +100,21 @@ class ViewConverterBuilder {
 	}
 
 	/**
+	 * Set one or more dispatchers which the built converter will be attached to.
+	 *
+	 * @chainable
+	 * @param {...engine.conversion.ViewConversionDispatcher} dispatchers One or more dispatchers.
```
oskarwrobel: Missing `@returns {engine.conversion.ViewConverterBuilder}`

PR #534, opened 2016-07-18
Moved default observers from EditingController to view document.

### LINKED ISSUE #453: Move default observers to view.Document
Now default observers are defined in the `EditingController`. But the view Document needs all of them anyway: `SelectionObserver` needs `MutationObserver`, jump-over-inline-filler needs `KeyObserver` and now `focusedEditable` property needs `FocusObserver`. If `Document` needs these observers to work properly, it may make sense it add them in the Document constructor instead of `EditingController`.

szymonkups:
Fixes #453.

### DIFF
--- a/src/editingcontroller.js
+++ b/src/editingcontroller.js
@@ -4,11 +4,6 @@
  */
 
 import ViewDocument from './view/document.js';
-import MutationObserver from './view/observer/mutationobserver.js';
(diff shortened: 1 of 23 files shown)


scofalik:
I rebased the branch with actual master and also pushed some minor docs fixes.

Issue #473, opened 2016-06-06
InsertOperation should deeply clone passed nodes.


jodator:
The problem is that NodeList have references to the same nodes.

As a result getting history of simple type operations duplicates text - found in classic editor.

Test Scenario I:
1. Type abc into empty editor (only with `<p></p>` inside).
2. Get deltas from `document.editor.history`

Result:
- first delta's operation reference whole text "abc" instead of "a" only

Test Scenario I:
1. Type abcd into empty editor (only with `<p></p>` inside)
2. Place cursor between b and c ( ab|cd )
3. Press enter key
4. Get deltas from `document.editor.history`

Result:
- last delta (SplitDelta) has two operations:
    - insert operation of a paragraph with text nodes ("cd")  - a probably bug
    - move operation of text nodes "cd" inside a new paragraph

scofalik:
In second case it's not really a bug (if I understand correctly).
The `SplitDelta` has an `InsertOperation` of `paragraph`. When the delta was created the `paragraph` was empty and only after it got its children. I am not sure right now but AFAIR we need to reference the original node to have *something* working correctly. But as I wrote, I am not sure now.

The first case is interesting though.

We will have to look into it because this is a difficult problem.

jodator:
The socond case is bug in case of collaborative editing - when appling the same deltas on a different document you get:
```
ab
cdcd
```
and you get mismatch of data from both editors.

Also any reference to a node that can change later can result in collaborative editing mismatch.

pjasiun:
Fixed by #537.

Issue #321, opened 2016-04-05
Model: Rename TextProxy.commonParent to .parent


Reinmar:
It's confusing that it's now named `.commonParent` because other kinds of nodes (real or no) have `.parent`.

pjasiun:
Agreed the same way `DocumentFragment` has `children` even if it is not an element.

PR #537, opened 2016-07-19
T/480 Refactor engine model text handling.

### LINKED ISSUE #480: Text nodes simplification

scofalik:
Total refactor in how text data is handled `engine.model`. Also introduces several methods and properties that were not available in `engine.model` but are in `engine.view` (of course as long as they made sense as some of them are related strictly to `engine.view` things.

Fixes #480, #511, #473, #351, #321.

PS. I really tried to get to 100 files changed but couldn't do it without including less related issues in this PR, and then PJ would scream at me.


scofalik:
Related PRs in other repos:
https://github.com/ckeditor/ckeditor5/pull/248
https://github.com/ckeditor/ckeditor5-undo/pull/22

Reinmar:
Files changed... 99. Have you changed line endings in all files or what?

Reinmar:
There are conflicts with my recent changes, but fortunately easy to resolve. Make sure to run `gulp lint` on the code.

PR #538, opened 2016-07-20
T/165 engine.view API unified with model

### LINKED ISSUE #165: Tree view and tree model APIs should be simmilar
Both data structures use the thee structures which allows attributes and child nodes manipulating. We should ensure that if both have `insertChildren` method, then both of these methods have the same order of parameters.

AFAIR, beside https://github.com/ckeditor/ckeditor5-core/issues/76, we need to introduced `element.appendChildren` to the tree model element.

scofalik:
Fixes #165.

This PR bases on #480, PR #537 -- it includes that PR. Please review it only after #537 is merged.

### DIFF
--- a/src/model/node.js
+++ b/src/model/node.js
@@ -242,6 +242,27 @@ export default class Node {
 	}
 
 	/**
+	 * Returns ancestors array of this node.
+	 *
+	 * @param {Object} options Options object.
+	 * @param {Boolean} [options.includeNode=false] When set to `true` this node will be also included in parent's array.
+	 * @param {Boolean} [options.parentFirst=false] When set to `true`, array will be sorted from node's parent to root element,
+	 * otherwise root element will be the first item in the array.
+	 * @returns {Array} Array with ancestors.
+	 */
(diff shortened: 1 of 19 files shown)


scofalik:
For now I decided to skip getters/setters unification across model and view. It can be done in #518. I only brought missing API parts but some of them are getters in model and methods in view and vice versa.

I also wasn't really sure what to do about `Node#getPath()` in model vs `Node#getAncestors()` in view. I ended up not adding `getAncestors` to model and `getPath` to view. I assumed they do similar thing, and since model positions bases on paths while view positions bases on element+offset it makes sense.

scofalik:
Since `t/480` got merged into `master`, I rebased this branch with `master` too. Now it's just one commit.

pjasiun:
We do not need `Node#getPath()` in the view, there is nothing like path used in model position. But we may need `Node#getAncestors()` in the model.

scofalik:
I've added `engine.model.Node#getAncestors`.

PR #545, opened 2016-07-25
T/542 Follow-ups after refactor engine model text handling

### LINKED ISSUE #542: Follow-ups after refactor engine model text handling.
Follow-ups after https://github.com/ckeditor/ckeditor5-engine/pull/537.

- [x] `NodeList` has `totalOffset` and `Element` and `DocumentFragment` have `getMaxOffset`, it shoud be normalized.
- [x] `NodeList.indexToOffset`  and `offsetToIndex` fix offset if it is out of the range. They should throw instead.
- [x] `NodeList._nodeAtOffset` is a memory consuming optimalization which can couses other issues, we should avoid optimalization untill we know we need it.
- [x] `Position` should mention that it use offset instead of indexes.
- [x] `TextProxy.isPartial` method should be introduced.
- [x] `Element` and `DocumentFragment` constructors should transform string to the text nodes (in both model and view).
- ~~`InsertOperation` should copy nodes in contructor, not in `execute` method.~~

scofalik:
Fixes #542

### DIFF
--- a/src/model/nodelist.js
+++ b/src/model/nodelist.js
@@ -93,10 +93,10 @@ export default class NodeList {
 	}
 
 	/**
-	 * Converts index "position" to offset "position".
+	 * Converts index to offset in node list.
 	 *
-	 * Returns starting offset of a node that is at given index. If given index is too low, `0` is returned. If
-	 * given index is too high, {@link engine.model.NodeList#getMaxOffset last available offset} is returned.
+	 * Returns starting offset of a node that is at given index. Throws {@link utils.CKEditorError CKEditorError}
+	 * `nodelist-index-out-of-bounds` if given index is less than `0` or more than {@link engine.model.NodeList#length}.
 	 *
 	 * @param {Number} index Node's index.
 	 * @returns {Number} Node's starting offset.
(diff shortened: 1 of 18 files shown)


src/model/nodelist.js
```diff
@@ -95,29 +87,36 @@ export default class NodeList {
...
-		return offset == -1 ? null : offset;
+		return index === null ? null : this._nodes.slice( 0, index ).reduce( ( sum, node ) => sum + node.offsetSize, 0 );
 	}
 
 	/**
 	 * Converts index "position" to offset "position".
 	 *
 	 * Returns starting offset of a node that is at given index. If given index is too low, `0` is returned. If
-	 * given index is too high, {@link engine.model.NodeList#totalOffset last available offset} is returned.
+	 * given index is too high, {@link engine.model.NodeList#getMaxOffset last available offset} is returned.
```
pjasiun: This comment seems to miss updated.

pjasiun:
Please fix the outdated docs and you can merge this PR.

src/model/nodelist.js
```diff
@@ -131,15 +130,26 @@ export default class NodeList {
 	 * @returns {Number} Index of a node that occupies given offset.
 	 */
```
pjasiun: Same here.

Issue #527, opened 2016-07-14
Inconsistent errors naming


scofalik:
We have a problem with errors naming consistency in the engine. Some errors are named with namespace prefixes, i.e. `view-selection-range-intersects` in `view/selection.js`, others do not have that prefix i.e. `position-after-root` in `view/position.js` or most (all?) things in model.

We should decide whether we want to add those prefixes or not and refactor code accordingly.

Reinmar:
I think we'll not be able to define some strict rules and we should not try to. It's going to be the getters/properties battle again. I'd follow common sense in all cases. In this one adding view/model denominator for every error code sounds good.

scofalik:
I do not want to have a complex rule here. Right now error names follow more-or-less this convention:
`<className>-<methodName>-<errorCode>` where `errorCode` is two-three word description of the error. If we want a rule it has to be simple and enforced everywhere. We can add `<package>` before `<className>`.

Another idea is to make error names follow the same naming style as, how you call them, package paths? I.e. if error is thrown by `engine.model.Element#getChildIndex()` maybe the error name should be same as path but without package name? So it will be something like:

`model.Element#getChildIndex-error-description` or `model-element-getChildIndex-error-description`. That would be an easy and clear rule without place for doubts.

Right now it is `element-getChildIndex-error-description`.

Reinmar:
> We can add `<package>` before `<className>`.

You mean the namespace, right? But it doesn't always matter. E.g. in `ckeditor5-core` there's (will be, in fact) `command/attributecommand.js`. In such case adding `command` to the name makes no sense. That's why I said that I wouldn't try to create any strict rules. This could be sufficient:

* In most cases use `<className>-<methodName>-<errorCode>`.
* You can use the namespace fragment as a denominator in case of multiple classes having the same name within on package, e.g. in the engine: `<namespaceFragment>-<className>-<methodName>-<errorCode>`.
* Some error codes can be shared between multiple methods or even classes and then need to use module name and namespace to create a properly specific name.

Is it enough?

scofalik:
Yes, I've meant `<namespace>`.

Anyway what you described is exactly what I wanted avoid. You propose different error naming concepts in different cases. You say "don't create rule" then you propose how to name them (rule 1) and how to name them in special case (rule 2). Versus my one strict rule which is same everywhere.

pjasiun:
[...]

PR #547, opened 2016-08-01
T/527 Fixes inconsistent error names

maxbarnas:
Fixes https://github.com/ckeditor/ckeditor5-engine/issues/527.


maxbarnas:
I need to update tests. Please do not close this task right now.

scofalik:
The tests passes, but they should be update. Since we have 100% CC, there is at least one test for each error. Those tests check if the error is thrown and use regexp to check if correct error has been thrown. For example: https://github.com/ckeditor/ckeditor5-engine/blob/master/tests/model/history.js#L133.

Please fix all tests for error throwing, in which error names has been changed.

PR #550, opened 2016-08-03
T/478d Multi-byte character support

### LINKED ISSUE #478: Multi-byte characters support
At the moment, neither the engine (except some fragments) nor the features support multi-byte characters in the content (you can read about them in https://mathiasbynens.be/notes/javascript-unicode).

This means that if we insert e.g. `'\u{1F4A9}'` into the model and press backspace after it only half of that character will be deleted.

The model features `CharacterProxy` class but other pieces of code don't expect the character to be of length 2. The same knowledge will need to be spread in the view.

scofalik:
Fixes #478. See the issue for details and discussion about implementation.


scofalik:
Related PRs:
https://github.com/ckeditor/ckeditor5-utils/pull/54
https://github.com/ckeditor/ckeditor5-typing/pull/34

scofalik:
Note: because of changes introduced by this feature, I had to comment out one test in `ckeditor5-undo` which started to fail. It testes an incomplete feature, in an incomplete way. There is already an issue for this incomplete feature in `ckeditor5-undo`: https://github.com/ckeditor/ckeditor5-undo/issues/12.

Before this feature, the commented test tested just model tree state, but did not test the selection. This feature introduces a callback whenever `engine.model.Document#selection` `change:range` event is fired. Callback checks some things in model tree. However, since the selection was in wrong place, error was thrown in the test.

I tried to fix it but it proved to be more time consuming, so I decided that for now it is better just to comment out the test. This, unfortunately, led to a minor code coverage decrease in `ckeditor5-undo`.

src/model/text.js
```diff
@@ -33,7 +33,7 @@ export default class Text extends Node {
 		 *
 		 * @type {String}
 		 */
-		this.data = data || '';
+		this.data = data ? data.normalize() : '';
```
pjasiun: Won't it break a composition?

src/model/composer/modifyselection.js
```diff
@@ -6,21 +6,44 @@
 import Position from '../position.js';
 import TreeWalker from '../treewalker.js';
 import Range from '../range.js';
+import { isInsideSurrogatePair, isInsideCombinedSymbol } from '../../../utils/unicode.js';
 
+/* jshint -W100 */
```
pjasiun: ??

src/model/composer/modifyselection.js
```diff
@@ -6,21 +6,44 @@
...
 /**
  * Modifies the selection. Currently the supported modifications are:
  *
- * * Extending. The selection focus is moved in the specified `options.direction` with a step specified in `options.unit`
- * (defaults to `'character'`, other values are not not yet supported).
- * Note: if you extend a forward selection in a backward direction you will in fact shrink it.
+ * * Extending. The selection focus is moved in the specified `options.direction` with a step specified in `options.unit`.
+ * Possible values for `unit` are:
+ *  * `'character'` (default) - moves selection by one user-perceived character. In most cases this means moving by one
+ *  character in {String} sense. However, unicode also defines "combing marks". These are special symbols, that combines
```
pjasiun: {String} -> `String`

src/model/composer/modifyselection.js
```diff
@@ -6,21 +6,44 @@
...
+ *  selection between "base character" and "combining mark", because "combining marks" have their own unicode code points.
+ *  However, for technical reasons, unicode code points with values above `UxFFFF` are represented in native {String} by
+ *  two characters, called "surrogate pairs". Halves of "surrogate pairs" have a meaning only when placed next to each other.
+ *  For example `𨭎` is represented in {String} by `\uD862\uDF4E`. Both `\uD862` and `\uDF4E` do not have any meaning
+ *  outside the pair (are rendered as ? when alone). Position between them would be incorrect. In this case, selection
+ *  extension will include whole "surrogate pair".
+ *
+ * **Note:** if you extend a forward selection in a backward direction you will in fact shrink it.
+ *
+ * **Note:** you may use `CKEditor5 Graphemes` feature available at https://github.com/ckeditor/ckeditor5-graphemes
```
pjasiun: Can we remove it for now since we are unsure we want to support this plugin?

src/view/text.js
```diff
@@ -28,7 +28,7 @@ export default class Text extends Node {
 		 * @private
 		 * @member {String} engine.view.Text#_data
 		 */
-		this._data = data;
+		this._data = data ? data.normalize() : '';
```
pjasiun: Same as above.

tests/model/composer/modifyselection.js
```diff
@@ -123,6 +124,64 @@ describe( 'Delete utils', () => {
...
+				//	'<p>நி<selection>லை</selection>க்கு</p>',
+				//	'<p>நி<selection>லைக்</selection>கு</p>'
+				//);
+				//
+				//test(
+				//	'unicode support - backward',
+				//	'<p>நி<selection backward>லை</selection>க்கு</p>',
+				//	'<p><selection backward>நிலை</selection>க்கு</p>',
+				//	{ direction: 'backward' }
+				//);
```
pjasiun: Shouldn't you remove these?

tests/model/composer/modifyselection.js
```diff
@@ -218,21 +277,100 @@ describe( 'Delete utils', () => {
...
+			//	'<p>நி<selection>லைக</selection>்கு</p>',
+			//	{ unit: 'codePoint' }
+			//);
+			//
+			//test(
+			//	'shrinks one unicode code point backward (combining mark case) ',
+			//	'<p>நி<selection>லைக்</selection>கு</p>',
+			//	'<p>நி<selection>லைக</selection>்கு</p>',
+			//	{ unit: 'codePoint', direction: 'backward' }
+			//);
```
pjasiun: And these?

src/model/text.js
scofalik: It may break composition and it proved to be painful during tests. After all, I think we should not normalize this. The only problem will be for features that scan texts for given value (i.e. "search feature"). Because user might try to search for normalized version and there could be not-normalized version in text.

But I think it's better to not normalize.

src/model/composer/modifyselection.js
scofalik: Ok

tests/model/composer/modifyselection.js
scofalik: Yeah, sorry for these.

pjasiun:
> Note: because of changes introduced by this feature, I had to comment out one test in ckeditor5-undo which started to fail. It testes an incomplete feature, in an incomplete way. There is already an issue for this incomplete feature in ckeditor5-undo: ckeditor/ckeditor5-undo#12.

It is fine for now, but we should have 100% coverage when we will close iteration 2 so the issue should be fixed soon.

src/model/composer/modifyselection.js
scofalik: We needed it before, because there was an incorrect sign in the comment (half of surrogate pair, rendered) and linter was throwing an error because it won't be properly displayed in HTML :).

Then the comment changed and we do not need it anymore.

PR #554, opened 2016-08-09
Refactored model utils.

oskarwrobel:
Fixes https://github.com/ckeditor/ckeditor5-engine/issues/306

Related to:
- https://github.com/ckeditor/ckeditor5-enter/pull/22
- https://github.com/ckeditor/ckeditor5-typing/pull/35
- https://github.com/ckeditor/ckeditor5-undo/pull/33
- https://github.com/ckeditor/ckeditor5-basic-styles/pull/18
- https://github.com/ckeditor/ckeditor5-headings/pull/20

Temporary branches configuration available here: https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-engine/306

### DIFF
--- a/src/dataprocessor/xmldataprocessor.js
+++ b/src/dataprocessor/xmldataprocessor.js
@@ -11,7 +11,7 @@ import { NBSP_FILLER } from '../view/filler.js';
  * XmlDataProcessor class.
  * This data processor implementation uses XML as input/output data.
  * This class is needed because unlike HTML, XML allows to use any tag with any value.
- * E.g. `<link>Text</link>` is a valid XML byt invalid HTML.
+ * E.g. `<link>Text</link>` is a valid XML but invalid HTML.
  *
  * @memberOf engine.dataProcessor
  * @implements engine.dataProcessor.DataProcessor
--- a/src/dataprocessor/xmldataprocessor.js
+++ b/src/dataprocessor/xmldataprocessor.js
@@ -30,10 +30,10 @@ export default class XmlDataProcessor {
 		 * E.g. Registering namespaces [ 'attribute', 'container' ] allows to use `<attirbute:tagName></attribute:tagName>` and
 		 * `<container:tagName></container:tagName>` input. It is mainly for debugging.
 		 *
-		 * @private
-		 * @member {DOMParser} engine.dataProcessor.XmlDataProcessor#_namespaces
+		 * @public
+		 * @member {DOMParser} engine.dataProcessor.XmlDataProcessor#namespaces
 		 */
-		this._namespaces = options.namespaces || [];
+		this.namespaces = options.namespaces || [];
 
 		/**
 		 * DOMParser instance used to parse XML string to XMLDocument.
@@ -100,7 +100,7 @@ export default class XmlDataProcessor {
 	 */
 	_toDom( data ) {
 		// Stringify namespaces.
-		const namespaces = this._namespaces.map( nsp => `xmlns:${ nsp }="nsp"` ).join( ' ' );
+		const namespaces = this.namespaces.map( nsp => `xmlns:${ nsp }="nsp"` ).join( ' ' );
 
 		// Wrap data into root element with optional namespace definitions.
 		data = `<xml ${ namespaces }>${ data }</xml>`;
--- a/src/dataprocessor/xmldataprocessor.js
+++ b/src/dataprocessor/xmldataprocessor.js
@@ -10,6 +10,8 @@ import { NBSP_FILLER } from '../view/filler.js';
 /**
  * XmlDataProcessor class.
  * This data processor implementation uses XML as input/output data.
+ * This class is needed because unlike HTML, XML allows to use any tag with any value.
+ * E.g. `<link>Text</link>` is a valid XML byt invalid HTML.
  *
  * @memberOf engine.dataProcessor
  * @implements engine.dataProcessor.DataProcessor
@@ -17,8 +19,22 @@ import { NBSP_FILLER } from '../view/filler.js';
 export default class XmlDataProcessor {
 	/**
 	 * Creates a new instance of the XmlDataProcessor class.
+	 *
+	 * @param {Object} options Configuration options.
+	 * @param {Array<String>} [options.namespaces=[]] List of namespaces allowed to use in XML input.
 	 */
-	constructor() {
+	constructor( options = {} ) {
+		/**
+		 * List of namespaces allowed to use in XML input.
+		 *
+		 * E.g. Registering namespaces [ 'attribute', 'container' ] allows to use `<attirbute:tagName></attribute:tagName>` and
+		 * `<container:tagName></container:tagName>` input. It is mainly for debugging.
+		 *
+		 * @private
+		 * @member {DOMParser} engine.dataProcessor.XmlDataProcessor#_namespaces
+		 */
+		this._namespaces = options.namespaces || [];
+
 		/**
 		 * DOMParser instance used to parse XML string to XMLDocument.
 		 *
@@ -36,7 +52,8 @@ export default class XmlDataProcessor {
 		this._domConverter = new DomConverter( { blockFiller: NBSP_FILLER } );
 
 		/**
-		 * BasicHtmlWriter instance used to convert DOM elements to HTML string.
+		 * BasicHtmlWriter instance used to convert DOM elements to XML string.
+		 * There is no need to use dedicated for XML writer because BasicHtmlWriter works well in this case.
 		 *
 		 * @private
 		 * @member {engine.dataProcessor.BasicHtmlWriter} engine.dataProcessor.HtmlDataProcessor#_htmlWriter
@@ -54,7 +71,8 @@ export default class XmlDataProcessor {
 		// Convert view DocumentFragment to DOM DocumentFragment.
 		const domFragment = this._domConverter.viewToDom( viewFragment, document );
 
-		// Convert DOM DocumentFragment to HTML output.
+		// Convert DOM DocumentFragment to XML output.
+		// There is no need to use dedicated for XML serializing method because BasicHtmlWriter works well in this case.
 		return this._htmlWriter.getHtml( domFragment );
 	}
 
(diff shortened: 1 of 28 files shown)


src/dataprocessor/xmldataprocessor.js
```diff
@@ -0,0 +1,104 @@
...
+		/**
+		 * DOM converter used to convert DOM elements to view elements.
+		 *
+		 * @private
+		 * @member {engine.view.DomConverter} engine.dataProcessor.XmlDataProcessor#_domConverter.
+		 */
+		this._domConverter = new DomConverter( { blockFiller: NBSP_FILLER } );
+
+		/**
+		 * BasicHtmlWriter instance used to convert DOM elements to HTML string.
```
pjasiun: A comment is needed why XML DP use HTML writer.

tests/_utils-tests/model.js
```diff
@@ -195,14 +200,14 @@ describe( 'model test utils', () => {
 		it( 'writes element attributes', () => {
 			root.appendChildren(
 				new Element( 'a', { foo: true, bar: 1, car: false }, [
-					new Element( 'b', { fooBar: 'x y', barFoo: { x: 1, y: 2 } } )
+					new Element( 'b', { fooBar: 'x y', barFoo: JSON.stringify( { x: 1, y: 2 } ) } )
```
pjasiun: That's strange. It should be possible to set object as an attribute. It should be `parse` who stringify attributes.

tests/_utils-tests/model.js
```diff
@@ -256,32 +261,33 @@ describe( 'model test utils', () => {
...
 			} );
 
 			it( 'writes selection collapsed at the text right boundary', () => {
 				selection.collapse( elB, 'before' );
 
 				expect( stringify( root, selection ) ).to.equal(
-					'<a></a>foo<$text bold=true>bar</$text><selection bold=true /><b></b>'
+					'<a></a>foo<$text bold="true">bar</$text>[]<b></b>'
 				);
+				expect( selection.getAttribute( 'bold' ) ).to.true;
```
pjasiun: You do not check if there is no other selection attributes. Add check to the number of attributes.

tests/_utils-tests/model.js
```diff
@@ -291,25 +297,17 @@ describe( 'model test utils', () => {
...
 			} );
 
 			it( 'writes selection collapsed selection in a text with attributes', () => {
 				selection.collapse( root, 5 );
 
 				expect( stringify( root, selection ) ).to.equal(
-					'<a></a>foo<$text bold=true>b<selection bold=true />ar</$text><b></b>'
+					'<a></a>foo<$text bold="true">b[]ar</$text><b></b>'
 				);
+				expect( selection.getAttribute( 'bold' ) ).to.true;
```
pjasiun: As above.

tests/_utils-tests/model.js
```diff
@@ -460,61 +453,61 @@ describe( 'model test utils', () => {
...
-			} ).to.throw( Error, 'Parse error - unexpected closing tag.' );
+			} ).to.throw( Error );
 		} );
 
 		it( 'throws when missing closing tag for text', () => {
 			expect( () => {
 				parse( '<$text>' );
-			} ).to.throw( Error, 'Parse error - missing closing tags: $text.' );
+			} ).to.throw( Error );
 		} );
```
pjasiun: Do we need all these tests now? Aren't they check native behavior now? Maybe one will be enough then.

tests/_utils-tests/model.js
```diff
@@ -523,58 +516,77 @@ describe( 'model test utils', () => {
...
 			} );
 
 			test( 'sets selection within a text', {
-				data: 'x<selection bold=true>y</selection>z'
+				data: 'x[y]z'
 			} );
 
-			test( 'sets selection within a text with different attributes', {
-				data: '<$text bold=true>fo<selection bold=true>o</$text>ba</selection>r'
+			it( 'sets a backward selection containing an element', () => {
```
pjasiun: "backward", you said... interesting...

tests/_utils/view.js
```diff
@@ -255,13 +266,17 @@ export function stringify( node, selectionOrPositionOrRange = null, options = {}
  * When set to `null` root element will be created automatically. If set to
  * {@link engine.view.Element Element} or {@link engine.view.DocumentFragment DocumentFragment} - this node
  * will be used as root for all parsed nodes.
+ * @param {Boolean} [options.sameSelectionCharacters=false] When set to `true` then selection inside text should be marked using `{` and `}`
+ * and selection outside text using `[` and `]`. When set to `false` then both should be marked with `[` and `]` only.
```
pjasiun: Isn't it opposite?

src/dataprocessor/xmldataprocessor.js
```diff
@@ -0,0 +1,104 @@
...
+ * For licensing, see LICENSE.md.
+ */
+
+import BasicHtmlWriter from './basichtmlwriter.js';
+import DomConverter from '../view/domconverter.js';
+import { NBSP_FILLER } from '../view/filler.js';
+
+/**
+ * XmlDataProcessor class.
+ * This data processor implementation uses XML as input/output data.
```
pjasiun: Explanation what is the difference between  XML and HTML data processor would be nice, maybe with some examples. I believe this is a question everybody will ask, checking this class: why we need both HTML and XML DP.

src/dataprocessor/xmldataprocessor.js
```diff
@@ -0,0 +1,104 @@
...
+	/**
+	 * Converts XML String to its DOM representation. Returns DocumentFragment, containing nodes parsed from
+	 * provided data.
+	 *
+	 * @private
+	 * @param {String} data
+	 * @returns {DocumentFragment}
+	 */
+	_toDom( data ) {
+		data = `<xml xmlns:attribute="foo" xmlns:container="foo">${ data }</xml>`;
```
pjasiun: Do we need this `foo` here? Can't it be empty? Also `attribute:x` and `container:x` is a specific testing utils notation. Is should not be here, but in these utils. Additionally we need a comment why we add this tag.

src/dataprocessor/xmldataprocessor.js
```diff
@@ -0,0 +1,104 @@
...
+	 * Converts provided {@link engine.view.DocumentFragment DocumentFragment} to data format - in this case XML string.
+	 *
+	 * @param {engine.view.DocumentFragment} viewFragment
+	 * @returns {String} XML string.
+	 */
+	toData( viewFragment ) {
+		// Convert view DocumentFragment to DOM DocumentFragment.
+		const domFragment = this._domConverter.viewToDom( viewFragment, document );
+
+		// Convert DOM DocumentFragment to HTML output.
```
pjasiun: `to HTML` or `to XML`? Why `getHtml` to get "XML string"?

tests/_utils/model.js
```diff
@@ -3,20 +3,41 @@
  * For licensing, see LICENSE.md.
  */
 
-import TreeWalker from '/ckeditor5/engine/model/treewalker.js';
+import Mapper from '/ckeditor5/engine/conversion/mapper.js';
+import Document from '/ckeditor5/engine/model/document.js';
```
pjasiun: You could use `ModelDocument`, `ModelRange` and `ModelPosition` not to confuse with native one, especially if you have `ModelElement` below.

tests/_utils/model.js
```diff
@@ -3,20 +3,41 @@
...
+import viewWriter from '/ckeditor5/engine/view/writer.js';
+
+import count from '/ckeditor5/utils/count.js';
+import { parse as viewParse, stringify as viewStringify } from '/tests/engine/_utils/view.js';
+import { convertRangeSelection, convertCollapsedSelection } from '/ckeditor5/engine/conversion/model-selection-to-view-converters.js';
+
+// Test utils uses `<$text foo="bar">Lorem ipsum</$text>` notation to create text with attributes, but `$text` is not
+// valid XML element name, so needs to be parsed before conversion to view.
+const VIEW_TEXT_WITH_ATTRIBUTES_ELEMENT = 'view-text-with-attributes';
+const DATA_STRING_TEXT_WITH_ATTRIBUTES_ELEMENT = '$text';
```
pjasiun: I don't find this `const` useful, but it might be me ;)

tests/_utils/model.js
```diff
@@ -138,367 +179,223 @@ export function stringify( node, selectionOrPositionOrRange = null ) {
...
-	for ( let patternName in patterns ) {
-		match = data.match( patterns[ patternName ] );
+function insertElement() {
+	return ( evt, data, consumable, conversionApi ) => {
+		consumable.consume( data.item, 'insert' );
 
-		if ( match ) {
-			data = data.slice( match[ 0 ].length );
+		const viewPosition = conversionApi.mapper.toViewPosition( data.range.start );
+		const viewElement = new ViewElement( data.item.name, data.item.getAttributes() );
```
pjasiun: Use can use generic `insertElement` and add view element constructor as a parameter. Something like this:
```
insertElement( data => new ViewElement( data.item.name, data.item.getAttributes() ) );
```

tests/_utils/model.js
```diff
@@ -138,367 +179,223 @@ export function stringify( node, selectionOrPositionOrRange = null ) {
...
-
-	throw new Error( 'Parse error - unexpected token: ' + data + '.' );
+		conversionApi.mapper.bindElements( data.item, viewElement );
+		viewWriter.insert( viewPosition, viewElement );
+	};
 }
 
-function parseAttributes( attrsString ) {
-	attrsString = attrsString.trim();
+function insertText() {
```
pjasiun: You should be able to use default `insertText`

tests/_utils/model.js
```diff
@@ -138,367 +179,223 @@ export function stringify( node, selectionOrPositionOrRange = null ) {
...
-		ret += writeItem( value, selection, { selection: withSelection } );
+	modelToView.on( 'insert:$text', insertText() );
+	modelToView.on( 'insert', insertElement() );
+	modelToView.on( 'selection', convertRangeSelection() );
+	modelToView.on( 'selection', convertCollapsedSelection() );
 
-		lastPosition = value.nextPosition;
-	}
+	// Convert model to view.
+	modelToView.convertInsert( range );
```
pjasiun: https://github.com/ckeditor/ckeditor5-engine/pull/560
```
convertInsert -> convertInsertion
```

src/dataprocessor/xmldataprocessor.js
oskarwrobel: - I've just checked and there must to be some value `xmlns:attribute: Empty XML namespace is not allowed`.
- I'll move this namespaces to config.

tests/_utils-tests/model.js
oskarwrobel: I'll fix it working on https://github.com/ckeditor/ckeditor5-engine/issues/556.

PR #568, opened 2016-08-11
Added parsing values of model element attributes + little test cleanup.

### LINKED ISSUE #556: Parsing boolean attribute values in view
FolowUp: https://github.com/ckeditor/ckeditor5-engine/issues/306

After changing dataprocessor in view from `HTML` to `XML` is not possible to set attribute with boolean value `attribute=true`. It must be `attribute="true"` and value is always a string.

oskarwrobel:
Fixes #556

### DIFF
--- a/tests/_utils/model.js
+++ b/tests/_utils/model.js
@@ -3,15 +3,6 @@
  * For licensing, see LICENSE.md.
  */
 
-import Mapper from '/ckeditor5/engine/conversion/mapper.js';
-import { parse as viewParse, stringify as viewStringify } from '/tests/engine/_utils/view.js';
-import {
-	convertRangeSelection,
-	convertCollapsedSelection,
-	convertSelectionAttribute
-} from '/ckeditor5/engine/conversion/model-selection-to-view-converters.js';
-import { insertText, insertElement, wrap } from '/ckeditor5/engine/conversion/model-to-view-converters.js';
-
 import RootElement from '/ckeditor5/engine/model/rootelement.js';
 import ModelDocument from '/ckeditor5/engine/model/document.js';
 import ModelRange from '/ckeditor5/engine/model/range.js';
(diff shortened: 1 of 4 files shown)


tests/_utils/model.js
```diff
@@ -365,3 +379,35 @@ function convertToModelText( withAttributes = false ) {
...
+
+// Loop trough attributes map and parse each value by passed parser.
+function parseAttributes( attributes, parser ) {
+	let result = new Map();
+
+	for ( let attribute of attributes ) {
+		result.set( attribute[ 0 ], parser( attribute[ 1 ] ) );
+	}
+
+	return result;
```
pjasiun: Can't you make it a generator?

tests/_utils/model.js
```diff
@@ -365,3 +379,35 @@ function convertToModelText( withAttributes = false ) {
 		evt.stop();
 	};
 }
+
+// Typecast string value to the original type.
+// 		`'true'` => `true`
+// 		`'1'` => `1`
+//		`'{"x":1,"y":2}'` => `{ x: 1, y: 2 }`
```
pjasiun: You could add a sample with a string which can not be parsed and explain why you return an attribute in case of error. "Typecast string value to the original type." is not clear enought.

tests/_utils/model.js
```diff
@@ -365,3 +379,35 @@ function convertToModelText( withAttributes = false ) {
...
+function parseValue( attribute ) {
+	try {
+		return JSON.parse( attribute );
+	} catch ( e ) {
+		return attribute;
+	}
+}
+
+// When value is an Object stringify it.
+function stringifyObject( data ) {
```
pjasiun: `stringifyAttributeValue`?

tests/_utils/model.js
```diff
@@ -365,3 +379,35 @@ function convertToModelText( withAttributes = false ) {
 		evt.stop();
 	};
 }
+
+// Typecast string value to the original type.
+// 		`'true'` => `true`
+// 		`'1'` => `1`
+//		`'{"x":1,"y":2}'` => `{ x: 1, y: 2 }`
+function parseValue( attribute ) {
```
pjasiun: `parseAttributeValue`?

tests/_utils/model.js
```diff
@@ -365,3 +379,35 @@ function convertToModelText( withAttributes = false ) {
...
+function stringifyObject( data ) {
+	if ( isPlainObject( data ) ) {
+		return JSON.stringify( data );
+	}
+
+	return data;
+}
+
+// Loop trough attributes map and parse each value by passed parser.
+function parseAttributes( attributes, parser ) {
```
pjasiun: Maybe `convertAttributes` and `converter`. The fact that `parser` can be `stringify` is misleading.

PR #578, opened 2016-08-25
Introduced ClickObserver.

### LINKED ISSUE #577: Introduce click observer
Simple click observer is needed by link feature.

oskarwrobel:
Fixes #577.

### DIFF
--- a/src/view/observer/clickobserver.js
+++ b/src/view/observer/clickobserver.js
@@ -8,8 +8,8 @@ import DomEventObserver from './domeventobserver.js';
 /**
  * {@link engine.view.Document#click Click} event observer.
  *
- * Note that this observer is not attached by the {@link engine.view.Document} and is not available by default.
- * To make it available it needs to be added to {@link engine.view.Document} by {@link engine.view.Document#addObserver} method.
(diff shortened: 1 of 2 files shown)


src/view/observer/clickobserver.js
```diff
@@ -0,0 +1,26 @@
...
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import DomEventObserver from './domeventobserver.js';
+
+/**
+ * {@link engine.view.Document#click Click} event observer.
+ *
+ * Note that this observer is not attached by the {@link engine.view.Document} and is not available by default.
```
Reinmar: If we're saying this it would be much more useful if we said how then to enable this observer.

oskarwrobel: Right, done.

oleq: The only thing missing here is rich documentation, similar to [which comes with the `KeyObserver`](https://github.com/ckeditor/ckeditor5-engine/blob/t/577/src/view/observer/keyobserver.js#L39-L65).

oleq:
Docs needed. Besides that, it looks yummy.

PR #584, opened 2016-09-07
T/583

### LINKED ISSUE #583: Changes required by List feature.
This is an overall issue for changes required in engine by List feature. It also covers bugs discovered and fixed during implementing List feature. For complete list of changes look at commits in related PR.

#266, #267 and #367 will be handled in this issue (and related PR), however discussion may be continued in those issues.

scofalik:
Fixes #583


src/conversion/mapper.js
```diff
@@ -37,6 +49,28 @@ export default class Mapper {
 		 * @member {WeakMap} engine.conversion.Mapper#_viewToModelMapping
 		 */
 		this._viewToModelMapping = new WeakMap();
+
+		// Add default callback for model to view position mapping.
+		this.on( 'modelToViewPosition', ( evt, modelPosition, data ) => {
+			let viewContainer = this._modelToViewMapping.get( modelPosition.parent );
```
pjasiun: Can't in be a local function now?

src/view/element.js
```diff
@@ -123,6 +123,17 @@ export default class Element extends Node {
 	}
 
 	/**
+	 * Length of this view element in model if it is mapped. This value can be overwritten in classes extending
+	 * {@link engine.view.Element Element} to provide correct position mapping in {@link engine.conversion.Mapper Mapper}.
+	 *
+	 * @readonly
+	 * @type {Number}
+	 */
+	get modelLength() {
```
pjasiun: WTF? Why view knows anything about the model? What model?

src/conversion/mapper.js
```diff
@@ -104,33 +172,35 @@ export default class Mapper {
...
-		while ( !modelParent ) {
-			viewBlock = viewBlock.parent;
-			modelParent = this._viewToModelMapping.get( viewBlock );
-		}
+		const data = {
+			modelPosition: null
+		};
 
-		let modelOffset = this._toModelOffset( viewPosition.parent, viewPosition.offset, viewBlock );
+		this.fire( 'viewToModelPosition', viewPosition, data );
```
pjasiun: Can't we put `viewPosition` also in `data`? Then some listener can change it and  pass updated position to the default listener. Does it make sense?

src/view/element.js
pjasiun: Also, if it is always 1 why you need to implement it?

src/conversion/model-to-view-converters.js
```diff
@@ -311,19 +305,49 @@ export function move() {
...
+ * and their view counterparts.
+ *
+ * @external engine.conversion.modelToView
+ * @function engine.conversion.modelToView.rename
+ * @returns {Function}
+ */
+export function rename() {
+	return ( evt, data, consumable, conversionApi ) => {
+		if ( consumable.test( data.element, 'rename' ) ) {
+			// Create fake element in model, needed for conversion purposes.
```
pjasiun: "needed for conversion purposes" does not look like a good explanation.

src/conversion/modelconversiondispatcher.js
```diff
@@ -252,6 +271,50 @@ export default class ModelConversionDispatcher {
...
+		const consumable = new Consumable();
+		consumable.add( element, 'rename' );
+
+		// Callbacks added to `rename` event should either provide totally custom handling of rename conversion
+		// or create "fake element" in model that is needed to fire default rename conversion.
+		const data = { element, oldName };
+		this.fire( 'rename:' + element.name + ':' + oldName, data, consumable, this.conversionApi );
+
+		// Default callback should provide `data.fakeElement` model element that is needed for firing conversion.
+		if ( data.fakeElement ) {
```
pjasiun: Yep, it's hacky. The whole code should be in one place.

src/conversion/modelconversiondispatcher.js
```diff
@@ -416,6 +475,29 @@ export default class ModelConversionDispatcher {
 	 */
 
 	/**
+	 * Fired for renamed element.
+	 *
+	 * **Note:** rename event is a specific event because it requires to have a specific callback. The callback should either
```
pjasiun: Can we make this event simpler?

src/model/liverange.js
```diff
@@ -88,11 +88,15 @@ function bindWithDocument() {
...
 		( event, type, changes ) => {
-			transform.call( this, type, changes.range, changes.sourcePosition );
+			if ( supportedTypes.has( type ) ) {
+				transform.call( this, type, changes.range, changes.sourcePosition );
+			}
 		}
 	);
 }
 
+const supportedTypes = new Set( [ 'insert', 'move', 'remove', 'reinsert' ] );
```
pjasiun: Comment needed.

src/view/writer.js
```diff
@@ -72,6 +74,57 @@ export function breakRange( range ) {
...
 /**
+ * Breaks {@link engine.view.ContainerElement container view element} into two, at the given position. Position
+ * has to be directly inside container element and cannot be in root. Does not break if position is at the beginning
+ * or at the end of it's parent element.
+ *
+ *		<p>foo^bar</p> -> <p>foo</p><p>bar</p>
+ *		<div><p>foo</p>^<p>bar</p></div> -> <div><p>foo</p></div><div><p>bar</p></div>
+ *		<p>^foobar</p> -> ^<p>foobar</p>
+ *		<p>foobar^</p> -> <p>foobar</p>^
+ *
```
pjasiun: It would be nice to crosslink with `breakAt` and `breakRange` to mention the difference between these methods.

pjasiun: Also, we may consider renaming `breakAt` to `breakAttributes` now to stress the difference.

src/view/writer.js
```diff
@@ -144,6 +197,38 @@ export function mergeAt( position ) {
 }
 
 /**
+ * Merges two {@link engine.view.ContainerElement container view elements} that are before and after given position.
+ * Precisly, the element after the position is removed and it's contents are moved to element before the position.
+ *
+ *		<p>foo</p>^<p>bar</p> -> <p>foo^bar</p>
+ *		<div>foo</div>^<p>bar</p> -> <div>foo^bar</div>
+ *
```
pjasiun: As above.

src/view/element.js
scofalik: Okay so I don't like this one too. I was supposed to think more of it but I forgot. I don't like mentioning mapping and model in the view classes and I am all for doing it some other way.

Unfortunately, we still need that `modelLength` thing, even if we use callbacks in `Mapper`. It is needed for "default" mapper algorithm, because we don't fire custom callbacks "deeply" (after each position change when looking for correct position).

I set it to `1` and implemented it so there is a track in documentation that such thing is available and you can overwrite it.

Since this is bad, here is my other proposal:
`Mapper.registerViewElementLength( ViewElementClass, lengthCallback );`

This would map custom view element classes with functions that will calculate and return length of given view element in model. This could be either instance method or static method. BTW. I was thinking whether make position mapping callback per instance or static. Per instance = better flexibility, but it often ends up with just adding same thing in multiple places instead of one.

So, anyway, right now, in `List` feature I have:
```javascript
class ViewListItemElement extends ViewContainerElement {
	// ... cut ...

	get modelLength() {
		let length = 1;

		for ( let child of this.getChildren() ) {
			if ( child.name == 'ul' || child.name == 'ol' ) {
				for ( let item of child.getChildren() ) {
					length += item.modelLength;
				}
			}
		}

		return length;
	}
}
```
It will be changed to:
```javascript
mapper.registerViewElementLength( ViewListItemElement, ( element ) => {
	let length = 1;

	for ( let child of this.getChildren() ) {
		if ( child.name == 'ul' || child.name == 'ol' ) {
			for ( let item of child.getChildren() ) {
				length += item.modelLength;
			}
		}
	}

	return length;
} );
```
WDYT?

src/conversion/mapper.js
scofalik: At the beginning, I thought about it and did it like you said. But then I thought that it is clearer this way. Still, if you really want you can change parent/offset of `viewPosition` but it is much less explicit that this variable can be changed. If you really want I will move `viewPosition` to `data`.

src/conversion/model-to-view-converters.js
scofalik: Reminder: in F2F talk we agreed that code from `ModelConversionDispatcher#convertRename` will be put here making the whole event more clear. The drawback is that we will need to expose `ModelConversionDispatcher` instance to the `conversionApi` or as another parameter as we will need `convertInsert`, `convertMove` and `convertRemove` methods.

src/conversion/modelconversiondispatcher.js
scofalik: This will be simplified and will look like the other methods in `ModelConversionDispatcher`.

src/conversion/mapper.js
pjasiun: The question, as always, is: what if we will have more parameters here, similar to `viewPosition`? If everything is in `data` it is clear.

src/view/element.js
pjasiun: Yep, `Mapper.registerViewElementLength( ViewElementClass, lengthCallback );` is fine.

scofalik: Static or instance?

src/conversion/mapper.js
scofalik: Makes sense.

scofalik:
All reported issues fixed.

src/conversion/mapper.js
```diff
@@ -37,6 +49,37 @@ export default class Mapper {
...
 		this._viewToModelMapping = new WeakMap();
+
+		/**
+		 * A map containing callbacks between view element classes and functions evaluating length of view elements
+		 * in model.
+		 *
+		 * @private
+		 * @member {Map} engine.conversion.Mapper#_viewToModelLengthCallbacks
+		 */
+		this._viewToModelLengthCallbacks = new Map();
```
pjasiun: Why is it called `Callbacks`?

pjasiun:
It's R+ from me now, but I did not run the code. So ask someone else to check tests and check the coverage, or do it twice and merge the branch :)

scofalik:
Hopefully everything is alright....

Issue #436, opened 2016-05-23
view-converter-builder #fromAttribute() doesn't cover all characters


Reinmar:
The `/.*/` regexp will not pass line terminators: `\n`, `\r`, `\u2028` or `\u2029`.

We used to define such patterns as `/[\s\S]*/`.

Reinmar:
There's a PR, but it requires tests.

pjasiun:
Closed by https://github.com/ckeditor/ckeditor5-engine/issues/1239.

Reinmar:
It wasn't:

```js
/workspace/ckeditor5/packages/ckeditor5-engine/src/conversion/buildviewconverter.js:
  139  	 * @returns {module:engine/conversion/buildviewconverter~ViewConverterBuilder}
  140  	 */
  141: 	fromAttribute( key, value = /.*/ ) {
  142  		const pattern = {};
```

pjasiun:
Now it is closed. After the engine refactoring, we don't have `fromAttribute` method anymore.

PR #587, opened 2016-09-12
Changed pattern to collect all characters

pomek:
Fixes: #436

Unfortunately, the tests are missing because I don't have enough knowledge how it should look. 

OTOH, I talked F2F @pjasiun and we decided - if creating tests will be problematic, we can skip it.

### DIFF
--- a/src/conversion/buildviewconverter.js
+++ b/src/conversion/buildviewconverter.js
@@ -135,7 +135,7 @@ class ViewConverterBuilder {
 	 * @param {String|RegExp} [value] View attribute value.
 	 * @returns {engine.conversion.ViewConverterBuilder}
 	 */
-	fromAttribute( key, value = /.*/ ) {
+	fromAttribute( key, value = /[\s\S]*/ ) {
 		let pattern = {};
 		pattern[ key ] = value;
 


Reinmar:
We shouldn't skip writing any tests. I don't think that this isn't testable at all – perhaps it requires some knowledge. So I'd recommend passing this issue to someone else.

pjasiun:
> OTOH, I talked F2F @pjasiun and we decided - if creating tests will be problematic, we can skip it.

But then we realised that this parameter is not checked at all, so it is a bigger bug, isn't it? If so, changing the value does not fix anything.

pomek:
I had a test like this:

```js
// add attribute 'data-attr' to textAttributes array
// allow to have attribue 'data-attr' for span element (in schema)

it( 'should convert from view attribute to model attribute (2)', () => {
	const dataAttrValue = `Lorem\n\tipsum\tdolor\t\tsit\n\namet.\t.\n\t.`;

	buildViewConverter().for( dispatcher ).fromElement( 'span' ).toElement( 'span' );

	buildViewConverter().for( dispatcher )
		.fromAttribute( 'data-attr' /*, /[0-9]+/*/ )
		.toAttribute( ( viewElement ) => ( {
			key: 'data-attr',
			value: viewElement.getAttribute( 'data-attr' )
		} ) );

	const result = dispatcher.convert( new ViewContainerElement( 'span', { 'data-attr': dataAttrValue }, new ViewText( 'foo' ) ), objWithContext );
	modelRoot.appendChildren( result );

	expect( modelToString( result ) ).to.equal( `<span data-attr="${ dataAttrValue }">foo</span>` );
} );
```

I tried to modify the second parameter for `buildViewConverter.fromAttribute()` and nothing happens. Above tests always passed.

pjasiun:
The test looks fine. It means that this parameter is, most probably not checked at all. And most probably it is not passed to the `Matcher` correctly. 

It needs to be debugged. You can report a separate issue or add a comment to this one and expand the scope of it (and rename it). Anyway, changing the default value of the parameter which is not checked in the first place is pointless.

pomek:
I've created a new one ticket for it - https://github.com/ckeditor/ckeditor5-engine/issues/588

pjasiun:
I'm closing this PR for now, since we need to fix a different bug which we are not planning to work on right now.

PR #605, opened 2016-09-21
t/604: Introduced DataCotroller methods: stringify, toView and optional context parameter to parse.

### LINKED ISSUE #604: Improve data controller API
In the data controller we could:
 - extract part of the `get` as `stringify` methods, the same way we have `parse` extracted from `set`,
 - since model to view conversion there is a little bit tricky (https://github.com/ckeditor/ckeditor5-engine/blob/master/src/datacontroller.js#L121-L127), it could be extracted too, as a `toView` method,
 - parse could get context as an optional parameter instead of having hardcoded context: https://github.com/ckeditor/ckeditor5-engine/blob/master/src/datacontroller.js#L173

pjasiun:
Fixes #604.

### DIFF
--- a/src/datacontroller.js
+++ b/src/datacontroller.js
@@ -115,19 +115,44 @@ export default class DataController {
 	 */
 	get( rootName = 'main' ) {
 		// Get model range.
-		const modelRoot = this.model.getRoot( rootName );
-		const modelRange = ModelRange.createIn( modelRoot );
+		return this.stringify( this.model.getRoot( rootName ) );
+	}
 
+	/**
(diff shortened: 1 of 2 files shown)


src/datacontroller.js
```diff
@@ -163,16 +188,16 @@ export default class DataController {
 	 *
 	 * @see engine.DataController#set
 	 * @param {String} data Data to parse.
+	 * @param {String} [context='$root''] Base context in which view will be converted to the model. See:
```
scofalik: [context='$root''] <- double quote.

PR #616, opened 2016-09-23
T/583b Changes to the model required by List feature

scofalik:
A second batch of changes. Can be treated as a followup to #583. Mostly fixes. Look at commig messages for changelog.

### DIFF
--- a/src/conversion/mapper.js
+++ b/src/conversion/mapper.js
@@ -275,7 +275,7 @@ export default class Mapper {
 		let modelOffset = 0;
 
 		for ( let i = 0; i < viewOffset; i++ ) {
-			modelOffset += this._getModelLength( viewParent.getChild( i ) );
+			modelOffset += this.getModelLength( viewParent.getChild( i ) );
 		}
 
 		return modelOffset;
@@ -309,11 +309,10 @@ export default class Mapper {
 	 *		<b>foo</b>              -> 3 // Length of an element which is not mapped is a length of its children.
 	 *		<div><p>x</p><p>y</p>   -> 2 // Assuming that <div> is not mapped and <p> are mapped.
 	 *
-	 * @private
 	 * @param {engine.view.Element} viewNode View node.
 	 * @returns {Number} Length of the node in the tree model.
 	 */
-	_getModelLength( viewNode ) {
+	getModelLength( viewNode ) {
 		if ( this._viewToModelMapping.has( viewNode ) ) {
 			const callback = this._viewToModelLengthCallbacks.get( viewNode.name );
 
@@ -324,7 +323,7 @@ export default class Mapper {
 			let len = 0;
 
 			for ( let child of viewNode.getChildren() ) {
-				len += this._getModelLength( child );
+				len += this.getModelLength( child );
 			}
 
 			return len;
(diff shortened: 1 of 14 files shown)


src/conversion/mapper.js
```diff
@@ -307,11 +309,10 @@ export default class Mapper {
 	 *		<b>foo</b>              -> 3 // Length of an element which is not mapped is a length of its children.
 	 *		<div><p>x</p><p>y</p>   -> 2 // Assuming that <div> is not mapped and <p> are mapped.
 	 *
-	 * @private
 	 * @param {engine.view.Element} viewNode View node.
```
pjasiun: Why it is not private anymore?

src/conversion/model-to-view-converters.js
```diff
@@ -307,17 +322,46 @@ export function move() {
...
+		viewWriter.remove( viewRange );
+		conversionApi.mapper.unbindModelElement( data.item );
 	};
 }
 
+// Helper function for `remove` and `move` converters. It returns a ViewRange that starts at ViewPosition `start` and
+// includes `size` characters.
+// This method is used to find a ViewRange basing on ModelPosition and ModelTextProxy item size in `move` and `remove`
+// converters where it is impossible to just map positions because those positions already are invalid in model
+// (because they got moved or removed).
```
pjasiun: Can't you simply use size and `Range.createFromPositionAndShift`?

scofalik: No, because `Range.createFromPositionAndShift` is "dumb", meaning that it will create a range starting in given position and ending in a position in same parent but changed offset.

Unfortunately this is wrong, because characters in view can be spread among multiple `AttributeElement`s. Consider this:
model: `^foobaruuu` with `foo` having bold, `bar` having italic and `uuu` having underline.
view: `<p>^<b>foo</b><i>bar</i><u>uuu</u></p>`
Now, if the `ModelText` has `offsetSize` equal to `3` and starting position is mapped as it is, creating a `ViewRange` using `createFromPositionAndShift` will lead to wrong range -- containing whole text, instead of only `foo`.

src/conversion/mapper.js
scofalik: I found it useful to use in `List` feature converters. It looks like function like this may be useful in creating more complicated converters. After all this is not some kind of magic, it just returns view element's length in model. It's even easy to describe.

https://github.com/ckeditor/ckeditor5-list/blob/t/1b/src/converters.js#L542

src/conversion/model-to-view-converters.js
pjasiun: So that's fine for me. We can consider making `Range.createFromPositionAndShift` smarter in the future, but for now it is fine as it is.

src/dataprocessor/xmldataprocessor.js
```diff
@@ -89,7 +89,7 @@ export default class XmlDataProcessor {
 		const domFragment = this._toDom( data );
 
 		// Convert DOM DocumentFragment to view DocumentFragment.
-		return this._domConverter.domToView( domFragment );
+		return this._domConverter.domToView( domFragment, { keepOriginalCase: true } );
```
Reinmar: This change was done because XmlDP is used in the dev utils to parse textual representation of the model. We use elements like <listItem> and it should be converted to listItem in the view and then in the model. Not the cleanest change, but perhaps it makes sense that XmlDP works this way (comparing to HtmlDP's lowercasing everything).

Issue #562, opened 2016-08-10
View writer does not merge attributes if range starts in text


pjasiun:
Input: 
```
<container:p>foo{<attribute:b view-priority="1" foo="1"></attribute:b>]</container:p>
```
Wrap attribute:
```
<attribute:b view-priority="1" bar="2"></attribute:b>
```
Expected:
```
<container:p>foo{<attribute:b view-priority="1" bar="2" foo="1"></attribute:b>]</container:p>
```
Acctual:
```
<container:p>foo[<attribute:b view-priority="1" bar="2"><attribute:b view-priority="1" foo="1"></attribute:b></attribute:b>
```

Reinmar:
Why iteration 3? Does this have some significant effect on the editor?

pjasiun:
This is the reason why testing tool does not work as expected in some cases.

Reinmar:
@scofalik wrote some guides on how this should be resolved: https://github.com/ckeditor/ckeditor5-engine/pull/618#issuecomment-250694843.

PR #618, opened 2016-09-29
View writer: Attributes will be merged if range starts in text.

pomek:
Fixes: #562

### DIFF
--- a/src/view/writer.js
+++ b/src/view/writer.js
@@ -394,6 +394,14 @@ export function wrap( range, attribute ) {
 		return range;
 	}
 
+	const isStartTextParent = range.start.parent instanceof Text;
+	const isEndAttributeBefore = range.end.nodeBefore instanceof AttributeElement;
+
+	// Range starts in text and indicates on an attribute.
+	if ( isStartTextParent && isEndAttributeBefore && wrapAttributeElement( attribute, range.end.nodeBefore ) ) {
+		return range;
+	}
+
 	// Range around one element.
 	if ( range.end.isEqual( range.start.getShiftedBy( 1 ) ) ) {
 		const node = range.start.nodeAfter;
--- a/tests/view/writer/wrap.js
+++ b/tests/view/writer/wrap.js
@@ -270,5 +270,13 @@ describe( 'writer', () => {
 				'</container:p>'
 			);
 		} );
+
+		it( 'should merge different attributes with the same priority into one', () => {
+			test(
+				'<container:p>foo{<attribute:b view-priority="1" foo="1"></attribute:b>]</container:p>',
+				'<attribute:b view-priority="1" bar="2"></attribute:b>',
+				'<container:p>foo{<attribute:b view-priority="1" bar="2" foo="1"></attribute:b>]</container:p>'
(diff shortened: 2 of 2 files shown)


scofalik:
Proposed solution fails for this case:

```javascript
test(
	'<container:p>fo{o<attribute:b view-priority="1" foo="1"></attribute:b>]</container:p>',
	'<attribute:b view-priority="1" bar="2"></attribute:b>',
	'<container:p>fo[<attribute:b view-priority="1" bar="2">o<attribute:b view-priority="1" foo="1"></attribute:b></attribute:b>]</container:p>'
);
```

I understand what was the reasoning behind proposed condition (`isStartTextParent && isEndAttributeBefore`) but:
1) it was incorrectly implemented (the issue says about position at the end of text node),
2) is too weak (even if it was "at the end of text node", it would still work incorrectly, for example `foo{<i>bar</i><b>bom</b>]`. Also what about `foo{<b>bar</b>}bom`? Those examples should also work and all the cases are similiar).

The quickest and simplest solution that comes to my mind is "normalizing" positions by checking whether they are at the end of the beginning view `Text` node and if so, moving them to before/after or that `Text` node. I don't know if that's all that is needed to be done but it is first step nevertheless. When doing it, check if such function or such code already exists so you won't repeat existing code.

General advice - try to think on general solutions, not trying to fix one special case.

On the other hand, I don't like how `wrapChildren` works. I am not sure, but I have a feeling that the top-down merging (<b foo=1><b bar=2></b></b> => <b foo=1 bar=2></b>) should be done there. This might spare us some special case conditions, like:
`if ( range.end.isEqual( range.start.getShiftedBy( 1 ) ) ) {`
`if ( rangeSpansOnAllChildren( range ) && wrapAttributeElement( attribute, range.start.parent ) ) {`
This is just a wild guess but I have a feeling that `wrapChildren` should somehow use `wrapAttributeElement`.

Okay, I checked it.
This:
```javascript
test(
	'<container:p>foo[<attribute:i>bar</attribute:i><attribute:b view-priority="1" foo="1"></attribute:b>]</container:p>',
	'<attribute:b view-priority="1" bar="2"></attribute:b>',
	'<container:p>foo[<attribute:b view-priority="1" bar="2"><attribute:i view-priority="10">bar</attribute:i></attribute:b><attribute:b view-priority="1" bar="2" foo="1"></attribute:b>]</container:p>'
);
```
Fails. The actual result is `<container:p>foo[<attribute:b view-priority="1" bar="2"><attribute:i view-priority="10">bar</attribute:i><attribute:b view-priority="1" foo="1"></attribute:b></attribute:b>]</container:p>` which shows that attributes are not merged even though they could.
However, merging in `wrapChildren` is more difficult scenario because there are many possible cases.

To sum up:
- prepare correct solution so it won't introduce regressions and will cover also similar cases,
- you can try to solve the issue with the last example (when solved correctly, I hope that you will not have to make position normalization I proposed and we will also be able to get rid of some already existing special case conditions) however it will be trickier than it looks like so if it takes you too much time just create followup issue and we will discuss how and when we will solve it.

PR #620, opened 2016-09-29
T/379

### LINKED ISSUE #379: View: Rendering and handling (in mutations) spaces at block boundaries and multiple spaces.
We need to figure out what to do with https://github.com/ckeditor/ckeditor5-design/issues/61

Imagine that your model contains paragraph with `"foo "`. This needs to be rendered as `<p>foo&nbsp;</p>`.

Another case: user is typing at the end of a paragraph – "foo bar". We have "foo", then user presses space. That space must be rendered as `&nbsp;` (we actually get `&nbsp;` in a mutation). Then, two things may happen:

1. If user moved caret somewhere else and back to that place or we rerendered the selection, then browser *may* insert "b" without turning that nbsp into a normal space.
2. Typically (at least Chrome) converts that nbsp into normal space once "b" was typed. I checked that I can move selection around or load such data and the conversion takes place. I'm not sure however that it's going to work for all browsers and in all cases, so it's safe to assume that it won't.

So:

* A space typed at the end of a paragraph should be inserted as a normal space into the model. We may add a feature that <kbd>Alt + Space</kbd> (or whichever keystroke we'll choose from https://en.wikipedia.org/wiki/Non-breaking_space#Keyboard_entry_methods) will insert an `nbsp` to the model. Existence of such a feature shows that the model must always contain clean data.
* Multiple subsequent spaces must be rendered correctly.
* Mutations should rather contain un-quirked text content. So we assume that a space typed at the end of block, even though inserted as an nbsp by the browser should really be a normal space.
* However, the above should not be true for characters which were added as nbsps into the model and rendered as such. DOM converter must not convert them to normal spaces. This may be tricky ;/.

Option – we may for now use the CSS property which I described in ckeditor/ckeditor5-design#61 and do not care about this at all. The editor should work perfectly.

scofalik:
Fixes #379 .

### DIFF
--- a/src/model/documentfragment.js
+++ b/src/model/documentfragment.js
@@ -84,6 +84,16 @@ export default class DocumentFragment {
 	}
 
 	/**
+	 * Artificial parent of `DocumentFragment`. Returns `null`. Added for compatibility reasons.
+	 *
+	 * @readonly
+	 * @type {null}
+	 */
+	get parent() {
+		return null;
+	}
+
+	/**
 	 * Gets the child at the given index. Returns `null` if incorrect index was passed.
 	 *
 	 * @param {Number} index Index of child.
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -3,7 +3,7 @@
  * For licensing, see LICENSE.md.
  */
 
-/* globals Range, Node */
+/* globals Range, Node, NodeFilter */
 
 import ViewText from './text.js';
 import ViewElement from './element.js';
@@ -11,9 +11,13 @@ import ViewPosition from './position.js';
 import ViewRange from './range.js';
 import ViewSelection from './selection.js';
 import ViewDocumentFragment from './documentfragment.js';
+import ViewContainerElement from './containerelement.js';
+import ViewTreeWalker from './treewalker.js';
 import { BR_FILLER, INLINE_FILLER_LENGTH, isBlockFiller, isInlineFiller, startsWithFiller, getDataWithoutFiller } from './filler.js';
 
 import indexOf from '../../utils/dom/indexof.js';
+import getAncestors from '../../utils/dom/getancestors.js';
+import getCommonAncestor from '../../utils/dom/getcommonancestor.js';
 
 /**
  * DomConverter is a set of tools to do transformations between DOM nodes and view nodes. It also handles
@@ -69,6 +73,20 @@ export default class DomConverter {
 		 * @member {WeakMap} engine.view.DomConverter#_viewToDomMapping
 		 */
 		this._viewToDomMapping = new WeakMap();
+
+		/**
+		 * Tag names of DOM `Element`s which are considered pre-formatted elements.
+		 *
+		 * @type {Array.<String>}
+		 */
+		this.preNodes = [ 'pre' ];
+
+		/**
+		 * Tag names of DOM `Element`s which are considered block elements.
+		 *
+		 * @type {Array.<String>}
+		 */
+		this.blockNodes = [ 'p', 'div', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ];
 	}
 
 	/**
@@ -110,7 +128,9 @@ export default class DomConverter {
 	 */
 	viewToDom( viewNode, domDocument, options = {} ) {
 		if ( viewNode instanceof ViewText ) {
-			return domDocument.createTextNode( viewNode.data );
+			const textData = this._processDataFromViewText( viewNode );
+
+			return domDocument.createTextNode( textData );
 		} else {
 			if ( this.getCorrespondingDom( viewNode ) ) {
 				return this.getCorrespondingDom( viewNode );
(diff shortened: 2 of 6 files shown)


scofalik:
Related PRs:
https://github.com/ckeditor/ckeditor5-utils/pull/86 (already closed)
https://github.com/ckeditor/ckeditor5-typing/pull/50

src/model/documentfragment.js
```diff
@@ -84,6 +84,16 @@ export default class DocumentFragment {
 	}
 
 	/**
+	 * Artificial parent of `DocumentFragment`. Returns `null`. Added for compatibility reasons.
+	 *
+	 * @readonly
+	 * @type {null}
```
Reinmar: I'm afraid you need to use the whole `@member` notation.

src/model/documentfragment.js
```diff
@@ -84,6 +84,16 @@ export default class DocumentFragment {
 	}
 
 	/**
+	 * Artificial parent of `DocumentFragment`. Returns `null`. Added for compatibility reasons.
```
Reinmar: What compatibility reasons do you mean?

src/view/domconverter.js
```diff
@@ -69,6 +73,20 @@ export default class DomConverter {
 		 * @member {WeakMap} engine.view.DomConverter#_viewToDomMapping
 		 */
 		this._viewToDomMapping = new WeakMap();
+
+		/**
+		 * Tag names of DOM `Element`s which are considered pre-formatted elements.
+		 *
+		 * @type {Array.<String>}
```
Reinmar: Again, let's be consistent and use `@member`.

src/view/domconverter.js
```diff
@@ -651,4 +673,173 @@ export default class DomConverter {
...
+		const direction = getNext ? 'nextNode' : 'previousNode';
+		const document = node.ownerDocument;
+		const treeWalker = document.createTreeWalker( document.childNodes[ 0 ], NodeFilter.SHOW_TEXT );
+
+		treeWalker.currentNode = node;
+
+		const touchingNode = treeWalker[ direction ]();
+
+		if ( touchingNode !== null ) {
+			const lca = getCommonAncestor( node, touchingNode );
```
Reinmar: What's "l" in "lca"?

scofalik: Lowest: https://en.wikipedia.org/wiki/Lowest_common_ancestor

src/model/documentfragment.js
scofalik: Compability with other view/model classes, like `Element`. AFAIR it came up at some point. I think that I was checking elements parents until I found something or got `null`. The parent of `DocumentFragment` was `undefined`. I know I could just do `if ( !element.parent ) {` but I think having this unified is better.

src/view/domconverter.js
```diff
@@ -266,7 +286,9 @@ export default class DomConverter {
 			if ( isInlineFiller( domNode ) ) {
 				return null;
 			} else {
-				return new ViewText( getDataWithoutFiller( domNode ) );
+				const textData = this._processDataFromDomText( domNode );
+
+				return textData === '' ? null : new ViewText( textData );
```
scofalik: Unrelated improvement. `domToView` is used in `domChildrenToView`, I just didn't want empty `ViewText` in view. However, the docs (`@returns` sentence) should be updated. Of course the drawback now is that if you want to just convert a dom Text instance and it is empty, you'll get `null` instance empty `ViewText`.

src/model/documentfragment.js
Reinmar: I've just checked that the native doc frag also has a parrentNode and parentElement properties, so it makes some sense here.

src/view/domconverter.js
Reinmar: You can get `null` already if the node is a filler, so I guess this is fine... I'll align the docs.

Reinmar: But... I don't see a test for this change.

Reinmar:
`/ckeditor5/engine/view/domconverter.js.html`:

screen shot 2016-10-06 at 17 25 31

`/ckeditor5/engine/model/liveposition.js.html`:

![screen shot 2016-10-06 at 17 25 51]
(https://cloud.githubusercontent.com/assets/156149/19158712/0df44e06-8bea-11e6-9dfe-f75b02e8c46f.png)

Reinmar:
I found one bug – `editor.setData( '<p>x &nbsp; x</p>' );` produce a content with a single space between "x"es.

Reinmar:
The same happens with `editor.setData( '<p>x&nbsp;&nbsp;x</p>' );` and in fact every nbsp which I try to load into the editor.

scofalik:
I pushed improvements for `DomConverter` which fixed incorrect case you posted. CC is 100% now, too.

tests/view/domconverter/view-to-dom.js
```diff
@@ -164,10 +166,192 @@ describe( 'DomConverter', () => {
...
+					}
+
+					const domElement = converter.viewToDom( viewElement, document );
+					const data = domElement.innerHTML.replace( /&nbsp;/g, '_' );
+
+					expect( data ).to.equal( output );
+				}
+
+				// At the beginning.
+				test( ' x', '_x' );
```
Reinmar: Those `test()` function should be moved out of `it()`. They should call their own `it()` with proper names. Otherwise you won't be able to know what's actually failing. And it's super frustraiting if e.g. you know that 3rd cases fails but you don't know which of the next ones fails too, because first failed one stops test execution.

PR #623, opened 2016-10-05
Changed: introduced selection direction handling in DOM<->view selection conversion.

### LINKED ISSUE #401: Support selection direction
To simplify the first implementation `Renderer` and `DomConverter` now ignore selection direction. They should handle it.

See https://github.com/ckeditor/ckeditor5-engine/issues/574 for awesome screencasts.

scofalik:
Fixes #401.

### DIFF
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -360,12 +360,15 @@ export default class DomConverter {
 	domSelectionToView( domSelection ) {
 		const viewSelection = new ViewSelection();
 
+		const isBackward = this.isDomSelectionBackward( domSelection );
+
 		for ( let i = 0; i < domSelection.rangeCount; i++ ) {
+			// DOM Range have correct start and end, no matter what is the DOM Selection direction. So we don't have to fix anything.
 			const domRange = domSelection.getRangeAt( i );
 			const viewRange = this.domRangeToView( domRange );
 
 			if ( viewRange ) {
-				viewSelection.addRange( viewRange );
+				viewSelection.addRange( viewRange, isBackward );
 			}
 		}
 
@@ -672,6 +675,31 @@ export default class DomConverter {
 	 */
 	isDocumentFragment( node ) {
 		return node && node.nodeType == Node.DOCUMENT_FRAGMENT_NODE;
+	}
+
+	/**
+	 * Returns `true` if given selection is a backward selection, that is, if it's `focus` is before `anchor`.
+	 *
+	 * @param {Selection} DOM Selection instance to check.
+	 * @return {Boolean}
+	 */
+	isDomSelectionBackward( selection ) {
+		if ( selection.isCollapsed ) {
+			return false;
+		}
+
+		// Since it takes multiple lines of code to check whether a "DOM Position" is before/after another "DOM Position",
+		// we will use the fact that range will collapse if it's end is before it's start.
+		const range = new Range();
+
+		range.setStart( selection.anchorNode, selection.anchorOffset );
+		range.setEnd( selection.focusNode, selection.focusOffset );
+
+		const backward = range.collapsed;
+
+		range.detach();
+
+		return backward;
 	}
 
 	/**
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -681,7 +681,7 @@ export default class DomConverter {
 	 * Returns `true` if given selection is a backward selection, that is, if it's `focus` is before `anchor`.
 	 *
 	 * @param {Selection} DOM Selection instance to check.
-	 * @return {Boolean}
+	 * @returns {Boolean}
 	 */
 	isDomSelectionBackward( selection ) {
 		if ( selection.isCollapsed ) {
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -422,7 +422,7 @@ export default class Renderer {
 	_updateSelection() {
 		// If there is no selection - remove it from DOM elements that belongs to the editor.
 		if ( this.selection.rangeCount === 0 ) {
-			this._removeDomSelction();
+			this._removeDomSelection();
 
 			return;
 		}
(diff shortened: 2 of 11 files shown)


src/view/renderer.js
```diff
@@ -421,7 +421,7 @@ export default class Renderer {
 	_updateSelection() {
 		// If there is no selection - remove it from DOM elements that belongs to the editor.
 		if ( this.selection.rangeCount === 0 ) {
-			this._removeDomSelction();
+			this._removeDomSelection();
```
scofalik: I fixed typo.

src/view/selection.js
```diff
@@ -117,8 +117,8 @@ export default class Selection {
 	 * @type {engine.view.EditableElement|null}
 	 */
 	get editableElement() {
-		if ( this.rangeCount ) {
-			return this.getFirstPosition().editableElement;
+		if ( this.anchor ) {
+			return this.anchor.editableElement;
```
scofalik: To be honest, I am not sure how `Selection` will work if there are ranges from different view tree roots. `getFirstPosition()` returns first position "place-wise", that means a position which `isBefore()` any other position. It does not to be last focused editable, etc.

I think it's better and more predictible to get `editableElement` from `anchor`. Even more, since I changed `Renderer` to use `anchor` and `focus` during conversion, rather than all ranges.

tests/view/renderer.js
```diff
@@ -918,26 +920,6 @@ describe( 'Renderer', () => {
 			expect( domSelection.getRangeAt( 0 ).collapsed ).to.equal( true );
 		} );
 
-		it( 'should not add ranges if different editable is selected', () => {
```
Reinmar: 100%CC is just a help for us to understand whether some case wasn't missed by test. Here, you clearly add a feature and write 0 tests for it which means that if this feature breaks in the future (or disappears) there's a high chance that it won't be noticed by tests.

So don't head for 100%CC but for tests which validate feature correctness. Then you'll have 100%CC as well.

BTW, manual test is missing as well. There are clear cases like dragging from right to left which were broken. They should work now. I propose to create a single, selection rendering test for it and check multiple scenarios there.

src/view/domconverter.js
```diff
@@ -651,4 +654,29 @@ export default class DomConverter {
 	isDocumentFragment( node ) {
 		return node && node.nodeType == Node.DOCUMENT_FRAGMENT_NODE;
 	}
+
+	/**
+	 * Returns `true` if given selection is a backward selection, that is, if it's `focus` is before `anchor`.
+	 *
+	 * @param {Selection} DOM Selection instance to check.
+	 * @return {Boolean}
```
Reinmar: returns

src/view/domconverter.js
```diff
@@ -338,12 +338,15 @@ export default class DomConverter {
...
+		const isBackward = this.isDomSelectionBackward( domSelection );
+
 		for ( let i = 0; i < domSelection.rangeCount; i++ ) {
+			// DOM Range have correct start and end, no matter what is the DOM Selection direction. So we don't have to fix anything.
 			const domRange = domSelection.getRangeAt( i );
 			const viewRange = this.domRangeToView( domRange );
 
 			if ( viewRange ) {
-				viewSelection.addRange( viewRange );
+				viewSelection.addRange( viewRange, isBackward );
```
Reinmar: Hm... this isn't entirely correct, cause only the last range can be backward. But I guess this shouldn't cause any problems.

src/view/selection.js
Reinmar: @pjasiun?

Reinmar:
I made initial review, but without tests I can't fully verify the implementation.

However, I found two issues:

* one test is failing in http://tests.ckeditor.dev:1030/build/modules/amd/tests/engine/view/renderer
* the inf-loop selection warning is kicking in immediately which means that something totally doesn't work

tests/view/renderer.js
scofalik: You are right that manual test is missing, and unit test can be done too, checking whether selection is properly converted to backward selection.

However in the case of commented test it's prefectly valid to remove it, as I kinda removed the functionallity here. Before, theoretically, the code supported rendering mutli-range view selection to DOM. This test checked on of the cases of that scenario. Since this scenario is now gone (now only view selection anchor and focus is rendered) the test is incorrect.

src/view/domconverter.js
```diff
@@ -651,4 +654,29 @@ export default class DomConverter {
...
 		return node && node.nodeType == Node.DOCUMENT_FRAGMENT_NODE;
 	}
+
+	/**
+	 * Returns `true` if given selection is a backward selection, that is, if it's `focus` is before `anchor`.
+	 *
+	 * @param {Selection} DOM Selection instance to check.
+	 * @return {Boolean}
+	 */
+	isDomSelectionBackward( selection ) {
```
scofalik: Hmm, I thought that I commented on it, but there are multpile methods in `DomConverter` that actually should be moved to `uitls/dom`. This included.

src/view/domconverter.js
scofalik: I could push all ranges to one array and then use `setRanges` but what's the point? The code is more messy and you have one more variable. When you add a range it does not matter how the other were selected, both from API point of view, DOM point of view and logical point of view.

scofalik:
> the inf-loop selection warning is kicking in immediately which means that something totally doesn't work

What does it mean immediately? I am aware of problem of infinite loop kicking in at the edge of `<strong>` element but not immediately. Also, I didn't want to solve this problem in this PR, because provided code solves problem described in connected issue and there is another issue about infinite selection loop.

scofalik:
> one test is failing in http://tests.ckeditor.dev:1030/build/modules/amd/tests/engine/view/renderer

It's all green for me so we might have "wrong branch" issue here. Could you please review #620 (and related PR)? I right now have `t/379` merged to `t/401` locally and I am on `t/49` in `typing` repo and it works for me.

scofalik:
What error do you get in that test? I checkedout to `master` on typing and rebased this branch with `master` and it still works for me.

scofalik:
I've added missing tests. I don't know why all commits got "refreshed", I just rebased with `t/401`.

src/view/domconverter.js
Reinmar: It's definitely premature, but I know that browsers hate ranges because they are "live". So I'm really worried that creating a range instance must create many listeners and will be slow, which in case of selection rendering may have a negative impact on the performance, but very hard to estimate. Another thing which is hard to estimate is memory pressure caused by ranges. 

But when I started writing down how an algorithm using `compareDocPos` would look like I understood why you used ranges :D. Let's not optimise it now.

Reinmar:
> What does it mean immediately? I am aware of problem of infinite loop kicking in at the edge of <strong> element but not immediately. Also, I didn't want to solve this problem in this PR, because provided code solves problem described in connected issue and there is another issue about infinite selection loop.

OK, I've got this error yesterday so quickly that I thought that the rendering must be broken :D. Sorry for the lack of trust.

Reinmar:
> What error do you get in that test? I checkedout to master on typing and rebased this branch with master and it still works for me.

http://tests.ckeditor.dev:1030/build/modules/amd/tests/engine/view/renderer#render%20should%20not%20add%20inline%20filler%20after%20text%20node

I ran `gulp pull` and merged master into `t/401`. The test still fail.

scofalik:
It seems that the bug occurs only on OSX Chrome and OSX Safari.

scofalik:
The bug is connected with how `Selection#getRangeAt()` and `Selection#anchor` / `Selection#offset` works. It seems that Chrome stores originally added range (i.e. next to a text node) and returns it on `getRangeAt` but `anchor` and `offset` are correct (in text node). Since `getRangeAt` returns original range, the test was okay.

On Safari / OSX though, `getRangeAt` returns "normalized" range. I leave it here as a note. It will have to be fixed at some point.

As far as the test itself is concerned, it will be fixed so it still check something but not base on DOM selection.

scofalik:
I rebased with master, fixed conflicts, checked coverage and fixed mistakes mentioned in review.

Reinmar:
There's a regression (https://github.com/ckeditor/ckeditor5-engine/issues/629) which we'll need to fix ASAP, but let's merge the code since it works fine and we've got a stack of other PRs waiting.

PR #625, opened 2016-10-06
Fixed: LiveRange throwing after executing RenameOperation.

### LINKED ISSUE #624: LivePosition throws error when `RenameOperation` is used
When `RenameOperation` is a cause of `change` event on document, it makes `LivePosition` throw an error. This is because `RenameOperation` does not return `range` property needed by `LivePosition`. This issue was already addressed in `LiveRange` but it was ommited in `LivePosition`. Same fix will apply.

scofalik:
Fixes #624

### DIFF
--- a/src/model/liveposition.js
+++ b/src/model/liveposition.js
@@ -123,12 +123,16 @@ export default class LivePosition extends Position {
  */
 function bindWithDocument() {
 	/*jshint validthis: true */
+	// Operation types handled by LivePosition (these are operations that change model tree structure).
+	const supportedTypes = new Set( [ 'insert', 'move', 'remove', 'reinsert' ] );
 
 	this.listenTo(
 		this.root.document,
 		'change',
 		( event, type, changes ) => {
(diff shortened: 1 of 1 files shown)


src/model/liveposition.js
```diff
@@ -123,12 +123,16 @@ export default class LivePosition extends Position {
  */
 function bindWithDocument() {
 	/*jshint validthis: true */
+	// Operation types handled by LivePosition (these are operations that change model tree structure).
+	const supportedTypes = new Set( [ 'insert', 'move', 'remove', 'reinsert' ] );
```
maxbarnas: Is there a reason to move it up and make it a property of LivePosition class?

scofalik: It might look better (if it is static), but I don't want to create new `Set` or `Array` each time it is called so it would have to be something like:

```javascript
static get supportedTypes() {
    return supportedTypes;
};
// ... after class declaration
const supportedTypes = new Set( ... );
```

I think we are fine with how it looks like now.

Issue #399, opened 2016-05-11
Selection in Mutation


pjasiun:
This one is tricky.

View and view events should build an abstraction laver over the DOM. In case of the mutation it means that old and new children are view nodes as well selection should be a view selection.

The problem is that new nodes are not yet in the view tree, they are only listed as new children of the parent element which is in the view tree. But that parent element does not have new children in the list of the child node (the change have to go to the model and back to be applied on the view document).

Now the selection. It usually land in that new node. It means that it need to convert DOM selection positions to the positions in the node which have no corresponding view elements in the view tree. It means that we can not use standard `DomConverter` method which assume that DOM and View are similar, or return null if it is not possible to return corresponding position. We need to find the way to put the position in the new view element which are only on the list of the mutations.

Fortunately we should not use this position very often. In most cases model will know better where to put selection after changes. On the other hand this information might be useful for instance for the typing: we may want to put the caret after inserted text on typing, but have selected word in case of autocorrection. Or if the autocorrection changed only one letter in the word the whole word should be selected or the caret should be after that word. To some extent, we are able to guess where the selection should be put based on the change which is done, but without the information where the native selection was we may have some issues.

pjasiun:
There are more and more cases in which it will be tricky to predict where the selection should be put. The simplest case is: `foo []bar` and user press space so we have `foo  bar`. It is not obvious which of these spaces is the result of the mutation and where selection should be put (`foo [] bar` or `foo  []bar`).

Reinmar:
Actually, when you type the second space one of them is usually turned into `nbsp` (by a browser ofc), and second is a normal space. That makes things even more complex :D.

Plus, of course, this is not only about spaces, but about all possible repeated characters.

pjasiun:
This is why we should fix https://github.com/ckeditor/ckeditor5-engine/issues/379 too. `nbsp`s should stay in the browser and you should have regular spaces on the View level.

scofalik:
From what I take after toying a bit with `MutationObserver`, `SelectionObserver`, and `Typing` feature I see that we don't need to set up selection in `Typing`. It's set automatically after character is inserted, thanks to `LivePosition`s.

The only problem with this approach is `diff`. Now, when there are multiple same characters in two diffed words, we don't know which character was inserted. This affects position on which new character is inserted, which in return affects how `LivePosition` in `LiveSelection` behaves.

[...]

PR #626, opened 2016-10-10
ViewSelection instance is passed with viewMutations in mutations events, if possible.

scofalik:
This implements some ideas and fixes some issues described in #399.

This PR introduces `viewSelection` parameter to `mutations` event. `viewSelection` is instance of `engine.view.Selection` created basing on actual DOM selection. `viewSelection` is set only if DOM selection got correctly mapped to view. This means that it is available only when after mutations selection is placed in an already existing DOM node.

For now I feel this is enough. It solves an issue in `Typing` and right now we don't know whether anything else will use mutations.

Additionally, if DOM selection ends up in non-existing DOM node, we need to add a node in model anyway. This means that we should be able to handle setting model selection at that point.

### DIFF
--- a/src/view/observer/mutationobserver.js
+++ b/src/view/observer/mutationobserver.js
@@ -238,7 +238,7 @@ export default class MutationObserver extends Observer {
  * Array of mutations.
  * For mutated texts it will be {@link engine.view.Document~MutatatedText} and for mutated elements it will be
  * {@link engine.view.Document~MutatatedElement}. You can recognize the type based on the `type` property.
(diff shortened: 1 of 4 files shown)


Reinmar:
screen shot 2016-10-17 at 14 01 42

Reinmar:
Please correct the CC and and I'm merging this.

Reinmar:
One more thing is missing – definition of the new param here: https://github.com/ckeditor/ckeditor5-engine/pull/626/files#diff-57e194ad013563b5752afa147f447d67R237

Issue #590, opened 2016-09-12
Filler (?) is infinitely created in a specific case


Reinmar:
1. Open https://ckeditor5.github.io/.
2. Open dev tools and inspect a paragraph in which you'll type.
3. Make a collapsed selection.
4. Press ctrl+b.
5. Type one letter.

See that some text node is created infinitely.

http://giphy.com/gifs/3o6ZsVwMmFFLZuJSRW

pjasiun:
Related: https://github.com/ckeditor/ckeditor5-engine/issues/462, https://github.com/ckeditor/ckeditor5-engine/issues/613.

Reinmar:
OK, what @pjasiun meant by "related" is that this bug is actually caused by #462. So #462 must be resolved first and this is a pretty big topic.

pjasiun:
The proper order of fixing is: https://github.com/ckeditor/ckeditor5-engine/issues/259 -> https://github.com/ckeditor/ckeditor5-engine/issues/462 -> this.

Reinmar:
Seems to be fixed by https://github.com/ckeditor/ckeditor5-engine/pull/634.

Issue #633, opened 2016-10-17
Caret moved to the end of paragraph when filling chars removed.


scofalik:
After fixing issue #259 editor finally is able to create empty inline tags (like `<em></em>`) with filling characters, basing on `model.Document#selection` attributes.

However, there is a bug when user types a letter at such position:
`<p>foo<em>&#8203;&#8203;&#8203;&#8203;&#8203;&#8203;&#8203;^</em>bar</p>`. After typing one character at such position, all filling characters are (correctly) removed. But the selection is moved to the end of paragraph.

This is a followup issue to #259 and #462. It was a hidden bug.

Reinmar:
I think it's a DUP of #590.

Reinmar:
And of https://github.com/ckeditor/ckeditor5-typing/issues/47.

scofalik:
It is a DUP of #47

Issue #603, opened 2016-09-21
Selection change:attribute and change:range are fired far too often


Reinmar:
They are fired multiple times whatever you do. This makes them useless, because even if e.g. attributes don't really change, the event is fired, so https://github.com/ckeditor/ckeditor5-typing/issues/21 cannot use this event.

This issue is also related to https://github.com/ckeditor/ckeditor5-engine/issues/259.

pjasiun:
I think it is not related to #259. It seems to be a dup of #259.

scofalik:
What we know is that selection events are mess, so this is not a matter of specific issue, it should be re-thought as a whole.

One thing is how this is solved in model. Now we know it is wrong that `change:attribute` is fired even when attribute has not really changed -- it is in fact changed each time we fire `change:range`. We should not fire `change:attribute` if the attribute key is same. Maybe we should not even fire `LiveSelection#_updateAttributes` for every `change:range`. For example, we might not fire this function if selection parent is same, but it gets complicated when you think of multiple ranges selection.

The other thing is how often model selection events are fired because of DOM/view. On one hand it seems wrong that `change:range` is fired after each letter is typed. OTOH, however, features may expect that, because, all in all, selection has actually changed. Unfortunately, IDK how DOM<->view<->model selection update works in details. Maybe DOM selection is changed too often? We know that we have so-called "selection infinite loops". Maybe we fire too many view selection changes?

I'd not only review model selection implementation but also DOM and view selection conversion/handling. What I know is that typing a letter should fire just one `change:range` event at most.

maxbarnas:
While working on https://github.com/ckeditor/ckeditor5-typing/issues/21 I made fix prototype. The thing that was needed was a listener for `changesDone` in ckeditor5-typing/input.js: https://github.com/ckeditor/ckeditor5-typing/commit/25e0429adffb8e0c07e7099c9336dd3afb914713#diff-8052e4286a369903aa5d7f6c2d8b7190R49.

Without this I was unable to wrap all `change:attribute` events with my lock mechanism in `changebuffer.js`. I noticed `changesDone` handler in https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/document.js#L112 and used it.

This makes handling all `change:attribute` events impossible from `input.js` and needs fix.

maxbarnas:
Also what do you thing about creating new, more granular events to help with situation described in this issue?

Reinmar:
[...]

PR #634, opened 2016-10-17
Handling attributes, ranges and their changes in (Live)Selection.

### LINKED ISSUE #259: Check whether it's optimal that selection updates its own attributes on #changesDone and #change:range
Quote from https://github.com/ckeditor/ckeditor5-core/pull/234

> I'm not sure on what events the selection should update its attributes itself. This should not conflict with the situation when the feature sets the selection and its attributes (e.g. bold command when executed will add attributes to the selection and that will, if I understand the flow correctly, trigger #changesDone).

scofalik:
Fixes #259. See the issue for the list of changes.

Followup issue: #633

Related PR: https://github.com/ckeditor/ckeditor5-core/pull/31

### DIFF
--- a/src/model/liveselection.js
+++ b/src/model/liveselection.js
@@ -158,8 +158,6 @@ export default class LiveSelection extends Selection {
 			this._storeAttribute( key, value );
 		}
 
-		// This uses private internal `_setAttribute` method which uses `directChange`.
-		// It's done like that to not document `directChange` in documentation because it's internal `LiveSelection` stuff.
 		if ( this._setAttribute( key, value ) ) {
 			// Fire event with exact data.
 			const attributeKeys = [ key ];
@@ -176,8 +174,6 @@ export default class LiveSelection extends Selection {
 			this._removeStoredAttribute( key );
 		}
 
-		// This uses private internal `_removeAttribute` method which uses `directChange`.
-		// It's done like that to not document `directChange` in documentation because it's internal `LiveSelection` stuff.
 		if ( this._removeAttribute( key ) ) {
 			// Fire event with exact data.
 			const attributeKeys = [ key ];
@@ -195,8 +191,6 @@ export default class LiveSelection extends Selection {
 			this._setStoredAttributesTo( attrs );
 		}
 
-		// This uses private internal `_setAttributesTo` method which uses `directChange`.
-		// It's done like that to not document `directChange` in documentation because it's internal `LiveSelection` stuff.
 		const changed = this._setAttributesTo( attrs );
 
 		if ( changed.size > 0 ) {
@@ -253,7 +247,7 @@ export default class LiveSelection extends Selection {
 	}
 
 	/**
-	 * Internal method for setting `LiveSelection` attribute. Supports attributes priorities (through `directChange`
+	 * Internal method for setting `LiveSelection` attribute. Supports attribute priorities (through `directChange`
 	 * parameter).
 	 *
 	 * @private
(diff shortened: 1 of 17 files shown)


src/model/schema.js
```diff
@@ -178,6 +178,11 @@ export class SchemaItem {
 
 			// Now we have to check every item name from the path to check.
 			for ( let checkName of checkPath ) {
+				// Don't check items that are not registered in schema.
+				if ( !this._schema.hasItem( checkName ) ) {
```
Reinmar: Missing test for this change.

Reinmar:
I found one regression: https://github.com/ckeditor/ckeditor5-engine/issues/637.

Reinmar:
The rest is fine – 100% CC, manual tests pass, etc. #637 needs to be changed in other piece of code, so it's independent.

Reinmar:
The other tickets which will be closed or needs to be checked:

* https://github.com/ckeditor/ckeditor5-engine/issues/589
* https://github.com/ckeditor/ckeditor5-engine/issues/603 (needs to be checked)
* https://github.com/ckeditor/ckeditor5-engine/issues/462

Reinmar:
I pushed manual test for #603 and checked that it's fully fixed by this PR. Cool!

src/model/liveselection.js
```diff
@@ -42,10 +42,41 @@ export default class LiveSelection extends Selection {
...
+		 * @protected
+		 * @member {engine.model.Document} engine.model.LiveSelection#_document
 		 */
 		this._document = document;
+
+		/**
+		 * Keeps mapping of attribute name to priority with which the attribute got modified (added/changed/removed)
+		 * last time. Possible values of priority are: `'low'` and `'normal'`.
+		 *
+		 * Priorities are used by internal `LiveSelection` mechanisms. All attributes set using `LiveSelection`
```
scofalik: `low` is being used when attributes are changed indirectly, that is through `_updateAttributes`.

Really, if you think is is overengineered I'll try to close this "logic" (that `_updateAttributes` clears only attributes set by itself) in `_updateAttributes`. IMHO it is over-engineered, but we may resolve this issue in ... spearate issue.

src/model/liveselection.js
```diff
@@ -42,10 +42,41 @@ export default class LiveSelection extends Selection {
...
+		 * @private
+		 * @member {Map} engine.model.LiveSelection#_attributePriority
+		 */
+		this._attributePriority = new Map();
+
+		// Whenever selection range changes, if the change comes directly from selection API (direct user change).
+		this.on( 'change:range', ( evt, data ) => {
+			if ( data.directChange ) {
+				// Reset attributes on selection (clear attributes and priorities).
+				//this._resetAttributes();
```
scofalik: Te commented code is a mistake. Sorry for that.

As far as `_updateAttributes` is concerned, this issue may be resolved by simplifying `LiveSelection` internal logic and getting rid of priorities. I thought priorities, as a general solution, will be nice, but they got too clunky and even though I have those priorities, I still have to write some special code (like in `_updateAttributes`).

scofalik:
I've pushed test for `Schema` and removed commented line of code.
As I understand #603 is fine white #589 and #462 are to-be-closed.
#637 is a different case and will be handled separately.

Does this mean that this PR is ready to be closed?

Reinmar:
Another ticket which seemed to be fixed: https://github.com/ckeditor/ckeditor5-engine/issues/590.

PR #650, opened 2016-10-25
Fake selection

### LINKED ISSUE #627: Fake selection
Fake selection is a feature which lets us work around browser limitations. Various browsers support only limited types of selections. For example, most browsers do not support multi-range selections (which are needed to represent table cell selections) and Blink and Webkit don't support selections anchored in elements which makes it impossible to have e.g. this:

```html
<p>Foo.</p>
[<figure contentEditable=false>...</figure>]
<p>Bar.</p>
```

In CKEditor 4 we worked around this by styling the selected object so it looks like selected and hiding the real selection in a hidden container. Table cells selection works a bit differently and only marks the cells, leaving the native selection somewhere in the table.

## Proposal

1. We're free to place whatever selection we want in the data model.
2. While conversion, a specific feature must tell how to render a selection on itself (e.g. add a class) and mark it as "fake".
3. The DOM renderer will then render such fake selection in a way it wants, handling all possible DOM quirks.

Most likely, we'll also need to types of fake selections – "on objects" and something to handle tables. Alternatively, the table feature will mark all cells (by adding classes) and cancel rendering 2+ ranges of the selection. But I don't know how this could work while you make selection using DnD (because it shouldn't be re-rendered then).

szymonkups:
Fixes #627 
- added marking selection as fake in view selection,
- added fake selection rendering to view renderer,
- created `FakeSelectionObserver` with basic arrow keys handling when selection is fake,
- added `FakeSelectionObserver` to default view document observers,
- added mapping between DOM element and fake view selection to `DomConverter`,
- added `clearFakeSelection` converter to  `model-selection-to-view-converters.js`,

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -466,7 +466,7 @@ export default class Renderer {
 			this._fakeSelectionContainer = domDocument.createElement( 'div' );
 			this._fakeSelectionContainer.style.position = 'fixed';
 			this._fakeSelectionContainer.style.top = 0;
-			this._fakeSelectionContainer.style.left = '-1000px';
+			this._fakeSelectionContainer.style.left = '-9999px';
 		}
 
 		// Add fake container if not already added.
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -462,30 +462,22 @@ export default class Renderer {
 		const domDocument = domRoot.ownerDocument;
 
 		// Create fake selection container if one does not exist.
-		if ( this._fakeSelectionContainer === null ) {
+		if ( !this._fakeSelectionContainer ) {
 			this._fakeSelectionContainer = domDocument.createElement( 'div' );
 			this._fakeSelectionContainer.style.position = 'fixed';
 			this._fakeSelectionContainer.style.top = 0;
 			this._fakeSelectionContainer.style.left = '-9999px';
+			this._fakeSelectionContainer.appendChild( domDocument.createTextNode( '\u00A0' ) );
 		}
 
 		// Add fake container if not already added.
-		if ( this._fakeSelectionContainer.parentElement === null ) {
+		if ( !this._fakeSelectionContainer.parentElement ) {
 			domRoot.appendChild( this._fakeSelectionContainer );
 		}
 
 		// Update contents.
 		const content = this.selection.fakeSelectionLabel || '\u00A0';
-		const textNode = this._fakeSelectionContainer.firstChild;
-
-		if ( !textNode || content !== textNode.textContent ) {
-			if ( textNode ) {
-				this._fakeSelectionContainer.removeChild( textNode );
-			}
-
-			const newTextNode = domDocument.createTextNode( content );
-			this._fakeSelectionContainer.appendChild( newTextNode );
-		}
+		this._fakeSelectionContainer.firstChild.data = content;
 
 		// Update selection.
 		const domSelection = domDocument.getSelection();
@@ -552,9 +544,8 @@ export default class Renderer {
 	_removeFakeSelection() {
 		const container = this._fakeSelectionContainer;
 
-		if ( container !== null ) {
+		if ( container ) {
 			container.remove();
-			container.textContent = '';
 		}
 	}
 
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -450,6 +450,12 @@ export default class Renderer {
 		}
 	}
 
+	/**
+	 * Updates fake selection.
+	 *
+	 * @private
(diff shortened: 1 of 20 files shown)


src/view/selection.js
```diff
@@ -45,6 +45,55 @@ export default class Selection {
...
+		 *
+		 * @private
+		 * @member {String} engine.view.Selection#_fakeSelectionLabel
+		 */
+		this._fakeSelectionLabel = '';
+	}
+
+	/**
+	 * Sets this selection instance to be marked as `fake`. This means only that selection will not be rendered
+	 * as real selection.
```
pjasiun: I believe we need some better explanation what the fake selection is.

src/view/selection.js
```diff
@@ -45,6 +45,55 @@ export default class Selection {
...
+	 * Additionally fake's selection label can be provided. It will be used to describe fake selection in DOM (and be
+	 * properly handled by screen readers).
+	 *
+	 * @param {Boolean} [value=true] If set to true selection will be marked as `fake`.
+	 * @param {Object} [options] Additional options.
+	 * @param {String} [options.label=''] Fake selection label.
+	 */
+	setFake( value = true, options = {} ) {
```
pjasiun: And if we remove the label, I believe we could also remove `options` and change `setFake` and `isFake ` to simple `fake` getter/setter, or even a property. KISS.

Reinmar: I'd keep `setFake()`. Feel more future-safe and there's nothing wrong with it. And definitely not `fake` getter/setter because it's against our conventions. If anything, it could be `isFake` setter/getter.

src/view/renderer.js
```diff
@@ -418,24 +428,77 @@ export default class Renderer {
...
+	 */
+	_updateFakeSelection( domRoot ) {
+		const domDocument = domRoot.ownerDocument;
+
+		// Create fake selection container if one does not exist.
+		if ( this._fakeSelectionContainer === null ) {
+			this._fakeSelectionContainer = domDocument.createElement( 'div' );
+			this._fakeSelectionContainer.style.position = 'fixed';
+			this._fakeSelectionContainer.style.top = 0;
+			this._fakeSelectionContainer.style.left = '-1000px';
```
Reinmar: May not be enough nowadays. 9999 is better :P

src/view/renderer.js
```diff
@@ -418,24 +428,77 @@ export default class Renderer {
...
+			domRoot.appendChild( this._fakeSelectionContainer );
+		}
 
-		if ( !domRoot ) {
-			return;
+		// Update contents.
+		const content = this.selection.fakeSelectionLabel || '&nbsp;';
+
+		if ( content !== this._fakeSelectionContainer.innerHTML ) {
+			this._fakeSelectionContainer.innerHTML = content;
```
Reinmar: Imagine what happens if someone set alt of their image to `""`. That alt will be used to represent the image o it will be passed straight to `innerHTML`. Better to create a text node and set its data.

src/view/renderer.js
```diff
@@ -418,24 +428,77 @@ export default class Renderer {
...
 		}
 
+		// Update selection.
+		const domSelection = domDocument.getSelection();
+		domSelection.removeAllRanges();
```
Reinmar: I'll feel safer with https://developer.mozilla.org/en-US/docs/Web/API/Selection/selectAllChildren. Every time we remove all ranges the browser may be confused what to do now.

szymonkups: It looks like this method would help us shorten this code, unfortunately I've just tested that it fires `selectionchange` event with empty selection at the end :disappointed:

Reinmar:
The automated tests pass and there's 100% CC. I made a small change in the manual test though and it fails now. It's pretty strange and I hope it's not my fault, but you'll need to check it. After preventing `mousedown` on the button the fake selection isn't created any more. Or rather – it's immediately removed. 

Preventing the mouse down event simulates how the thing would normally be selectable if that was caused by the editor button. After committing that, I even thought that a better test would be to listen on clicks on the element on which you want to make the fake selection. And that it should rather have `contenteditable=false` if we want to be sure about all this.

So I propose to change the test even further, to simulate focusing widgets by clicking on them. And `cE=false` may change something as well, so better add it.

szymonkups:
> The automated tests pass and there's 100% CC. I made a small change in the manual test though and it fails now. It's pretty strange and I hope it's not my fault, but you'll need to check it.

After removing `viewDocument.focus();` there is no rendering performed of the changed view elements and selection (because `focus()` calls it internally). I've added `viewDocument.render()` there.

Also as requested - changed focusing to happen after clicking on element with `cE=false`.

Issue #637, opened 2016-10-17
Composer.deleteContents() should clean selection attributes


Reinmar:
This issue is visible after https://github.com/ckeditor/ckeditor5-engine/pull/634.
1. Put caret at the end of a bolded word.
2. Delete all the word.
3. Type a letter.

This letter is now bolded, while it shouldn't. We agreed that deleting whole content of a bolded word should also remove the bold from the selection.

I'm not 100% sure whether this should be implemented by the `composer.deleteContents()` or the delete feature, but composer makes more sense.

BTW, @scofalik, what would be the easiest way to implement that? How can I now refresh selection attributes so it reads all of them from the content?

scofalik:
Honestly, this should work like you described. During character deletion, caret is moved so `model.Document.selection#_updateAttributes( true )` should be fired. This means that all attributes should be cleared and only attributes from caret neighbourhood. It's weird it doesn't work like this. There should be no need to refresh attributes.

Anyway, right now, there is no method to do so. Even setting selection to same position as it is won't work. So you'd have to collapse selection at different selection and then move it back. Or just use `_updateAttributes( true )` (but it should work right now for deleteing, so IDK).

Reinmar:
Unfortunately, this doesn't work as expected because of... how smart we made all this :D.

After `deleteContents()` run `batch.remove()` the live selection moves automatically to the place where we expect it to be, so setting it there explicitly in the next line doesn't trigger `change:range`.

So it seems that we'll have to update those attributes there anyway. And this leads to my next question – how to update selection attributes? Can we make the `_updateAttributes( true )` method public?

Reinmar:
Fixed by https://github.com/ckeditor/ckeditor5-engine/pull/653.

PR #653, opened 2016-10-25
T/331 Selection range should not end up in graveyard

### LINKED ISSUE #331: Selection should not end up in the graveyard
Test:

``` html
f<selection>o</selection>o
```

``` js
batch.remove( selection.getFirstRange() );
```
- Expected: selection collapsed between `f^o`.
- Actual: selection in the graveyard.

scofalik:
Fixes #331.

### DIFF
--- a/src/model/liveposition.js
+++ b/src/model/liveposition.js
@@ -111,6 +111,13 @@ export default class LivePosition extends Position {
 	 * @param {engine.model.Position} position
 	 * @returns {engine.model.LivePosition}
 	 */
+
+	/**
+	 * Fired when `LivePosition` instance is changed due to changes on {@link engine.model.Document}.
+	 *
+	 * @event engine.model.LivePosition#change
+	 * @param {Position} oldPosition Position equal to this live position before it got changed.
+	 */
 }
 
 /**
@@ -178,8 +185,14 @@ function transform( type, range, position ) {
 			break;
 	}
 
-	this.path = transformed.path;
-	this.root = transformed.root;
+	if ( !this.isEqual( transformed ) ) {
+		const oldPosition = Position.createFromPosition( this );
+
+		this.path = transformed.path;
+		this.root = transformed.root;
+
+		this.fire( 'change', oldPosition );
+	}
 }
 
 mix( LivePosition, EmitterMixin );
--- a/src/model/liveposition.js
+++ b/src/model/liveposition.js
@@ -116,7 +116,7 @@ export default class LivePosition extends Position {
 	 * Fired when `LivePosition` instance is changed due to changes on {@link engine.model.Document}.
 	 *
 	 * @event engine.model.LivePosition#change
-	 * @param {Position} oldPosition Position equal to this live position before it got changed.
+	 * @param {engine.model.Position} oldPosition Position equal to this live position before it got changed.
 	 */
 }
 
--- a/src/model/liveposition.js
+++ b/src/model/liveposition.js
@@ -30,7 +30,7 @@ export default class LivePosition extends Position {
 	 * @see engine.model.Position
 	 * @param {engine.model.RootElement} root
 	 * @param {Array.<Number>} path
-	 * @param {engine.model.PositionStickiness} [stickiness] Defaults to `'STICKS_TO_NEXT'`.
+	 * @param {engine.model.PositionStickiness} [stickiness] Defaults to `'sticksToNext'`.
 	 * See {@link engine.model.LivePosition#stickiness}.
 	 */
 	constructor( root, path, stickiness ) {
@@ -65,7 +65,7 @@ export default class LivePosition extends Position {
 		 *
 		 * @member {engine.model.PositionStickiness} engine.model.LivePosition#stickiness
 		 */
-		this.stickiness = stickiness || 'STICKS_TO_NEXT';
+		this.stickiness = stickiness || 'sticksToNext';
 
 		bindWithDocument.call( this );
 	}
@@ -155,7 +155,7 @@ function transform( type, range, position ) {
 
 	switch ( type ) {
 		case 'insert':
-			let insertBefore = this.stickiness == 'STICKS_TO_NEXT';
+			let insertBefore = this.stickiness == 'sticksToNext';
 			transformed = this._getTransformedByInsertion( range.start, howMany, insertBefore );
 			break;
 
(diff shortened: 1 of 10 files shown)


scofalik:
I've made several changes in tests for `deleteContents`. In my opinion the tests were wrong. In tests, selection was given attributes when it shouldn't. The selection takes attributes automatically from it's surrounding, and passing attributes to selection when it is non-collapsed is just wrong. Non-collapsed selection cannot have "it's own" attributes. Even if we can argue that it might be a valid case, the expected results in the changed tests did not match the input.

Reinmar:
Does it work the same way when selection is being made by the user or set from code? That its attributes are always reset then?

scofalik:
No, it will read attributes from it's (new) surroundings.

Let me explain how it works:

1. Change is done on document using `Batch` API.
2. `model.Document#event:change` is fired.
3. `LiveRange` listens to `Document#event:change`, updates itself, ends up in graveyard and fires it's own `LiveRange#event:change`.
4. `LiveSelection` listens to `LiveRange#event:change`. In callback, it checks if the range ended up in graveyard. If so, it fixes it:
  * finds a new, "correct", closest collapsed range,
  * removes range that is in graveyard,
  * splices "correct" range in that range's place (this is to prevent changing anchor and focus if the range that ended up in graveyard was not the last added range).
5. Additonally, the callback fires `LiveSelection#change:range` event.
6. `LiveSelection` listens to it's own `change:range` event and fires `_updateAttributes()`.

As you can see, the process is almost the same no matter if changed range ended up in graveyard or not. The only difference is in point 4. So you may think that doing a change on model that moves range to the graveyard has the same outcome as doing a change on model that collapses a selection somewhere near the original selection.

Reinmar:
I didn't mean the attributes behaviour in context of moving to the graveyard but rather this:

>  In my opinion the tests were wrong. In tests, selection was given attributes when it shouldn't. The selection takes attributes automatically from it's surrounding, and passing attributes to selection when it is non-collapsed is just wrong. Non-collapsed selection cannot have "it's own" attributes.

Since you changed the code I wanted to know whether the non-collapsed selections **never** have any attributes. If that should be a rule, then we should enforce this in the Selection API by throwing errors when someone tries to set attrs of non-collapsed selection and resetting attributes when selection becomes non-collapsed.

scofalik:
It's not enforced in the code. You can still set attributes on non-collapsed selection. What I meant is that there aren't any real scenarios when that would make sense. We can make a rule out of it or just leave it to not introduce problems in future.

This way or the other, the tests were incorrect (it's enough said that it's hard to understand what is the expect behavior/result when non-collapsed selection has different attributes than selected content).

Reinmar:
Could you report a followup for those attributes? I feel that it may be confusing and as a result, cause some bugs.

tests/model/composer/deletecontents.js
```diff
@@ -99,31 +97,25 @@ describe( 'Delete utils', () => {
...
-				setData( document, '<p>x</p><p>[<$text bold="true">foo</$text>]</p><p>y</p>', { selectionAttributes: {
-					bold: true
-				} } );
-
-				deleteContents( document.batch(), document.selection );
-
-				expect( getData( document ) ).to.equal( '<p>x</p><p>[]</p><p>y</p>' );
-				expect( document.selection.getAttribute( 'bold' ) ).to.undefined;
-			} );
+			//it( 'clears selection attrs when emptied content', () => {
```
Reinmar: What's with this test?

Reinmar:
I presume this PR fixes also https://github.com/ckeditor/ckeditor5-list/issues/16?

scofalik:
It should fix the issue in list repo.

Reinmar:
Unfortunately, this PR is breaking tests in the typing package because the `refreshAttributes()` method is not available in `Selection` but only in `LiveSelection`.

Since the `delete()` and `insert()` method should work with any selection, not only live selection, this new method should be moved to the `Selection` class.

Reinmar:
Related ticket: https://github.com/ckeditor/ckeditor5-engine/issues/658

src/model/liveposition.js
```diff
@@ -165,28 +173,34 @@ function transform( type, range, position ) {
...
+		this.fire( 'change', oldPosition );
+	}
 }
 
 mix( LivePosition, EmitterMixin );
 
 /**
  * Enum representing how position is "sticking" with their neighbour nodes.
- * Possible values: `'STICKS_TO_NEXT'`, `'STICKS_TO_PREVIOUS'`.
+ * Possible values: `'sticksToNext'`, `'sticksToPrevious'`.
```
Reinmar: STICKS_TO_PREVIOUS is used in one place in the `controller/insert.js` and needs to be updated now.

Reinmar:
I've been trying to test these changes manually, but the problem with missing `refreshAttributes()` breaks things a bit too much to be sure. However, I can confirm that the list problem seems to be gone.

Reinmar:
This PR has a chance to fix https://github.com/ckeditor/ckeditor5-engine/issues/637 as well.

src/model/liveselection.js
```diff
@@ -544,6 +588,29 @@ export default class LiveSelection extends Selection {
...
+		const newPath = oldRange.start.path.slice( 0, newPathLength );
+		newPath[ newPath.length - 1 ] -= gyPath[ 1 ];
+
+		const newPosition = new Position( oldRange.root, newPath );
+		const newRange = this._prepareRange( new Range( newPosition, newPosition ) );
+
+		const index = this._ranges.indexOf( gyRange );
+		this._ranges.splice( index, 1, newRange );
+
+		gyRange.detach();
```
scofalik: You are right.

scofalik:
> Since the delete() and insert() method should work with any selection, not only live selection, this new method should be moved to the Selection class.

This means that we need move `_updateAttributes` and `_getSurroundingAttributes` to the `Selection` too.

scofalik:
Since this is a part of `_getSurroundingAttributes`:
```javascript
// 6. If not found, selection should retrieve attributes from parent.
if ( !attrs ) {
	attrs = this._getStoredAttributes();
}
```

We would also need to move all the logic connected with storing attributes.

scofalik:
Instead of moving `refreshAttributes` to `Selection` (which would trigger moving a lot of logic), I moved calling `refreshAttributes` in `LiveSelection` from `change:range` event to `addRange`, `setRanges` and `removeAllRanges`. This way `refreshAttributes` is called every time `setRanges` or similar is called. This means that there should be no need to manually fire `refreshAttributes`.

I hope this will fix your issues. Let me know if it works for you.

Reinmar:
Works super fine. I'm running a CC test right now, but I don't expect any problems.

Issue #652, opened 2016-10-25
Symmetrical handling of the first and last blocks in getSelectedContent() and insertContent()


Reinmar:
Make the following selection:

```html
<p>Fo[o</p>
<p>Bar</p>
<p>B]om</p>
```

Copy and paste this.

Expected is that nothing has changed. The current behaviour will be this:

```html
<p>Fo</p>
<p>o</p>
<p>Bar</p>
<p>B</p>
<p>om</p>
```

The reason is that the blocks are inserted "as is" without merging with the currently existing ones. In the clipboard we have:

```html
<p>o</p>
<p>Bar</p>
<p>B</p>
```

So e.g. when we paste this into:

```html
<p>X^X</p>
```

We get:

```html
<p>X</p>
<p>o</p>
<p>Bar</p>
<p>B</p>
<p>X</p>
```

We need to decide whether we want to always merge the first and pasted lines with the paste location or only in special cases where on copy the first and last blocks weren't fully selected.

This is, unfortunately, quite tricky. In the past we had both behaviours in CKE4 (it changed around CKE4.5, with the introducion of custom copy/cut IIRC) and none was perfect. There are many edge cases depending on the type of blocks where we've got the selection and what we paste. In #648 I'm going to implement the simplest possible behaviour and we'll need to discuss this matter before I'll spend more time on it.

Reinmar:
This ticket will be quite important for one more reason – to not waste time on something which is likely to change soon, I'm not going to work on selection position after blocks were pasted, so the selection will most likely end up at the beginning of the block after the pasted content. This certainly isn't a good UX.

fredck:
I'm for:

 * Always merging first and last blocks if content is available before/after the selection boundaries. The caret stays blinking at the end of the pasted content.
 * Append all blocks before as is, if the user is in an empty block. The caret stays blinking at the same place it was (that empty block).

Reinmar:
> Append all blocks before as is, if the user is in an empty block. The caret stays blinking at the same place it was (that empty block).

You mean this?

```html
<p>^</p> + <p>x</p> => <p>x</p><p>^</p>
```

I think it's wrong, cause this would be strange:

```html
<p>[a</p><p>b]</p> + <p>x</p> => <p>x</p><p>^</p>
```

So if the block after calling `deleteContent()` is empty, it should not be left empty after inserting the content. So, basically, I wouldn't make an exception here.

Reinmar:
[...]

PR #656, opened 2016-10-26
Introduces DataController.insertContent

### LINKED ISSUE #648: Implement DataController.insertContent()
Needed for ckeditor/ckeditor5-clipboard#1.

It seems that it must accept *view* document fragment, because it will need to convert it to the model using the context of insertion. And this becomes part of the "insert content" logic, so should most likely be done inside this method.

BTW, it made me think that this method would make sense on the `DataController`. And I kinda like this idea. We've been thinking where to move the composer and I can see that `modifySelection()` makes for the `EditingController` and `deleteContent()` together with `insertContent()` could perhaps end in the the `DataController`. Note also that we'll need something like `getSelectedContent()` for the copy and cut operations (so `deleteContent()` would become `deleteSelectedContent()`).

OTOH, I know what you're going to say – those methods look weird on the `DataController`. I agree... I'd have to get used to them. But consider this – let's say we'd like to move `insertContent()` to the `Document`. But `insertContent()` requires (as explained above) a knowledge about `DataController`. So how could it be implemented on the `Document`?

tl;dr:

* `DataController#insertContent()` – paste and I'm sure it'll be used in many ways in many cases by developers,
* `DataController#deleteSelectedContent()` – delete feature,
* `DataController#getSelectedContent()` – copy, cut and this has been also requested many times in the past for various integration scenariosm
* `EditingController#modifySelection()`

Does it make any sense or no sense at all?

Reinmar:
Fixed #648. Covers big part of #652.

What can I say? Not the prettiest code in the world – written in 2.5 days (including tests). Certainly, we'll have to work on this more, but it's ok for now. While working on the tests I decided to cover also big part of #652. We'll see with time whether I missed something important

You can test this code on https://github.com/ckeditor/ckeditor5-clipboard/tree/t/1.

### DIFF
--- a/src/datacontroller.js
+++ b/src/datacontroller.js
@@ -227,3 +227,16 @@ export default class DataController {
 }
 
 mix( DataController, EmitterMixin );
+
+/**
+ * Event fired when {@link engine.DataController#insertContent} method is called.
+ * The {@link engine.dataController.insertContent default action of the composer} is implemented as a
+ * listener to that event so it can be fully customized by the features.
+ *
+ * @event engine.DataController#insertContent
+ * @param {Object} data
+ * @param {engine.model.Batch} data.batch Batch to which deltas will be added.
+ * @param {engine.model.Selection} data.selection Selection into which the content should be inserted.
+ * @param {engine.view.DocumentFragment} data.content The content to insert.
+ */
+
--- a/src/datacontroller.js
+++ b/src/datacontroller.js
@@ -3,6 +3,9 @@
  * For licensing, see LICENSE.md.
  */
 
+import mix from '../utils/mix.js';
+import EmitterMixin from '../utils/emittermixin.js';
+
 import Mapper from './conversion/mapper.js';
 
 import ModelConversionDispatcher from './conversion/modelconversiondispatcher.js';
@@ -16,6 +19,8 @@ import ViewDocumentFragment from './view/documentfragment.js';
 import ModelRange from './model/range.js';
 import ModelPosition from './model/position.js';
 
+import { stringify as stringifyModel } from '../engine/dev-utils/model.js';
+
 /**
  * Controller for the data pipeline. The data pipeline controls how data is retrieved from the document
  * and set inside it. Hence, the controller features two methods which allow to {@link engine.DataController#get get}
@@ -27,6 +32,7 @@ import ModelPosition from './model/position.js';
  * * {@link engine.conversion.ViewConversionDispatcher view to model} converters.
  *
  * @memberOf engine
+ * @mixes utils.EmitterMixin
  */
 export default class DataController {
 	/**
@@ -104,6 +110,17 @@ export default class DataController {
 		this.viewToModel.on( 'text', convertText(), { priority: 'lowest' } );
 		this.viewToModel.on( 'element', convertToModelFragment(), { priority: 'lowest' } );
 		this.viewToModel.on( 'documentFragment', convertToModelFragment(), { priority: 'lowest' } );
+
+		this.on( 'insertContent', ( evt, data ) => insertContent.call( this, data.batch, data.selection, data.content ) );
+
+		// TMP!
+		// Create an "all allowed" context in the schema for processing the pasted content.
+		// Read: https://github.com/ckeditor/ckeditor5-engine/issues/638#issuecomment-255086588
+
+		const schema = model.schema;
+
+		schema.registerItem( '$clipboardHolder', '$root' );
+		schema.allow( { name: '$text', inside: '$clipboardHolder' } );
 	}
 
 	/**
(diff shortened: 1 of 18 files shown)


src/datacontroller.js
```diff
@@ -104,6 +110,8 @@ export default class DataController {
 		this.viewToModel.on( 'text', convertText(), { priority: 'lowest' } );
 		this.viewToModel.on( 'element', convertToModelFragment(), { priority: 'lowest' } );
 		this.viewToModel.on( 'documentFragment', convertToModelFragment(), { priority: 'lowest' } );
+
+		this.on( 'insertContent', ( evt, data ) => insertContent( this, data.batch, data.selection, data.content ) );
```
pjasiun: Are you sure it should have a default priority? As a default behavior, it should be `lowest` or at least `low`.

src/datacontroller.js
```diff
@@ -204,4 +212,31 @@ export default class DataController {
...
+
+	/**
+	 * See {@link engine.model.composer.insertContent}.
+	 *
+	 * @fires engine.DataController#insertContent
+	 * @param {engine.model.Batch} batch Batch to which deltas will be added.
+	 * @param {engine.model.Selection} selection Selection into which the content should be inserted.
+	 * @param {engine.view.DocumentFragment} content The content to insert.
+	 */
+	insertContent( batch, selection, content ) {
```
pjasiun: To make it similar to other methods here we could rename it to simple `insert`. It will create then a nice API: `editor.data.insert()`. Also, the order of parameters should be same as other methods: what, where, other. Note that `content` is the only parameter which for sure will not be optional. So: `insert( content, selection, batch )`

Moreover, we may make the API more consistent: all `parse`, `set` and `insert` could get data as string or view and call `data = this.processor.toView( data );` if data is a string. 

Also, we have a different policy when it comes to the `batch` in `set` and `insert`. I think both could take `batch` as an optional parameter and create a batch if it is not passed.

It would be a nice and useful API, not only for clipboard: `editor.data.insert( '<p>foo</p>', editor.document.selection )`.

pjasiun:
Because we have more `controller` code, all `insertcontent.js`, `datacontroller.js` and `editingcontroller.js` should be moved to the `controller` folder.

tests/datacontroller/insertcontent.js
```diff
@@ -0,0 +1,606 @@
...
+				'<paragraph>f[]oo</paragraph>',
+				'<paragraph>fxxx</paragraph><paragraph>yyy</paragraph><paragraph>zzz[]oo</paragraph>'
+			);
+
+			test(
+				'filters out disallowed objects',
+				'<disallowedWidget>xxx</disallowedWidget>',
+				'<paragraph>f[]oo</paragraph>',
+				'<paragraph>f[]oo</paragraph>'
+			);
```
pjasiun: You know that "helpers better then magic" is true also in tests? Such tests are very hard  to maintain. You already had to create a `options.beforeCallback` method which is not very readable. In fact, these tests are not readable at all. And I think that it's worth to make code even longer if it is more readable thanks to it, but in this case, code is even shorter:

```js
it( 'filters out disallowed objects', () => {
	setData( doc, '<disallowedWidget>xxx</disallowedWidget>' );
	insertHelper( '<paragraph>f[]oo</paragraph>' );
	expect( getData( doc ) ).to.equal( '<paragraph>f[]oo</paragraph>' );
} );

function insertHelper( content ) {
		if ( typeof content == 'string' ) {
			content = parse( content, doc.schema, {
				context: [ '$clipboardHolder' ]
			} );
		}

		if ( !( content instanceof ModelDocumentFragment ) ) {
			content = new ModelDocumentFragment( [ content ] );
		}

		// Override the convertion so we get exactly the model that we defined in the content param.
		// This way we avoid the need to write converters for everything we want to test.
		dataController.viewToModel.convert = () => {
			return content;
		};

		insertContent( dataController, doc.batch(), doc.selection, new ViewDocumentFragment() );
}
```

src/datacontroller/insertcontent.js
```diff
@@ -0,0 +1,411 @@
...
+			this.batch.merge( mergePosLeft );
+
+			this.position = Position.createFromPosition( position );
+			position.detach();
+		}
+
+		if ( mergeRight ) {
+			let position;
+
+			/* istanbul ignore next */
```
pjasiun: Why you had `/* istanbul ignore if */` before and `/* istanbul ignore next */ here?

src/datacontroller.js
Reinmar: I copied this from the composer. Can be `low` but it's hard to tell whether people will want to do something before or after. Since it's 50/50, default was my choice.

src/datacontroller.js
Reinmar: I agree with I think everything here, but I want to work on a cleanup when we'll be moving `deleteContent()` here. In fact – should it be `delete()` then too? I think this would be confusing so the `Content` part is good in both cases.

pjasiun: `delete()` then. We use it as `editor.data.delete()`. Data and content are synonyms, in this case, so `editor.data.deleteContent()` would be repeating yourself.

tests/datacontroller/insertcontent.js
Reinmar: I've changed the tests as you asked and IMO they lost on readability A LOT. The input + output patterns don't begin with the same indentation now so it's hard to compare what has changed.

Reinmar: Besides, the lines are longer so it's harder to fit some patterns.

Issue #661, opened 2016-10-28
Move to tests content from HTML files


pomek:
In order to run tests using Karma, we need to remove all HTML files.

Required DOM should be created directly through tests which require additional HTML.

scofalik:
I understand that this should be corrected in all the repos?

Reinmar:
Yes, we need to get rid of HTML for automated tests in all repos. Kamil is working on that.

PR #662, opened 2016-10-28
T/661: Removed HTML files from automatic tests

pomek:
Fixes: #661.


Reinmar:
All fine, but we should be using `beforeEach` and `afterEach`. Otherwise we may be creating connected tests.

PR #668, opened 2016-10-31
T/629 Fix infinite selection change loop

### LINKED ISSUE #629: Fix infinite selection change loop introduced in #623
We agreed to merge #623 with a small regression – the infinite loop check kicks in when moving selection over boundaries of inline elements. It needs to be fixed ASAP.

This is a broader issue, though, and includes also:
- https://github.com/ckeditor/ckeditor5-engine/issues/612
- https://github.com/ckeditor/ckeditor5-engine/issues/535

It's a bit unclear now whether all of them are caused by the same issue or they are separate. In this ticket we must focus on the regression (if it's regression at all). But perhaps, at the same time we'll fix also the rest.

scofalik:
Fixes #629.

The problem with selection was that during conversion chain: `DOM -> view (1) -> model -> view (2) -> DOM` view (1) selection was different than view (2) selection because there was a difference how model -> view and DOM -> view selection is converted. This caused an infinite loop of selection fixing.

I changed one of conditions that was responsible for breaking the loop. The condition looked whether view (1) and view (2) are equal. Now I check whether they are touching. If they are, the selection is not converted any further.

To do so, I needed to implement `isTouching` in model range, selection and view position, range selection.

### DIFF
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -341,52 +341,72 @@ export default class Position {
 	 * @returns {Boolean} True if positions touch.
 	 */
 	isTouching( otherPosition ) {
+		if ( this.root != otherPosition.root ) {
+			return false;
+		}
+
 		let left = null;
 		let right = null;
-		let compare = this.compareWith( otherPosition );
+		let compare = compareArrays( this.path, otherPosition.path );
+		let diffAt = null;
 
 		switch ( compare ) {
 			case 'same':
 				return true;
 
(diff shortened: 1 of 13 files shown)


scofalik:
I provided an alternative solution in #671

Reinmar:
Since we're betting more on https://github.com/ckeditor/ckeditor5-engine/pull/671 I'm closing this.

Issue #535, opened 2016-07-18
Selection check infinite loop kicks in too often


Reinmar:
1. Go to https://ckeditor5.github.io/
2. Put selection in the first heading.
3. Start pressing up, down, up, down arrows quickly.

If you do it frequently enough and repeat enough times you got:

```
selectionchange-infinite-loop: Selection change observer detected an infinite rendering loop. undefined
```

and the seleciton stops refreshing (the label of the headings dropdown stops updating).

**Now, god bless our marvellous idea to log stuff like this to the console!!!** Imagine the hours spent on debugging this security check if it ended up with a silent errror.

Reinmar:
This is even easier to trigger (which makes it pretty critical). Just keep pressing backspace for a longer time. The warning will be logged pretty quickly.

scofalik:
After fixing #401 selection infinite loop kicks in also happen at the edge of `<strong>` element.

scofalik:
> This is even easier to trigger (which makes it pretty critical). Just keep pressing backspace for a longer time. The warning will be logged pretty quickly.

I can't reproduce this. However in this case:

> Go to https://ckeditor5.github.io/
Put selection in the first heading.
Start pressing up, down, up, down arrows quickly.

The warning/check kicks in incorrectly. When you comment out the whole if nothing breaks or crashes (no infinite loop).

scofalik:
Okay it appears that the `_loopbackCounter` in `SelectionObserver` is not cleared on user action which is just wrong. The infinite loop condition should kick in only if there were multiple dom selection changes without user interaction. Do you have an idea how to check it without having interaction between other observers?

Reinmar:
I think that what we want to check is the number of selection changes in a period of time. Like whether 100 last changes happened in a second. If that's true, then we have loop.

The first way to implement this which comes to my mind is having an array of last change timestamps. You add the current one and checks the one that you remove (like in a FIFO queue). I guess there may be smarter ways too :D.

Anyway, this should not fail us. In fact, this is how I thought it worked.

Reinmar:
PS. The numbers must be chosen in such a way that it won't fail in automated tests :D. But it should be fine if in every `beforeEach()` we create a new observer. If not, then this is test what's broken.

Reinmar:
Fixed by https://github.com/ckeditor/ckeditor5-engine/pull/671.

PR #671, opened 2016-11-02
T/629b Alternative fix infinite selection loop.

scofalik:
This is an alternative way of solving #629 (previous solution is in #688)


src/model/selection.js
```diff
@@ -125,25 +125,29 @@ export default class Selection {
...
+		if ( this.rangeCount != otherSelection.rangeCount ) {
+			return false;
 		}
 
-		return this.isBackward === otherSelection.isBackward;
+		// Every range from this selection...
+		return Array.from( this.getRanges() ).every( ( rangeA ) => {
+			// ...Has a range in other selection...
+			return Array.from( otherSelection.getRanges() ).some( ( rangeB ) => {
+				// That it is equal to.
```
Reinmar: `// ... which it is equal to.`

src/view/observer/selectionobserver.js
```diff
@@ -102,6 +102,9 @@ export default class SelectionObserver extends Observer {
 		}
 
 		domDocument.addEventListener( 'selectionchange', () => this._handleSelectionChange( domDocument ) );
+		domDocument.addEventListener( 'keydown', () => this._clearInfiniteLoop() );
```
Reinmar: As I commented in https://github.com/ckeditor/ckeditor5-engine/issues/535#issuecomment-258168263, I prefer the other solution.

src/view/selection.js
```diff
@@ -283,33 +283,39 @@ export default class Selection {
...
-		for ( let i = 0; i < this.rangeCount; i++ ) {
-			if ( !this._ranges[ i ].isEqual( otherSelection._ranges[ i ] ) ) {
-				return false;
-			}
+		if ( !this.anchor.isEqual( otherSelection.anchor ) || !this.focus.isEqual( otherSelection.focus ) ) {
+			return false;
 		}
 
-		return this._lastRangeBackward === otherSelection._lastRangeBackward;
+		// Every range from this selection...
```
Reinmar: Do I see correctly that the ranges can be in a different order? Why so? Then, the `_lastRangeBackward` property would need a more precise check. Besides, if we don't expect this to happen frequently, I'd consider these selections different.

tests/view/observer/selectionobserver.js
```diff
@@ -162,71 +162,90 @@ describe( 'SelectionObserver', () => {
...
-		} );
+		for ( let i = 0; i < 4; i++ ) {
+			changeDomSelection();
+		}
 
-		changeDomSelection();
+		setTimeout( () => {
+			expect( spy.called ).to.be.false;
+			done();
+		}, 1500 );
```
Reinmar: If there's a chance to avoid (or shorten) such tests, then please do.

Reinmar:
I'm trying to match this code to the comment that you left me on priv. Is this really comparing the selection as you described it – after "unifying" it by conversion to model and back?

In general, I like the idea very much. I'm not if it will be perfect (there are tricky cases, especially with block widgets), but the improved inf loop check should catch them. So I'm +1, but I'd like to understand how this code works :D.

PS. It's not review ready, right? You wrote something about the tests. If you were worried whether it makes sense to work on them, then for me it does.

scofalik:
This solution prevents creating infinite loop by not setting model selection if it has not changed. This results in not firing renderer and breaks the loop. It's done here.
https://github.com/ckeditor/ckeditor5-engine/pull/671/files#diff-977f0f36f33d59ceb01cc1aaafaa45faR42

This was problem number one and it actually caused an infinite loop if not for the `if` that breaks it. The problem number two was that current infinite loop check is too restrictive. It does not take into account time in which changes happen or their source. This means that you could slowly press left arrow key and right arrow key and still get a warning.

Your idea with checking time is good to. Conceptually it's similar to what I've done (we both want to check if something happens between selection changes) but I like your idea better.

As far as tests are concerned, I feel like testing DOM selection behavior in unit test might be a mistake due to asynchronous nature of selectonchange event. I had to pick correct params for tests (timeouts) so they won't fail (just like I did on physics labs when I was studying 🎓 ). I think I'd rather have some manual tests for those cases.

Reinmar:
OK, now all it's clear. So the only thing that I'd like to improve before the review is the way how we test this. I agree with you that this deserves a manual test. Let's try to avoid such complicated (and slow) automated tests.

scofalik:
I've changed `SelectionObserver` so that it clears infinite loop counter in time interval, rather after DOM event.
I've also added manual test, removed unit test.

src/view/observer/selectionobserver.js
```diff
@@ -179,12 +179,23 @@ export default class SelectionObserver extends Observer {
 			this._loopbackCounter = 0;
 		}
 
-		if ( this._loopbackCounter > 10 ) {
+		if ( this._loopbackCounter > 50 ) {
```
Reinmar: Deserves some comment that you empirically tested that making more than 50 selection changes in 2s is not possible. BTW. What about making selections by dragging over the text?

Reinmar:
I merged master to this branch and fixed a manual test (which now should always be inside `manual/` directory to be properly filtered out). There are some red tests. Please check then and let's try to finalize this ticket. I'll check your solution manually tomorrow morning.

Reinmar:
Interesting. This is my status:

```

SUMMARY:
✔ 2198 tests completed
✖ 5 tests failed

FAILED TESTS:
  ✖ "before each" hook
    Chrome 54.0.2840 (Mac OS X 10.12.1)
  Error: Uncaught TypeError: Cannot read property 'isEqual' of null (tests/engine/controller/editingcontroller.js:21316)

  SelectionObserver
    ✖ should fire selectionChange when it is the only change
      Chrome 54.0.2840 (Mac OS X 10.12.1)
    Error: Uncaught TypeError: Cannot read property 'isEqual' of null (tests/engine/controller/editingcontroller.js:21316)

    ✖ "before each" hook
      Chrome 54.0.2840 (Mac OS X 10.12.1)
    Error: Uncaught TypeError: Cannot read property 'isEqual' of null (tests/engine/controller/editingcontroller.js:21316)

    ✖ "after each" hook for "should throw error if one try to create positions before root"
      Chrome 54.0.2840 (Mac OS X 10.12.1)
    TypeError: Cannot read property 'call' of undefined

    ✖ "after each" hook for "should throw error if one try to create positions before root"
      Chrome 54.0.2840 (Mac OS X 10.12.1)
    TypeError: Cannot read property 'call' of undefined


=============================== Coverage summary ===============================
Statements   : 95.44% ( 4541/4758 )
Branches     : 93.21% ( 2594/2783 )
Functions    : 97.77% ( 963/985 )
Lines        : 95.4% ( 4502/4719 )
================================================================================
```

On Travis there's less red tests, but the coverage is very low ;|

Reinmar:
I've got this on master now:

```
Finished in 13.196 secs / 5.218 secs

SUMMARY:
✔ 2656 tests completed

=============================== Coverage summary ===============================
Statements   : 100% ( 4749/4749 )
Branches     : 99.89% ( 2772/2775 )
Functions    : 100% ( 979/979 )
Lines        : 100% ( 4711/4711 )
================================================================================
```

So it seems that some failing things outside of tests (in `beforeEach`  as the first error says) prevents other tests from running... but it breaks a huge number of tests for some strange reason  :D. Anyway, if you fix that, then other tests should start being executed.

Reinmar:
OK, the trick here is that the order of running tests on CI and locally may be a bit different. I've noticed e.g. that when I executed `--files=engine/controller,engine/view` I've got some failing tests and in the reverse order everything is fine.

This means that some tests are not cleaning up after themselves correctly.

Reinmar:
BTW. running `engine/view/observer` still logs out a lot of warnings about inf loop.

Reinmar:
OK, I fixed some smaller things in the tests, but there's one, huge issue which is this https://github.com/ckeditor/ckeditor5/issues/114#issuecomment-259695367.

Run:

```
gulp test:server -ws --files=engine/controller,engine/view
```

Open the debug mode (if you go to the Chrome window opened by Karma there's "debug mode" button), open console and run the tests there. See that there are multiple errors thrown. They are thrown because some old selection observers are not destroyed and affect other tests.

There's no other solution, IMO, than working on view document destroy chain which at least should destroy observers which at least should remove all native event listeners.

To do that we need DomEmitterMixin in ckeditor5-utils. It will need to be mixed into Observer class I guess. Then, using `listenTo( nativeNode )` you can listen to native events. Calling `stopListening()` in `Observer`'s `destroy()` methods should clean up everything nicely. 

The last step is to ensure that `ViewDocument` instances are always being destroyed in `afterEach()` in all existing tests. This will ensure that engine tests pass nicely. But this is only the beginning if one wants to ensure that all package tests work together, because then we also need to destroy whole editors and work on their destroy chains :D But this is future.

src/view/observer/selectionobserver.js
```diff
@@ -103,6 +105,8 @@ export default class SelectionObserver extends Observer {
 
 		domDocument.addEventListener( 'selectionchange', () => this._handleSelectionChange( domDocument ) );
 
+		setInterval( () => this._clearInfiniteLoop(), 2000 );
```
Reinmar: This interval needs to be cleaned on `destroy()`.

src/view/selection.js
scofalik: Backward/forward selection check is indirectly implemented when we check `anchor` and `focus`. Basically, if `anchor` and `focus` are equal and ranges are equal, this means that selections are equal. If the last range was added in different direction, `anchor` and `focus` would differ.

src/view/observer/selectionobserver.js
scofalik: > BTW. What about making selections by dragging over the text?

The selection has to be same as previous or pre-previous selection, so you would have to drag precisly, which means that it would be probably impossible to do 50 changes in 2 secs.

scofalik:
I've made changes to the commented parts of code and also worked on destroy chain for `EditingController` -> `ViewDocument` -> observers -> dom events.

I've also, hopefully, got rid of inf-selection-loop warning in tests for good :).

src/view/selection.js
Reinmar: I was rather asking why is this method trying to find a range in the other selection which matches range from this selection... for each range in this selection? Moreover, this implementation is very ineffective, cause it creates new array on every call of `every()` callback and makes huge number of comparisons (in a bit random way). And last but not least – since we're checking anchor and focus first, then checking ranges again will for a collapsed selection tripple the amount of work and for a single range non collapsed selection double it.

Therefore, I've been asking about the order of ranges (whether it shouldn't be the same in both selections) because the easiest possible implementation is to loop through both selections' ranges and compare the pairs. Not only this will be much faster, but also cleaner.

tests/controller/editingcontroller.js
```diff
@@ -274,6 +277,8 @@ describe( 'EditingController', () => {
 			} );
 
 			expect( spy.called ).to.be.false;
+
+			editing.destroy();
 		} );
```
Reinmar: I miss a test that destroy() destroys the view.

Reinmar:
Manually all this works perfectly.

The remaining issues are:

* https://github.com/ckeditor/ckeditor5-engine/pull/671#discussion_r88046379
* https://github.com/ckeditor/ckeditor5-engine/pull/671#discussion_r88045725
* missing docs for new destroy() methods.

src/view/selection.js
scofalik: Okay, I see.

Ranges are not sorted upon insertion. This means that we have to check whether every range in selection A has a matching range in selection B the way we do. If the ranges are ordered that would be easier.

It is a matter of us deciding whether the order range is important. Being strict, I think that we should not care about the ranges order, because from "outside" selection: `[ 1, 4 ], [ 5, 7 ], [ 8, 9 ] forward` behaves exactly the same as `[ 5, 7 ], [ 1, 4 ], [ 8, 9 ] forward`.

I can change the implementation but... To be honest selection will usually have just one range, except of tables where it may be more but in 99% scenarios it will will be less than 10-20 ranges. So I propose just changing so the arrays are not created.

> And last but not least – since we're checking anchor and focus first, then checking ranges again will for a collapsed selection triple.
So, instead one operation we have three? And it is me who get complaints about premature optimization? :)

scofalik: >  was rather asking why is this method trying to find a range in the other selection which matches range from this selection... for each range in this selection

BTW. I don't know if you haven't got too confused...

For each range in selection A, we check if there is a matching range in selection B. The only way we can "speed" this up is if we already have sorted arrays...

scofalik:
I've added requested test and changed implementation of `isEqual` a bit. If you are fine with it, please close the review. If you really want to not care about ranges order, I will simplify the implementation further.

PR #672, opened 2016-11-04
Added EmptyElement and WidgetElement to view

### LINKED ISSUE #657: Create WidgetElement and EmptyElement
We need special view elements for widgets and empty elements.
`WidgetElement` will be used to represent widgets inside the view.
`EmptyElement` will be used to represent elements that cannot be divided and cannot contain any elements inside (for example ``).

Things to have in mind:
- View writer should handle these new elements,
- both new element types should provide proper `getFillerOffset` method,
- both should extend view's `Element` class.

szymonkups:
Fixes #657.


src/dev-utils/view.js
```diff
@@ -914,12 +951,12 @@ function _convertElementNameAndPriority( viewElement ) {
...
 }
 
-// Checks if element's type is allowed. Returns `attribute`, `container` or `null`.
+// Checks if element's type is allowed. Returns `attribute`, `container`, `empty`, `widget` or `null`.
 //
 // @param {String} type
 // @returns {String|null}
 function _convertType( type ) {
-	if ( type == 'container' || type == 'attribute' ) {
+	if ( type == 'container' || type == 'attribute' || type == 'empty' || type == 'widget' ) {
```
pjasiun: There could be an array with namespaces names and constructors reused in all these conditions instead of checking each type separately.

src/view/widgetelement.js
```diff
@@ -0,0 +1,24 @@
...
+
+	/**
+	 * Returns `null` because block filler is not needed.
+	 *
+	 * @returns {null}
+	 */
+	getFillerOffset() {
+		return null;
+	}
+}
```
pjasiun: I'm thinking how we will add other widget parts, like drag handler. If it will not be a child in the view and it will appear only in the DOM (added in the rendering/DOM conversion), we may not add `contenteditable` attribute here too. On the other hand, I like that this attribute is here because it's more transparent.

Reinmar:
Before I'll dig more into this PR, I'd like to ask one question. Don't you feel that it was too hard to add these two new element types? It means that the list of them is very closed. We've got container element, attribute element, empty element and widget element. If we'll see a need for a new one (e.g. nested editable) or that one for the created ones doesn't fulfil its role entirely (how widget element works inline?), then we'll have a problem. 

Another thing is that we expect that if anyone wants to change some behaviour a bit the right way to go will be to extend one of the types. E.g. a list item may need to override the allowed filler positions inside itself. But how much can you actually change the existing element's behaviour?

For instance, I can see that the new elements definitions are short... while the changes in other places are long. It should be the opposite. And this situation is very dangerous (I'm not joking here).

So, if you agree, I'd report this and keep in mind. Before doing next changes to this structure, I'd work on finding out what can be moved our of the writer and the view utils to the elements.

szymonkups:
@Reinmar +1 from me. It would be great if we could add new element type and implement everything inside it, without touching utils and writer.

pjasiun:
On one hand, I also agree, it would be nice. On the other, it makes writer far more complicated, I'm against such change. From the very begging there suppose to be 3 types: `ContainerElement`, `AttributeElement` and `EmptyElement` + `DocumentFragment` which is not an element. Rest should be children of these 3 types, so, when introduced, there will be no need to change writer.

Note that if `WidgetElement` would be a `ContainerElement`, there would be no changes in writer needed. In fact, we did not call `ContainerElement` `BlockElement`, because we assumed that inline widget will still be an `ContainerElement`.

Reinmar:
We've talked about this with PJ and we found a middle ground. It is tricky, because we found a few new types of elements which mix different "type", "filler" and "what's allowed in me" options.

E.g. – textarea (can have text, but no filler and no attribute elements inside), non-editable paragraph (like container, but non-editable and should render filler for simplicity), style element (like textarea)  and I guess we'll know more in the future.

However, I can agree with PJ that we could implement only 3 types of elements – container, attribute, empty. Let those be the only `Element` subclasses in the first line. Then we also need a `RootEditable` and `EditableElement`. But on this we should end.

This means that we should remove the widget element. We created it to override the filler behaviour, but let's treat filler's behaviour (`getFillerOffset()`) as a strategy pattern with a default strategy defined in the base classes. So a widget should be a mixin which you'll apply to a specific element and that mixin will change this strategy on the fly.

szymonkups:
Reverted commits which add `WidgetElement`. Should I include widget mixin in this PR, or we will extract it with all needed functionalities when image feature will be completed?

Reinmar:
Am I right that `widgetize()` will do two things:

1. set cE=false
2. overwrite `getFillerOffset`?

If that's it then let keep this code together with the rest of widgets API. This can be in this repo or outside, as you wish. If in the engine, then I'd put these functions in `view/widget/*`.

szymonkups:
> Am I right that widgetize() will do two things:

Yes, for now. I would wait to finish initial image feature implementation and see if anything else will be needed, and then move it into proper place.

PR #676, opened 2016-11-09
T/675

### LINKED ISSUE #675: OT: Stabilize operational transformation.
The goal of this issue is to make a research on bugs and problems directly in operational transformation algorithms and also in related parts of engine.

After this issue is done, operational transformation in engine should be considered stable. This basically means that editor won't blow up on most set of deltas to transform, apply to model and render.

Still it is expected that there may be some more complicated/edgy scenarios that will result in an error or incorrect document state.

scofalik:
Fixes #675 

A number of bugs or missed cases has been found and fixed:

1. `ModelConversionDispatcher` should not dispatch changes that happen in graveyard. Two conditions has been added to filter out that kind of changes.

2. `deleteContents` should use `RemoveDelta` if merged element is empty. This is more correct semantically and provides better OT results.

3. Recently, we fixed the issue that model selection could end up in graveyard. This solution has been enhanced. Previously, selection could end up, for example, between paragraphs. Now the selection is collapsed at the nearest position where text can be placed.

4. Found an issue in `DomConverter#viewPositionToDom`, in a corner case scenario. This was view before applying changes to model: `<p>abc<b>bold</b>FILLER{}`. Due to changes in model, a text node (AFAIR) was added after `<b>` and selection was placed after that text. In `viewPositionToDom` it was expected that either `viewPosition` has offset `0` or it has a view node before it that is bound to DOM (and then dom position was taken using that node before). However in this case both were not true, which resulted in incorrect behavior of scripts that based on `viewPositionToDom`.

5. `AttributeDelta#range` was incorrectly implemented. Before, it checked first and last operation to get start and end of range. However ranges in `AttributeDelta` are not sorted, which resulted in wrong `#range` value.

6.  During OT, when `AttributeOperation` was transformed by `AttributeOperation` and both those operations changed the same attribute from same key to same value, both those operations were converted to `NoOperation`. This resulted in faulty results in given scenario:
```
attrOp1 key 'foo', 'old' -> 'new'
attrOp2 key 'foo', 'new' -> 'old'
```
transformed by:
```
attrOp3 key 'foo', 'old' -> 'new'
```
Let's assume that second side is strong, which means that what `attrOp3` sets should be set. If that operation get's changed to `NoOperation` it will be `NoOperation` "forever". This means that on first side, element will have attribute value `old`. On second side, `attrOp3` will become `NoOperation` too, which will result in `attrOp2` being applied as is. Though both sides will have same `foo` value, it is the value from "weak" side which caused problems in further transformations. Instead of changing `attrOp3` to `NoOperation`, after it is transformed by `attrOp1` it is changed to `new -> new` and then after transformed by `attrOp2` it is transformed to `old -> new`. On the second side, it stays as it is while incoming operations becomes `NoOperation`s. This guarantees that "stronger" side is actually "stronger".

The same applies for `RenameOperation` which is very similar in how it works in OT.

7. `InsertOperation` when cloned did not clone it's `#nodes` which resulted in incorrect behavior during transformation.

8. `MoveOperation#movedRangeStart` was a property that was set in constructor and was based on `#targetPosition`. However, `#targetPosition` might change (in OT) and each time this happens, `#movedRangeStart` would have to be updated. That defeated the purpose of existing of this property (having to write less code). `#movedRangeStart` was changed to `#getMovedRangeStart()` which is evaluated on the fly basing on current value of `#targetPosition`.

9.  `MoveOperation#isSticky` is a property that is not set in a constructor but can be set after operation is created and this is used in some deltas. The error was that it wasn't used in `#fromJSON`. The result was that sticky move operation converted to JSON and parsed back was not sticky anymore.

10. `RemoveOperation` is setting it's `#targetPosition` in constructor basing on graveyard state. This is convenient when creating `RemoveOperation` instances. However, when `RemoveOperation` is created from JSON object, we expect that it will have the same `#targetPosition` as JSON-ized one. This wasn't true. Now `RemoveOperation#fromJSON` sets `RemoveOperation#targetPosition` after the operation is created, using the value from JSON object.

11. `RenameOperation` was not added to `OperationFactory`, which meant that it wasn't able to create it from JSON string.

12. Multiple enhancements have been introduced in operations transformations.

13. New special case transformations has been added for `SplitDelta`, `AttributeDelta` and `RemoveDelta`.

Even though a lot has been improved, there are some known errors that would need to be fixed later on. Also, I have some general thoughts on this part of engine. I will include all of this in a followup, umbrella ticket.


scofalik:
Related PR https://github.com/ckeditor/ckeditor5-enter/pull/35

src/view/domconverter.js
```diff
@@ -278,6 +278,12 @@ export default class DomConverter {
 				domAfter = domParent.childNodes[ 0 ];
 			} else {
 				domBefore = this.getCorrespondingDom( viewPosition.nodeBefore );
+
+				if ( !domBefore ) {
```
Reinmar: I guess you tried to fix cases like https://github.com/ckeditor/ckeditor5-autoformat/issues/13#issuecomment-259352223. I'm generally ok with this, because the only alternative is that `viewPositionToDom` will be called in a `try-catch` which smells really bad. So if we agree that `viewPositionToDom` may return `null`, then we should mention it in the documentation.

But then we should perhaps handle this more holistically in this method. E.g. `getCorrespondingDom()` may return `null` also in other cases here, so those would need to be handled as well.

cc @pjasiun

src/view/renderer.js
```diff
@@ -304,10 +304,12 @@ export default class Renderer {
 			return false;
```
Reinmar: If we agree to do https://github.com/ckeditor/ckeditor5-engine/pull/676/files#r87222594, then this won't be needed – that case will be handled inside `viewPositionToDom()`.

src/model/operation/moveoperation.js
```diff
@@ -95,13 +79,31 @@ export default class MoveOperation extends Operation {
...
+	 * inside the same parent but {@link engine.model.operation.MoveOperation#targetPosition targetPosition}
+	 * is after {@link engine.model.operation.MoveOperation#sourcePosition sourcePosition}.
+	 *
+	 *		 vv              vv
+	 *		abcdefg ===> adefbcg
+	 *		     ^          ^
+	 *		     targetPos	movedRangeStart
+	 *		     offset 6	offset 4
+	 *
+	 * @type {engine.model.Position}
```
Reinmar: s/type/returns

src/model/operation/moveoperation.js
```diff
@@ -95,13 +79,31 @@ export default class MoveOperation extends Operation {
 	}
 
 	/**
+	 * Position of the start of the moved range after it got moved. This may be different than
```
Reinmar: "Returns a position".

src/model/operation/attributeoperation.js
```diff
@@ -133,6 +134,14 @@ export default class AttributeOperation extends Operation {
 					{ node: item, key: this.key }
 				);
 			}
+
+			// If value to set is same as old value, don't do anything.
+			// By not returning `undefined`, this operation will be seen as `NoOperation` - that means
```
Reinmar: But it does return `undefined`.

src/view/domconverter.js
pjasiun: > I'm generally ok with this, because the only alternative is that viewPositionToDom will be called in a try-catch which smells really bad.

Yes, most of the view DOM conversion methods can return null, to avoid throwing. Also, I agree with you that methods, which use then, should be prepared for such scenario. But note that not in all cases such null is possible. For instance, in if you just rendered content you should be able to find a corresponding element/position without expecting a null. So we should check if the function return null in many cases, but not in all of them.

Reinmar:
To sum up what needs to be done in this PR:

1. `viewPositionToDom` needs to be reviewed. In each case when it may throw, it should return `null` instead. We need to update its documentation and its usage in renderer in that specific place which was problematic (https://github.com/ckeditor/ckeditor5-engine/pull/676#discussion_r87222913)
2. https://github.com/ckeditor/ckeditor5-engine/pull/676#discussion_r87224927 can be moved to a separate ticket. It's not necessary now.
3. Some typos.
4. Merging master into this branch and checking that everything works fine.

Reinmar:
I'm not sure whether this is relevant to this PR, but I tried to break undo/redo on this branch and weren't able. The editor was super stable despite a lot of random operations in random order. I know that this was a problem until recently, so, if not this PR, then one of the recent PRs have really stabilised the whole system. Great job anyway :)

scofalik:
I was aware that there were some problems with undo/redo but mostly they were connected with undoing/redoing selection in model, and most of fixes didn't touch it. Anyway, I remember there was at least one case of undo/redo bug that I found and fixed, however I don't remember in what case it happened and there were so many different buggy scenarios that I can't remember them all.

PR #685, opened 2016-11-17
Added MouseObserver class

### LINKED ISSUE #683: Create MouseObserver
During creation of image feature there was `MousedownObserver` needed. We decided that it would be better for it to land in the engine. Also it should be named `MouseObserver` and contain `mousedown`, `mouseup`, and `mousemove` listeners.

For now it will introduce `mousedown` event only, rest of the mouse events can be added later when there will be a need for them.

szymonkups:
Fixes #683.

### DIFF
--- a/tests/view/observer/mouseobserver.js
+++ b/tests/view/observer/mouseobserver.js
@@ -17,6 +17,10 @@ describe( 'MouseObserver', () => {
 		observer = viewDocument.addObserver( MouseObserver );
 	} );
 
+	afterEach( () => {
+		viewDocument.destroy();
+	} );
+
 	it( 'should define domEventType', () => {
 		expect( observer.domEventType ).to.equal( 'mousedown' );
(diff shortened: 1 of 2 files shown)


tests/view/observer/mouseobserver.js
```diff
@@ -0,0 +1,38 @@
...
+/* bender-tags: view, browser-only */
+
+import MouseObserver from 'ckeditor5/engine/view/observer/mouseobserver.js';
+import ViewDocument from 'ckeditor5/engine/view/document.js';
+
+describe( 'MouseObserver', () => {
+	let viewDocument, observer;
+
+	beforeEach( () => {
+		viewDocument = new ViewDocument();
```
Reinmar: You need to destroy the viewDocument in afterEach(). Otherwise it stays listening the events what affects other tests.

PR #687, opened 2016-11-17
Add findAncestor method to view.Element

### LINKED ISSUE #686: Introduce view.Element#findAncestor()
When dealing with mouse/click observers there is a need to find out if `DomEventData#target` element is placed inside some specified ancestor. For example when we want to find if clicked image is placed inside a widget.

Instead of moving it to some utils I would like to introduce `findAncestor()` on `view.Element` as it might be useful in other use cases.

It will use [view.Matcher](https://github.com/ckeditor/ckeditor5-engine/blob/master/src/view/matcher.js), and will take patterns accepted by it as paramter. Similar approach could be used for https://github.com/ckeditor/ckeditor5-engine/issues/640.

szymonkups:
Fixes #686.

### DIFF
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -575,14 +575,14 @@ export default class Element extends Node {
 	 * Provided patterns should be compatible with {@link engine.view.Matcher Matcher} as it is used internally.
 	 *
 	 * @see engine.view.Matcher
-	 * @param {Object|String|RegExp|function} patterns Patterns used to match correct ancestor. See {@link engine.view.Matcher}.
-	 * @return {engine.view.Element|null} Found element or `null` if no matching ancestor was found.
+	 * @param {Object|String|RegExp|Function} patterns Patterns used to match correct ancestor. See {@link engine.view.Matcher}.
+	 * @returns {engine.view.Element|null} Found element or `null` if no matching ancestor was found.
 	 */
 	findAncestor( ...patterns ) {
 		const matcher = new Matcher( ...patterns );
 		let parent = this.parent;
 
-		while ( parent !== null ) {
+		while ( parent ) {
 			if ( matcher.match( parent ) ) {
 				return parent;
 			}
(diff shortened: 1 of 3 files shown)


src/view/element.js
```diff
@@ -568,6 +569,31 @@ export default class Element extends Node {
...
 		property.forEach( name => this._styles.delete( name ) );
 	}
+
+	/**
+	 * Returns ancestor element that match specified pattern.
+	 * Provided patterns should be compatible with {@link engine.view.Matcher Matcher} as it is used internally.
+	 *
+	 * @see engine.view.Matcher
+	 * @param {Object|String|RegExp|function} patterns Patterns used to match correct ancestor. See {@link engine.view.Matcher}.
+	 * @return {engine.view.Element|null} Found element or `null` if no matching ancestor was found.
```
Reinmar: `@returns`

src/view/element.js
```diff
@@ -568,6 +569,31 @@ export default class Element extends Node {
...
+	 */
+	findAncestor( ...patterns ) {
```
Reinmar: Iterable `findAncestors()` would be more future safe, but perhaps it's an overkill for now. It'd also be harder to use. So I was considering it for a while, but let's not do this.

src/view/element.js
```diff
@@ -568,6 +569,31 @@ export default class Element extends Node {
 		this._fireChange( 'attributes', this );
 		property.forEach( name => this._styles.delete( name ) );
 	}
+
+	/**
+	 * Returns ancestor element that match specified pattern.
+	 * Provided patterns should be compatible with {@link engine.view.Matcher Matcher} as it is used internally.
+	 *
+	 * @see engine.view.Matcher
+	 * @param {Object|String|RegExp|function} patterns Patterns used to match correct ancestor. See {@link engine.view.Matcher}.
```
Reinmar: `Function`

src/view/element.js
```diff
@@ -568,6 +569,31 @@ export default class Element extends Node {
...
+	 * @return {engine.view.Element|null} Found element or `null` if no matching ancestor was found.
+	 */
+	findAncestor( ...patterns ) {
+		const matcher = new Matcher( ...patterns );
+		let parent = this.parent;
+
+		while ( parent !== null ) {
```
Reinmar: `while ( parent )`

PR #694, opened 2016-11-21
Introduced getSelectedContent()

### LINKED ISSUE #684: Implement DataController.getSelectedContent()
The pair for `DataController.insertContent()`. Required for https://github.com/ckeditor/ckeditor5-clipboard/issues/7.

Reinmar:
Fixes #684.

### DIFF
--- a/src/model/element.js
+++ b/src/model/element.js
@@ -198,6 +198,17 @@ export default class Element extends Node {
 		return nodes;
 	}
 
+	/**
+	 * Returns a descendant node by its path relative to this element.
+	 *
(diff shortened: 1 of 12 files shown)


src/model/element.js
```diff
@@ -198,6 +198,16 @@ export default class Element extends Node {
 		return nodes;
 	}
 
+	getNodeByPath( relativePath ) {
```
szymonkups: Docs missing.

Reinmar: I missed that I added this method completely. Good catch :) Tests are needed too.

PR #695, opened 2016-11-23
Custom properties and getFillerOffset

### LINKED ISSUE #693: Widget representation in the view
Currently we have three specialized types of elements in the view:
- `AttributeElement`,
- `ContainerElement`,
- `EmptyElement`.

Now we want to introduce widgets. Things we know so far that are needed to properly represent them in the view:

- we need to be able to tell if given view element is a widget,
- we need to be able to tell what kind of widget it is (image, table, etc.),
- we need to be able to represent widgets differently for data and editing pipelines.

### WidgetElement class

First idea was to introduce `WidgetElement` class extending `ContainerElement`.
It would be used as a base class for all future widgets. It could add proper `getFillerOffset`,  CSS classes and set `contenteditable` attribute to `false`. Finding if element is a widget and what kind of widget it is would be very easy, by using `instanceof` operator.

Problem with this approach is that `WidgetElement` class would only be valid for editing pipeline. For data pipeline each widget is just a `ContainerElement`.

### widgetize() decorator

In this approach we could define widgets using `ContainerElements`, and represent them as they should appear in data pipeline. For editing pipeline, widgets would be decorated with `widgetize` function:
```
createImageWidget(); // <figure class="image"><img src="..." alt="..." /></figure>
widgetize( createImageWidget() ); // <figure class="image ck-widget" contenteditable="false"><img src="..." alt="..." /></figure>
```
This way we could avoid creation of new sub-classes, and show that representation in editing pipeline just "decorates" data representation.

But how to distinguish widget elements from each other? `widgetize()` could add some custom properties to decorated element:
```
function widgetize( viewElement ) {
    viewElement.setAttribute( 'contenteditable', false );
    viewElement.addClass( 'ck-widget' );
    viewElement.isWidget = true;
}
```
but those custom properties will not be present in cloned elements, so this approach is not totally valid. 

Maybe we could add some meta-information to `ContainerElements` about widget type as some special attributes, that would be cloned as regular attributes but will not be rendered?

szymonkups:
Fixes #693.

### DIFF
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -163,6 +163,11 @@ export default class Element extends Node {
 		// Clone custom properties.
 		cloned._customProperties = new Map( this._customProperties );
 
+		// Clone filler offset method if one exists.
+		if ( this.getFillerOffset ) {
+			cloned.getFillerOffset = this.getFillerOffset;
+		}
+
 		return cloned;
 	}
 
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -645,7 +645,7 @@ export default class Element extends Node {
 	 *
 	 * @returns {Iterable.<*>}
 	 */
-	*getCustomProperties( ) {
+	*getCustomProperties() {
 		yield* this._customProperties.entries();
 	}
 }
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -614,7 +614,8 @@ export default class Element extends Node {
 	}
 
 	/**
-	 * Sets a custom property.
+	 * Sets a custom property. Unlike attributes, custom properties are not rendered to the DOM,
+	 * so they can be used to add special data to elements.
 	 *
 	 * @param {String|Symbol} key
 	 * @param {*} value
@@ -654,7 +655,7 @@ export default class Element extends Node {
 	}
 
 	/**
-	 * Returns block {@link module:engine/view/filter~Filter filler} offset or `null` if block filler is not needed.
+	 * Returns block {@link module:engine/view/filler~Filler filler} offset or `null` if block filler is not needed.
 	 *
 	 * @abstract
 	 * @method module:engine/view/element~Element#getFillerOffset
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -105,6 +105,15 @@ export default class Element extends Node {
 			parseInlineStyles( this._styles, this._attrs.get( 'style' ) );
 			this._attrs.delete( 'style' );
 		}
+
+		/**
+		 * Map of custom properties.
+		 * Custom properties can be added to element instance, will be cloned but not rendered into DOM.
+		 *
+		 * @protected
+		 * @memeber {Map} engine.view.Element#_customProperties.
+		 */
+		this._customProperties = new Map();
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/view/element.js
```diff
@@ -151,6 +160,14 @@ export default class Element extends Node {
 		cloned._classes = new Set( this._classes );
 		cloned._styles = new Map( this._styles );
 
+		// Clone custom properties.
+		cloned._customProperties = new Map( this._customProperties );
+
+		// Clone filler offset method if one exists.
+		if ( this.getFillerOffset ) {
```
Reinmar: Don't all elements need this method? The check shouldn't be needed if `Element` class defines this abstract method, because the code outside will try to use it anyway.

src/view/element.js
```diff
@@ -597,6 +614,45 @@ export default class Element extends Node {
...
+	 *
+	 * @param {String|Symbol} key
+	 * @returns {Boolean} Returns true if property was removed.
+	 */
+	removeCustomProperty( key ) {
+		return this._customProperties.delete( key );
+	}
+
+	/**
+	 * Returns iterator that iterates over this element's custom properties.
```
Reinmar: "an iterator which"

src/view/element.js
```diff
@@ -597,6 +614,45 @@ export default class Element extends Node {
...
+	 *
+	 * @param {String|Symbol} key
+	 * @returns {*}
+	 */
+	getCustomProperty( key ) {
+		return this._customProperties.get( key );
+	}
+
+	/**
+	 * Removes custom property stored under given key.
```
Reinmar: "the custom property", "the given key"

src/view/element.js
```diff
@@ -597,6 +614,45 @@ export default class Element extends Node {
...
+	 *
+	 * @param {String|Symbol} key
+	 * @param {*} value
+	 */
+	setCustomProperty( key, value ) {
+		this._customProperties.set( key, value );
+	}
+
+	/**
+	 * Returns custom property value for given key.
```
Reinmar: "the custom property", "the given key"

src/view/element.js
```diff
@@ -597,6 +614,45 @@ export default class Element extends Node {
 
 		return null;
 	}
+
+	/**
+	 * Sets custom property.
```
Reinmar: "a" (I hope)

src/view/element.js
Reinmar: Besides that, some explanation that this is iterator of key->value pairs.

src/view/emptyelement.js
```diff
@@ -29,6 +29,18 @@ export default class EmptyElement extends Element {
 		if ( arguments.length > 2 ) {
 			throwCannotAdd();
 		}
+
+		/**
+		 * Returns `null` because filler is not needed.
+		 * This method is implemented in a constructor because it is treated as an injected behaviour and it might be
+		 * changed during execution (in elements representing widgets for example).
+		 * When {@link module:engine/view/element~Element.clone element is cloned}, this method will be referenced from
```
Reinmar: Unnecessary explanation from the API docs perspective (the whole piece about where it's defined and that it will be cloned). It's defined as a method, so it will be cloned and it will be available in all instances of this class.

szymonkups:
Pushed changes and merged with latest master. Just a reminder to close with this PR: https://github.com/ckeditor/ckeditor5-list/pull/26.

PR #700, opened 2016-11-29
T/643 Markers in engine

### LINKED ISSUE #643: Markers
It seems to be the last big piece of this puzzle.
# What is the problem?

There are some fancy features, changes on the document one may want to apply which are not regular model attributes:
- showing selections of other users in collaborative editing,
- marking comments (similar to google docs),
- showing find results (Chome style),
- underline misspelled words.

Let's call the "markers".
# Markers vs attributes

What is the main difference between markers and attributes (like bold or link)?

Attributes are assigned to letters. It does not matter if one broke paragraph or moved part of the text: bolded letters should still be bolded. Markers, on the other hand, need to be consistent, part of the marker should not be moved in the different part of the document.

Moreover, when an attribute is applied it does not matter what will happen with it later. Finding all texts with a certain attribute is an edge case and can be slow. At the same time, for markers, we need to be able to observe changes on then and react on these changes. Finding marker in the document should be very quick and should not require whole document scanning.
# Implementation

It looks that `LiveRange`s and `LivePosition`s will fit perfectly here. Document should have a collection of named markers, which are pairs of the name and `LiveRange`/`LivePosition`:

``` js
document.markers.set( 'findResult', new LiveRange( ... ) );
```

`document.markers` should fire events whenever a marker is set, changed or removed.

Then, it should be possible to easily create converters from markers to view attributes, for instance:

``` js
converterBuilder.for( dipatcher ).fromMarker( 'findResult' ).to( 'span', { color: 'yellow' ) );
converterBuilder.for( dipatcher ).fromCollapsedMarker( 'extraSelection' ).to( 'span', { color: 'red' ) );
```

These converters should handle inserting, removing and moving content from the marked fragment. They should also handle updating marker (creating a new marker with the same name).
# Additional requirements
## Should work with OT

It should be possible to transform markers the same way operations are transformed. When the marker is created on one document it should be possible to recreate it on the different state of the document. Then marker should keep the synchronization with the document (it should happen automatically thanks to `LiveRanges`).
## Groups of markers

It should be possible to create markers names using namespaces: (`selection:john`, `selection:marry`, ...) and listen to the generic event (`markerCreate:selection`).
## Markers should not break text nodes

This is not an obligated requirement. 

Because markers do not have a semantic information about the content, it would be nice if they would not break text nodes in the DOM and not break the native spell checking.

scofalik:
This PR introduces markers implementation in CKEditor5 engine. Fixes #643.

Other than that some changes and fixes has been done:
* `Mapper` position mapping events have changed, now callbacks receive position mapped by default algorithm (previously default mapping was a lowest priority callback). This way default mapping can be used in custom callbacks.
* `LiveRange` transformation has been changed. `LiveRange` is not sticky anymore (which was weird from UX perspective).
* `Range.createFromRanges` was introduced, it glues multiple ranges into one. Useful after a range is transformed.
* `view.Writer.move` and `ModelConversionDispatcher#convertMove` were fixed.
* other slight fixes.

Related PRs and follow-up issues will be linked in comments below.

### DIFF
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -84,21 +84,21 @@ export default class EditingController {
 		 * @private
 		 * @member {utils.EmitterMixin} #_listenter
 		 */
-		this._listenter = Object.create( EmitterMixin );
+		this._listener = Object.create( EmitterMixin );
 
 		// Convert changes in model to view.
-		this._listenter.listenTo( this.model, 'change', ( evt, type, changes ) => {
+		this._listener.listenTo( this.model, 'change', ( evt, type, changes ) => {
 			this.modelToView.convertChange( type, changes );
 		}, { priority: 'low' } );
 
 		// Convert model selection to view.
-		this._listenter.listenTo( this.model, 'changesDone', () => {
+		this._listener.listenTo( this.model, 'changesDone', () => {
 			this.modelToView.convertSelection( model.selection );
 			this.view.render();
 		}, { priority: 'low' } );
 
 		// Convert view selection to model.
-		this._listenter.listenTo( this.view, 'selectionChange', convertSelectionChange( model, this.mapper ) );
+		this._listener.listenTo( this.view, 'selectionChange', convertSelectionChange( model, this.mapper ) );
 
 		// Attach default content converters.
 		this.modelToView.on( 'insert:$text', insertText(), { priority: 'lowest' } );
@@ -146,6 +146,6 @@ export default class EditingController {
 	 */
 	destroy() {
 		this.view.destroy();
-		this._listenter.stopListening();
+		this._listener.stopListening();
 	}
 }
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -8,6 +8,7 @@
  */
 
 import ViewDocument from '../view/document.js';
+import ModelRange from '../model/range.js';
 import Mapper from '../conversion/mapper.js';
 import ModelConversionDispatcher from '../conversion/modelconversiondispatcher.js';
 import { insertText, remove, move, rename } from '../conversion/model-to-view-converters.js';
@@ -97,6 +98,15 @@ export default class EditingController {
 			this.view.render();
 		}, { priority: 'low' } );
 
+		// Convert model markers changes.
+		this._listener.listenTo( this.model.markers, 'addMarker', ( evt, name, range ) => {
+			this.modelToView.convertMarker( 'addMarker', name, range );
+		} );
+
+		this._listener.listenTo( this.model.markers, 'removeMarker', ( evt, name, range ) => {
+			this.modelToView.convertMarker( 'removeMarker', name, range );
+		} );
+
 		// Convert view selection to model.
 		this._listener.listenTo( this.view, 'selectionChange', convertSelectionChange( model, this.mapper ) );
 
@@ -105,6 +115,33 @@ export default class EditingController {
 		this.modelToView.on( 'remove', remove(), { priority: 'low' } );
 		this.modelToView.on( 'move', move(), { priority: 'low' } );
 		this.modelToView.on( 'rename', rename(), { priority: 'low' } );
+
+		// Attach markers converters.
+		this.modelToView.on( 'insert', ( evt, data ) => {
+			const pos = data.range.start;
+
+			for ( let [ range, name ] of this.model.markers._markerNames ) {
+				if ( range.containsPosition( pos ) || range.start.isEqual( pos ) ) {
+					this.modelToView.convertMarker( 'addMarker', name, data.range );
+				}
+			}
+		}, { priority: 'lowest' } );
+
+		this.modelToView.on( 'move', ( evt, data ) => {
+			const pos = data.sourcePosition._getTransformedByInsertion( data.targetPosition, data.item.offsetSize );
+
+			for ( let [ range, name ] of this.model.markers._markerNames ) {
+				if ( range.start.isEqual( pos ) || range.end.isEqual( pos ) ) {
+					const movedRange = ModelRange.createOn( data.item );
+
+					if ( range.containsPosition( data.targetPosition ) ) {
+						this.modelToView.convertMarker( 'addMarker', name, movedRange );
+					} else {
+						this.modelToView.convertMarker( 'removeMarker', name, movedRange );
+					}
+				}
+			}
+		}, { priority: 'lowest' } );
 
 		// Attach default selection converters.
 		this.modelToView.on( 'selection', clearAttributes(), { priority: 'low' } );
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -8,10 +8,16 @@
  */
 
 import ViewDocument from '../view/document.js';
-import ModelRange from '../model/range.js';
 import Mapper from '../conversion/mapper.js';
 import ModelConversionDispatcher from '../conversion/modelconversiondispatcher.js';
-import { insertText, remove, move, rename } from '../conversion/model-to-view-converters.js';
+import {
+	insertText,
+	remove,
+	move,
+	rename,
+	insertIntoMarker,
+	moveInOutOfMarker
+} from '../conversion/model-to-view-converters.js';
 import { convertSelectionChange } from '../conversion/view-selection-to-model-converters.js';
 import {
 	convertRangeSelection,
@@ -99,11 +105,11 @@ export default class EditingController {
 		}, { priority: 'low' } );
 
 		// Convert model markers changes.
-		this._listener.listenTo( this.model.markers, 'addMarker', ( evt, name, range ) => {
+		this._listener.listenTo( this.model.markers, 'add', ( evt, name, range ) => {
 			this.modelToView.convertMarker( 'addMarker', name, range );
 		} );
 
-		this._listener.listenTo( this.model.markers, 'removeMarker', ( evt, name, range ) => {
+		this._listener.listenTo( this.model.markers, 'remove', ( evt, name, range ) => {
 			this.modelToView.convertMarker( 'removeMarker', name, range );
 		} );
 
@@ -116,32 +122,9 @@ export default class EditingController {
 		this.modelToView.on( 'move', move(), { priority: 'low' } );
 		this.modelToView.on( 'rename', rename(), { priority: 'low' } );
 
-		// Attach markers converters.
-		this.modelToView.on( 'insert', ( evt, data ) => {
-			const pos = data.range.start;
-
-			for ( let [ range, name ] of this.model.markers._markerNames ) {
-				if ( range.containsPosition( pos ) || range.start.isEqual( pos ) ) {
-					this.modelToView.convertMarker( 'addMarker', name, data.range );
-				}
-			}
-		}, { priority: 'lowest' } );
-
-		this.modelToView.on( 'move', ( evt, data ) => {
-			const pos = data.sourcePosition._getTransformedByInsertion( data.targetPosition, data.item.offsetSize );
-
-			for ( let [ range, name ] of this.model.markers._markerNames ) {
-				if ( range.start.isEqual( pos ) || range.end.isEqual( pos ) ) {
-					const movedRange = ModelRange.createOn( data.item );
-
-					if ( range.containsPosition( data.targetPosition ) ) {
-						this.modelToView.convertMarker( 'addMarker', name, movedRange );
-					} else {
-						this.modelToView.convertMarker( 'removeMarker', name, movedRange );
-					}
-				}
-			}
-		}, { priority: 'lowest' } );
+		// Attach default markers converters.
+		this.modelToView.on( 'insert', insertIntoMarker( this.model.markers ), { priority: 'lowest' } );
+		this.modelToView.on( 'move', moveInOutOfMarker( this.model.markers ), { priority: 'lowest' } );
 
 		// Attach default selection converters.
 		this.modelToView.on( 'selection', clearAttributes(), { priority: 'low' } );
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -12,7 +12,9 @@ import {
 	setAttribute,
 	removeAttribute,
 	wrap,
-	unwrap
+	unwrap,
+	wrapMarker,
+	unwrapMarker
 } from './model-to-view-converters.js';
 
 import { convertSelectionAttribute } from './model-selection-to-view-converters.js';
@@ -30,7 +32,7 @@ import ViewContainerElement from '../view/containerelement.js';
  * {@link module:engine/conversion/model-to-view-converters}, {@link module:engine/conversion/modelconsumable~ModelConsumable},
  * {@link module:engine/conversion/mapper~Mapper}.
  *
- * Using this API it is possible to create three kinds of converters:
+ * Using this API it is possible to create four kinds of converters:
  *
  * 1. Model element to view element converter. This is a converter that takes the model element and represents it
  * in the view.
@@ -50,6 +52,14 @@ import ViewContainerElement from '../view/containerelement.js';
  * set on model text nodes is converter to `strong` view element.
  *
  *		buildModelConverter().for( dispatcher ).fromAttribute( 'bold' ).toElement( 'strong' );
+ *
+ * 4. Model marker to view element converter. This is a converter that converts markers from given group to view attribute element.
+ * Markers, basically, are {@link engine.model.LiveRange} instances, that are named. In this conversion, model range is
+ * converted to view range, then that view range is wrapped (or unwrapped, if range is removed) in a view attribute element.
+ * To learn more about markers, see {@link engine.model.MarkersCollection}.
+ *
+ *		const viewSpanSearchResult = new ViewAttributeElement( 'span', { class: 'search-result' } );
+ *		buildModelConverter().for( dispatcher ).fromMarker( 'searchResult' ).toElement( viewSpanSearchResult );
  *
  * It is possible to provide various different parameters for
  * {@link module:engine/conversion/buildmodelconverter~ModelConverterBuilder#toElement}
@@ -137,6 +147,23 @@ class ModelConverterBuilder {
 	}
 
 	/**
+	 * Registers what type of marker should be converted.
+	 *
+	 * @chainable
+	 * @param {String} markerName Name of marker to convert.
+	 * @returns {engine.conversion.ModelConverterBuilder}
+	 */
+	fromMarker( markerName ) {
+		this._from = {
+			type: 'marker',
+			name: markerName,
+			priority: null
+		};
+
+		return this;
+	}
+
+	/**
 	 * Changes default priority for built converter. The lower the number, the earlier converter will be fired.
 	 * Default priority is `10`.
 	 *
@@ -190,7 +217,8 @@ class ModelConverterBuilder {
 	 * This method creates the converter and adds it as a callback to a proper
 	 * {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher conversion dispatcher} event.
 	 *
-	 * @param {String|module:engine/view/element~Element|Function} element Element created by converter.
+	 * @param {String|module:engine/view/element~ViewElement|Function} element Element created by converter or
+	 * a function that returns view element.
 	 */
 	toElement( element ) {
 		const priority = this._from.priority === null ? 'normal' : this._from.priority;
@@ -201,7 +229,7 @@ class ModelConverterBuilder {
 				element = typeof element == 'string' ? new ViewContainerElement( element ) : element;
 
 				dispatcher.on( 'insert:' + this._from.name, insertElement( element ), { priority } );
-			} else {
+			} else if ( this._from.type == 'attribute' ) {
 				// From model attribute to view element -> wrap and unwrap.
 				element = typeof element == 'string' ? new ViewAttributeElement( element ) : element;
 
@@ -210,6 +238,12 @@ class ModelConverterBuilder {
 				dispatcher.on( 'removeAttribute:' + this._from.key, unwrap( element ), { priority } );
 
 				dispatcher.on( 'selectionAttribute:' + this._from.key, convertSelectionAttribute( element ), { priority } );
+			} else {
+				// From marker to view element -> wrapRange and unwrapRange.
+				element = typeof element == 'string' ? new ViewAttributeElement( element ) : element;
+
+				dispatcher.on( 'addMarker:' + this._from.name, wrapMarker( element ), { priority } );
+				dispatcher.on( 'removeMarker:' + this._from.name, unwrapMarker( element ), { priority } );
 			}
 		}
 	}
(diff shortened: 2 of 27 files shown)


scofalik:
Related PRs:
https://github.com/ckeditor/ckeditor5-list/pull/28
https://github.com/ckeditor/ckeditor5-undo/pull/53
https://github.com/ckeditor/ckeditor5-autoformat/pull/16

scofalik:
Follow-ups:
#701 - Selection is incorrect after creating intersecting marker.
#702 - Selection break view elements created by markers. This issue should be solved first as it may solve other #701 and #703 
#703 - Non-collapsed selection sometimes is not created when it starts inside marker
#704 - Refactoring/enhancement in `EditingController`. Not a bug, not critical.
#705 - Refactoring/enhancement in conversion eco-system. Not a bug, not critical.

Also in undo:
https://github.com/ckeditor/ckeditor5-undo/issues/54
https://github.com/ckeditor/ckeditor5-undo/issues/55

src/conversion/model-to-view-converters.js
```diff
@@ -288,6 +289,157 @@ export function unwrap( elementCreator ) {
...
+ *
+ *		modelDispatcher.on( 'addMarker:searchResult', wrapMarker( new ViewAttributeElement( 'span', { class: 'searchResult' } ) ) );
+ *
+ * @external engine.conversion.modelToView
+ * @function engine.conversion.modelToView.wrapMarker
+ * @param {engine.view.AttributeElement|Function} elementCreator View attribute element, or function returning
+ * a view attribute element, which will be used for wrapping.
+ * @returns {Function} Add marker converter.
+ */
+export function wrapMarker( elementCreator ) {
```
pjasiun: What is the difference between `wrap` and `wrapMarker`? From the view point of view, there is nothing like `marker` to be wrapped.

src/conversion/model-to-view-converters.js
```diff
@@ -288,6 +289,157 @@ export function unwrap( elementCreator ) {
...
+}
+
+// Starting from given `viewPosition`, moving forward, finds the furthest position in a view, which is
+// separated only by elements with zero model length (which in most cases are empty `ViewAttributeElement`s).
+// Using `viewPosition` and returned position, one can build a view range, which start and position are mapped to
+// the same position in model.
+// @param {engine.view.Position} viewPosition Position to start from when looking for furthest zero length position.
+// @param {engine.conversion.Mapper} mapper Mapper to use when looking for furthest zero length position.
+// @returns {engine.view.Position} Furthest zero length position.
+function findFurthestZeroLengthPoint( viewPosition, mapper ) {
```
pjasiun: Enlarge?

Reinmar:
Remember to fix API docs. This PR has been made before we reformatted them.

src/controller/editingcontroller.js
```diff
@@ -84,28 +85,64 @@ export default class EditingController {
...
+					const movedRange = ModelRange.createOn( data.item );
+
+					if ( range.containsPosition( data.targetPosition ) ) {
+						this.modelToView.convertMarker( 'addMarker', name, movedRange );
+					} else {
+						this.modelToView.convertMarker( 'removeMarker', name, movedRange );
+					}
+				}
+			}
+		}, { priority: 'lowest' } );
```
pjasiun: These two functions (for insert and move) should be moved to the separate functions like all other.

src/controller/editingcontroller.js
```diff
@@ -84,28 +85,64 @@ export default class EditingController {
...
+
+			for ( let [ range, name ] of this.model.markers._markerNames ) {
+				if ( range.containsPosition( pos ) || range.start.isEqual( pos ) ) {
+					this.modelToView.convertMarker( 'addMarker', name, data.range );
+				}
+			}
+		}, { priority: 'lowest' } );
+
+		this.modelToView.on( 'move', ( evt, data ) => {
+			const pos = data.sourcePosition._getTransformedByInsertion( data.targetPosition, data.item.offsetSize );
```
pjasiun: Why do you need to transform this position here?

src/model/markerscollection.js
```diff
@@ -0,0 +1,142 @@
...
+export default class MarkersCollection {
+	/**
+	 * Creates a markers collection.
+	 */
+	constructor() {
+		/**
+		 * Stores range to marker name bindings for added ranges.
+		 *
+		 * @private
+		 * @member {Map} #_markerNames
```
pjasiun: Why this is called "names" if it contains ranges?

src/controller/editingcontroller.js
```diff
@@ -84,28 +85,64 @@ export default class EditingController {
...
 		this.modelToView.on( 'insert:$text', insertText(), { priority: 'lowest' } );
 		this.modelToView.on( 'remove', remove(), { priority: 'low' } );
 		this.modelToView.on( 'move', move(), { priority: 'low' } );
 		this.modelToView.on( 'rename', rename(), { priority: 'low' } );
 
+		// Attach markers converters.
+		this.modelToView.on( 'insert', ( evt, data ) => {
+			const pos = data.range.start;
+
+			for ( let [ range, name ] of this.model.markers._markerNames ) {
```
pjasiun: You iterate here over the private property. I believe `this.model.markers` should be iterable and return `[ name, range ]`.

src/controller/editingcontroller.js
```diff
@@ -84,28 +85,64 @@ export default class EditingController {
...
+				if ( range.containsPosition( pos ) || range.start.isEqual( pos ) ) {
+					this.modelToView.convertMarker( 'addMarker', name, data.range );
+				}
+			}
+		}, { priority: 'lowest' } );
+
+		this.modelToView.on( 'move', ( evt, data ) => {
+			const pos = data.sourcePosition._getTransformedByInsertion( data.targetPosition, data.item.offsetSize );
+
+			for ( let [ range, name ] of this.model.markers._markerNames ) {
```
pjasiun: As above.

src/conversion/modelconversiondispatcher.js
```diff
@@ -328,6 +328,13 @@ export default class ModelConversionDispatcher {
 		}
 	}
 
+	convertMarker( type, name, range ) {
```
pjasiun: Docs missing.

src/conversion/model-to-view-converters.js
```diff
@@ -288,6 +289,157 @@ export function unwrap( elementCreator ) {
...
+
+	return viewPosition;
+}
+
+// Breaks given `viewRange` on "flat" view ranges, where "flat" means that the range is contained within
+// one container element. After `viewRange` is broken, it's "pieces" can be used in ViewWriter (which expects
+// that passed ranges are contained within one container element).
+// @param {engine.view.Range} viewRange Range to break.
+// @returns {Array.<engine.view.Range>} Flat ranges that combines into passed `viewRange`.
+function breakViewRangeOnFlat( viewRange ) {
```
scofalik: For now, I will just move it. I thought about making this a default behavior of methods like `viewWriter.remove` or `viewWriter.wrap` but those methods return values and I think it would be too many changes in this PR.

src/conversion/modelconversiondispatcher.js
```diff
@@ -360,7 +367,7 @@ export default class ModelConversionDispatcher {
 	 * @param {String} type Consumable type.
 	 * @returns {module:engine/conversion/modelconsumable~ModelConsumable} Values to consume.
 	 */
-	_createConsumableForRange( range, type ) {
+	_createConsumableForType( range, type ) {
```
pjasiun: Why rename?

src/conversion/modelconversiondispatcher.js
```diff
@@ -389,6 +396,14 @@ export default class ModelConversionDispatcher {
 		return consumable;
 	}
 
+	_createRangeConsumable( range ) {
+		const consumable = new Consumable();
+
+		consumable.add( range, 'range' );
```
pjasiun: There should be an action to consume, not `range`, so for instance `makerCreate`, `makerRemove`.

src/conversion/model-to-view-converters.js
scofalik: I will rename it to `enlargeViewPosition` and it will return a `view.Range`.

src/conversion/buildmodelconverter.js
```diff
@@ -246,8 +280,8 @@ class ModelConverterBuilder {
 	 * @param {*} [value] Attribute value.
 	 */
 	toAttribute( keyOrCreator, value ) {
-		if ( this._from.type == 'element' ) {
-			// Converting from model element to view attribute is unsupported.
+		if ( this._from.type != 'attribute' ) {
+			// Only converting from attribute to attribute is supported.
```
scofalik: I will throw `CKEditorError`.

src/conversion/modelconversiondispatcher.js
```diff
@@ -389,6 +396,14 @@ export default class ModelConversionDispatcher {
 		return consumable;
 	}
 
+	_createRangeConsumable( range ) {
```
pjasiun: This function name is misleading.

src/model/markerscollection.js
scofalik: Because it keeps range => name mapping. It makes sense when you write:
`const name = this._markerNames.get( range );`.

src/model/markerscollection.js
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * Throws, if given {@link module:engine/model/liverange~LiveRange LiveRange} instance is already added to the collection.
+	 *
+	 * `MarkersCollection` instance listens to `change` event of all live ranges added to it and fires `removeMarker`
+	 * and `addMarker` events when the live range changes.
+	 *
+	 * @param {module:engine/model/liverange~LiveRange} liveRange Live range to be added as a marker to markers collection.
+	 * @param {String} markerName Name to be associated with given `liveRange`.
+	 */
+	addRange( liveRange, markerName ) {
```
pjasiun: It's markers collection why I'm adding `range`? I'm adding range with a name so marker (even `addMarker` event is fired). So `addMarker` would be better. But since there is nothing else I can add maybe simple `add`?

Also, it's more natural to have a name as the first parameter, so:

`add( name, range )`?

src/controller/editingcontroller.js
scofalik: I'll make `MarkersCollection` iterable and it will return an object with `name` and `range` properties.

src/model/markerscollection.js
```diff
@@ -0,0 +1,142 @@
...
+			/**
+			 * Added range is not an instance of LiveRange.
+			 *
+			 * @error markers-collection-add-range-not-live-range
+			 */
+			throw new CKEditorError( 'markers-collection-add-range-not-live-range: Added range is not an instance of LiveRange.' );
+		}
+
+		this._markerNames.set( liveRange, markerName );
+		this.fire( 'addMarker', markerName, Range.createFromRange( liveRange ) );
```
pjasiun: As above, maybe simple `add` event like in the `Collection` class?

src/model/markerscollection.js
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	/**
+	 * Removes a live range from markers collection and makes markers collection stop listening to that live range
+	 * `change` event.
+	 *
+	 * @param {module:engine/model/liverange~LiveRange} liveRange Live range to be removed from markers collection.
+	 * @returns {Boolean} `true` is passed `liveRange` was found and removed from the markers collection, `false` otherwise.
+	 */
+	removeRange( liveRange ) {
```
pjasiun: As above. It would be also nice to remove the marker by name.

src/model/markerscollection.js
```diff
@@ -0,0 +1,142 @@
...
+		if ( markerName ) {
+			this.removeRange( oldLiveRange );
+			this.addRange( newLiveRange, markerName );
+
+			return true;
+		}
+
+		return false;
+	}
+
```
pjasiun: The `get` method is missing.

src/conversion/modelconversiondispatcher.js
scofalik: I added docs for the method and also missing docs for events.

src/conversion/modelconversiondispatcher.js
scofalik: In F2F discussion we agreed to keep the old name.

src/conversion/modelconversiondispatcher.js
scofalik: I'll change to `_createMarkerConsumable` because it will be only for one marker at a time.

scofalik: Also added docs.

src/conversion/modelconversiondispatcher.js
scofalik: I'll change it to `addMarker` and `removeMarker`.

src/model/markerscollection.js
scofalik: Okay.

src/model/markerscollection.js
```diff
@@ -0,0 +1,142 @@
...
+	 * This method is basically a wrapper for using {@link module:engine/model/markerscollection~MarkersCollection#removeRange removeRange}
+	 * followed by using {@link module:engine/model/markerscollection~MarkersCollection#addRange addRange}.
+	 *
+	 * **Note**: this method does not change properties of `oldLiveRange`.
+	 *
+	 * @param {module:engine/model/liverange~LiveRange} oldLiveRange Live range to be changed.
+	 * @param {module:engine/model/liverange~LiveRange} newLiveRange Live range to be added.
+	 * @returns {Boolean} `true` if `oldLiveRange` was found and changed, `false` otherwise.
+	 */
+	updateRange( oldLiveRange, newLiveRange ) {
```
scofalik: Will be `update( nameOrRange, newLiveRange )`

src/model/markerscollection.js
scofalik: Will simplify.

src/model/markerscollection.js
scofalik: Will add `get( name )`

src/conversion/model-to-view-converters.js
scofalik: As far as name is concerned, I agree that the name could be different, i.e. `wrapRange` (while `wrap` could be changed to `wrapItem`, same for `unwrap`s).

The functions (`wrap` and `wrapMarker`) could be merged. But there are couple of differences between them:
1. Consumable type. It's taken from `eventNameToConsumableType( evt.name )` and would be invalid for `wrapMarker`. This would have to change.
2. `data` object passed to the converter is different. We can try to unify them a bit though. `wrap` gets: `item`, `range`, `attributeKey`, `attributeOldValue` and `attributeNewValue`. `wrapMarker` gets `range` and `name`. We could add `data.item` to `wrapMarker` `data` and make it equal to `data.range`, but that's it.
3. There's difference if view element is created using "element creator function":
`wrap`: `elementCreator( data.attributeNewValue, data, consumable, conversionApi );`
`wrapMarker`: `elementCreator( data, consumable, conversionApi );`
It would have to be something like:
`elementCreator( data.attributeNewValue || data.name, data, consumable, conversionApi );` and still, `elementCreator` functions would have to be different for converting items and ranges which means that we need to write about it in docs and it might be confusing.
4. Both converters have special behavior that are individual for them. `wrap` has to "unwrap" part of view if attribute changed. `wrapMarker` has to support collapsed ranges. Those would be just two `if`s standing next to each other, operating on different values and "active" depending on `data` content. We can put them together in one function, but on the same basis we could put everything into one class.

All in all the merging would not be as seamless as it looks on first sight so I am against it. Let's change `wrapMarker` to `wrapRange` and `wrap` to `wrapItem`, same for `unwrap` and keep them separated.

pjasiun: You're right. Maybe we should call them explicit: `wrapInAttribute` and `wrapInMarker`?

pjasiun: And the documentation should stress the difference.

src/controller/editingcontroller.js
scofalik: To make conversion easier, `data.sourcePosition` for `move` change is in state "before" operation happened. However, both `data.targetPosition` and markers state are already "after" operation. To check whether moved part of document was inside a range or was moved to the range, we need to transform `sourcePosition`.

It's only transforming by insertion, because move `sourcePosition` (in other words, start of moved range) could not change due to removing that range:
`ab[cde]f` <- if `cde` is moved, source position is after `b` and will stay there. Only change can happen if range is moved before letter `b`.

scofalik: Buuuuut, when I look at the conditions I am not sure why exactly I check `if ( marker.range.start.isEqual( pos ) || marker.range.end.isEqual( pos ) ) {`. I remember this condition went through some iterations.

pjasiun: Can you leave a comment there?

src/controller/editingcontroller.js
pjasiun: Maybe the array is better to keep it consistent with https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/node.js#L299-L309

src/controller/editingcontroller.js
scofalik: Checking with the start is because when you have a (marker) range and then you "cut" something out of it, the transformed range always starts or ends on sourcePosition. Examples:

`ab[cd{e]f}g` -> move `{}` -> `ab[cd]^g` -> `sourcePosition` will be after `d`
`a{b[c}de]fg` -> move `{}` -> `a^[de]fg` -> `sourcePosition` will be after `a`

That's why it makes sense to check `marker.range.start` and `marker.range.end`. But I will take a look if that loop cannot be simplified.

src/controller/editingcontroller.js
scofalik: Okay...

src/model/markerscollection.js
scofalik: Since we'll implement getting, removing and updating markers by name, it makes sense to store `name` => `range` mapping now. I'll rename this property to `_nameToRange`.

scofalik:
I have corrected all raised issues and updated docs.

tests/manual/tickets/643/1.js
```diff
@@ -0,0 +1,106 @@
...
+} );
+
+const ranges = [];
+
+function addHighlight( color ) {
+	model.enqueueChanges( () => {
+		const range = LiveRange.createFromRange( model.selection.getFirstRange() );
+
+		ranges.push( range );
+		model.markers.addRange( range, 'highlight:' + color );
```
pjasiun: It looks like the manual tests is not updated.

src/conversion/model-to-view-converters.js
```diff
@@ -272,6 +271,127 @@ export function unwrap( elementCreator ) {
...
+ * will be cloned and the copy will become the wrapping element. If `Function` is provided, it is passed all the parameters of the
+ * {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher#event:addMarker addMarker event}. It's expected
+ * that the function returns a {@link module:engine/view/attributeelement~AttributeElement}. The result of the function will be the
+ * wrapping element. When provided `Function` does not return element, then will be no conversion.
+ *
+ * The converter automatically consumes corresponding value from consumables list, stops the event (see
+ * {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher}).
+ *
+ *		modelDispatcher.on( 'addMarker:searchResult', wrapRange( new ViewAttributeElement( 'span', { class: 'searchResult' } ) ) );
+ *
```
pjasiun: I would like to see here the difference between `wrapItem` and this, so no one in the future asks again can we merge these two converters. I would be nice to have it written directly: "In contrast to `wrapItem`...".

src/model/markerscollection.js
```diff
@@ -0,0 +1,190 @@
...
+ * Markers can be converted to view by adding appropriate converters for
+ * {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher#event:addMarker addMarker} and
+ * {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher#event:removeMarker removeMarker}
+ * events, or by building converters for {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher}
+ * using {@link module:engine/conversion/buildmodelconverter~buildModelConverter model converter builder}.
+ *
+ * Markers are similar to adding and converting attributes on nodes. The difference is that attribute is connected to
+ * a given node (i.e. a character is bold no matter if it gets moved or content around it changes). Markers on the
+ * other hand are continuous ranges (i.e. if a character from inside of marker range is moved somewhere else, marker
+ * range is shrunk and the character does not have any attribute or information that it was in the marked range).
```
pjasiun: Also, what is most important, markers let you find quickly the marked range without searching through the whole document.

src/model/markerscollection.js
```diff
@@ -0,0 +1,190 @@
...
+
+	/**
+	 * Destroys markers collection.
+	 */
+	destroy() {
+		this.stopListening();
+	}
+
+	/**
+	 * Returns range for given name.
```
pjasiun: Returns name for given range.

src/model/markerscollection.js
```diff
@@ -0,0 +1,190 @@
...
+	 *
+	 * @private
```
pjasiun: Maybe this method could be public, it might be useful. Then `update` and `remove` could simply get a `name`.

pjasiun: Or we could keep it simple and drop the possibility of updating or removing marker by the range and remove this method. We copy range very ofter and it is a potential problem that someone uses a copy of a range to remove or update a marker. KISS, let's use names.

scofalik: I don't agree. I understand that you requested additional possibility to pass name instead of range. That was backed up by arguments and made sense. But why remove existing feature that works, is tested and it's not like this is 100LOC.

src/view/writer.js
```diff
@@ -271,6 +273,36 @@ export function mergeContainers( position ) {
 }
 
 /**
+ * Breaks given `range` on a set of {@link module:engine/view/range~Range ranges}, that each are contained within a
+ * {@link module:engine/view/containerelement~ContainerElement container element}. After `range` is broken, it's "pieces" can
+ * be used by {@link module:engine/view/writer~writer} (which expects that passed ranges are contained within on container element).
```
pjasiun: > be used by {@link module:engine/view/writer~writer}

*writer.wrap?

> (which expects that passed ranges are contained within on container element).

*one?

scofalik: Well, it applies to all methods in writer, so first one is correct, but the other is not.

pjasiun: >  all methods in writer

Others then this one.

src/model/markerscollection.js
pjasiun: Still, range instance is a terrible ID. Also, it is not a point to introduce more features that we need at this stage. Then, we need to maintain them, what is usually more costly them introducing them. If we are unsure we will need this feature we should not bring it.

scofalik:
I fixed one more thing and merged with master. I also merged related branches in `List`, `Undo` and `Autoformat` features. I will prepare just one more followup and this PR, hopefuly, will be ready to merge.

scofalik:
When testing this with `--files=**/*` please make sure that you have all the repos updated! Since I merged this with master, this includes changes that could brake some other repos, if they are not updated. This includes `ckeditor5-core` and maybe others. If you have an error in code from non-engine repo, please update that repo first.

scofalik:
Two followups: #740 and #741

PR #709, opened 2016-11-30
Added objects handling to modifySelection method

### LINKED ISSUE #708: DataController#extendSelection should not move selection to the inside of an object
Extracted from #696 @Reinmar's comment:

General rule of thumb – selection should not be moved to the inside of an object. E.g.:

`<obj></obj><p>^x</p>` + backspace => the selection must not be moved to `<obj>^</obj>`

This means that extending the selection to the left in the case mentioned above should not put the focus inside the object. The focus must be placed before the object.

* `<obj></obj><p>^x</p>` + extend left -> `[<obj></obj><p>]x</p>`,
* `<obj></obj>[<obj></obj>]` + extend left -> `[<obj></obj><obj></obj>]`,

So in other words, the selection outputed from the `extendSelection` should not cross object boundaries.

szymonkups:
Fixes #708.

### DIFF
--- a/tests/controller/modifyselection.js
+++ b/tests/controller/modifyselection.js
@@ -4,18 +4,26 @@
  */
 
 import Document from 'ckeditor5/engine/model/document.js';
+import DataController from 'ckeditor5/engine/controller/datacontroller.js';
 import Selection from 'ckeditor5/engine/model/selection.js';
 import modifySelection from 'ckeditor5/engine/controller/modifyselection.js';
 import { setData, stringify } from 'ckeditor5/engine/dev-utils/model.js';
 
 describe( 'DataController', () => {
-	let document;
+	let document, dataController;
 
 	beforeEach( () => {
 		document = new Document();
+		dataController = new DataController( document );
+		document.schema.registerItem( 'obj' );
+		document.schema.allow( { name: 'obj', inside: '$root' } );
+		document.schema.objects.add( 'obj' );
+		document.schema.registerItem( 'inlineObj', '$inline' );
+		document.schema.objects.add( 'inlineObj' );
 		document.schema.registerItem( 'p', '$block' );
 		document.schema.registerItem( 'x', '$block' );
 		document.schema.registerItem( 'img', '$inline' );
+		document.schema.registerItem( 'b', '$inline' );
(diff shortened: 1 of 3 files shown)


tests/controller/modifyselection.js
```diff
@@ -4,18 +4,26 @@
...
+		dataController = new DataController( document );
+		document.schema.registerItem( 'obj' );
+		document.schema.allow( { name: 'obj', inside: '$root' } );
+		document.schema.objects.add( 'obj' );
+		document.schema.registerItem( 'inlineObj', '$inline' );
+		document.schema.objects.add( 'inlineObj' );
 		document.schema.registerItem( 'p', '$block' );
 		document.schema.registerItem( 'x', '$block' );
 		document.schema.registerItem( 'img', '$inline' );
+		document.schema.registerItem( 'b', '$inline' );
```
Reinmar: This is confusing. We don't have inline containers in the model. `<b>` should rather be an attribute. This test:

```html
<p>foo<b>[]</b><inlineObj>bar</inlineObj></p>
```

have no sense. In the paragraph you can either have text or inline objects.

tests/controller/modifyselection.js
```diff
@@ -375,12 +383,55 @@ describe( 'DataController', () => {
...
 
 				expect( stringify( document.getRoot(), document.selection ) ).to.equal( '<p>[\uD83D\uDCA9]</p>' );
 				expect( document.selection.isBackward ).to.true;
 			} );
 		} );
+
+		describe( 'objects handling', () => {
+			test(
+				'extends over next object element when at the end of an element',
+				'<p>foo[]</p><obj>bar</obj>',
```
Reinmar: I miss a test case for `<blockquote><p>foo[]</p></blockquote><obj></obj>`.

tests/controller/modifyselection.js
Reinmar: BTW. It's better to have many schema definitions – one per some `describe()` block. It's then easier to remember all setting for each block. E.g. you can move the additional setting to the newly added `describe()` (and keep the rest here for all tests).

tests/controller/modifyselection.js
szymonkups: This looks like a bigger case and it will be handled separately.

PR #714, opened 2016-12-07
AttributeOperation should use isEqual to check attribute value.

### LINKED ISSUE #713: AttributeOperation: attribute value incorrectly check
At the moment, `AttributeOperation` checks, whether the changed `ModelItem` has the same value of given attribute as stated by `AttributeOperation#oldValue`.

Unfortunately, the check is done using `===` operator so it does not work for non-primitive values, like arrays or objects. This in return breaks OT.

scofalik:
Fixes #713

### DIFF
--- a/src/model/operation/attributeoperation.js
+++ b/src/model/operation/attributeoperation.js
@@ -110,7 +110,7 @@ export default class AttributeOperation extends Operation {
 	_execute() {
 		// Validation.
 		for ( let item of this.range.getItems() ) {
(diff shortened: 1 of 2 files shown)


Reinmar:
Please add tests.

scofalik:
I will, I will. Also I haven't added one important change and Travis failed.

pjasiun:
> Please add tests.

I told same think. :D

PR #717, opened 2016-12-08
getNearestSelectionPosition changes

### LINKED ISSUE #712: Improvements for model document.getNearestSelectionPosition
Some improvements are needed in order to correctly support object elements:
* it should return `Range` as it might return correct position in text or selected object,
* it should support direction: `forward`, `backward` or `both` - it will be especially needed when dealing with arrow keys
* it should have more tests and tests should use use our test utilities.

szymonkups:
Fixes #712.

### DIFF
--- a/src/controller/insertcontent.js
+++ b/src/controller/insertcontent.js
@@ -52,18 +52,19 @@ export default function insertContent( dataController, content, selection, batch
 
 	const newRange = insertion.getSelectionRange();
 
-	// We are not testing else because it's a safe check for unpredictable edge cases: an insertion without proper range
-	// to select.
 	/* istanbul ignore else */
 	if ( newRange ) {
 		selection.setRanges( [ newRange ] );
 	} else {
+		// We are not testing else because it's a safe check for unpredictable edge cases:
+		// an insertion without proper range to select.
+
 		/**
-		 * Cannot determine proper range after insertion.
+		 * Cannot determine a proper selection range after insertion.
 		 *
-		 * @warning controller-insert-content-no-range
+		 * @warning insertcontent-no-range
 		 */
-		log.warn( 'controller-insert-content-no-range: Cannot determine proper range after insertion.' );
+		log.warn( 'insertcontent-no-range: Cannot determine a proper selection range after insertion.' );
 	}
 }
 
--- a/src/controller/insertcontent.js
+++ b/src/controller/insertcontent.js
@@ -52,9 +52,18 @@ export default function insertContent( dataController, content, selection, batch
 
 	const newRange = insertion.getSelectionRange();
 
+	// We are not testing else because it's a safe check for unpredictable edge cases: an insertion without proper range
+	// to select.
 	/* istanbul ignore else */
 	if ( newRange ) {
 		selection.setRanges( [ newRange ] );
+	} else {
+		/**
+		 * Cannot determine proper range after insertion.
+		 *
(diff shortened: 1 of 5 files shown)


src/controller/insertcontent.js
```diff
@@ -50,7 +50,12 @@ export default function insertContent( dataController, content, selection, batch
 		isLast: true
 	} );
 
-	selection.setRanges( insertion.getSelectionRanges() );
+	const newRange = insertion.getSelectionRange();
+
+	/* istanbul ignore else */
```
Reinmar: I'd like to have every case like this commented. We're not testing else because it's a safe check for some unpredictable edge cases. Also, having a `log.warning()` would be good there, like in the other cases in this method.

src/model/document.js
```diff
@@ -280,51 +280,55 @@ export default class Document {
...
+	 * {@link module:engine/model/schema~Schema schema}.
 	 *
-	 * @param {module:engine/model/position~Position} position Reference position where selection position should be looked for.
-	 * @returns {module:engine/model/position~Position|null} Nearest selection position.
+	 * Direction of searching for nearest correct selection range can be specified as:
+	 * * `both` - searching will be performed in both ways,
+	 * * `forward` - searching will be performed only forward,
+	 * * `backward` - searching will be performed only backward.
+	 *
+	 * When valid selection range cannot be found, `null` value is returned.
```
Reinmar: "`null` is returned"

src/model/document.js
```diff
@@ -280,51 +280,55 @@ export default class Document {
...
-			let step = backwardWalker.next();
+		if ( direction == 'both' || direction == 'forward' ) {
+			forwardWalker = new TreeWalker( { startPosition: position } );
+		}
 
-			if ( !step.done ) {
-				if ( this.schema.check( { name: '$text', inside: step.value.nextPosition } ) ) {
-					return step.value.nextPosition;
-				}
+		for ( let data of getWalkersData( backwardWalker, forwardWalker ) ) {
```
Reinmar: Walker has no data. Walker walks ;). Maybe `combineWalkers()`?

szymonkups:
Added requested fixes.

tests/model/document/document.js
```diff
@@ -276,49 +275,160 @@ describe( 'Document', () => {
...
+		test(
+			'should find range before when searching both ways when it is closer',
+			'<paragraph></paragraph><image></image>[]<image></image><image></image><paragraph></paragraph>',
+			'both',
+			'<paragraph>[]</paragraph><image></image><image></image><image></image><paragraph></paragraph>'
+		);
+
+		test(
+			'should return null if there is no valid range',
+			'<image></image>',
```
Reinmar: OK... Image is not an object in this test suite. This is confusing. Could we name it like... damn, I can't find a name, because this case shouldn't really occur :D. Perhaps we can call it `<emptyBlock>` then?

szymonkups: This case should be changed indeed. Initial selection should be added.

szymonkups:
Added requested changes.

Reinmar:
There's a warning logged by the tests:

```
WARN: 'insertcontent-no-range: Cannot determine a proper selection range after insertion.', undefined
```

See https://travis-ci.org/ckeditor/ckeditor5-engine/builds/183855169

If it's expected that it's logged, it should be stubbed. If not, then it means that either we don't need that istanbul comment or that something works incorrectly.

PR #718, opened 2016-12-08
Implemented DC.toModel

### LINKED ISSUE #688: Implement DataController.toModel()
There's already `toView()`. The `toModel()` method will be super simple:

```js
toModel( viewFragmentOrElement, additionalData ) {
    return this.viewToModel( viewFragmentOrElement, additionalData );
}
```

but will make for a nice API parity.

Reinmar:
Fixes #688.

### DIFF
--- a/tests/controller/datacontroller.js
+++ b/tests/controller/datacontroller.js
@@ -14,6 +14,8 @@ import ModelDocumentFragment from 'ckeditor5/engine/model/documentfragment.js';
 import ModelText from 'ckeditor5/engine/model/text.js';
 import ModelSelection from 'ckeditor5/engine/model/selection.js';
 
+import ViewDocumentFragment from 'ckeditor5/engine/view/documentfragment.js';
+
 import { getData, setData, stringify, parse } from 'ckeditor5/engine/dev-utils/model.js';
 
 import count from 'ckeditor5/utils/count.js';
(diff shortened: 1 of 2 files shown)


Reinmar:
Besides adding the new method I also added tests for the other ones, to confirm that they work with document fragments. We used doc frag convertion already in the clipboard, so the engine code was outdated.

tests/controller/datacontroller.js
```diff
@@ -293,26 +337,57 @@ describe( 'DataController', () => {
...
+		it( 'should convert a document fragment', () => {
+			const modelDocumentFragment = parseModel( '<paragraph>foo</paragraph><paragraph>bar</paragraph>', modelDocument.schema );
+
+			const viewDocumentFragment = data.toView( modelDocumentFragment );
+
+			expect( viewDocumentFragment ).to.be.instanceOf( ViewDocumentFragment );
+			expect( viewDocumentFragment ).to.have.property( 'childCount', 2 );
 
-			const viewElement = data.toView( modelElement ).getChild( 0 );
+			const viewElement = viewDocumentFragment.getChild( 0 );
```
scofalik: It would be more complete if you check both children.

Reinmar: Agree. But this is not a test for conversion itself, but for methods utilising conversion.

Reinmar: I mean – there are a lot more things which could be broken below, but it'd be unwise to try to test them so late. I'm checking the number of children and the first child to verify that enough elements were converted and that they make some sense.

PR #720, opened 2016-12-08
Add: Introduce SplitDelta x RemoveDelta special case transformations.

### LINKED ISSUE #719: Transformation: RemoveDelta x SplitDelta special case error
`RemoveDelta` x `SplitDelta` transformation is giving incorrect results for one edge case - when `RemoveOperation` in `RemoveDelta` removes a range where the last `Element` is split. Since `SplitDelta` is `InsertOperation` + `MoveOperation`, `InsertOperation` is inserting split `Element` after the range in mentioned `RemoveOperation`. This means that a part of element which was removed is still in a tree which leads to errors.

scofalik:
Fixes #719


pjasiun:
Please add follow-up about missing tests.

scofalik:
#721

Issue #721, opened 2016-12-08
Missing code coverage in deltas transformations.


scofalik:
This is a followup to #719 (merged in #720).

PR #724, opened 2016-12-12
Changed: Range#_getTransformedByMove better UX for transforming collapsed range.

### LINKED ISSUE #723: Collapsed ModelRange move transformation enhancement.
Assume this model state:
`<paragraph>foobar[]</paragraph>`
Where `[]` is a collapsed live range.

Let's say that the paragraph is broken after `foo`:
`<paragraph>foo</paragraph><paragraph>bar[]</paragraph>`
This is what we would expect.

Breaking a paragraph is two steps: making a new paragraph and performing such move operation:
`<paragraph>foo{bar}[]</paragraph><paragraph>^</paragraph>`
Where `{bar}` is moved at `^`.

As we can see, _yes there are two ways you can go now_. Moving `{bar}` should either move range `[]` with it or leave it. Both behaviors could be backed with arguments. In `LivePosition`, there is a `stickiness` flag, which would describe how position at `[]` should be moved.

Something has to be done with this. At this moment, `[]` is not sticking with moved `{bar}`. It's bad from UX point of view. If `[]` was selection, we end up with wrong result -- selection is set at the "break position":
`<paragraph>foo[]</paragraph><paragraph>bar</paragraph>`

For now, I propose just going with keeping `LiveRange` sticking to the moved range in such case. If this backfires somehow, we can make a proper flag somewhere and make it configurable.

scofalik:
Fixes #723.

### DIFF
--- a/src/model/range.js
+++ b/src/model/range.js
@@ -442,6 +442,12 @@ export default class Range {
 	 * @returns {Array.<module:engine/model/range~Range>} Result of the transformation.
 	 */
 	_getTransformedByMove( sourcePosition, targetPosition, howMany, spread, isSticky = false ) {
+		if ( this.isCollapsed ) {
+			const newPos = this.start._getTransformedByMove( sourcePosition, targetPosition, howMany, true, true );
+
+			return [ new Range( newPos ) ];
+		}
+
 		let result;
 
 		const moveRange = new Range( sourcePosition, sourcePosition.getShiftedBy( howMany ) );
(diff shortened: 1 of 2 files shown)


Reinmar:
Without https://github.com/ckeditor/ckeditor5-engine/issues/721 it's hard to understand whether the CC stayed the same. I'll merge this PR, but before you make a new one please fix the CC.

scofalik:
This PR concerns only one `if` block and as you can see in CC, it is covered.

Reinmar:
I know. But I need to keep complaining about missing CC :P

PR #733, opened 2016-12-19
Schema check algorithm should be more precise

### LINKED ISSUE #732: Schema checks don't try to match entire paths
```js
it( 'should return false if given element is allowed in root but not deeper', () => {
	schema.registerItem( 'paragraph', '$block' );

	expect( schema.check( { name: 'paragraph', inside: [ '$root', 'paragraph' ] } ) ).to.be.false;
} );
```

This test fails because the matching algorithm allow ignoring "inside" items which are not matching items from the paths where the element being checked is allowed. This is fine when the left side is ignored (e.g. so you can do `inside: [ 'paragraph' ]`, without specifying `$root`), but the right sides should always match.

Reinmar:
Fixes #732.

I'll be now fixing failing tests in other repos. In most cases (if not all) I should be able to commit straight to master, because fixed tests should also pass with the previous check.

Related PRs:

* [ ] https://github.com/ckeditor/ckeditor5-core/pull/53

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -505,15 +505,15 @@ export class SchemaItem {
 	 *
 	 * @protected
 	 * @param {String} type Paths' type. Possible values are `allow` or `disallow`.
-	 * @param {Array.<String>} checkPath Path to check.
+	 * @param {Array.<String>} pathToCheck Path to check.
 	 * @param {String} [attribute] If set, only paths registered for given attribute will be checked.
 	 * @returns {Boolean} `true` if item has any registered matching path, `false` otherwise.
 	 */
-	_hasMatchingPath( type, checkPath, attribute ) {
-		const allowedPaths = this._getPaths( type, attribute );
+	_hasMatchingPath( type, pathToCheck, attribute ) {
+		const registeredPaths = this._getPaths( type, attribute );
 
-		for ( const allowedPath of allowedPaths ) {
-			if ( matchPaths( this._schema, checkPath, allowedPath ) ) {
+		for ( const registeredPathPath of registeredPaths ) {
+			if ( matchPaths( this._schema, pathToCheck, registeredPathPath ) ) {
 				return true;
 			}
 		}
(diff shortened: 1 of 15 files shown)


src/model/schema.js
```diff
@@ -510,38 +510,10 @@ export class SchemaItem {
 	 * @returns {Boolean} `true` if item has any registered matching path, `false` otherwise.
 	 */
 	_hasMatchingPath( type, checkPath, attribute ) {
```
scofalik: To make things less confusing we could rename `checkPath` to `pathToCheck` or `inputPath` or something. Right now when reading code it was difficult to keep in mind which paths were which.

src/model/schema.js
```diff
@@ -581,3 +553,45 @@ export class SchemaItem {
...
+// checkPath: A, B, C
+// allowedPath: A, B, C, D
+// result: NOK
+//
+// Note – when matching paths, element extension chains (inheritance) are taken into consideration.
+//
+// @param {Schema} schema
+// @param {Array.<String>} checkPath
+// @param {Array.<String>} allowedPath
+function matchPaths( schema, checkPath, allowedPath ) {
```
scofalik: As above, consider using more clear names.

src/model/schema.js
```diff
@@ -510,38 +510,10 @@ export class SchemaItem {
...
-				// Since our paths have to match in given order, we always check against first item from item path.
-				// So, if item path is: B D E
-				// And checked path is: A B C D E
-				// It will be matching (A won't match, B will match, C won't match, D and E will match)
-				if ( chain.indexOf( itemPath[ i ] ) > -1 ) {
-					// Move pointer as we found element under index `i`.
-					i++;
-				}
-			}
+		const allowedPaths = this._getPaths( type, attribute );
```
Reinmar: Makes sense. I'll change this.

src/model/schema.js
Reinmar: I didn't want to touch too much code, but I can take another look on that.

tests/controller/datacontroller.js
```diff
@@ -339,7 +339,10 @@ describe( 'DataController', () => {
 	describe( 'stringify', () => {
 		beforeEach( () => {
 			modelDocument.schema.registerItem( 'paragraph', '$block' );
-			modelDocument.schema.registerItem( 'div', '$block' );
+			modelDocument.schema.registerItem( 'div' );
+
+			modelDocument.schema.allow( { name: '$block', inside: 'div' } );
+			modelDocument.schema.allow( { name: 'div', inside: '$root' } );
```
Reinmar: IIRC, that would mean that it's both a block in which text is allowed and it would need to allow  blocks inside it. So, a schema like this is more realistic. Div is a container which can contain blocks.

Reinmar: Besides, I needed to change this test because it didn't pass ofc.

src/model/schema.js
```diff
@@ -581,3 +553,45 @@ export class SchemaItem {
...
+//
+// @param {Schema} schema
+// @param {Array.<String>} checkPath
+// @param {Array.<String>} allowedPath
+function matchPaths( schema, checkPath, allowedPath ) {
+	// Start checking from the right end of both tables.
+	let allowedPathIndex = allowedPath.length - 1;
+	let checkPathIndex = checkPath.length - 1;
+
+	// And finish once reaching an end of the shorter table.
```
scofalik: Okay it took me a while to analyze this. At first sight it seemed wrong to me that any path can be shorter. For `checkPath` = `C D E` both `allowedPath` = `A B C D E` and `allowedPath` = `D E` are right. This rises a question whether `allowedPath` should be matched in whole? In first `allowedPath` it seems that somebody explicitly want to set that `E` is allowed in `A B C D`, that is that one of ancestors is `A`.

OTOH, this will not be a problem if in `checkPath` we will always pass a "full path", that is a path that has `$root` as left-most element. This would "solve" the "issue" described above. Consider again:
`checkPath` = `$root C D E`, `allowedPath` = `A B C D E` and `allowedPath` = `$root A B C D E`.

Now, check will fail for both `allowedPath`s (which allows us to use shorter and simpler form without `$root`).

And now back to the original problem. If we suppose that `checkPath` always has `$root` as first element, a check should fail for every `checkPath` shorter than `allowedPath`. That is because there will be a moment when we will check `$root` from `checkPath` against non-`$root` from `allowedPath`.

So there are a few solutions:
1. We leave code that is here as it is and stress in documentation that `checkPath` has to be full. Drawback: it's prone to generate bugs. OTOH, I expect that in most cases, inside will be specified as `model.Position`.
2. Automatically add `$root` as first element to `checkPath` if first element is not `$root`. Drawback: this may backfire if, for a reason, really want to check a path that starts from something else than `$root`.
3. Return `false` if `checkPath` is shorter than `allowedPath`. This feels natural, because it means that it is impossible to match every item from explicitly stated `allowedPath` that should require, that the last item in path is exactly in those parents. Drawbacks: IDK?

PR #769, opened 2017-01-18
Added useCapture member to DomEventObserver

### LINKED ISSUE #768: FocusObserver should use capturing phase for its events
To get focus events from nested editables `FocusObserver` should add its listener on the `capturing` phase. 

We should allow somehow to pass information about phase to the `observe()` method of `DomEventObserver`. 

Ideas how it could be implemented:
* add `useCapturing` property that will set capturing phase for all events of that observer,
* add `setCapturing( eventName, useCapture )` method to `DomEventObserver` that will allow to set a phase to each event separately,
* change `domEventType` to allow use objects, so we can configure observer's events like this:
```
this.domEventType = [ 'focus', 'blur' ]; 
```
or
```
this.domEventType = [ { 'focus': true }, { 'blur': true } ];
```

szymonkups:
Fixes #768.

### DIFF
--- a/src/view/observer/domeventobserver.js
+++ b/src/view/observer/domeventobserver.js
@@ -52,6 +52,17 @@ export default class DomEventObserver extends Observer {
 	 * @method #onDomEvent
 	 */
 
+	constructor( document ) {
+		super( document );
+		/**
+		 * If set to `true` DOM events will be listened on capturing phase.
+		 * Default value is `false`.
+		 *
+		 * @member {Boolean} #useCapture
+		 */
+		this.useCapture = false;
+	}
+
 	/**
 	 * @inheritDoc
 	 */
@@ -63,7 +74,7 @@ export default class DomEventObserver extends Observer {
 				if ( this.isEnabled ) {
 					this.onDomEvent( domEvent );
 				}
-			} );
+			}, { useCapture: this.useCapture } );
 		} );
 	}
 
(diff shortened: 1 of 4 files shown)


src/view/observer/domeventobserver.js
```diff
@@ -52,6 +52,17 @@ export default class DomEventObserver extends Observer {
 	 * @method #onDomEvent
 	 */
 
+	constructor( document ) {
+		super( document );
+		/**
```
Reinmar: Missing empty line before the comment.

src/view/observer/domeventobserver.js
```diff
@@ -52,6 +52,17 @@ export default class DomEventObserver extends Observer {
...
+		 * If set to `true` DOM events will be listened on capturing phase.
+		 * Default value is `false`.
+		 *
+		 * @member {Boolean} #useCapture
```
Reinmar: You don't need to specify the name AFAIR.

src/view/observer/domeventobserver.js
```diff
@@ -52,6 +52,17 @@ export default class DomEventObserver extends Observer {
 	 * @method #onDomEvent
 	 */
 
+	constructor( document ) {
+		super( document );
+		/**
+		 * If set to `true` DOM events will be listened on capturing phase.
```
Reinmar: "on the capturing phase."

PR #771, opened 2017-01-19
EditableElement.isFocus is set before actual rendering

### LINKED ISSUE #770: EditableElement can change isFocused before rendering
`EditableElement` is changing its `isFocus` property after actual rendering. 
Rendering should not change anything in view's selection and focus state of the view document, so `EditableElement` can change `isFocus` before rendering.
This will be convenient for setting some CSS classes depending on `EditableElement.isFocus` - we will be able to set the the CSS class before rendering.

szymonkups:
Fixes #770.

### DIFF
--- a/src/view/editableelement.js
+++ b/src/view/editableelement.js
@@ -80,11 +80,11 @@ export default class EditableElement extends ContainerElement {
 			( isFocused ) => isFocused && document.selection.editableElement == this
 		);
 
-		// Update focus state after each rendering. Selection might be moved to different editable before rendering,
-		// but this does not mean that editable has focus - it will be placed there after rendering.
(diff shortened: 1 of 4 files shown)


Reinmar:
Something's wrong with http://localhost:8125/ckeditor5-engine/tests/view/manual/focusobserver.html

The background doesn't update after 2nd+ clicks.

This isn't a regression caused by this PR, but I have a feeling that this is related to what we're trying to fix now. Please check this test. Maybe it's something wrong with the test itself, but, in both cases, we need to figure this out.

szymonkups:
I've fixed the manual test: now document is rendered each time selection is changed. Before actual rendering `isFocus` property of editables is changed and styles are updated.

Issue #761, opened 2017-01-10
Markers add/set should accept range


pjasiun:
I should not be forced to transform range to the live range whenever I want to create a marker.
I should not need to care about destroying live range when I'm using markers.
I should be able to create a marker based on the selection, these APIs should be compatible: `doc.markers.add( name, doc.selection.getFirstRange() );`

pjasiun:
I fact, I think that people using markers may have no idea what `LiveRange` is.

pjasiun:
And even if the marker is created based on the `LiveRange` it should clone it and handle internally. Otherwise, there might be problems with destroying.

scofalik:
**tl;dr:** I agree that `Range` should be transformed to `LiveRange`. I don't agree about cloning `LiveRange`s though. `LiveRange`s that are created by `MarkerCollection` (upon adding them to it) should be then destroyed by `MarkerCollection` when the collection is destroyed. Explanation below.

-----

The idea to force `LiveRange` seemed good at first, for two reasons:

1. If all ranges passed to `MarkerCollection` are `LiveRange`s, `MarkerCollection` does not need to care about destroying ranges -- code that created the `LiveRange` should `detach` it.
2. It raises awareness. `LiveRange` is a costly resource. `LiveRange` creates event listener that is fired every time there is an action in the editor: selection gets moved, letter is typed, etc. So in this case, I think it's better that users exactly know what happens.

However, when I was using `MarkersCollection`, I also came to conclusion that it is cumbersome to import `LiveRange` every time and use `LiveRange.createFromRange( myRange )`. So we need some kind of solution.

1. I am fine with automatic `Range` to `LiveRange` conversion.
2. I am not fine with cloning `LiveRange` when it's passed `#add` method. This is for efficiency reasons. It creates twice as many `LiveRange`s. It is also annoying when you debug stuff in `LiveRange` transformation, because when you set a breakpoint inside transformation function you get there multiple times. Currently, when I debugged some stuff there I already got annoyed because `LiveRange`s are also in `LiveSelection`. It's hard to say "which" `LiveRange` you are transforming at the moment, etc. Long story short: it's better to have minimal amount of `LiveRange`s.

pjasiun:
The problem is destroying these live ranges. If `MarkerCollection` reuses the live range, you can not remove it as soon as you do not need it, because `MarkerCollection` may still need it. It means that it's hard to guess when you can destroy it and you won't destroy it, just in case, to avoid bugs.

[...]

Issue #751, opened 2017-01-02
MarkersCollection returns inconsistent data type


oskarwrobel:
`MarkersCollection` fires `add` event which provides instance of `Range` created from marker `LiveRange`, but `MarkersCollection#get()` returns `LiveRange` instance. AFAIK every collection in CKE5 returns consistent data for `add` event and `get()` method.

There is an idea to create `Marker` class to represent single Marker instead of pure `LiveRange` / `Range`. This should solve a problem with decision what `MarkersCollection` should return.

pjasiun:
I agree that the instance of the `Marker` would be nice. You will be able to get the name of the maker from it. We may also introduce some functions which take a marker as a parameter. From the logical POV marker is something more then the range.

scofalik:
Yes, it's done this way to not create a new instance of `LiveRange` which would be super annoying as you'd have to destroy such instance each time you're done working with it.

Having a `Marker` class would solve the issue. I've been thinking about it but, I had two problems with it:
* that would be just an object with no methods and only two properties,
* users would have to import a new class and create new instance just to "mark" a range as "marker".

OTOH, the next requirement for markers is to provide a way to state that some markers are "local" and some should be updated between clients. This could also be solved by "marker wrapper object".

So, we either could make `Marker` class, which would have two (three) properties and no methods other than constructor. Or we may do it internally in `MarkerCollection`. `MarkerCollection#add` could take three parameters (`name`, `range`, `type`) but then `add` `event` will be fired with plain object (containing `LiveRange` instance). Also `get` method would return an object with `LiveRange` instance.

WDYT?

scofalik:
`Marker` class will be introduced, as a simple API over `LiveRange`. `Marker` instances will be created when a `Range` is added to `MarkerCollection`. `Marker` instances will be returned by `MarkerCollection#get` method. `Marker` instances will be used in events fired by `MarkerCollection`.

PR #773, opened 2017-01-20
Changed: Markers refactor

scofalik:
* introduced `MarkerOperaton` and `MarkerDelta` with OT #765
* renamed `MarkersCollection` to `MarkerCollection` #749
* changed `MarkerCollection#add` to `#set`, removed `#update`, added `#has` #760
* changed `MarkerCollection` to accept `Range` and create and destroy own `LiveRange`s #761
* introduced `Marker` class and used it in `MarkerCollection` #751
* changed MarkerCollection events to pass Marker instances #751


src/controller/editingcontroller.js
```diff
@@ -105,12 +105,12 @@ export default class EditingController {
 		}, { priority: 'low' } );
 
 		// Convert model markers changes.
-		this._listener.listenTo( this.model.markers, 'add', ( evt, name, range ) => {
-			this.modelToView.convertMarker( 'addMarker', name, range );
+		this._listener.listenTo( this.model.markers, 'add', ( evt, marker ) => {
+			this.modelToView.convertMarker( 'addMarker', marker.name, marker.getRange() );
```
pjasiun: Why not `Marker` instance?

src/controller/editingcontroller.js
```diff
@@ -105,12 +105,12 @@ export default class EditingController {
...
 		} );
 
-		this._listener.listenTo( this.model.markers, 'remove', ( evt, name, range ) => {
-			this.modelToView.convertMarker( 'removeMarker', name, range );
+		this._listener.listenTo( this.model.markers, 'remove', ( evt, marker ) => {
+			this.modelToView.convertMarker( 'removeMarker', marker.name, marker.getRange() );
```
pjasiun: As above.

src/model/delta/markerdelta.js
```diff
@@ -0,0 +1,118 @@
...
+ * This is useful if the marker is created before it can be added to document history (e.g. a feature creating the marker
+ * is waiting for additional data, etc.). In this case, the marker may be first created directly through
+ * {@link module:engine/model/markercollection~MarkerCollection MarkerCollection API} and only later added using `Batch` API.
+ *
+ * @chainable
+ * @method module:engine/model/batch~Batch#setMarker
+ * @param {module:engine/model/markercollection~Marker|String} markerOrName Marker to update or marker name to add or update.
+ * @param {module:engine/model/range~Range} [range] Marker range.
+ */
+register( 'setMarker', function( markerOrName, range ) {
```
pjasiun: Why not arrow function?

src/controller/editingcontroller.js
scofalik: Because this function is not always used exactly with marker. This is more like saying "perform conversion of given part of data like it is in marker XXX". Take a look at `moveInOutOfMarker` and `insertMarkerOrIntoMarker` in `model-to-view-converters.js`.

src/controller/editingcontroller.js
scofalik: As above.

src/model/delta/markerdelta.js
scofalik: 🤷‍♂️

src/model/markercollection.js
```diff
@@ -0,0 +1,287 @@
...
+	 * set is different, the marker in collection is removed and then new marker is added. If the range was same, nothing
+	 * happens and `false` is returned.
+	 *
+	 * @fires {module:engine/model/markercollection~MarkerCollection#event:add}
+	 * @fires {module:engine/model/markercollection~MarkerCollection#event:remove}
+	 * @param {String|~Marker} markerOrName Name of marker to add or Marker instance to update.
+	 * @param {module:engine/model/range~Range} range Marker range.
+	 * @returns {~Marker} `Marker` instance added to the collection.
+	 */
+	set( markerOrName, range ) {
```
pjasiun: If I pass Marker as a first parameter only the name is taken from it or the range as well? If only the name I would keep it simple and accept only a name then it would be clear.

scofalik: Only name, why would it take `range`? You need to pass `range` to update the marker. Of course we can keep it simple because there's not much more to for developer if we restrict this only to marker name. It's just `.set( marker.name, range )` instead `.set( marker, range )`.

src/model/markercollection.js
```diff
@@ -0,0 +1,287 @@
...
+	 * @param {~Marker} marker The removed marker.
+	 */
+}
+
+mix( MarkerCollection, EmitterMixin );
+
+/**
+ * `Marker` is a continuous parts of model (like a range), is named and represent some kind of information about marked
+ * part of model document. In contrary to {@link module:engine/model/node~Node nodes}, which are bits of data, markers are
+ * kind of bookmarks, marking part of a model document between marker start and marker end.
```
pjasiun: If you are a CKE4 developer reading this doc, you can think that you should use markers for cases you used bookmarks in CKE4, what is wrong. You should use live ranges for these cases.

src/model/markercollection.js
```diff
@@ -0,0 +1,287 @@
...
+ * Markers can be converted to view by adding appropriate converters for
+ * {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher#event:addMarker addMarker} and
+ * {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher#event:removeMarker removeMarker}
+ * events, or by building converters for {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher}
+ * using {@link module:engine/conversion/buildmodelconverter~buildModelConverter model converter builder}.
+ *
+ * Markers are similar to adding and converting attributes on nodes. The difference is that attribute is connected with
+ * a given node (e.g. a character is bold no matter if it gets moved or content around it changes). Markers on the
+ * other hand are continuous ranges (e.g. a character inside of marker doesn't have any special property itself, when it
+ * is moved outside of marker it stops being "special" and the marker is shrunk).
```
pjasiun: Also, to find attribute you need to scan the whole document, while with a marker you always know which range is marked.

src/model/markercollection.js
```diff
@@ -0,0 +1,287 @@
...
+ *
+ * Another upside of markers is that finding marked part of document is fast and easy. Using attributes to mark some nodes
+ * and then trying to find that part of document would require traversing whole document tree. Marker gives instant access
+ * to the {@link ~Marker#range range} which it is marking at the moment.
+ *
+ * `Marker` instances are created and destroyed only by {@link ~MarkerCollection MarkerCollection}.
+ */
```
pjasiun: This description is really great!

scofalik: Not sure if trolling...

src/model/operation/markeroperation.js
```diff
@@ -0,0 +1,152 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	_execute() {
+		// If old range and new range are "same", operation should not do anything.
+		if ( this.oldRange === null && this.newRange === null ) {
+			return;
+		} else if ( this.oldRange !== null && this.newRange !== null && this.oldRange.isEqual( this.newRange ) ) {
+			return;
+		}
```
pjasiun: As we found out, now if you add existing marker to the document the change event won't fire, what is wrong.

src/controller/editingcontroller.js
pjasiun: I see. That's fine then.

src/controller/editingcontroller.js
pjasiun: As above.

src/model/delta/markerdelta.js
scofalik: Okay I know why. When arrow function is used, the function has wrong context.

scofalik: `function` is used in other deltas too.

scofalik:
Fixed

src/model/operation/markeroperation.js
```diff
@@ -94,19 +94,17 @@ export default class MarkerOperation extends Operation {
...
-		}
+		// Make a change in Document#markers only when something actually changes, that is
+		// `this.oldRange` and `this.newRange` are different.
+		const changeInMarkersCollection =
+			!( this.oldRange === null && this.newRange === null ) &&
+			!( this.oldRange !== null && this.newRange !== null && this.oldRange.isEqual( this.newRange ) );
 
-		const document = ( this.oldRange || this.newRange ).root.document;
-		let type;
+		const document = changeInMarkersCollection && ( this.oldRange || this.newRange ).root.document;
```
pjasiun: Can't you move this like to `changeInMarkersCollection` and simplified the code?

PR #774, opened 2017-01-20
T/762 Fire marker conversion if marker is reinserted from graveyard

scofalik:
Fixes #762 

Additional change: do not fire marker conversion if marker is in graveyard or not in model root.

Bases on PR: https://github.com/ckeditor/ckeditor5-engine/pull/773

PR #775, opened 2017-01-24
T/702 Selection breaks ViewAttributeElement created by marker conversion

### LINKED ISSUE #702: Markers: Selection breaks markers ViewAttributeElement.
The issue is simple. Consider some letters in a paragraph and a marker on them. They could be converter to something like this in view:
`<p>ab<span class="marker">cde</span>fg</p>`.
Now, selection is placed after `d`:
`<p>ab<span class="marker">cd</span><fillingChar>|<span class="marker">e</span>fg</p>`.
This happens because selection does not have marker's attributes.

Also weird thing happens next to space:
```
<p>Lorem ipsu<span class="h-yellow">m</span> <fillingChar>|<span class="h-yellow"> dolor s</span>it amet.</p>
```
This makes the space disappear. Somehow, this bug does not happen in following scenario
```
<p><strong>abc def ghi</strong></p>
```
When I place cursor after `c` or before `d` and "unbold" the space is correctly moved to left or right `em`.

I see two solutions:
1. Mark (pun not indented) view elements created by markers so they won't be broken by selection converter.
2. Use `ViewContainerElement` instead of `ViewAttributeElement` but this may have unexpected and incorrect impact on `view.Writer` and conversion. I'd advise against this.
3. Add attributes on selection (somehow).

scofalik:
Fixes #702 .
Bases on PR #774 .

Fixes also:
#703

### DIFF
--- a/src/conversion/model-selection-to-view-converters.js
+++ b/src/conversion/model-selection-to-view-converters.js
@@ -145,27 +145,54 @@ export function convertCollapsedSelection() {
  */
 export function convertSelectionAttribute( elementCreator ) {
 	return ( evt, data, consumable, conversionApi ) => {
-		const selection = data.selection;
-
-		if ( !selection.isCollapsed ) {
-			return;
-		}
-
-		if ( !consumable.consume( selection, 'selectionAttribute:' + data.key ) ) {
-			return;
-		}
-
-		let viewPosition = conversionApi.viewSelection.getFirstPosition();
-		conversionApi.viewSelection.removeAllRanges();
-
 		const viewElement = elementCreator instanceof ViewElement ?
-				elementCreator.clone( true ) :
-				elementCreator( data.value, data, selection, consumable, conversionApi );
+			elementCreator.clone( true ) :
+			elementCreator( data.value, data, data.selection, consumable, conversionApi );
 
-		viewPosition = viewWriter.wrapPosition( viewPosition, viewElement );
-
-		conversionApi.viewSelection.addRange( new ViewRange( viewPosition, viewPosition ) );
+		convertSelectionPosition( data.selection, conversionApi.viewSelection, viewElement, consumable, 'selectionAttribute:' + data.key );
 	};
+}
+
+/**
+ * Performs similar conversion as {@link ~convertSelectionAttribute}, but depends on a marker name of a marker in which
+ * collapsed selection is placed.
+ *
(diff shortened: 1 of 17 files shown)


pjasiun:
I don't like this set on the selection with all markers selection is in. It works only on collapsed selection and needs to be always synchronized. Also, as @scofalik note: it does not fit well for Non-Live (Dead?) Selection. In fact, since it's only used by converters which do not listen on any changes there but get markers on demand, it could be replaced with the method in `MarkersCollection` which returns all markers a position is in (such method might be useful in other cases). Then converter may use this method.

src/conversion/model-selection-to-view-converters.js
```diff
@@ -145,27 +145,54 @@ export function convertCollapsedSelection() {
...
-		conversionApi.viewSelection.removeAllRanges();
+		convertSelectionPosition( data.selection, conversionApi.viewSelection, viewElement, consumable, 'selectionMarker:' + data.name );
+	};
+}
 
-		const viewElement = elementCreator instanceof ViewElement ?
-				elementCreator.clone( true ) :
-				elementCreator( data.value, data, selection, consumable, conversionApi );
+// Helper function for `convertSelectionAttribute` and `convertSelectionMarker`, which perform similar task.
+function convertSelectionPosition( modelSelection, viewSelection, viewElement, consumable, consumableName ) {
```
pjasiun: Does not method convert "full" selection (with attribute and markers)? If not then the name is misleading. Maybe `wrapSelectionPosition`?

src/view/writer.js
```diff
@@ -887,6 +887,8 @@ function wrapChildren( parent, startOffset, endOffset, attribute ) {
 	let offsetChange = 0;
 
 	for ( let position of wrapPositions ) {
+		position.offset -= offsetChange;
```
pjasiun: Qu'est-ce?

scofalik: It is explained by commit message.

This got removed, probably due to mistake and was causing a bug. This exact line of code was there long time ago in repo history.

pjasiun: Please add comment then.

scofalik: But I don't know why this is like this. I just recovered a line of code that was mistakenly removed and added a test. What am I going to comment? `// Please do not remove it by mistake again.` ?

PR #777, opened 2017-01-24
T/754 MarkerCollection#getMarkersGroup

### LINKED ISSUE #754: Getting group of markers
When you add markers with name `foo:1`, `foo:2`, `foo:3` and `bar:1`, `bar:2`, `bar:3`, some way of getting group of markers (`bar` only) will be useful.

```
for ( const marker of markers.getStartsWith( 'bar' ) ) {
   ...
}
```

scofalik:
Fixes #754 

Bases on PR #775

### DIFF
--- a/src/model/markercollection.js
+++ b/src/model/markercollection.js
@@ -143,7 +143,7 @@ export default class MarkerCollection {
 	 */
 	*getMarkersGroup( prefix ) {
 		for ( let marker of this._markers.values() ) {
-			if ( marker.name.indexOf( prefix + ':' ) === 0 ) {
+			if ( marker.name.startsWith( prefix + ':' ) ) {
 				yield marker;
 			}
 		}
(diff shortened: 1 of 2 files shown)


src/model/markercollection.js
```diff
@@ -129,6 +129,27 @@ export default class MarkerCollection {
...
+	 *		const markerFooBarA = markersCollection.set( 'foobar:a', rangeFooBarA );
+	 *		Array.from( markersCollection.getMarkersGroup( 'foo' ) ); // [ markerFooA, markerFooB ]
+	 *		Array.from( markersCollection.getMarkersGroup( 'a' ) ); // []
+	 *
+	 * @param prefix
+	 * @returns {Iterator.<~Marker>}
+	 */
+	*getMarkersGroup( prefix ) {
+		for ( let marker of this._markers.values() ) {
+			if ( marker.name.indexOf( prefix + ':' ) === 0 ) {
```
pjasiun: Did you consider using startsWith? https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

PR #782, opened 2017-01-26
FocusObserver renders document after blur event

### LINKED ISSUE #781: Focus observer should re-render view document on blur
Currently, there is no rendering after editor is blurred. Some elements inside editing view may need to be re-rendered after editor is blurred. For example: nested editables that change CSS class when are focused and blurred.

szymonkups:
Fixes #781


szymonkups:
R- because I found that it breaks image alternate text balloon panel showing.

szymonkups:
Please review it after https://github.com/ckeditor/ckeditor5-image/pull/41.

Issue #679, opened 2016-11-14
Simplify batches/deltas and change event


pjasiun:
I do not believe anymore that users will create their own Deltas. It will not gonna happen. So we can simplify this piece of the API:
 - move the code from the `register` to the delta constructor,
 - introduce `ModelDocument.applyDelta` https://github.com/ckeditor/ckeditor5-engine/issues/500 (@scofalik will be happy).

I'm not sure about two thinks:
 - how relations between `Batch`, `Delta` and `applyDelta` should look like: do I need to pass `Batch` to the `Delta` constructor? Do I need to add `Delta` to the `Batch`?  In fact, I need to add delta to the batch and apply it at the same time so, maybe I should pass both `Delta` and `Batch` to the `applyDelta` and it will add delta to the batch? Or adding `Delta` to the `Batch` should apply it since `Batch` knows about the document?
 - if there will be a reach constructor in each  `Delta` class won't it be a problem for the OT where we need to create a delta in a custom way? The simplest solution would be to keep the constructor empty and move the `register` logic to the `init` method, which can not be called in the OT, but then we make the API more complicated again.

Reinmar:
https://github.com/ckeditor/ckeditor5-engine/issues/678 describes the bigger picture of this change.

Reinmar:
Quick thought, without much of an analysis: Each delta module should expose a function called like today's `Batch` methods. So instead of:

```js
model.batch().insert( what, where );
```

you'd do:

```js
const batch = model.createBatch();
insert( batch, what, where );
```

That `insert()` would do all the steps:

1. create delta,
2. add it to the batch,
3. apply the delta.

We have hundreds of `batch.xxx()` usage and this made me think that we must not complicate this too much. So the 3 steps above shouldn't be required to be done manually.

pjasiun:
Reviewing https://github.com/ckeditor/ckeditor5-engine/pull/793 I realized that we should also update the change event. Now it's very messy. It could be fired not per operation, but per delta and contains operations as internal. Fortunately, because big part of the code is listening on the dispatcher, not on the change event, it might be not that hard to change as it looks like.

Reinmar:
That's a part of a whole topic – https://github.com/ckeditor/ckeditor5-engine/issues/678.

pjasiun:
[...]

PR #793, opened 2017-01-30
T/772 Markers and LiveRange transformation incompatibility

### LINKED ISSUE #772: Markers: LiveRange and Markers transformation incompatibility.
Long story short: It's impossible to transform `LiveRange` in the same way as `MarkerOperation` is transformed. It leads to a problem where actual `LiveRange` after X operations done to document may have different parameters than a range in `MarkerOperation` that was transformed by those X operations.

This is not `MarkerOperation` fault. We wouldn't be able to do it anyway. It's a long story, so brace yourselves. Let's start from the beginning. A long time ago, in an editor, far, far away...

`LiveRange` updates itself according to the document changes. We wanted that functionality that when it is split (mostly due to pressing <kbd>Enter</kbd> inside it), we want the selection to be maintained:

```
<p>Lo[rem ip^su]m.</p>
```
```
<p>Lo[rem ip</p>
<p>su]m.</p>
```

Normally, when a part of `LiveRange` is moved, the range is shrunk. Enter `Position#isTouching`. `Position#isTouching` checks whether two position are "next to each other", that is whether there are no nodes between them. We use it to check whether the end of "old" part of `LiveRange` is "next to" the start of "moved part" of the `LiveRange`:

```
<p>Lo[rem ip]<p>
<p>{su}m.</p>
```

`]` and `{` are "touching". These are different positions but "close enough" that the range should not be broken. This is how we solve the issue with selection.

The problem is that `Position#isTouching` operates on model document. This means that we can't use it when we have a set of operations and we have to transform a range by them, because during transformations we operate on out-dated model document state. It works for `LiveRange` because `LiveRange` updates itself after each change and always have up-to-date model document state.

How to solve it?

First step is by introducing `MarkerDelta` x `SplitDelta` transformation. We can check if split happens inside marker range and fix the transformation result. But this still leaves us with (at least) two problems:

1. Any non-split operation that would move a part of `LiveRange` to a "touching" position, would cause a problem, for example:
```
<p>Foo bar.^</p>
<p>[{Foo} bar].</p>
```
Moving `{}` to `^` will keep `LiveRange` between `<p>`s, and it is impossible to fix during transformation, because we can't use `Position#isTouching`.

2. It still works bad for `Range#getTransformedByDeltas`. Even if we fix something in OT, we should fix it also here. Now this method is pretty simple. It takes all deltas, takes their operations and update the range operation-by-operation. We would need some hacks that "if delta is `SplitDelta` do something differently". This smells a bit.

So, what can we do?

1. Get rid of "keep LiveRange in whole" feature, at least for now. I really would not like to do this, though. It looks bad that if something splits your selection you end up only with the first part. It would work same for the markers too.

2. Enhance `Position`s. Right now `Position` have `#path` which is an array of offsets. If `Position` had also an array of max offsets on each level, we would be able to execute `#isTouching` because it wouldn't use document model. Right now position "reads" like: "in root X, get child 2, get child 3, get child 1, offset 2". After the change it would be: "in root X, get child 2 (of total 2), get child 3 (of total 5), get child 1 (of total 2), offset 2 (max 3)". The problem is that we will have to update this whenever we update `Position` - that is during any transformation. I suspect this may be a lot of work.

3. Do not use `#isTouching` in `LiveRange` updating algorithm. Instead, make there a special case for `SplitDelta`. If `LiveRange` sees that is changed because of `SplitDelta`, check if the split is inside the range and update the range using special case. We could keep `MarkerDelta` x `SplitDelta` transformation special case then and just update `Range#getTransformedByDeltas`. Since `LiveRange` transformation uses `Range#getTransformedByDeltas` anyway we might need to update only one piece of code. Unfortunately, as I said, it smells and will make the code look bad.

So, third solution is a middle ground. It's some work to do, code will be bad, but we will keep important feature. Second solution would be perfect, but I don't know if it is doable and how much time it will take.

Leaving things as they are is also some kind of short term solution, but this has to be fixed at some point.

scofalik:
Fixes #772 

Some additional information and explanation was given in the issue.

### DIFF
--- a/src/model/range.js
+++ b/src/model/range.js
@@ -639,8 +639,8 @@ export default class Range {
 	 * Combines all ranges from the passed array into a one range. At least one range has to be passed.
 	 * Passed ranges must not have common parts.
 	 *
-	 * The first range from the array is a reference range. If other ranges
-	 * {@link module:engine/model/position~Position#isTouching are touching} the reference range, they will get combined into one range.
+	 * The first range from the array is a reference range. If other ranges starts or ends on the exactly same position where
+	 * the reference range, they get combined into one range.
 	 *
 	 *		[  ][]  [    ][ ][  ref range  ][ ][]  [  ]  // Passed ranges, shown sorted. "Ref range" was the first range in original array.
 	 *		        [      returned range       ]  [  ]  // The combined range.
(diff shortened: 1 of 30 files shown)


pjasiun:
After this change, the change event looks very bad. I hope that after the API refactoring (https://github.com/ckeditor/ckeditor5-engine/issues/679) we will be able to improve it.

src/model/range.js
```diff
@@ -454,13 +443,21 @@ export default class Range {
 	 * @param {module:engine/model/position~Position} targetPosition Position before the first changed node.
 	 * @param {Number} howMany How many nodes has been changed.
 	 * @param {module:engine/model/position~Position} sourcePosition Source position of changes.
+	 * @param {module:engine/model/delta/delta~Delta} delta Delta that introduced the change.
```
pjasiun: Docs do not match to the code.

pjasiun:
Please fix docs and you can merge changes. :)

PR #798, opened 2017-01-31
Introduced UIElement and vie.writer.clear

### LINKED ISSUE #788: Introduce UIElement and view.writer.clear method
My first thought was to introduce an `UIElement`: similar to the `AttributeElement`, but which does not need to have content to exists. It would be an element to represent something that does not have a models representation, can not have a selection inside, like drag handler or collapsed selection of other users.

Then I realized that the only difference between `UIElement` and `EmptyElement` it that one have a length in the model 0 and the other 1. But it's mapper job to handle it, and mapper does not distinguish element types, it's based only on mapping.

This is why we may try to use `EmptyElement` for the cases we wanted to use `UIElement`. For sure we should try because it would be far simpler.

What we need, in such case, is the method to remove all empty element from the given range. It could be called `clear`, first, but if the concept will work we would have to work on names (EmptyElement does not sound good too). Also, the question is: should such method handle `Container` and `Attribute` elements as well. I think we should start with something simple and if it will work we can improve it later.

So the first approach is to create the `view.writer.clear` method which will take the range and the element and removes all `EmptyElement`s similar to the given element from the given range.

### LINKED ISSUE #803: DocumentFragment#getAncestors( { includeNode: true } ) returns empty array
`DocumentFragment#getAncestors` always returns empty array:
```
	getAncestors() {
		return [];
	}
```
but should return array with itself when `includeNode: true` option is set.

It causes: https://github.com/ckeditor/ckeditor5-engine/issues/802

### LINKED ISSUE #805: view.Position may end up in view.DocumentFragment
`view.Position` may end up in `view.DocumentFragment`. This is a fact. It is partially supported but we have to fix:

1. Docs - at the moment docs say that `view.Position#parent` is a `view.Node`.
2. Some things may fail and we need to check if given tests exist:
    - creating `view.Position` in `view.DocumentFragment`, getting node after it and node before it,
    - methods in `Position` that use `#parent` ( `#getAncestors()`, `#compareWith()` ),
    - iterating using `view.TreeWalker`, over `view.DocumentFragment`.

oskarwrobel:
Introduced `UIElement` for representing UI related stuff in the View. For now, it works exactly the same as `EmptyElement` but have a different semantic role. UIElement is not a content.

Besides new `view.writer.clear` method is added. This method removes every matching element from given range.

Also this PR removes `view.DocumentFragment#getAncestors()` method and fixes places where it was used (mainly in `view.Position`).

---

### Commit Message:

Feature: Introduced `view.UIElement` and `view.writer.clear` method. Closes #788.

BREAKING CHANGES: Removed `view.DocumentFragment#getAncestors`. Closes #803. Closes #805.
BREAKING CHANGES: `Position.getAncestors()` should return elements in the same order as `Node.getAncestors()`.


src/dev-utils/view.js
```diff
@@ -453,7 +457,7 @@ class RangeParser {
 				// Non-empty text node.
 				if ( text ) {
 					if (
-						( this.sameSelectionCharacters && !item.outer ) ||
+						this.sameSelectionCharacters ||
```
oskarwrobel: This condition was not needed and was not covered by tests. I've run tests for entire CKE5 after removing it and everything was green.

src/view/writer.js
```diff
@@ -399,6 +412,68 @@ export function remove( range ) {
...
+			rangeToRemove = Range.createOn( item );
+		// When range starts inside Text or TextProxy element.
+		} else if ( !current.nextPosition.isAfter( range.start ) && ( item instanceof Text || item instanceof TextProxy ) ) {
+			// We need to check if parent of this text matches to given element.
+			const parentElement = item.getAncestors().find( ( ancestor ) => {
+				return ancestor instanceof Element && element.isSimilar( ancestor );
+			} );
+
+			// If it is then create range inside this element.
+			if ( parentElement && parentElement ) {
```
scofalik: Funny condition.

src/view/writer.js
```diff
@@ -399,6 +412,68 @@ export function remove( range ) {
...
+				rangeToRemove = Range.createIn( parentElement );
+			}
+		}
+
+		// If we have found element to remove.
+		if ( rangeToRemove ) {
+			// We need to check if element range stick out of the given range and truncate if it is.
+			if ( !range.containsRange( rangeToRemove ) ) {
```
scofalik: Do we need this condition given below conditions?

scofalik: I mean, if `range` does contain `rangeToRemove`, neither of below conditions will be met and it is fine. You have no CPU gain because you have to perform `containsRange` check anyway, which also checks `isAfter` and `isBefore`. Additionally, this `if` makes the code less clean and understandable.

If you don't have any counterexample, check if any tests fail after removing this `if` and if not, remove it.

src/view/writer.js
```diff
@@ -399,6 +412,68 @@ export function remove( range ) {
...
+				if ( rangeToRemove.end.isAfter( range.end ) ) {
+					rangeToRemove.end = range.end;
+				}
+
+				if ( rangeToRemove.start.isBefore( range.start ) ) {
+					rangeToRemove.start = range.start;
+				}
+			}
+
+			// At the end we remove range with founded element.
```
scofalik: found

tests/view/writer/clear.js
```diff
@@ -0,0 +1,159 @@
...
+			const elementToRemove = new AttributeElement( 'b' );
+
+			test(
+				elementToRemove,
+				'<container:p>[b<attribute:b>a</attribute:b>r]</container:p>',
+				'<container:p>br</container:p>'
+			);
+		} );
+
+		it( 'should remove matches element from range when range is inside text node', () => {
```
scofalik: s/matches/matched

Here and below.

tests/view/writer/clear.js
```diff
@@ -0,0 +1,159 @@
...
+			const elementToRemove = new AttributeElement( 'b' );
+
+			test(
+				elementToRemove,
+				'<container:p>F{o<attribute:b>o</attribute:b> ba}r</container:p>',
+				'<container:p>Fo bar</container:p>'
+			);
+		} );
+
+		it( 'should remove matches elements from range', () => {
```
scofalik: Same test name as above?

scofalik: Oh sorry. Maybe should remove multiple matched elements?

tests/view/writer/clear.js
```diff
@@ -0,0 +1,159 @@
...
+			const elementToRemove = new AttributeElement( 'b' );
+
+			test(
+				elementToRemove,
+				'<container:p>[Fo<attribute:b>o</attribute:b> b<attribute:b>a</attribute:b>r]</container:p>',
+				'<container:p>Fo br</container:p>'
+			);
+		} );
+
+		it( 'should remove matches elements from range when range is inside text node', () => {
```
scofalik: As above.

tests/view/writer/clear.js
```diff
@@ -0,0 +1,159 @@
...
+			const elementToRemove = new EmptyElement( 'img' );
+
+			test(
+				elementToRemove,
+				'<container:p>f{oo<empty:img></empty:img>ba}r</container:p>',
+				'<container:p>foobar</container:p>'
+			);
+		} );
+
+		it( 'should remove UIElement', () => {
```
scofalik: What about removing ContainerElement? Is this a desired functionallity or not?

src/view/writer.js
```diff
@@ -1108,23 +1197,30 @@ function rangeSpansOnAllChildren( range ) {
...
 // {@link module:engine/view/attributeelement~AttributeElement AttributeElements} or
 // {@link module:engine/view/containerelement~ContainerElement ContainerElements}.
 //
 // @param Iterable.<module:engine/view/text~Text|module:engine/view/attributeelement~AttributeElement
 // |module:engine/view/containerelement~ContainerElement> nodes
 function validateNodesToInsert( nodes ) {
 	for ( let node of nodes ) {
-		if ( !( node instanceof Text || node instanceof AttributeElement || node instanceof ContainerElement || node instanceof EmptyElement ) ) {
+		if (
+			!( node instanceof Text || node instanceof AttributeElement || node instanceof ContainerElement ||
```
oskarwrobel: Maybe just
```javascript
if ( !( node instanceof Element ) )
```
?

src/view/writer.js
oskarwrobel: I like to be really sure :P
Fixing...

src/view/writer.js
oskarwrobel:
```javascript
if ( !( node instanceof Node  ) )
```

scofalik:
I talked with @pjasiun about implementation of `UIElement` and it extending `Element` (instead of extending `Node`). Now I understand that:

1. `UIElement` is made basing on `EmptyElement`.
2. View is supposed to be as similar to DOM as possible. "Empty", "UI", "Attribute" or "Container" are our additional requirements on top of `Element` and they are all just DOMElements.
3. Both `EmptyElement` and `UIElement` need attributes, classes and styles, which are implemented on `Element`, not `Node`.

Because of this, for now, You can dismiss my comments on that. However I will propose some refactoring in view structure, in separate issue.

tests/view/writer/clear.js
oskarwrobel: I think it should work for Containers as well, but  unfortunately because of https://github.com/ckeditor/ckeditor5-engine/issues/802 it doesn't. I'm on it.

src/view/writer.js
oskarwrobel: Here we can find some details about this condition: https://github.com/ckeditor/ckeditor5-engine/issues/377

scofalik:
I just realized that you did not exported `clear` function in default exported object `writer` (lines 30 and 45).

oskarwrobel:
Waiting for #802

src/view/position.js
```diff
@@ -202,8 +207,8 @@ export default class Position {
 		}
 
 		// Get path from root to position's parent element.
-		const path = this.parent.getAncestors( { includeNode: true } );
-		const otherPath = otherPosition.parent.getAncestors( { includeNode: true } );
+		const path = this.getAncestors().reverse();
```
Reinmar: Why `reverse()`?

Reinmar:
Does this mean that we close https://github.com/ckeditor/ckeditor5-engine/pull/804?

Reinmar:
BTW, please list all the breaking changes. This PR's message is outdated.

scofalik:
This PR makes https://github.com/ckeditor/ckeditor5-engine/pull/804 obsolete,
closes #805
fixes #802 #803
it also touches #807 where I discuss which parts of `view.DocumentFragment`should be kept, because the PR removes `getAncestors` method from that class.

scofalik:
I updated the first post. Please let me know if it is fine.

PR #800, opened 2017-02-01
Feature: Introduced `view.Document#selectionChangeDone` event. Resolves #791.

### LINKED ISSUE #791: Selection changeDone event
When the user drags the selection it fires the `change` event very very often. There are some actions, like showing balloon toolbar or sending the selection to the server, which should not be executed that ofter. They need an event when the selection is "stable".

There should be a `changeDone` event fired when the user stops changing the selection.

The very first implementation does not need to be perfect, can be based on the event throttling. Selection change is fired asynchronously anyway, so a timeout should not be a big issue here. Of course more precise we will recognize the selection change is done, the better, but the event can be improved with time.

oskarwrobel:
Fixes #791.

For now, additional triggers like `mouseup` or <kbd>Shift</kbd> up does not immediatelly fire `selectionChangeDone` event, but this should be very easy to implement because function wrapped by `lodash#debounce` provides `.flush()` method what is exatlly what we need.

---

Commit message:

Feature: Introduced `view.Document#selectionChangeDone` event. Closes: #791.


Reinmar:
I've talked with @pjasiun about this event and expressed my doubts regarding so specific implementation of it.

I totally see a need for an event which isn't fired when the user makes selection using mouse. It's a clear responsibility of such an event – fire once the user stopped changing the selection. It can be clearly documented and understood by the developers.

But here you got far more than that, based on the UI cases that you defined. You proposed building-in debouncing mechanism, because your specific cases require this. By doing this, you close the path for other use cases which might benefit from the knowledge about when selection stopped changing, but which must not be delayed. You also claim that you know all the use cases for this event which is wrong by default :P.

Besides, the fact of hardcoding a debouncing rate is quite ugly itslef and should warn you that something's seriously wrong in here.

So, how I see it is that we could have a `selectionChangeDone` event, but it should be focused on letting the developer know when the user stopped changing the selection. If, in your specific use case, you need debouncing, use debouncing. But do not build it in.

oskarwrobel:
> You also claim that you know all the use cases for this event which is wrong by default :P.

I don't think about mouse up and Shift up as every possible use case, but only as common use cases for which `selectionChangeDone` event might work a little bit more precisely.

> Besides, the fact of hardcoding a debouncing rate is quite ugly itslef and should warn you that something's seriously wrong in here.

You are right, it doesn't look pretty. I was trying to figure out some better solution but hardcoded debouced function makes it possible to `cancel()` this function (when is pending) while class is being destroyed.

oskarwrobel:
> You proposed building-in debouncing mechanism, because your specific cases require this. By doing this, you close the path for other use cases which might benefit from the knowledge about when selection stopped changing, but which must not be delayed.

I'm not sure if this debouncing mechanism closes the path for other use cases which must not be delayed. This hardcoded debounced function can be canceled or invoked without delay by `cancel()` and `flush()` methods.

Reinmar:
> I'm not sure if this debouncing mechanism closes the path for other use cases which must not be delayed. This hardcoded debounced function can be canceled or invoked without delay by cancel() and flush() methods.

If we'll clear it that on actions like `mouseup` or `keyup` there's always an immediate event, then perhaps there's a brighter future for this event. Then, the only debounced moment would be during dragging. Am I right?

oskarwrobel:
During dragging and when user holds <kbd>Shift</kbd> pressed and stopped changing selection by pressing arrow keys.

Reinmar:
> During dragging and when user holds Shift pressed and stopped changing selection by pressing arrow keys.

How many native `selectionchange` events can be fired for the latter? I have manually speeded up key repeat in MacOS to something like 30/s. The default is, I guess, around 5-10/s. Is it really something we must debounce?

Reinmar:
> The default is, I guess, around 5-10/s. Is it really something we must debounce?

Ekhm... I forgot about other OSes :D. Anyway, let's also not overexaggerate and let's debounce only things which need to be debounced. Key repeat, for me, is not such thing. Did you have performance issues already?

src/view/observer/selectionobserver.js
```diff
@@ -232,3 +252,21 @@ export default class SelectionObserver extends Observer {
  * @param {module:engine/view/selection~Selection} data.newSelection New View selection which is converted DOM selection.
  * @param {Selection} data.domSelection Native DOM selection.
  */
+
+/**
+ * Fired when selection stop changing.
```
Reinmar: `stops`

tests/view/observer/fakeselectionobserver.js
```diff
@@ -100,6 +100,60 @@ describe( 'FakeSelectionObserver', () => {
...
+
+		const spy = sinon.spy();
+
+		viewDocument.on( 'selectionChangeDone', spy );
+
+		// Change selection.
+		changeFakeSelectionPressing( keyCodes.arrowdown );
+
+		// Wait 100ms.
+		setTimeout( () => {
```
Reinmar: Please avoid timeouts in the tests if you can use fake timers (from Sinon).

src/view/observer/fakeselectionobserver.js
```diff
@@ -28,6 +29,15 @@ export default class FakeSelectionObserver extends Observer {
 	 */
 	constructor( document ) {
 		super( document );
+
+		/**
+		 * Fires `selectionChangeDone` event debounced. It use `lodash#debounce` method to delay function call.
```
Reinmar: "Fires debounce event" would be the right order. "It uses"

tests/view/observer/fakeselectionobserver.js
oskarwrobel: Using fake timers for lodash#debounce is not so easy: https://github.com/lodash/lodash/issues/304

I'll try to use `runInContext` for this.

Reinmar:
BTW, I’m working on a documentation for the commit names convention. It will take me a while, so if you need to commit anything now please refer to https://github.com/ckeditor/ckeditor5-dev/tree/master/packages/ckeditor5-dev-env#generating-changelog

In this case, it will be something like:

Feature: Introduced `view.Document#selectionChangeDone` event. Resolves #791.

tests/view/observer/fakeselectionobserver.js
oskarwrobel: Lodash debounce keeps `Date.now` in closure. I'm out of ideas how to test it using sinon fake timers.

oskarwrobel: Looks like there is a way to override lodash internal `Date.now` and `setTimeout` but to do it we have to use lodash `_` object instead of single import.

Reinmar: I don't understand. Doesn't http://sinonjs.org/docs/#clock work in this case?

oskarwrobel: It doesn't.

oskarwrobel: ### Tests:
```javascript
	it.only( 'should fire `selectionChangeDone` using sinon fake timers', () => {
		const clock = testUtils.sinon.useFakeTimers();
		const spy = testUtils.sinon.spy();

		viewDocument.on( 'selectionChangeDone', spy );

		changeDomSelection();

		clock.tick( 300 );

		expect( spy.calledOnce ).to.true;

		clock.restore();
	} );

	it.only( 'should fire `selectionChangeDone` using setTimeout', ( done ) => {
		const spy = testUtils.sinon.spy();

		viewDocument.on( 'selectionChangeDone', spy );

		changeDomSelection();

		setTimeout( () => {
			expect( spy.calledOnce ).to.true;
			done();
		}, 300 );
	} );
```
### Result:
```bash
  SelectionObserver
    ✖ should fire `selectionChangeDone` using sinon fake timers
    ✔ should fire `selectionChangeDone` using setTimeout

Finished in 0.68 secs / 0.31 secs @ 00:06:50 GMT+0100 (CET)

SUMMARY:
✔ 1 test completed
✖ 1 test failed

FAILED TESTS:
  SelectionObserver
    ✖ should fire `selectionChangeDone` using sinon fake timers
      Chrome 55.0.2883 (Mac OS X 10.12.2)
    AssertionError: expected false to be true
        at Context.it.only (packages/ckeditor5-engine/tests/view/observer/selectionobserver.js:12320:28)
```

Reinmar: Hm... pity. Perhaps Sinon controls timeouts but not the `Date()` class. Or you'd need to initialize the fake timers before the `viewDocument` is created. Did your try doing this in `beforeEach()`?

oskarwrobel: This will be more complicated, because `beforeEach` uses `setTimeout()`
```javascript
// Ensure selectionchange will not be fired.
setTimeout( () => done(), 100 );
```
I tried to use fake timers instead of this initial `setTimeout` but it didn't work. Besides, `setTimeout` is used few times in other test cases of this class and I'm not sure what is the reason, maybe fake timers are problematic for testing SelectionObserver? This may require deeper investigation.

Reinmar: For other tests we must certainly use real set timeouts. But ok, enough time for that – let's use real timeouts everywhere and only make sure that you require minimum delay – so e.g. write minimal number of tests.

PR #804, opened 2017-02-06
Fix: view.DocumentFragmen.getAncestors returns array with itself when requested. Closes: #803.

oskarwrobel:
### Commit message:

Fix: `view.DocumentFragmen.getAncestors` should return an array with itself when requested. Closes: #803.

### DIFF
--- a/src/view/documentfragment.js
+++ b/src/view/documentfragment.js
@@ -86,10 +86,12 @@ export default class DocumentFragment {
 	/**
 	 * Returns ancestor elements of `DocumentFragment`, which is an empty array. Added for compatibility reasons.
 	 *
-	 * @returns {Array}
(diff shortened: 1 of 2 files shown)


Reinmar:
Don't list ticket to closes after a comma. They must be separated – see https://github.com/ckeditor/ckeditor5-dev/issues/70.

Also, when documenting what has been fixed it's better to write what's the expected result rather than what was the old one. It's a bit more clear then.

Reinmar:
I'd like to understand https://github.com/ckeditor/ckeditor5-engine/issues/803#issuecomment-277740623 first.

Reinmar:
Replaced by https://github.com/ckeditor/ckeditor5-engine/pull/798

PR #810, opened 2017-02-09
T/792

### LINKED ISSUE #792: Markers to UI converters
When the `UIElement` will be introduced (https://github.com/ckeditor/ckeditor5-engine/issues/788) a Marker to UIElemnt converter is needed. It should take care about:
 - creating the UI Elemenet (duh!),
 - mapper, ensure that it will not crash because there is no model element for that view element (it might be needed to tell mapper what is the length of the element),
 - removing the UI Element (note that mapper will not return the exact position the UI element is and the converter needs to enlarge that position https://github.com/ckeditor/ckeditor5-engine/issues/789).

In the future we may consider also `attribute` to UI converter.

scofalik:
### Suggested merge commit message

Feature: Model converters from markers to `view.UIElement`. Closes #792.

---

### Additional information

`wrapRange` and `unwrapRange` converters does not properly convert collapsed markers anymore. Use introduced `insertUIElement` and `removeUIElement` instead. In most cases, given marker type will be either collapsed or non-collapsed, markers that have sense collapsed and non-collapsed will be rarity. Developer introducing functionality using such markers will have to bring it's own converter (probably will be able to use existing converters, though).

### DIFF
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -9,8 +9,10 @@
 
 import {
 	insertElement,
+	insertUIElement,
 	setAttribute,
 	removeAttribute,
+	removeUIElement,
 	wrapItem,
 	unwrapItem,
 	wrapRange,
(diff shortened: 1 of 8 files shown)


Reinmar:
Was:

Feature: model converters from markers to view.UIElement. Closes: #792.

Should be:

Feature: Model converters from markers to `view.UIElement`. Closes #792.

PR #813, opened 2017-02-12
T/808

### LINKED ISSUE #808: Engine debugging tools.
As some may already experienced, debugging engine may be difficult and tiring. Especially some more complicated parts like: operational transformation or model-view conversion could use a tool that will help in debugging.

In this issue I propose introducing a method, that when called, will enhance engine classes with additional functionalities. For now we could settle with calling the function manually (i.e. at the beginning of manual sample or unit test that we want to debug).

The list of features introduced by the debug tools is not closed, for now I propose:
* add `log` method to model deltas, operations, ranges and positions - the function will output clean and brief information about object so it is easier to check if, for example an operation has correct parameters,
* print model tree, view tree and DOM tree for given root,
* store model document state since last _n_ operations ("history" of model document states) - to be able to easier track what changed in the model after applying an operation, stored model document will have to be printable (and maybe diffable with current model?), I think that 5-10 operations history will be enough,
* logs in transforming operations (to see what operations/deltas where input and what was the transformation result) - possible addition, I am not sure if it will be doable this way,
* at some point we will also probably need some help debugging `LiveRange`s / `Marker`s but these will be added if and when needed.

scofalik:
### Suggested merge commit message

Feature: Introduced debugging tools for the engine. Closes #808.

### DIFF
--- a/src/dev-utils/enableenginedebug.js
+++ b/src/dev-utils/enableenginedebug.js
@@ -405,7 +405,7 @@ function enableDocumentTools() {
 	const _modelDocumentApplyOperation = ModelDocument.prototype.applyOperation;
 
 	ModelDocument.prototype.applyOperation = function( operation ) {
-		log( 'Applying : ' + operation );
+		log( 'Applying ' + operation );
 
 		_modelDocumentApplyOperation.call( this, operation );
 	};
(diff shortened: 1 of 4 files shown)


scofalik:
I was so much thinking that these are debug tools that I forgot to add comments.

scofalik:
I've added the docs.

Reinmar:
I asked @szymonkups to  review it.

szymonkups:
Code looks fine. I've pushed some minor docs fixes.

PR #814, opened 2017-02-13
t/789: View range enlarge method

### LINKED ISSUE #789: View range enlarge method
Because mapper does return the exact view position we need to have a way to get all possible positions. Fortunately, thanks to the `view.TreeWalker`, it should be much simpler than the method in the CKEditor 4 (which is my favorite CKEditor 4 method and the symbolic reason why we are writing CKE 5 from scratch: https://github.com/ckeditor/ckeditor-dev/blob/2db0ce3e25f9c97404e2be107e1250a78254ddd7/core/dom/range.js#L1201-L1715).

The enlarge will go forward as long as it does not meet a text node or an empty element, so enlarge the range including all non-semantic content (opening and closing attribute elements tags).

pjasiun:
## Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Feature: Introduced `view.Range#getEnlarged()`, `view.Range#getTrimmed()`, `view.Position#getLastMatchingPosition()`, `model.Position#getLastMatchingPosition()`, `view.TreeWalker#skip()`, `model.TreeWalker#skip()`. Closes #789.

### DIFF
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -270,7 +270,7 @@ export default class Position {
 	 *
 	 * For example:
 	 *
-	 * 		getPriorPosition( value => value.type == 'text' ); // <paragraph>[]foo</paragraph> -> <paragraph>foo[]</paragraph>
+	 * 		getPriorPosition( value => value.type == 'text' ); // <paragraph>foo[]</paragraph> -> <paragraph>[]foo</paragraph>
 	 * 		getPriorPosition( value => false ); // Do not move the position.
 	 *
 	 * @param {Function} skip Callback function. Gets {@link module:engine/model/treewalker~TreeWalkerValue} and should
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -9,6 +9,7 @@
 
 import DocumentFragment from './documentfragment';
 import Element from './element';
+import TreeWalker from './treewalker';
 import last from '@ckeditor/ckeditor5-utils/src/lib/lodash/last';
 import compareArrays from '@ckeditor/ckeditor5-utils/src/comparearrays';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
@@ -242,6 +243,44 @@ export default class Position {
 					return 'after';
 				}
 		}
+	}
+
+	/**
+	 * Use forward {@link module:engine/model/treewalker~TreeWalker TreeWalker} to get the farthest position which
+	 * matches the callback.
+	 *
+	 * For example:
+	 *
+	 * 		getFurtherPosition( value => value.type == 'text' ); // <paragraph>[]foo</paragraph> -> <paragraph>foo[]</paragraph>
+	 * 		getFurtherPosition( value => false ); // Do not move the position.
+	 *
+	 * @param {Function} skip Callback function. Gets {@link module:engine/model/treewalker~TreeWalkerValue} and should
+	 * return `true` if the value should be skipped or `false` if not.
+	 */
+	getFurtherPosition( skip ) {
+		const treeWalker = new TreeWalker( { startPosition: this } );
+		treeWalker.skip( skip );
+
+		return treeWalker.position;
+	}
+
+	/**
+	 * Use backward {@link module:engine/model/treewalker~TreeWalker TreeWalker} to get the farthest position which
+	 * matches the callback.
+	 *
+	 * For example:
+	 *
+	 * 		getPriorPosition( value => value.type == 'text' ); // <paragraph>[]foo</paragraph> -> <paragraph>foo[]</paragraph>
+	 * 		getPriorPosition( value => false ); // Do not move the position.
+	 *
+	 * @param {Function} skip Callback function. Gets {@link module:engine/model/treewalker~TreeWalkerValue} and should
+	 * return `true` if the value should be skipped or `false` if not.
+	 */
+	getPriorPosition( skip ) {
+		const treeWalker = new TreeWalker( { startPosition: this, direction: 'backward' } );
+		treeWalker.skip( skip );
+
+		return treeWalker.position;
 	}
 
 	/**
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -246,38 +246,30 @@ export default class Position {
 	}
 
 	/**
-	 * Use forward {@link module:engine/model/treewalker~TreeWalker TreeWalker} to get the farthest position which
-	 * matches the callback.
+	 * Gets the farthest position which matches the callback using
+	 * {@link module:engine/model/treewalker~TreeWalker TreeWalker}.
 	 *
 	 * For example:
 	 *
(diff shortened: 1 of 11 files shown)


src/model/treewalker.js
```diff
@@ -153,6 +153,34 @@ export default class TreeWalker {
...
+	 * For example:
+	 *
+	 * 		walker.skip( value => value.type == 'text' ); // <paragraph>[]foo</paragraph> -> <paragraph>foo[]</paragraph>
+	 * 		walker.skip( value => true ); // Move the position to the end: <paragraph>[]foo</paragraph> -> <paragraph>foo</paragraph>[]
+	 * 		walker.skip( value => false ); // Do not move the position.
+	 *
+	 * @param {Function} skip Callback function. Gets {@link module:engine/model/treewalker~TreeWalkerValue} and should
+	 * return `true` if the value should be skipped or `false` if not.
+	 */
+	skip( skip ) {
```
scofalik: Didn't you want to implement this functionality as additional option in `TreeWalker` configuration? This would allow making `TreeWalker` more powerful in normal usage.

pjasiun: Yea. Good luck ;) This was my first idea but then I realized how complex it would be. This way you can call `skip` multiple times, if you need, with different parameters and the code is simple. Adding more options to the `TreeWalker` we will make to code unmaintainable (in fact it's on the edge now).

src/view/range.js
```diff
@@ -82,6 +87,60 @@ export default class Range {
 	}
 
 	/**
+	 * Creates a new range with the same content, but containing all surrendering tags with no semantic meaning.
```
scofalik: I am not sure if this comment is correct, though I understand it is hard to explain how enlarge works :).

Anyway, I'd skip the second part of that because both `<p>` and `<b>` in examples have semantic meaning.

Maybe: "Creates a maximal range that has the same content as this range but is expanded in both ways (at the beginning and at the end).".

src/view/range.js
```diff
@@ -82,6 +87,60 @@ export default class Range {
...
+	 *
+	 * For example:
+	 *
+	 * 		<p>Foo</p><p><b>{Bar}</b></p> -> <p>Foo</p>[<p><b>Bar</b></p>]
+	 * 		<p><b>foo</b>{bar}<span></span></p> -> <p><b>foo[</b>bar<span></span>]</p>
+	 *
+	 * Note that in the sample above all:
```
scofalik: remove "all", I understand that sentence as "the sample that is above all samples".

src/view/range.js
scofalik: BTW. I understand that by "semantic" you meant "not being a data". Data elements are `view.Text` and `view.EmptyElement` right now, right? Others are aggregating elements (`ContainerElement`) styling elements (`AttributeElement`) and UI elements (`UIElement`).

src/model/treewalker.js
scofalik: I understand, I thought that was the case.

src/view/range.js
```diff
@@ -82,6 +87,60 @@ export default class Range {
...
+	 *
+	 * @returns {module:engine/view/range~Range} Shrink range.
+	 */
+	getShrinked() {
+		let start = this.start.getFurtherPosition( enlagreShrinkStartSkip );
+		let end = this.end.getPriorPosition( enlagreShrinkEndSkip );
+		let nodeAfterStart = start.nodeAfter;
+		let nodeBeforeEnd = end.nodeBefore;
+
+		if ( nodeAfterStart instanceof Text ) {
```
scofalik: I'll add a comment that "Technically, position at the beginning of a text node is after position before that node, so we are fixing this because TreeWalker does not enter text nodes"

scofalik: So others won't have to think why it's that.

src/view/range.js
```diff
@@ -349,3 +408,29 @@ export default class Range {
 		return this.createFromPositionAndShift( Position.createBefore( item ), 1 );
 	}
 }
+
+// Function used by getEnlagred and getShrinked methods.
+function enlagreShrinkStartSkip( value ) {
```
scofalik: Those two functions differ only in one string. Maybe we can do them as function generator function (like converters).

Reinmar:
I'm not sure that `getPriorPosition()` and `getFurtherPosition()` are correct method names. How are they different from previous and next positions?

src/model/position.js
```diff
@@ -245,6 +246,44 @@ export default class Position {
 	}
 
 	/**
+	 * Use forward {@link module:engine/model/treewalker~TreeWalker TreeWalker} to get the farthest position which
```
Reinmar: "Uses".

Also, I'm not sure whether mentioning that it uses the TW in the first sentence makes any sense. This is implementation detail.

Finally, "farthest position which matches the callback" is confusing. Is it the furthest position in the document? I guess nope. So perhaps the better comment would be "farthest position which still matches the callback"? Or "last position which still ..."?

Reinmar:
`getLastMatchingPosition()`, `getFirstMatchingPosition()`?

Or better:

`getLastMatchingPositionForward()` and `getLastMatchingPositionBackward()?

src/model/position.js
scofalik: That was my thought too.

scofalik:
The code is clean and solves generic case, which I like. Please check out the comments by me and @Reinmar and if you don't want to discuss them, for me, you can just fix and merge.

src/view/range.js
```diff
@@ -82,6 +87,60 @@ export default class Range {
...
+	 *
+	 * For example:
+	 *
+	 * 		<p>Foo</p>[<p><b>Bar</b></p>] -> <p>Foo</p><p><b>{Bar}</b></p>
+	 * 		<p><b>foo[</b>bar<span></span>]</p> -> <p><b>foo</b>{bar}<span></span></p>
+	 *
+	 * Note that in the sample above all:
+	 *  - `<p>` have type of {@link module:engine/view/containerelement~ContainerElement},
+	 *  - `<b>` have type of {@link module:engine/view/attributeelement~AttributeElement},
+	 *  - `<span>` have type of {@link module:engine/view/uielement~UiElement}.
```
Reinmar: Should be `UIElement`.

src/view/range.js
```diff
@@ -82,6 +87,60 @@ export default class Range {
...
+	 *
+	 * @returns {module:engine/view/range~Range} Enlarged range.
+	 */
+	getEnlarged() {
+		const start = this.start.getPriorPosition( enlagreShrinkStartSkip );
+		const end = this.end.getFurtherPosition( enlagreShrinkEndSkip );
```
Reinmar: enlarge

src/view/range.js
pjasiun: Yep.

src/view/range.js
pjasiun: Isn't it overengineering? These functions could be one-liners as well.

PR #816, opened 2017-02-13
Fixed issue when setData() didn't work with custom roots.

### LINKED ISSUE #815: Dev-utils: setData() doesn't work with a non-standard root
```js
it.only( 'should be false in a root which does not allow blocks at all', () => {
	doc.createRoot( 'paragraph', 'inlineOnlyRoot' );
	setData( doc, 'a[]b', { rootName: 'inlineOnlyRoot' } );
	expect( command.isEnabled ).to.be.false;
} );
```

I can't use it like that. The reason is that the `parse()` function is used with a wrong context.

Reinmar:

Fix: `dev-utils/model#setData()` should work with custom roots. Closes #815.

---

### Additional information

I'll run all editors tests because this issue meant that some issues in tests might be unnoticed.

### DIFF
--- a/src/dev-utils/model.js
+++ b/src/dev-utils/model.js
@@ -100,7 +100,8 @@ export function setData( document, data, options = {} ) {
 	// Parse data string to model.
 	const parsedResult = setData._parse( data, document.schema, {
 		lastRangeBackward: options.lastRangeBackward,
-		selectionAttributes: options.selectionAttributes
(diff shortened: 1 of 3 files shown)


Reinmar:
```
Finished in 39.333 secs / 11.218 secs

SUMMARY:
✔ 4900 tests completed
[15:07:24] Finished 'test' after 1.23 min
```

PR #817, opened 2017-02-13
Introduced Selection#getSelectedBlocks()

### LINKED ISSUE #811: Try to extract "get selected blocks" logic from the list and heading features to the engine
Issue starter: https://github.com/ckeditor/ckeditor5-engine/issues/532#issuecomment-278900072

The code repeats in 2 places now and I need in the 3rd place (message quote), so it's time to hoist it.

Reinmar:

Feature: Introduced `model.Selection#getSelectedBlocks()`. Closes #811.

---

### Additional information

Requires #815 (merged into this branch).

### DIFF
--- a/src/model/selection.js
+++ b/src/model/selection.js
@@ -564,10 +564,11 @@ export default class Selection {
 	/**
 	 * Gets elements of type "block" touched by the selection.
 	 *
-	 * This method's result can be used for example to apply block styling to all
-	 * blocks covered by this selection.
+	 * This method's result can be used for example to apply block styling to all blocks covered by this selection.
 	 *
-	 * In this case the function will return all 3 paragraphs:
+	 * *Note:* `getSelectedBlocks` always returns the deepest block.
+	 *
+	 * In this case the function will return exactly all 3 paragraphs:
 	 *
 	 *		<paragraph>[a</paragraph>
 	 *		<quote>
@@ -592,7 +593,7 @@ export default class Selection {
 			}
 
 			for ( const value of range.getWalker() ) {
-				if ( value.type == 'elementEnd' && checkIsBlockContainer( value.item, visited ) ) {
+				if ( value.type == 'elementEnd' && isUnvisitedBlockContainer( value.item, visited ) ) {
 					yield value.item;
 				}
 			}
@@ -708,7 +709,7 @@ mix( Selection, EmitterMixin );
 
 // Checks whether the given element extends $block in the schema and has a parent (is not a root).
 // Marks it as already visited.
-function checkIsBlockContainer( element, visited ) {
+function isUnvisitedBlockContainer( element, visited ) {
 	if ( visited.has( element ) ) {
 		return false;
 	}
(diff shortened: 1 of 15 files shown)


src/model/selection.js
```diff
@@ -562,6 +562,50 @@ export default class Selection {
 	}
 
 	/**
+	 * Gets elements of type "block" touched by the selection.
+	 *
+	 * This method's result can be used for example to apply block styling to all
+	 * blocks covered by this selection.
+	 *
+	 * In this case the function will return all 3 paragraphs:
```
scofalik: What exactly will be returned? Just 3 paragraphs? Or 3 paragraphs and quote? Clarification needed. If just 3 paragraphs it need to be clarified that quote is not seen as block.

src/model/selection.js
```diff
@@ -661,3 +705,30 @@ export default class Selection {
 }
 
 mix( Selection, EmitterMixin );
+
+// Checks whether the given element extends $block in the schema and has a parent (is not a root).
+// Marks it as already visited.
+function checkIsBlockContainer( element, visited ) {
```
scofalik: `function isUnvisitedBlockContainer( element, visited ) {`

src/model/selection.js
```diff
@@ -661,3 +705,30 @@ export default class Selection {
...
+
+	// TODO https://github.com/ckeditor/ckeditor5-engine/issues/532#issuecomment-278900072.
+	// This should not be a `$block` check.
+	return element.document.schema.itemExtends( element.name, '$block' ) && element.parent;
+}
+
+// Finds the lowest element in position's ancestors which is a block.
+// Marks all ancestors as already visited to not include any of them later on.
+function getParentBlock( position, visited ) {
+	const ancestors = position.parent.getAncestors( { parentFirst: true, includeNode: true } );
```
scofalik: `position.getAncestors()` should return same set of nodes

scofalik: Sorry, I've been looking at the wrong version of code. Currently, `Position#getAncestors` returns them in parent-last order.

scofalik:
What about such situation:
```
setData( doc, '<nestedBlock>a[<nestedBlock>b</nestedBlock>]</nestedBlock>' );
```
In this case, both `nestedBlock`s are returned. If we claim that nesting `$block`s is bad practice, then okay - as you've written in tests - this is edge case and it is not expected to work 100% correctly.

OTOH this might be fixable - we would have to check what are ranges start/end surroundings before deciding to push `range.start` or `range.end` parent to the results set.

scofalik:
I've pushed commit fixing what I pointed out. If you don't want to do anything with the case I described above, feel free to merge this PR.

src/model/selection.js
Reinmar: Hm... To my taste, this is too verbose, but I understand your decision and can live with it :D.

Reinmar:
> In this case, both nestedBlocks are returned. If we claim that nesting $blocks is bad practice, then okay - as you've written in tests - this is edge case and it is not expected to work 100% correctly.

Yeah, I knew that the cases I decided to cover were just the tip of the iceberg. There's a whole spectrum of weird schema configurations to cover and I decided to ignore them. It's never an easy choice because those things backfire sooner or later, so you need to trust yourself that the most common scenarios are covered.

PR #820, opened 2017-02-15
Tests: Added manual test featuring all features.

### LINKED ISSUE #819: One test to test them all.
I propose creating a manual test in `ckeditor5-engine` that will have all the official plugins and features enabled. The test will check editor integrity. It will be similar to what we have on https://ckeditor5.github.io but we will be able to test it when developing features.

scofalik:

Tests: Added a manual test that features all plugins and features. Closes #819.


Reinmar:
As mentioned in #819, it's not the perfect place to put integration tests. E.g. you forgot the deps in `package.json`. This test only works because the directory is places inside ckeditor5 which delivers all the other packages.

Unless this is critical to have this test ASAP, I'd recommend waiting until we have presets or tests in ckeditor5.

Reinmar:
Not needed any more. The presets have their manual tests (https://github.com/ckeditor/ckeditor5-presets/tree/master/tests) and the builds will have too.

Issue #710, opened 2016-12-01
Blockquote implementation


szymonkups:
This is an initial issue describing `Blockquote` feature implementation. 
There is no requirements list yet, but we already see that it will require some changes in the engine. Let's gather them here.

#### Add support for nested elements to `DataController#modifySelection`:
Extending selection in situation like this:
```html
<blockquote><p>foo[]</p></blockquote><p>bar</p>
```
should return:
```html
<blockquote><p>foo[</p></blockquote><p>]bar</p>
```

Reinmar:
I'm also curious how's `deleteContent()` dealing with such a cases. I remember that I didn't want to think too much on it back then so the implementation may be poor.

Reinmar:
Besides, it's tricky to understand which rules are specific to blockquote and which are general and should land in the default algorithms.

fredck:
Btw, be sure to check [Editor Recommendations on Quotes](http://ckeditor.github.io/editor-recommendations/features/quote.html).

Reinmar:
Good point. Do we want to have a caption support? If so, then it's a far more complex feature than what we have in CKEditor 4. It may need to be implemented with a widget or some smart model operations. The latter would be a very interesting test for our data model, but I'm afraid that we don't have time for that now.

We'd need to start this with a UX research too, cause I'm not sure how should the caption be reachable when vs how to escape blockquote. Perhaps it's a simple rule that enter in an empty paragraph in a blockquote removes that paragraph and moves selection out of the quote and below the caption.

Reinmar:
I talked with @fredck. We're going now with a CKEditor4 like implementation (but rendering to a `<figure class="quote">`). We'll rethink the captioning in the future.

szymonkups:
When modifying selection in situation described in  [first comment](https://github.com/ckeditor/ckeditor5-engine/issues/710#issue-192804232) will be implemented, it should also support objects and widgets should also use it during `delete` and `backspace` handling:
```html
<widget></widget><blockquote><p>[]</p></blockquote>
// backspace pressed
[<widget></widget>]
```
```html
<blockquote><p>[]</p></blockquote><widget></widget>
// delete pressed
[<widget></widget>]
```

Reinmar:
I can see that `modifySelection()` is covered by `describe( 'beyond element – skipping incorrect positions' )` and works as expected.

So the problem with https://github.com/ckeditor/ckeditor5-block-quote/issues/7 lays inside `deleteContent()`.

PR #825, opened 2017-02-20
Added Schema#limits and integrated it with data controller's modifySelection

### LINKED ISSUE #818: Integrate various algorithms with schema#limits
We need to add `Shema#limits` https://github.com/ckeditor/ckeditor5-engine/issues/532#issuecomment-278900072 to improve compatibility of our algorithms with nested editables like `ImageCaption`.
I will create separate issues in features that needs this improvement, I want to gather here a list of places where change will be needed:

- enter should not break `limit` elements: https://github.com/ckeditor/ckeditor5-enter/issues/38,
- backspace should not move selection outside empty `limit` elements
output

Any other ideas?

szymonkups:

Feature: Added `Schema#limits` and integrated it with data controller's modifySelection. Closes #818.

### DIFF
--- a/src/controller/deletecontent.js
+++ b/src/controller/deletecontent.js
@@ -21,6 +21,8 @@ import compareArrays from '@ckeditor/ckeditor5-utils/src/comparearrays';
  * @param {Boolean} [options.merge=false] Merge elements after removing the contents of the selection.
  * For example, `<h>x[x</h><p>y]y</p>` will become: `<h>x^y</h>` with the option enabled
  * and: `<h>x^</h><p>y</p>` without it.
+ * Note: {@link module:engine/model/schema~Schema#objects object} and {@link module:engine/model/schema~Schema#limits limit}
+ * elements will not be merged.
  */
 export default function deleteContent( selection, batch, options = {} ) {
 	if ( selection.isCollapsed ) {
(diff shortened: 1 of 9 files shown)


szymonkups:
Please close together with https://github.com/ckeditor/ckeditor5-enter/pull/39.

Reinmar:
I fixed the failing tests, but there are missing tests for `modifySelection()`:

image

Reinmar:
Also, one important thing to notice is that you check only the immediate parent of the selection (whether it's a block or not). We don't have yet a case for that but you may have:

```html
<caption><paragraph>^</paragraph></caption>
```

And the selection must not be extended to the left in this case.

Reinmar:
More things:

* This PR desc is wrong – the limits are integrated with all the DC's methods, not just `modifySelection()`.
* I'm able to paste a block content into an image caption if I paste at the end of it. If I paste in the middle the https://github.com/ckeditor/ckeditor5/issues/404 occurs. The first must be blocked – the insertion should not leave the limit element. The latter may be a bug in `insertContent()` too so you can give it a try fixing it.

Reinmar:
I'm going to take over the PR from here, @szymonkups.

Reinmar:
Warning for the future (and fun fact too :P). The added `modifySelection()` test looked like this:

```js
test(
	'should not extend to outside of inline limit element',
	'<inlineLimit>foo[]</inlineLimit>',
	'<inlineLimit>foo[]</inlineLimit>'
);
```

(it was meant to test that the selection won't leave a "limit" element)

It passed. At the same time, the CC dropped below 100%. So what's wrong here?

Two things actually.

First – if the test was written before the code was modified (TDD!) you'd know that it passes without any changes in the code.

Why? Because the test is too limited. It tests the behaviour in too narrow context. Namely – the selection has no place to go wrong – there's no text outside the `<inlineElement>` :D.

When writing such tests I'm very often adding some additional elements or text nodes at the both ends of the content. I also try many variations and one or two more complicated scenarios where more things could go wrong. This, even though might be a bit exaggerated at the beginning (because you can have 100% CC a lot earlier), makes sure the algorithm stays stable in the future, when some inevitable refactoring touches it.

Reinmar:
Okey dokey, I added a lot of tests for all 3 functions – `modifySelection()`, `insertContent()` and `deleteContent()`. I had to reimplement `modifySelection()` and I took this opportunity to change the semantics a bit. Now, the focus will be placed in the first position where text is allowed. This simplified the implementation a lot and handled a lot of cases, including some from #710. In case of other functions, I made sure that limit elements are never left or changed (e.g. merged with other elements in case of `deleteContent()`).

Based on my manual tests, all works fine now :).

PR #827, opened 2017-02-20
T/809

### LINKED ISSUE #809: Introduce Node/DF#is method
We had such method in CKE4, but it was only available for elements: `el.is( 'p' )`. It's just a bit nicer than `el.name == 'p'`, but a lot nicer than `el.name == 'p' || el.name == 'div'`.

What I've been thinking of is to make this method even more useful, by allowing checking whether the node is a text node or document fragment or whatever else.

There are many algorithms which need to verify what kind of node they are dealing with. We've seen e.g. the whole discussion around https://github.com/ckeditor/ckeditor5-engine/issues/807.

If we had a nicer API to check the type of the node or DF, then we'd be more pleased to use it.

Some ideas:

```js
// If called with one arg, checks if this is an element with a given name:
node.is( 'p' );
// Equals to:
node.is( 'element', 'p' );

// Other types:
node.is( 'documentFragment' );
node.is( 'text' );
node.is( 'textProxy' );
node.is( 'text', 'expected data?' ); // not sure about this

// Matching more elements names:
node.is( [ 'p', 'div' ] ); // I don't like passing an array, so maybe simply...
node.is( 'p,div' );
```

Such method will have a lot of pros:

1. Less code.
1. More readable code.
1. Less imports.
1. Less potential circular deps.

scofalik:

Feature: Introduced `is()` method for model and view tree nodes and document fragments. Closes #809.

### DIFF
--- a/src/model/element.js
+++ b/src/model/element.js
@@ -81,6 +81,37 @@ export default class Element extends Node {
 	}
 
 	/**
+	 * Checks whether given model tree object is of given type.
+	 *
+	 * This method is useful when processing model tree objects that are of unknown type. For example, a function
+	 * may return {@link module:engine/model/documentfragment~DocumentFragment} or {@link module:engine/model/node~Node}
+	 * that can be either text node or element. This method can be used to check what kind of object is returned.
+	 *
+	 *		obj.is( 'node' ); // true for any node, false for document fragment
+	 *		obj.is( 'documentFragment' ); // true for document fragment, false for any node
+	 *		obj.is( 'element' ); // true for any element, false for text node or document fragment
+	 *		obj.is( 'element', 'paragraph' ); // true only for element which name is 'paragraph'
+	 *		obj.is( 'paragraph' ); // shortcut for obj.is( 'element', 'paragraph' )
+	 *		obj.is( 'text' ); // true for text node, false for element and document fragment
+	 *		obj.is( 'textProxy' ); // true for text proxy object
+	 *
+	 * @param {'node'|'element'|'rootElement'|'text'|'textProxy'|'documentFragment'} type
+	 * @param {String} [name] Element name.
+	 * @returns {Boolean}
+	 */
+	is( type, name = null ) {
+		let matches = false;
+
+		if ( !name ) {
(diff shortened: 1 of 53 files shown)


src/model/element.js
```diff
@@ -81,6 +81,33 @@ export default class Element extends Node {
 	}
 
 	/**
+	 * Checks whether given model tree object is of given type.
```
Reinmar: Surperfluous comment. Can be `@inheritDoc`.

src/model/liveselection.js
```diff
@@ -651,7 +651,7 @@ export default class LiveSelection extends Selection {
 // Helper function for {@link module:engine/model/liveselection~LiveSelection#_updateAttributes}. It takes model item, checks whether
 // it is a text node (or text proxy) and if so, returns it's attributes. If not, returns `null`.
 //
-// @param {module:engine/model/item~Item}  node
+// @param {module:engine/model/item~Item|null}  node
 // @returns {Boolean}
 function getAttrsIfCharacter( node ) {
 	if ( node instanceof TextProxy || node instanceof Text ) {
```
Reinmar: `node && ( node.is( 'textProxy' ) || node.is( 'text' ) )`?

scofalik: There are many places where such replacement would be viable but I wanted to change only those where we are sure that we have some kind of element (because it is returned by the method that guarantees returning something or is a parameter that has to be for example a node or something will crash anyway).

Reinmar:
All fine. Perhaps we could just keep the documentation in one place. Full documentation could be in `Node#is` and the other would look like this:

```
@inheritDoc
```

or:

```
 Checks whether given model tree object is of given type.

Read more in {@link module:engine/model/node~Node#is}.

@param {module:engine/model/node~ModelItemTypeName}
@param {String} [name] Element name.
```

So this also assumes defining a typedef in the node module to not repeat the vals everywhere.

(Yes, I know that we're not going to change this code very often, so you can ignore my comment and keep it in mind for the next time you encounter similar situation ;P)

scofalik:
Hope you like it better now.

PR #834, opened 2017-02-22
T/830 view.Range.enlarge is not compatible with view.Writer.wrap

### LINKED ISSUE #830: view.Range.enlarge is not compatible with view.Writer.wrap
`view.Writer.wrap` expect that the whole range will be in the same container while `view.Range.enlarge` go outside the container. It means that calling wrap/unwrap on the enlarged range causes error.

The simplest solution seems to be changing `enlarge` not to go outside the container. It was not clear from the beginning if enlarge should do it.

pjasiun:

Other: `view.Range.enlarge()` and `view.Range.shrink()` should not pass the container limit because the `view.Writer` expects that the whole range is in the same container. Closes #830.

### DIFF
--- a/src/view/range.js
+++ b/src/view/range.js
@@ -98,8 +98,8 @@ export default class Range {
 	 * @returns {module:engine/view/range~Range} Enlarged range.
 	 */
 	getEnlarged() {
-		const start = this.start.getLastMatchingPosition( enlargeShrinkStartSkip, { direction: 'backward' } );
-		const end = this.end.getLastMatchingPosition( enlargeShrinkEndSkip );
+		const start = this.start.getLastMatchingPosition( enlargeShrinkSkip, { direction: 'backward' } );
+		const end = this.end.getLastMatchingPosition( enlargeShrinkSkip );
 
 		return new Range( start, end );
 	}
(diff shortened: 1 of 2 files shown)


Reinmar:
Isn't it a bug fix? The commit message is confusing.

Also, I improved the message because you haven't used `()` after method names.

scofalik:
I guess this is not really a bug fix, it's more an improvement.

Reinmar:
Even the message says that it fixes something – that something was expected but the code worked differently. So it is a bug fix. The issue label also indicates that.

Also, please do think about the other developers when writing these messages. This one doesn't explain what kind of issue was fixed. It says that some values are passed somewhere. What does it mean? I have no idea myself.

The task title is much better in describing this: "view.Range.enlarge is not compatible with view.Writer.wrap". This is some valuable information. If I had a bug in the past with `enlarge()` and `wrap()` I would now have an idea that this is a fix for it.

Of course, during the course of action it might've turned out that the issue is wider or a bit different, so then the description needs to be generalised. But still, **it should bring explanation, a parsed information, not a raw description of changes.**

pjasiun:
Here I was thinking about something different. This changes touch code which was introduced recently. We need to way to "amend" the recent change. In fact, in this case, the original post did not tell now new method behave, so what user get are new methods with already changed behavior. But it may happen that we need to change something we recently introduced to the changelog. We may need something like "Amend", what need to be reviewed manually before the release to have only the final change in the changelog.

Reinmar:
Good point. There should be no changelog entry for this change then, because it doesn't change the previous "Feature: Introduced ble ble". Hence, "Internal" seems to be the best option.

PR #836, opened 2017-02-23
Fixed: Wrong usage of `is()` method.

### LINKED ISSUE #835: Renderer updates removed `AttributeElement` and throws error.
0. Open console :)
1. Open http://localhost:8125/ckeditor5-basic-styles/tests/manual/basic-styles.html
2. Press enter at the end of first sentence to create new paragraph.
3. Press ctrl + b to activate bold.
4. Press any letter key.

Editor will crash with:

```
Uncaught TypeError: Cannot read property 'is' of null
    at AttributeElement.getFillerOffset
    at DomConverter.viewChildrenToDom
    at viewChildrenToDom.next
    at Function.from
    at Renderer._updateChildren
    at Renderer.render
    at Document.on
    at Document.fire
    at Document.render
    at Document.EditingController._listener.listenTo
```

The error is in `getFillerOffset` of `AttributeElement`. The function uses `this.parent` without checking if the parent exists. When tracked down, though, it appears that the element is marked to be synced and `Renderer` tries to update it.

The question is why removed element is marked to be synced? Was it marked first and then removed? Do we expect that this kind of situation may happen? Or is it caused by some other bug?

If we expect such situation to happen I guess that we should filter out removed elements from marked elements.

scofalik:

Internal: Used `element instanceof` instead of `element.is()` in `AttributeElement#getFillerOffset` because it is expected that `element` is `null`. Closes #835.

---

### Additional information

Internal because it reverts a change done in one of earlier PRs.

### DIFF
--- a/src/view/attributeelement.js
+++ b/src/view/attributeelement.js
@@ -110,7 +110,7 @@ function getFillerOffset() {
 	let element = this.parent;
 
 	// <p><b></b></p> needs filler -> <p><b><br></b></p>
-	while ( element.is( 'attributeElement' ) ) {
+	while ( element && element.is( 'attributeElement' ) ) {
 		if ( element.childCount > 1 ) {
 			return null;
 		}
--- a/tests/view/attributeelement.js
+++ b/tests/view/attributeelement.js
@@ -131,5 +131,11 @@ describe( 'AttributeElement', () => {
 
 			expect( attribute.getFillerOffset() ).to.be.null;
 		} );
+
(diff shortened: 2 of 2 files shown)


Reinmar:
To avoid this error in the future please use `is()` and null check.

Also, a test's missing.

scofalik:
I didn't want to include test because it was just reverting of old change but the test is easy to add, so I added it as you suggested.

PR #838, opened 2017-02-23
Conversion - get rid of `move` and `rename` events. 

### LINKED ISSUE #837: Conversion - get rid of `move` and `rename` events.
Refactor `ModelConversionDispatcher` so it does not fire `move` and `rename` events. Instead, move and rename conversion (for `MoveOperation` and `RenameOperation`) should fire `remove` + `insert` events.

This is a first step to [simplify conversion](https://github.com/ckeditor/ckeditor5-engine/issues/829).

scofalik:

Other: `ModelConversionDispatcher` now uses `remove` + `insert` events to convert `move` and `rename` changes, instead of dedicated `move` and `rename` events. Closes #837.

BREAKING CHANGES: `ModelConversionDispatcher` no longer fires `move` and `rename` events. This means that feature converters added as callbacks to those should be replaced by `remove` and `insert` converters.

### DIFF
--- a/src/conversion/modelconversiondispatcher.js
+++ b/src/conversion/modelconversiondispatcher.js
@@ -200,9 +200,7 @@ export default class ModelConversionDispatcher {
 				range: itemRange
 			};
 
-			const name = item.name || '$text';
-
-			this._testAndFire( 'insert', consumable, data, name );
+			this._testAndFire( 'insert', data, consumable );
 
 			// Fire a separate addAttribute event for each attribute that was set on inserted items.
 			// This is important because most attributes converters will listen only to add/change/removeAttribute events.
@@ -212,7 +210,7 @@ export default class ModelConversionDispatcher {
 				data.attributeOldValue = null;
 				data.attributeNewValue = item.getAttribute( key );
 
-				this._testAndFire( `addAttribute:${ key }`, consumable, data, name );
+				this._testAndFire( `addAttribute:${ key }`, data, consumable );
 			}
 		}
 
@@ -257,9 +255,8 @@ export default class ModelConversionDispatcher {
 	 * @param {module:engine/model/position~Position} sourcePosition Position from where the range has been removed.
 	 * @param {module:engine/model/range~Range} range Removed range (after remove, in
 	 * {@link module:engine/model/document~Document#graveyard graveyard root}).
-	 * @param {String} removeAs If passed, removed item is treated like it was an element of given name.
 	 */
-	convertRemove( sourcePosition, range, removeAs = null ) {
+	convertRemove( sourcePosition, range ) {
 		const consumable = this._createConsumableForRange( range, 'remove' );
 
 		for ( let item of range.getItems( { shallow: true } ) ) {
@@ -268,9 +265,7 @@ export default class ModelConversionDispatcher {
 				item: item
 			};
 
-			const name = removeAs || item.name || '$text';
-
-			this._testAndFire( `remove`, consumable, data, name );
+			this._testAndFire( `remove`, data, consumable );
 		}
 	}
 
@@ -304,9 +299,7 @@ export default class ModelConversionDispatcher {
 				attributeNewValue: newValue
 			};
 
-			const name = item.name || '$text';
-
-			this._testAndFire( `${ type }:${ key }`, consumable, data, name );
+			this._testAndFire( `${ type }:${ key }`, data, consumable );
 		}
 	}
 
@@ -321,11 +314,20 @@ export default class ModelConversionDispatcher {
 	 * @param {String} oldName Name of the renamed element before it was renamed.
 	 */
 	convertRename( element, oldName ) {
-		const sourcePosition = Position.createBefore( element );
-		const range = Range.createOn( element );
+		const fakeElement = element.clone( true );
+		fakeElement.name = oldName;
 
-		this.convertRemove( sourcePosition, range, oldName );
-		this.convertInsertion( range );
+		const viewElement = this.conversionApi.mapper.toViewElement( element );
+
+		this.conversionApi.mapper.bindElements( fakeElement, viewElement );
+
+		element.parent.insertChildren( element.index, fakeElement );
+
+		this.convertRemove( Position.createBefore( fakeElement ), Range.createOn( fakeElement ) );
+
+		fakeElement.remove();
+
+		this.convertInsertion( Range.createOn( element ) );
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


scofalik:
There will be related PR in `ckeditor5-list`.

src/conversion/modelconversiondispatcher.js
```diff
@@ -198,50 +210,40 @@ export default class ModelConversionDispatcher {
...
 
-			this._testAndFire( 'move', data, consumable );
+			// Check if inserted content contains a marker.
+			if ( range.containsRange( markerRange ) || range.isEqual( markerRange ) ) {
+				this.convertMarker( 'addMarker', marker.name, markerRange );
+			}
 		}
 	}
 
+	convertMove( sourcePosition, range ) {
```
Reinmar: Lacks docs.

scofalik:
I've re-pushed the branch. I've squashed two commits with similar name and removed commit that is on different branch (different PR).

Reinmar:
I've got this error when running message-quote tests:

image

It goes through:

```js
	convertMove( sourcePosition, range ) {
		this.convertRemove( sourcePosition, range );
		this.convertInsertion( range );
	}
```

and I thought that there may be a bug in this function, but now I realised that I don't even now how it all works :D We first remove the range... and insert the same range in the next step?

Reinmar:
The code I'm testing is here: https://github.com/ckeditor/ckeditor5-message-quote/tree/t/1.

src/conversion/modelconversiondispatcher.js
```diff
@@ -188,7 +198,9 @@ export default class ModelConversionDispatcher {
 				range: itemRange
 			};
 
-			this._testAndFire( 'insert', data, consumable );
+			const name = item.name || '$text';
+
+			this._testAndFire( 'insert', consumable, data, name );
```
pjasiun: Why name is not part of the data?

pjasiun: In fact the whole idea of the `name` parameter, when you have `item` is strange for me.

src/conversion/modelconversiondispatcher.js
```diff
@@ -250,7 +252,9 @@ export default class ModelConversionDispatcher {
 				item: item
 			};
 
-			this._testAndFire( 'remove', data, consumable );
+			const name = removeAs || item.name || '$text';
+
+			this._testAndFire( `remove`, consumable, data, name );
```
pjasiun: I don't like the fact that you fire `data.item`, which have the name, and  `name` which usually is the same, but developer need to know that in the case of rename it can be different. It does simplify nothing because developer still needs to know about rename and now it's hidden inside `remove`.

Since the `item` in `remove` is not in the document anyway (it's in the graveyard, so have no context), can't we pass a clone of the element with the old name in case of rename as item?

src/conversion/modelconversiondispatcher.js
scofalik: This is like this to support `removeAs` parameter in remove conversion (for rename purposes).

src/conversion/modelconversiondispatcher.js
scofalik: If we go with your solution, the "fake" old item has to be bind with original element in view (so mapper correctly maps ranges in remove converter). It was like this in old implementation, when "rename converter" was in `model-to-view-converters.js`. I will check how complicated the code will end up with this solution.

scofalik:
> I realised that I don't even now how it all works :D We first remove the range... and insert the same range in the next step?

First, you start remove-change conversion, stating that given range was removed in model and now that change should be represented in view.

Then, you start insert-change conversion, stating that given range was inserted in model and now that change should be represented in view.

This is just representing model changes in view and we use model just for data source. Nothing changes in model so the function should be fine.

scofalik:
I've updated docs and pushed alternative solution for rename conversion: https://github.com/ckeditor/ckeditor5-engine/pull/838/files#diff-2713f31db8799d747fab1e9c607e4a55

I don't like it because it changes model during conversion and this is hacking it, because all that MCD should do is send events with proper data and it doesn't feel right that model should be changed to do so. Also we are using `this.conversionApi` directly in MCD which we haven't done yet and we make a requirement to always include mapper in `conversionApi`. I know this is not-really-a-requirement because we always provide mapper anyway.

Other solution could be to do:
```javascript
convertRename( element, oldName ) {
	const sourcePosition = Position.createBefore( element );
	const range = Range.createOn( element );

	const newName = element.name;
	element.name = oldName;

	this.convertRemove( sourcePosition, range );

	element.name = newName;

	this.convertInsertion( range );
}
```

But I don't like it either, as once again we are changing model. Actually, we temporarily revert what original operation did, just for conversion purposes. Once again this does not feel right, but is at least cleaner.

So we could:
1. Use the first solution I proposed (with `removeAs`).
2. Go with the solution I just pushed.
3. Go with the solution I proposed here.
4. Try to go deeper, looking for a perfect solution.

You decide.

pjasiun:
Well, to be honest, I don't like any of these proposals. All of them looks more complicated than what we have now. Maybe, we could implement "RenameDelta" as a pair of remove and insert operations, but then OT will be more complicated. If we have nothing really simpler, I'm against this refactoring (nothing surprising).

scofalik:
But we can't. We did it like this because we need to keep the reference to the original node. Unless we introduce it as `RemoveOperation` + `RenameOperation` (in GY, not converted at all) + `ReinsertOperation`.

This is an interesting idea, honestly :D.

Reinmar:
Let me underline one thing which should not be forgotten – MDC's complexity is not a problem as long as conversion events are clear and useful.

For the rename conversion to be clear, the `remove` event must carry same data as it does in the case of normal remove. We agreed that if we're renaming an element the event could carry a fake element. I just checked that in case of a normal `remove` the parent of this element is always graveyard holder, which AFAIU means that it could also be `null` because it's not an important info. When you're converting remove operation you shouldn't expect to be able to check the model item's parent anyway because this item was already removed.

Hence, I don't understand the hacking that you needed to introduce @scofalik. Why inserting the fake element anyway?

pjasiun:
> Hence, I don't understand the hacking that you needed to introduce @scofalik. Why inserting the fake element anyway?

In fact, it's not clear for me too.

scofalik:
You are right, the solution could be simplified by not inserting the fake item into tree. If you are fine with that solution after fixing this, we can move on with it.

scofalik:
Okay, I pushed proposed changes.

Reinmar:
This PR fixes broken conversion in message quote. I wann it!

PR #839, opened 2017-02-23
t/833: Nothing should happen if creator return null in convertSelectionMarker

### LINKED ISSUE #833: Nothing should happen if creator return null in convertSelectionMarker
Now, there is an error if creator return `null` in `convertSelectionMarker`.

pjasiun:

Fix: You can now return `null` from the element creators in converters for selection attributes and markers. It does not crash the conversion anymore. Closes #833.

---

### Additional information


Reinmar:
Was:

> Fix: You can return null from the elements creators in selection attributes and markers converters, it do not crash conversion anymore. Closes #833.

Is:

> Fix: You can now return `null` from the element creators in converters for selection attributes and markers. It does not crash the conversion anymore. Closes #833.

PR #845, opened 2017-02-26
Introduced markers serialization

### LINKED ISSUE #787: Markers: data convertors should create simpler structures
Markers should be converted differently in the data pipeline.

Right now, they're converted in a way that satisfies the view pipeline purpose by creating spans with content inside. For example, if a marker spans across several paragraphs, this is a (simplified) output of the DOM:

```
<p><span marker1>...</span></p>
<p><span marker1>...</span></p>
<p><span marker1>...</span></p>
```

When it comes to the data pipeline instead, outputting markers like above is semantically wrong and senseless as it unnecessarily bloats up the data.

Markers are simply ranges within the document with no visual representation (which can be added on top of markers). Therefore it's data representation should be much simpler, just describing the start and end boundaries of the range.

For example, for the above case, the following data output makes much more sense:

```
<p><span marker1-start />...</p>
<p>...</p>
<p>...<span marker1-end /></p>
```

### LINKED ISSUE #846: Create diffrent opening and closing element in marker conversion
Follow-up: https://github.com/ckeditor/ckeditor5-engine/issues/787

See: https://github.com/ckeditor/ckeditor5-engine/pull/845#discussion_r103220919

oskarwrobel:

Feature: Introduced markers serialization. Closes #787. Closes #846.

BREAKING CHANGES: `BuildModelConverter#fromMarkerCollapsed` is removed. Use `BuildModelConverter#fromMarker` instead. Format of data returned by `ViewConversionDispatcher#convert` has changed from single model item to object with model item and list of static markers.

NOTE: `insertUIElement` model to view converter now supports collapsed and non-collapsed ranges.

---

Related to: https://github.com/ckeditor/ckeditor5-clipboard/pull/15

Follow-ups:
- Fixed https://github.com/ckeditor/ckeditor5-engine/issues/846 - Create diffrent opening and closing element in marker conversion

### DIFF
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -234,7 +234,7 @@ export default class DataController {
 	 * Element or document fragment which content will be converted.
 	 * @param {String} [context='$root'] Base context in which the view will be converted to the model. See:
 	 * {@link module:engine/conversion/viewconversiondispatcher~ViewConversionDispatcher#convert}.
-	 * @returns {module:engine/model/documentfragment~DocumentFragment} View item converted to Model DocumentFragment.
+	 * @returns {module:engine/model/documentfragment~DocumentFragment} Output document fragment.
 	 */
 	toModel( viewElementOrFragment, context = '$root' ) {
 		const { conversionResult, markers } = this.viewToModel.convert( viewElementOrFragment, { context: [ context ] } );
@@ -246,10 +246,8 @@ export default class DataController {
 			documentFragment = conversionResult;
 		}
 
-		// Set markers to the documentFragment.
-		for ( const marker of markers ) {
-			documentFragment.markers.set( marker[ 0 ], marker[ 1 ] );
-		}
+		// Set converted markers to the documentFragment.
+		documentFragment.markers = markers;
 
 		return documentFragment;
 	}
@@ -320,15 +318,6 @@ export default class DataController {
 }
 
 mix( DataController, EmitterMixin );
-
-/**
- * Result of parsing view to model.
- *
- * @typedef {Object} {module:engine/controller/datacontroller~ParsedModelData}
- * @property {module:engine/model/documentfragment~DocumentFragment|module:engine/model/node~Node} modelItem
- * Fragment of parsed model.
- * @property {Map<String, module:engine/model/range~Range>} markers List of parsed marker stamps.
- */
 
 /**
  * Event fired when {@link #insertContent} method is called.
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -205,7 +205,8 @@ export default class DataController {
 			this.model.selection.clearAttributes();
 
 			// Parse data to model and extract markers from parsed document fragment.
-			const { documentFragment, markersData } = extractMarkersFromModelFragment( this.parse( data ) );
+			const documentFragment = this.parse( data );
+			const markersData = extractMarkersFromModelFragment( documentFragment );
 
 			// Initial batch should be ignored by features like undo, etc.
 			const batch = this.model.batch( 'transparent' );
@@ -282,8 +283,8 @@ export default class DataController {
 	 * changes will be added to a new batch.
 	 */
 	insertContent( content, selection, batch ) {
-		const { documentFragment } = extractMarkersFromModelFragment( content );
-		this.fire( 'insertContent', { content: documentFragment, selection, batch } );
+		extractMarkersFromModelFragment( content );
+		this.fire( 'insertContent', { content, selection, batch } );
 	}
 
 	/**
@@ -335,55 +336,47 @@ export default class DataController {
 
 mix( DataController, EmitterMixin );
 
-// Traverses given DocumentFragment and searches elements which marks marker range. Founded element is removed from
+// Traverses given DocumentFragment and searches elements which marks marker range. Found element is removed from
 // DocumentFragment but path of this element is stored in Map.
 //
 // @param {module:engine/view/documentfragment~DocumentFragment} documentFragment Model DocumentFragment.
-// @returns {Object} Object with markers data and cleaned up document fragment.
+// @returns {Map} Map with markers data.
 function extractMarkersFromModelFragment( documentFragment ) {
-	const markersData = new Map();
+	const markerStamps = new Set();
 
-	// Creates ModelTreeWalker with given start position.
-	function walkFrom( position ) {
-		const walker = new ModelTreeWalker( {
-			startPosition: position,
-			ignoreElementEnd: true,
-			shallow: false
-		} );
+	// Create ModelTreeWalker.
+	const walker = new ModelTreeWalker( {
+		startPosition: ModelPosition.createAt( documentFragment, 0 ),
+		ignoreElementEnd: true,
+		shallow: false
+	} );
 
-		// Walk through DocumentFragment.
-		for ( const value of walker ) {
-			// Check if current element is a marker stamp.
-			if ( value.item.name == '$marker' ) {
-				const markerName = value.item.getAttribute( 'data-name' );
-				const currentPosition = ModelPosition.createBefore( value.item );
-
-				// When marker of given name is not stored it means that we have found the beginning of the range.
-				if ( !markersData.has( markerName ) ) {
-					markersData.set( markerName, { startPath: currentPosition.path } );
-				// Otherwise is means that we have found end of the marker range.
-				} else {
-					markersData.get( markerName ).endPath = currentPosition.path;
-				}
-
-				// Remove marker stamp element from DocumentFragment.
-				remove( ModelRange.createOn( value.item ) );
-
-				// Keep walking using new instance of TreeWalker but starting from last visited position.
-				// This is because after removing marker stamp element DocumentFragment structure might change
-				// and TreeWalker might omit some node.
-				walkFrom( currentPosition );
-
-				// Stop this walker, we have continued walk using new TreeWalker instance.
-				break;
-			}
+	// Walk through DocumentFragment and collect marker elements.
+	for ( const value of walker ) {
+		// Check if current element is a marker stamp.
+		if ( value.item.name == '$marker' ) {
+			markerStamps.add( value.item );
 		}
 	}
 
-	// Start traversing.
-	walkFrom( ModelPosition.createAt( documentFragment, 0 ) );
+	// Walk through collected marker elements store its path and remove its from the DocumentFragment.
+	return Array.from( markerStamps ).reduce( ( result, stamp ) => {
+		const markerName = stamp.getAttribute( 'data-name' );
+		const currentPosition = ModelPosition.createBefore( stamp );
 
-	return { markersData, documentFragment };
+		// When marker of given name is not stored it means that we have found the beginning of the range.
+		if ( !result.has( markerName ) ) {
+			result.set( markerName, { startPath: currentPosition.path } );
+		// Otherwise is means that we have found end of the marker range.
+		} else {
+			result.get( markerName ).endPath = currentPosition.path;
+		}
+
+		// Remove marker stamp element from DocumentFragment.
+		remove( ModelRange.createOn( stamp ) );
+
+		return result;
+	}, new Map() );
 }
 
 /**
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -205,7 +205,7 @@ export default class DataController {
 			this.model.selection.clearAttributes();
 
 			// Parse data to model and extract markers from parsed document fragment.
-			const { documentFragment, markersData } = extractMarkersDataFromModelElement( this.parse( data ) );
+			const { documentFragment, markersData } = extractMarkersFromModelFragment( this.parse( data ) );
 
 			// Initial batch should be ignored by features like undo, etc.
 			const batch = this.model.batch( 'transparent' );
@@ -282,7 +282,7 @@ export default class DataController {
 	 * changes will be added to a new batch.
 	 */
 	insertContent( content, selection, batch ) {
-		const { documentFragment } = extractMarkersDataFromModelElement( content );
+		const { documentFragment } = extractMarkersFromModelFragment( content );
 		this.fire( 'insertContent', { content: documentFragment, selection, batch } );
 	}
 
@@ -340,7 +340,7 @@ mix( DataController, EmitterMixin );
 //
 // @param {module:engine/view/documentfragment~DocumentFragment} documentFragment Model DocumentFragment.
 // @returns {Object} Object with markers data and cleaned up document fragment.
-function extractMarkersDataFromModelElement( documentFragment ) {
+function extractMarkersFromModelFragment( documentFragment ) {
 	const markersData = new Map();
 
 	// Creates ModelTreeWalker with given start position.
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -22,11 +22,14 @@ import ViewDocumentFragment from '../view/documentfragment';
 
 import ModelRange from '../model/range';
 import ModelPosition from '../model/position';
+import ModelTreeWalker from '../model/treewalker';
 
 import insertContent from './insertcontent';
 import deleteContent from './deletecontent';
 import modifySelection from './modifyselection';
 import getSelectedContent from './getselectedcontent';
+
+import { remove } from '@ckeditor/ckeditor5-engine/src/model/writer';
 
 /**
  * Controller for the data pipeline. The data pipeline controls how data is retrieved from the document
@@ -180,7 +183,10 @@ export default class DataController {
 
 	/**
 	 * Sets input data parsed by the {@link #processor data processor} and
-	 * converted by the {@link #viewToModel view to model converters}.
+	 * converted by the {@link #viewToModel view to model converters}. When markers where converted
+	 * from view to model as temporary {@link module:engine/model/element/~Element model elements} then those element
+	 * will be removed from parsed {@link module:engine/model/element/~DocumentFragment} and added to the
+	 * {@link module:engine/model/document~Document#markers markers collection}.
 	 *
 	 * This method also creates a batch with all the changes applied. If all you need is to parse data use
 	 * the {@link #parse} method.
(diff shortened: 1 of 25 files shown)


src/controller/datacontroller.js
```diff
@@ -180,7 +183,10 @@ export default class DataController {
 
 	/**
 	 * Sets input data parsed by the {@link #processor data processor} and
-	 * converted by the {@link #viewToModel view to model converters}.
+	 * converted by the {@link #viewToModel view to model converters}. When markers where converted
```
scofalik: If

src/controller/datacontroller.js
```diff
@@ -180,7 +183,10 @@ export default class DataController {
...
+	 * from view to model as temporary {@link module:engine/model/element/~Element model elements} then those element
```
scofalik: elements

src/controller/datacontroller.js
```diff
@@ -198,10 +204,28 @@ export default class DataController {
...
+			// Add markers to the document.
+			for ( const marker of markersData ) {
+				const markerName = marker[ 0 ];
+				const markerData = marker[ 1 ];
+				const range = new ModelRange(
+					new ModelPosition( modelRoot, markerData.startPath ),
+					markerData.endPath ? new ModelPosition( modelRoot, markerData.endPath ) : null
+				);
+
+				batch.setMarker( this.model.markers.set( markerName, range ) );
```
scofalik: AFAIR you can directly use `setMarker` - no need to add that marker to `MarkerCollection` first. `MarkerOperation`, upon execution, will do it.

src/controller/datacontroller.js
```diff
@@ -306,6 +335,57 @@ export default class DataController {
 
 mix( DataController, EmitterMixin );
 
+// Traverses given DocumentFragment and searches elements which marks marker range. Founded element is removed from
```
scofalik: Found

src/controller/datacontroller.js
```diff
@@ -306,6 +335,57 @@ export default class DataController {
...
+// DocumentFragment but path of this element is stored in Map.
```
scofalik: in a `Map`.

src/controller/datacontroller.js
```diff
@@ -306,6 +335,57 @@ export default class DataController {
...
+//
+// @param {module:engine/view/documentfragment~DocumentFragment} documentFragment Model DocumentFragment.
+// @returns {Object} Object with markers data and cleaned up document fragment.
+function extractMarkersFromModelFragment( documentFragment ) {
```
scofalik: To simplify this function and avoid refreshing walker, you could simply first find all elements to be extracted, and then remove them one by one after tree walker already did it's job:

```javascript
const markerStamps = new Set();
// ...
for ( const value of walker ) {
// ...
markerStamps.add( value.item )
// ...
}

for ( const stamp of markerStamps ) {
// ...
}
```

AFAIR, `Set` remembers the order in which items where stored.

scofalik: Also, since `documentFragment` is passed by reference, you only need to return the markers. It will be more elegant to use.

scofalik: Finally, this could be a private method too, but I also have this "problem" of using sometimes private "static-like" methods and sometimes not exported functions, so as you wish :)

oskarwrobel: > Also, since documentFragment is passed by reference, you only need to return the markers. It will be more elegant to use.

That's how it looked in my first implementation but I assumed that it'll be more readable to return data.

src/conversion/model-to-view-converters.js
```diff
@@ -115,9 +117,13 @@ export function insertUIElement( elementCreator ) {
...
 		}
 
-		const viewPosition = conversionApi.mapper.toViewPosition( data.range.start );
+		const mapper = conversionApi.mapper;
 
-		viewWriter.insert( viewPosition, viewElement );
+		viewWriter.insert( mapper.toViewPosition( data.range.start ), viewElement );
+
+		if ( !data.range.isCollapsed ) {
+			viewWriter.insert( mapper.toViewPosition( data.range.end ), viewElement.clone( true ) );
```
scofalik: We've been discussing that it would be nice if closing element is different than opening element, so this information would have to be send to creator and two view elements would have to be created.

If that is not too complicated, it would be nice to have this, but it is not a strict requirement. Take both model->view and view->model conversion into consideration.

scofalik:
Changed proposed commit message from

NOTE: BuildModelConverter#fromMarker and BuildModelConverter#fromCollapsedMarker are unified, insertUIElement method of model -> view converter is enhanced to supports collapsed and non-collapsed ranges.

to

BREAKING CHANGES: `BuildModelConverter#fromMarker` and `#fromCollapsedMarker` are unified, and the latter is removed.
NOTE: `insertUIElement` model->view converter now supports collapsed and non-collapsed ranges.

Reinmar:
I added missing empty line between both blocks :D

src/conversion/model-to-view-converters.js
oskarwrobel: FollowUp: https://github.com/ckeditor/ckeditor5-engine/issues/846

src/controller/datacontroller.js
```diff
@@ -306,6 +336,49 @@ export default class DataController {
...
+	// Walk through DocumentFragment and collect marker elements.
+	for ( const value of walker ) {
+		// Check if current element is a marker stamp.
+		if ( value.item.name == '$marker' ) {
+			markerStamps.add( value.item );
+		}
+	}
+
+	// Walk through collected marker elements store its path and remove its from the DocumentFragment.
+	return Array.from( markerStamps ).reduce( ( result, stamp ) => {
```
pjasiun: Iterators FTW!

scofalik: Loops FTW!

On serious note, this looks overcomplicated. Simple loop over `markerStamps` would be okay too.

src/controller/datacontroller.js
```diff
@@ -180,7 +183,10 @@ export default class DataController {
 
 	/**
 	 * Sets input data parsed by the {@link #processor data processor} and
-	 * converted by the {@link #viewToModel view to model converters}.
+	 * converted by the {@link #viewToModel view to model converters}. If markers where converted
+	 * from view to model as temporary {@link module:engine/model/element/~Element model elements} then those elements
+	 * will be removed from parsed {@link module:engine/model/element/~DocumentFragment} and added to the
+	 * {@link module:engine/model/document~Document#markers markers collection}.
```
pjasiun: This should not be here. All what `DataController` is, is a facade for the complex API. I should not have any complex logic because quickly it became uber-class.

pjasiun: Especially since now `parse` does not return any useful data, and the whole processing is done outside this function.

src/controller/datacontroller.js
```diff
@@ -247,13 +272,18 @@ export default class DataController {
...
+	 * data pipeline allows to set markers only by a {@link #set set method}.
+	 *
 	 * @fires insertContent
 	 * @param {module:engine/model/documentfragment~DocumentFragment} content The content to insert.
 	 * @param {module:engine/model/selection~Selection} selection Selection into which the content should be inserted.
 	 * @param {module:engine/model/batch~Batch} [batch] Batch to which deltas will be added. If not specified, then
 	 * changes will be added to a new batch.
 	 */
 	insertContent( content, selection, batch ) {
+		extractMarkersFromModelFragment( content );
```
pjasiun: Do we need to extract anything on insert content?

src/controller/datacontroller.js
Reinmar: For var i = 0; i < length; ++i FTW!

Reinmar:
This branch has conflicts! Git FTW!

src/controller/datacontroller.js
Reinmar: Why is it here? Any clean up should be done beforehand, using the clipboard events (`#paste` or `#clipboardInput`).

src/controller/datacontroller.js
pjasiun: Anyway, the whole function should be moved from the `DataController` into conversion view dispatcher or view converter helper.

oskarwrobel: I was really curious what will you say guys about this `reduce` :)

src/controller/datacontroller.js
```diff
@@ -198,10 +198,28 @@ export default class DataController {
 			this.model.selection.removeAllRanges();
 			this.model.selection.clearAttributes();
 
+			// Parse data to model and extract markers from parsed document fragment.
+			const { modelItem, markersData } = this.parse( data );
```
pjasiun: I was thinking a while about the return format. I don't like the code bellow needed to translate `markersData` to `document.makers`. It's not short nor simple, and you will have to write it every time you need a custom handling of parsed data. Of course, we could have a helper function, but where-ever we put such function it will be ugly. Then I thought that maybe markers should be returned as `MarkersCollection`.

At the same time, it's strange that you can not return one object representing a parsed fragment of the document, something like... `DocumentFragment`?

Markers are integrated part of the document data. They are not like selection or history, they are more like attributes or children, so maybe we should have `DocumentFragment.makers`? Then, when you add document fragment to another document, makers from it should be taken and added to the parent document, the same way like children are moved to the parent element. It could be done by the `model.writer` or `element.insert` (I hope soon we will get rid off `model.writer` anyway).

pjasiun: After this change, the code here will not change much.

pjasiun: I know I am usually the one who is against magic, but the possibility of having all data in one object is worth trying.

oskarwrobel: I like this idea. Converter still will be returning `Object` with conversion result and markers data, but `parser()` will merge it together into `DocumentFragment` with its own `MarkersCollection` ?

src/conversion/buildviewconverter.js
```diff
@@ -362,6 +363,86 @@ class ViewConverterBuilder {
...
+	 * converter should look like this:
+	 *
+	 * 		buildViewConverter().for( dispatcher ).from( { name: 'span', { attribute: foo: /^\w/ } } ).toMarker( ( data ) => {
+	 * 			return new Element( '$marker', { 'data-name': data.getAttribute( 'foo' ) } );
+	 * 		} );
+	 *
+	 * @param {Function} [creator] Creator function.
+	 */
+	toMarker( creator ) {
+		const eventCallbackGen = function( from ) {
```
pjasiun: Why not `function eventCallbackGen( from )`?

src/conversion/viewconversiondispatcher.js
```diff
@@ -221,6 +233,62 @@ export default class ViewConversionDispatcher {
...
+
+	if ( modelItem.is( 'text' ) ) {
+		return markersData;
+	}
+
+	// Create ModelTreeWalker.
+	const walker = new ModelTreeWalker( {
+		startPosition: ModelPosition.createAt( modelItem, 0 ),
+		ignoreElementEnd: true,
+		shallow: false
```
pjasiun: AFAIR shallow is false by default.

src/conversion/buildviewconverter.js
oskarwrobel: I'm also for named function instead of anonymous function assigned to const, but this is consistent with the other methods. I'll fix it.

src/controller/datacontroller.js
oskarwrobel: Creating the real `MarkersCollection` on `DocumentFragment` might be more compilated because `Marker` keeps `LiveRange` inside but `LiveRange` work only for `Document`. We can:
- try to improve `DocumentFragment` and `LiveRange` to work each other but it might be overengineering. 
- don't create the real `MarkersCollection` on `DocumentFragment`but only something to store markers data. I don't like it because every change on DocFrag will make markers positions out of date.

src/controller/datacontroller.js
```diff
@@ -236,7 +237,19 @@ export default class DataController {
 	 * @returns {module:engine/model/documentfragment~DocumentFragment} Output document fragment.
 	 */
 	toModel( viewElementOrFragment, context = '$root' ) {
-		return this.viewToModel.convert( viewElementOrFragment, { context: [ context ] } );
+		const { conversionResult, markers } = this.viewToModel.convert( viewElementOrFragment, { context: [ context ] } );
```
pjasiun: I think `convert` should simply return `documentFragmet` with markers which can be added instead of transforming it here. `DataController` should not contain any transformation logic, it should be as simple as possible.

src/conversion/buildmodelconverter.js
```diff
@@ -265,22 +244,76 @@ class ModelConverterBuilder {
...
+	 *			.fromMarker( 'search' )
+	 *			.toStamp( ( data ) => new UIElement( 'span', { 'data-name': data.marker.getName() ) );
+	 *
+	 * Creator function provides parameter `isOpening` which defined if currently converted element is a beginning or end of
+	 * the marker range. This makes possible to create different opening and closing stamp.
+	 *
+	 *		buildModelConverter().for( dispatcher )
+	 *			.fromMarker( 'search' )
+	 *			.toStamp( ( data, isOpening ) => {
+	 *				if ( isOpening ) {
```
pjasiun: `isOpening` should be a part of the data (`data.isOpening`).

src/conversion/buildmodelconverter.js
```diff
@@ -265,22 +244,76 @@ class ModelConverterBuilder {
...
+	 *
+	 * See how markers {module:engine/model/buildviewconverter~ViewConverterBuilder#toMarker view -> model serialization}
+	 * works to find out what view element format is the best for you.
+	 *
+	 * @param {String|module:engine/view/element~UIElement|Function} element UIElement created by converter or
+	 * a function that returns view element.
+	 */
+	toStamp( element ) {
+		for ( let dispatcher of this._dispatchers ) {
+			if ( this._from.type == 'element' || this._from.type == 'attribute' ) {
```
pjasiun: Why not `if ( this._from.type != 'marker' )`? If we will introduce a new type it most probably should not be transformed to stamp as well.

src/model/writer.js
```diff
@@ -63,12 +70,27 @@ export function insert( position, nodes ) {
...
+			// We need to transfer its markers and update position markers positions.
+			for ( const marker of nodes.markers ) {
+				const range = new Range( new Position( parent, marker[ 1 ].start.path ),  new Position( parent, marker[ 1 ].end.path ) );
+				position.root.document.markers.set( marker[ 0 ], range );
+			}
+		// Otherwise we need to show warning about losing markers.
+		} else {
+			log.warn( 'model-writer-insert-lose-markers: Element containing markers is set to element without MarkersCollection.' );
+		}
+	}
```
pjasiun: Will it work fine if I insert `DocumentFragment` with markers to another `DocumentFragment`?

oskarwrobel: I did a small fix and wrote test for it and it works.

src/conversion/buildmodelconverter.js
oskarwrobel: Done

src/conversion/buildmodelconverter.js
oskarwrobel: Done.

src/controller/datacontroller.js
oskarwrobel: Done.

src/conversion/viewconversiondispatcher.js
```diff
@@ -128,14 +134,31 @@ export default class ViewConversionDispatcher {
...
 		this.fire( 'viewCleanup', viewItem );
 
 		const consumable = ViewConsumable.createFrom( viewItem );
+		const conversionResult = this._convertItem( viewItem, consumable, additionalData );
 
-		return this._convertItem( viewItem, consumable, additionalData );
+		// When conversion output is not a Node or Element we just return it.
+		if ( !( conversionResult instanceof ModelNode || conversionResult instanceof ModelDocumentFragment ) ) {
+			return conversionResult;
+		}
```
pjasiun: I don't get this condition. Is it for text nodes? I think that there is nothing wrong if we will wrap everything in the document fragment.

src/conversion/viewconversiondispatcher.js
```diff
@@ -128,14 +134,31 @@ export default class ViewConversionDispatcher {
...
+	 * wrapped by DocumentFragment.
 	 */
 	convert( viewItem, additionalData = {} ) {
 		this.fire( 'viewCleanup', viewItem );
 
 		const consumable = ViewConsumable.createFrom( viewItem );
+		const conversionResult = this._convertItem( viewItem, consumable, additionalData );
 
-		return this._convertItem( viewItem, consumable, additionalData );
+		// When conversion output is not a Node or Element we just return it.
```
pjasiun: This comment is useless. It tells exactly what the next line do. I can read code instead. What comment should do is tell me why this code is needed.

src/controller/datacontroller.js
```diff
@@ -224,10 +224,10 @@ export default class DataController {
 	}
 
 	/**
-	 * Returns the content of the given {@link module:engine/view/element~Element view element} or
+	 * Returns wrapped by {module:engine/model/documentfragment~DocumentFragment} result of the given
+	 * {@link module:engine/view/element~Element view element} or
 	 * {@link module:engine/view/documentfragment~DocumentFragment view document fragment} converted by the
-	 * {@link #viewToModel view to model converters} to a
-	 * {@link module:engine/model/documentfragment~DocumentFragment model document fragment}.
+	 * {@link #viewToModel view to model converters}.
```
pjasiun: You could add a note to `toModel` and `viewToModel.convert` that the document fragment may contain markers.

src/model/documentfragment.js
```diff
@@ -25,6 +29,15 @@ export default class DocumentFragment {
 	 */
 	constructor( children ) {
 		/**
+		 * DocumentFragment static markers list. This is a list of names and {@link module:engine/model/range~Range ranges}
```
pjasiun: It's map, not list.

src/conversion/viewconversiondispatcher.js
oskarwrobel: This is because conversion output doesn't have to be a node. Theoretically it can be [any value](https://github.com/ckeditor/ckeditor5-engine/blob/master/tests/conversion/viewconversiondispatcher.js#L75) but we use it to provide converted element. @scofalik know the details :) I know that it might be confusing especially when there is no explanation in docs.

src/model/documentfragment.js
oskarwrobel: Fixed.

src/controller/datacontroller.js
oskarwrobel: Done.

src/conversion/viewconversiondispatcher.js
oskarwrobel: Fixed.

src/conversion/viewconversiondispatcher.js
pjasiun: Yea, I know. It was even my idea, but it was a very theoretical idea. You can remove these tests because at this point it's pretty clear that we will not use dispatcher this way, soon.

oskarwrobel: Follow-up: https://github.com/ckeditor/ckeditor5-engine/issues/860

PR #848, opened 2017-03-01
T/842

### LINKED ISSUE #842: view.TreeWalker changes position on constructor
It took me half of the day.

Because of this hack, https://github.com/ckeditor/ckeditor5-engine/blob/d0b99809d24a40cb017f777acb6a7997722e0768/src/view/treewalker.js#L133 tree walker position is moved when tree walker is created. At the same time boundaries stays. It means that if someone creates a collapsed range at the end of the text node and ask for values the position is moved after boundaries and tree walker is not limited, so it gets all values from the end of that range to the end of the document.

I think that moving the position when the developer creates walker is in general wrong. It was totally unexpected that position is not where I put it, even before I moved it. The hack should be moved to the `_next`/`_previous` methods which are expected to move the position.

scofalik:

Internal: Moved `view.TreeWalker` starting position fix from constructor to `#_next()` and `#_previous()` methods. Closes #842.

### DIFF
--- a/src/view/range.js
+++ b/src/view/range.js
@@ -98,8 +98,17 @@ export default class Range {
 	 * @returns {module:engine/view/range~Range} Enlarged range.
 	 */
 	getEnlarged() {
-		const start = this.start.getLastMatchingPosition( enlargeShrinkSkip, { direction: 'backward' } );
-		const end = this.end.getLastMatchingPosition( enlargeShrinkSkip );
+		let start = this.start.getLastMatchingPosition( enlargeShrinkSkip, { direction: 'backward' } );
+		let end = this.end.getLastMatchingPosition( enlargeShrinkSkip );
+
+		// Fix positions, in case if they are in Text node.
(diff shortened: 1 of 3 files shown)


Reinmar:
I guess it was mentioned as an internal change, even though it's fixing a bug, because that bug was introduced in this iteration, yes? Would be good to mention such situations in additional notes or something because I was confused.

Also, when merging the branch, the message was not copied.

Reinmar:
Anyway, the message was still rather poor (:P) because it repeated what the change was (which is obvious once you check the code), but not what the change was about (it fixed a bug).

scofalik:
I made it "Internal" because it changes (You won't guess) internals xD. That is, fix that was in private method used in constructor was moved to other method.

I wasn't sure if it is a fix or it is internal, but when I think of it now, I think it should be fix after all. It was fixing a bug with `TreeWalker` when it was set with collapsed boundary range in `Text` node.

PR #849, opened 2017-03-01
Fixed: wrapRange model-to-view converter should not convert if range is collapsed.

### LINKED ISSUE #832: Marker to AttributeElement converter should not call wrap, if the range is collapsed
If developer creates a converter from marker to the `AttributeElement` it should do nothing if the marker is collapsed, because there should not be an empty `AttributeElement`.

scofalik:

Fix: `wrapRange()` and `unwrapRange()` default model-to-view converters will not consume nor convert changes if the range to convert is collapsed. Closes #832.

### DIFF
--- a/src/conversion/model-to-view-converters.js
+++ b/src/conversion/model-to-view-converters.js
@@ -351,11 +351,16 @@ export function wrapRange( elementCreator ) {
 			return;
 		}
 
+		const viewRange = conversionApi.mapper.toViewRange( data.range );
+
+		if ( viewRange.isCollapsed ) {
+			return;
+		}
+
 		if ( !consumable.consume( data.range, 'addMarker' ) ) {
 			return;
 		}
 
-		const viewRange = conversionApi.mapper.toViewRange( data.range );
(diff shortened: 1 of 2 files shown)


src/conversion/model-to-view-converters.js
```diff
@@ -351,11 +351,16 @@ export function wrapRange( elementCreator ) {
 			return;
 		}
 
+		const viewRange = conversionApi.mapper.toViewRange( data.range );
+
+		if ( viewRange.isCollapsed ) {
+			return;
+		}
+
```
pjasiun: Yea, it was my first idea to fix it. But then I realized that nothing should happen if I do not return because https://github.com/ckeditor/ckeditor5-engine/blob/master/src/view/writer.js#L517-L519 This code for sure should be moved before `validateRangeContainer`, but when I did it new bugs appeared.

pjasiun: But now it can be fine after `TreeWalker` fixes.

pjasiun:
We decided not to introduce new code since we are not sure if we need it.

Reinmar:
Ohh... pity, cause the commit message was great this time :P

PR #851, opened 2017-03-02
Fixed view Matcher.getElementName to not return function name.

### LINKED ISSUE #850: View matcher.getElementName returns name of a function
`Matcher.getElementName` returns `this._patterns[ 0 ].name` and checks only if that name is not a `RegExp`. Pattern might be defined as a function. When function is a named one `name` property returns it's name. This was fun debugging :smile:

szymonkups:

Fix: `view.Matcher#getElementName()` now returns proper value when named function is used as a pattern. Closes #850.

### DIFF
--- a/src/view/matcher.js
+++ b/src/view/matcher.js
@@ -203,9 +203,14 @@ export default class Matcher {
 	 * @returns {String|null} Element name trying to match.
 	 */
 	getElementName() {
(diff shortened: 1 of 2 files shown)


scofalik:
Changed commit message, was:

> Fix: View Matcher.getElementName returns proper value when named function is used as a pattern. Closes #850.

PR #853, opened 2017-03-02
Changed: collapsed model.Range does not stick with it's neighbour node when that node is moved.

### LINKED ISSUE #852: model.Range move transformation incorrect for collapsed range
I noticed this bug when I was dealing with markers. This is what happens:

1. You have collapsed marker and selection at the same position: `<p>foo^[]bar</p>`.
2. You press backspace to remove a letter before selection.
3. Expected: `<p>fo^[]bar</p>`.

However, with current implementation, marker will stick with removed letter and will go to graveyard.

As much as I know, this is the first time we deal with move-transforming of collapsed ranges. When implementing this functionality we didn't know what will be the better behaviour. After changing the behaviour only two tests fail - the ones that checked exactly if marker sticks with removed element.

scofalik:

Fix: Collapsed `model.Range` no longer sticks to it's neighbour node when that node is moved. Closes #852.

---

### Additional information

According to tests, no code used this functionality yet (no tests failed after changing it, other than tests checking exactly that functionality).

### DIFF
--- a/tests/model/range.js
+++ b/tests/model/range.js
@@ -18,7 +18,8 @@ import {
 	getMoveDelta,
 	getRemoveDelta,
 	getRenameDelta,
-	getSplitDelta
+	getSplitDelta,
+	getMergeDelta
 } from '../../tests/model/delta/transform/_utils/utils';
 
 describe( 'Range', () => {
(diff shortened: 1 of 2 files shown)


tests/model/range.js
```diff
@@ -593,20 +593,22 @@ describe( 'Range', () => {
 			expect( transformed[ 0 ].end.path ).to.deep.equal( [ 4, 7 ] );
 		} );
 
-		it( 'should stick to moved range, if the transformed range is collapsed #1', () => {
+		it( 'should not stick to moved range, if the transformed range is collapsed #1', () => {
```
pjasiun: Are you sure that is should not move, if we already have tests checking the opposite behavior?

scofalik: As I stated above, those tests were probably to meet 100% CC and we have to decide on something. I don't remember any situation where we did depend on this behaviour.

OTOH, when I think of it now I have some doubts, so I will rethink it.

scofalik:
I've added fix for one scenario where "sticking" to moved nodes makes sense.

PR #856, opened 2017-03-03
T/841

### LINKED ISSUE #841: Selection position is lost when unwrapping range
```js
setModelData(
	doc,
	'<heading>a[bc</heading>' +
	'<messageQuote><paragraph>x]x</paragraph></messageQuote>' +
	'<paragraph>def</paragraph>'
);

batch.unwrap( quote );

console.log( getModelData( doc ) ); // <heading>a[bc</heading><paragraph>xx</paragraph>]<paragraph>def</paragraph>
```

Reinmar:

Fix: Fixed a bug where `LiveRange` position would be lost when using wrap and unwrap deltas. Closes #841.

### DIFF
--- a/src/model/range.js
+++ b/src/model/range.js
@@ -455,14 +455,26 @@ export default class Range {
 			if ( deltaType == 'split' && this.containsPosition( sourcePosition ) ) {
 				ranges[ 0 ].end = ranges[ 1 ].end;
 				ranges.pop();
-			} else if ( type == 'move' && deltaType == 'wrap' ) {
+			}
+			// Don't ask. Just debug.
+			// Like this: https://github.com/ckeditor/ckeditor5-engine/issues/841#issuecomment-282706488.
+			else if ( type == 'move' && deltaType == 'wrap' ) {
 				// <p>a[b</p><w></w><p>c]d</p> -> <w><p>a[b</p></w><p>c]d</p>
 				if ( this.containsPosition( targetPosition ) ) {
 					return [ new Range( ranges[ 2 ].start, ranges[ 1 ].end ) ];
 				}
-				// <p>a[b</p><p>c]d</p><w></w> -><p>a[b</p><w><p>c]d</p></w>
+				// <p>a[b</p><p>c]d</p><w></w> -> <p>a[b</p><w><p>c]d</p></w>
 				else if ( this.containsPosition( sourcePosition ) ) {
 					return [ new Range( ranges[ 0 ].start, ranges[ 1 ].end ) ];
+				}
+			} else if ( type == 'move' && deltaType == 'unwrap' ) {
+				// <w><p>a[b</p></w><p>c]d</p> -> <p>a[b</p><w></w><p>c]d</p>
+				if ( this.containsPosition( sourcePosition.getShiftedBy( howMany ) ) ) {
+					return [ new Range( ranges[ 1 ].start, ranges[ 0 ].end ) ];
+				}
+				// <p>a[b</p><w><p>c]d</p></w> -> <p>a[b</p><p>c]d</p><w></w>
+				else if ( this.containsPosition( sourcePosition ) ) {
+					return [ new Range( ranges[ 0 ].start, ranges[ 2 ].end ) ];
 				}
 			}
 
--- a/src/model/range.js
+++ b/src/model/range.js
@@ -447,40 +447,55 @@ export default class Range {
 	 * @returns {Array.<module:engine/model/range~Range>}
 	 */
 	_getTransformedByDocumentChange( type, deltaType, targetPosition, howMany, sourcePosition ) {
+		// IMPORTANT! Every special case added here has to be reflected in MarkerDelta transformations!
+		// Check /src/model/delta/basic-transformations.js.
 		if ( type == 'insert' ) {
 			return this._getTransformedByInsertion( targetPosition, howMany, false, false );
 		} else {
 			const ranges = this._getTransformedByMove( sourcePosition, targetPosition, howMany );
 
-			if ( deltaType == 'split' && this.containsPosition( sourcePosition ) ) {
-				// Special case for splitting element inside range.
-				// <p>f[ooba]r</p> -> <p>f[oo</p><p>ba]r</p>
-				ranges[ 0 ].end = ranges[ 1 ].end;
-				ranges.pop();
-			} else if ( deltaType == 'merge' && type == 'move' && this.isCollapsed && ranges[ 0 ].start.isEqual( sourcePosition ) ) {
-				// Special case when collapsed range is in merged element.
-				// <p>foo</p><p>[]bar{}</p> -> <p>foo[]bar{}</p>
-				ranges[ 0 ] = new Range( targetPosition.getShiftedBy( this.start.offset ) );
-			}
 			// Don't ask. Just debug.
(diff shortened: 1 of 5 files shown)


src/model/range.js
```diff
@@ -462,6 +462,27 @@ export default class Range {
 				// <p>foo</p><p>[]bar{}</p> -> <p>foo[]bar{}</p>
 				ranges[ 0 ] = new Range( targetPosition.getShiftedBy( this.start.offset ) );
 			}
+			// Don't ask. Just debug.
+			// Like this: https://github.com/ckeditor/ckeditor5-engine/issues/841#issuecomment-282706488.
+			else if ( type == 'move' && deltaType == 'wrap' ) {
+				// <p>a[b</p><w></w><p>c]d</p> -> <w><p>a[b</p></w><p>c]d</p>
```
scofalik: It would be nice to expand this example a little bit more. State before wrapping, which element is wrapped, what conversion without the fix would look like, etc. (why we are taking start/end of exactly those ranges).

tests/model/liverange.js
```diff
@@ -361,6 +362,147 @@ describe( 'LiveRange', () => {
...
+				setData( doc, '<w><p>a[b</p><p>x</p></w><p>c]d</p>' );
+
+				live = new LiveRange( doc.selection.getFirstPosition(), doc.selection.getLastPosition() );
+
+				doc.batch().unwrap( root.getChild( 0 ) );
+
+				expect( stringify( root, live ) ).to.equal( '<p>a[b</p><p>x</p><p>c]d</p>' );
+			} );
+
+			it( 'its end is intersecting with the wrapper to remove', () => {
```
scofalik: add "(multiple elements)" in test name.

tests/model/liverange.js
```diff
@@ -361,6 +362,147 @@ describe( 'LiveRange', () => {
...
+				setData( doc, '<p>a[b</p><w><p>x</p><p>c]d</p></w>' );
+
+				live = new LiveRange( doc.selection.getFirstPosition(), doc.selection.getLastPosition() );
+
+				doc.batch().unwrap( root.getChild( 1 ) );
+
+				expect( stringify( root, live ) ).to.equal( '<p>a[b</p><p>x</p><p>c]d</p>' );
+			} );
+		} );
```
scofalik: Please add a test for `<p>x[x</p><w><p>foo</p><p>bar</p></w><p>y]y</p>`

tests/model/liverange.js
```diff
@@ -361,6 +362,147 @@ describe( 'LiveRange', () => {
...
+			it( 'its end is intersecting with the wrapped range (multiple elements)', () => {
+				setData( doc, '<p>a[b</p><p>x</p><p>c]d</p>' );
+
+				live = new LiveRange( doc.selection.getFirstPosition(), doc.selection.getLastPosition() );
+
+				// [<p>x</p><p>cd</p>]
+				doc.batch().wrap( new Range( new Position( root, [ 1 ] ), new Position( root, [ 3 ] ) ), 'w' );
+
+				expect( stringify( root, live ) ).to.equal( '<p>a[b</p><w><p>x</p><p>c]d</p></w>' );
+			} );
```
scofalik: Please add a test for `<p>x[x</p><p>foo</p><p>bar</p><p>y]y</p>` where `<p>foo</p><p>bar</p>` is wrapped.

scofalik: Or you could add such a test in `range.js`.

tests/model/liverange.js
scofalik: Or you could add such a test in `range.js`.

src/model/range.js
```diff
@@ -462,6 +462,27 @@ export default class Range {
 				// <p>foo</p><p>[]bar{}</p> -> <p>foo[]bar{}</p>
 				ranges[ 0 ] = new Range( targetPosition.getShiftedBy( this.start.offset ) );
```
scofalik: It'd be nice to change this to `return [ new Range( ... ) ]`, to fit versions below.

scofalik: Same above, but I cannot comment there.

scofalik:
Unfortunately, whenever we change something here, we need to introduce transformation case in deltas transformations: https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/delta/basic-transformations.js#L100-L126

I also forgot about this.

Honestly, this escalated quickly. Adding special transformation in `_getTransformedByDocumentChange` for `split` (the first special case there) was very hacky and I didn't like this solution, but we were unable to do it in a better way. No we treated it like "we can throw anything there", because there already is something (I am also guilty of this).

So, first of all, we need a comment in this part of code, that all added special cases need to be reflected in `MarkerDelta` transformation. Then we need to write those transformations for missing cases.

scofalik:
Funny enough, special case for `merge` does not have to have it's special `MarkerDelta` transformation, because it will work like this already _(this is because `MoveOperation` in `MergeDelta` is `sticky`, if you wondered :P)_.

scofalik:
Herp derp, it appeared, that it wasn't working for `MarkerDelta` x `MergeDelta` after all! I'm glad I wrote this test. I guess I will hijack this PR for a while...

scofalik:
I fixed all the issues I commented on and added tests for when transformed range contains wrapped range / unwrapped element (which wasn't working property too).

I also changed how `MarkerDelta` transformation is handled. Actually, it was very stupid to provide separate handling in that part of code. Instead, now it indirectly uses `Range#_getTransformedByDocumentChange` through `Range#getTransformedByDelta`. It makes sense, as what we are trying to achieve is the situation where `(Live)Range`s are transformed in the same way both by changes applied in the document and through OT.

So there goes that.

Now, I don't really like how the functions looks like. If you look closely, you realise that all those cases for wrap and unwrap are similar. As you know I am fan of generalised solutions...
```html
[<w>]<p>a(b</p>){</w><p>c}d</p> -> <w><p>a[b</p></w><p>c]d</p>
<p>a[b</p>]<w>{<p>c}d</p></w> -> <p>a[b</p><w><p>c]d</p></w>
<p>a{b</p>}<w>[</w><p>c]d</p> -> <p>a[b</p><w></w><p>c]d</p>
<p>a[b</p>](<p>c)d</p>{<w>}</w> -> <p>a[b</p><p>c]d</p><w></w>
```
As you can see all that needs to be done is:
1. Remove "empty" ranges (like `[<w>]`).
2. Join other ranges together (also when they are separated by opening tag).

Now, the sad thing is that it would all work, if we had old implementation that used `.isTouching()`. But the good thing is that we can check this scenario: `[<w>]` just by comparing paths (what we are unable to check by comparing paths is closing tag: `[</w>]` because we don't know if the starting position is the last position in it's parent by looking at the path).

On one hand it would be nice to have such general solution because we don't know what kind of cases are still there, and we depend on the order of ranges returned by `_getTransformedByMove`. On the other hand, we covered all "complicated" deltas (split, merge, wrap, unwrap). There may be something more for `MergeDelta`, but for other deltas I think that we have them covered. So maybe there's no use to waste time on it. And that general solution do not cover `SplitDelta` and `MergeDelta` special cases.

Reinmar:
> So maybe there's no use to waste time on it. And that general solution do not cover SplitDelta and MergeDelta special cases.

Let's wait. If we really covered all complicated deltas then we're fine for now. Let's see what the time will bring.

PR #861, opened 2017-03-08
T/460 Selection rendering breaks composition

### LINKED ISSUE #460: Selection rendering breaks composition
This check should prevent changing selection on composition: https://github.com/ckeditor/ckeditor5-engine/blob/cc5c39ec02a8096a92b246a331d9240eef1e272d/src/view/renderer.js#L428

It works for the regular navigation, but because of some reason during the composition, when I press left arrow to change the character value, the offsets do not match, selection is rerendered and composition is broken.

f1ames:
Fix: Rendering is suspended during composition. Closes #460.

---

### Additional information

The fix introduces `CompositionObserver` which listens to native composition events. If composition is in progress the `renderer` will not update the DOM.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -105,7 +105,7 @@ export default class Renderer {
 		this.isFocused = false;
 
 		/**
-		 * Indicates if composition takes places inside view document. View will not be rendered durning composition.
+		 * Indicates if composition takes places inside view document. View will not be rerendered during composition.
 		 *
 		 * @member {Boolean}
 		 */
@@ -166,8 +166,8 @@ export default class Renderer {
 	 * if needed updates the selection.
 	 *
 	 * Renderer tries not to break text composition (e.g. IME) and x-index of the selection,
-	 * so it does as little as it is needed to update the DOM. During the composition all rendering is blocked
-	 * in order not to break the composition. Render is called after composition ends.
+	 * so it does as little as it is needed to update the DOM. During the composition all rendering is suspended
+	 * in order not to break the composition.
 	 *
 	 * For attributes it adds new attributes to DOM elements, updates values and removes
 	 * attributes which do not exist in the view element.
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -105,6 +105,13 @@ export default class Renderer {
 		this.isFocused = false;
 
 		/**
+		 * Indicates if composition takes places inside view document. View will not be rendered durning composition.
+		 *
+		 * @member {Boolean}
+		 */
+		this.isComposing = false;
+
+		/**
 		 * DOM element containing fake selection.
 		 *
 		 * @private
@@ -159,7 +166,8 @@ export default class Renderer {
 	 * if needed updates the selection.
 	 *
 	 * Renderer tries not to break text composition (e.g. IME) and x-index of the selection,
-	 * so it does as little as it is needed to update the DOM.
+	 * so it does as little as it is needed to update the DOM. During the composition all rendering is blocked
+	 * in order not to break the composition. Render is called after composition ends.
 	 *
 	 * For attributes it adds new attributes to DOM elements, updates values and removes
 	 * attributes which do not exist in the view element.
@@ -177,6 +185,10 @@ export default class Renderer {
 	 * removed as long selection is in the text node which needed it at first.
 	 */
 	render() {
+		if ( this.isComposing ) {
+			return;
+		}
+
 		let inlineFillerPosition;
 
 		// There was inline filler rendered in the DOM but it's not
--- a/src/view/document.js
+++ b/src/view/document.js
@@ -18,6 +18,7 @@ import SelectionObserver from './observer/selectionobserver';
 import FocusObserver from './observer/focusobserver';
 import KeyObserver from './observer/keyobserver';
 import FakeSelectionObserver from './observer/fakeselectionobserver';
+import CompositionObserver from './observer/compositionobserver';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
 import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
 
@@ -91,6 +92,18 @@ export default class Document {
 		this.set( 'isFocused', false );
 
 		/**
+		 * True if composition is in progress inside the document.
+		 *
+		 * This property is updated by the {@link module:engine/view/observer/compositionobserver~CompositionObserver}.
+		 * If the {@link module:engine/view/observer/compositionobserver~CompositionObserver} is disabled this property will not change.
+		 *
+		 * @readonly
+		 * @observable
+		 * @member {Boolean} module:engine/view/document~Document#isComposing
+		 */
+		this.set( 'isComposing', false );
+
+		/**
 		 * Instance of the {@link module:engine/view/document~Document#renderer renderer}.
 		 *
 		 * @readonly
@@ -98,6 +111,7 @@ export default class Document {
 		 */
 		this.renderer = new Renderer( this.domConverter, this.selection );
 		this.renderer.bind( 'isFocused' ).to( this, 'isFocused' );
+		this.renderer.bind( 'isComposing' ).to( this, 'isComposing' );
 
 		/**
 		 * Map of registered {@link module:engine/view/observer/observer~Observer observers}.
@@ -113,6 +127,7 @@ export default class Document {
 		this.addObserver( FocusObserver );
 		this.addObserver( KeyObserver );
 		this.addObserver( FakeSelectionObserver );
+		this.addObserver( CompositionObserver );
 
 		injectQuirksHandling( this );
 
--- a/src/view/observer/compositionobserver.js
+++ b/src/view/observer/compositionobserver.js
@@ -30,9 +30,6 @@ export default class CompositionObserver extends DomEventObserver {
 
 		document.on( 'compositionend', () => {
 			document.isComposing = false;
-
-			// Re-render the document to update view elements.
-			document.render();
 		} );
 	}
 
--- a/src/view/observer/compositionobserver.js
+++ b/src/view/observer/compositionobserver.js
@@ -1,0 +1,84 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module engine/view/observer/compositionobserver
+ */
+
+import DomEventObserver from './domeventobserver';
+
+/**
+ * {@link module:engine/view/document~Document#event:compositionstart Compositionstart},
+ * {@link module:engine/view/document~Document#event:compositionupdate compositionupdate} and
+ * {@link module:engine/view/document~Document#event:compositionend compositionend} events observer.
+ *
+ * Note that this observer is attached by the {@link module:engine/view/document~Document} and is available by default.
+ *
+ * @extends module:engine/view/observer/domeventobserver~DomEventObserver
+ */
+export default class CompositionObserver extends DomEventObserver {
+	constructor( document ) {
+		super( document );
+
+		this.domEventType = [ 'compositionstart', 'compositionupdate', 'compositionend' ];
+
+		document.on( 'compositionstart', () => {
+			document.isComposing = true;
+		} );
+
+		document.on( 'compositionend', () => {
+			document.isComposing = false;
+
+			// Re-render the document to update view elements.
+			document.render();
+		} );
+	}
+
+	onDomEvent( domEvent ) {
+		this.fire( domEvent.type, domEvent );
+	}
+}
+
+/**
+ * Fired when composition starts inside one of the editables.
+ *
+ * Introduced by {@link module:engine/view/observer/compositionobserver~CompositionObserver}.
+ *
+ * Note that because {@link module:engine/view/observer/compositionobserver~CompositionObserver} is attached by the
+ * {@link module:engine/view/document~Document}
+ * this event is available by default.
+ *
+ * @see module:engine/view/observer/compositionobserver~CompositionObserver
+ * @event module:engine/view/document~Document#event:compositionstart
+ * @param {module:engine/view/observer/domeventdata~DomEventData} data Event data.
+ */
+
+/**
+ * Fired when composition is updated inside one of the editables.
+ *
+ * Introduced by {@link module:engine/view/observer/compositionobserver~CompositionObserver}.
+ *
+ * Note that because {@link module:engine/view/observer/compositionobserver~CompositionObserver} is attached by the
+ * {@link module:engine/view/document~Document}
+ * this event is available by default.
+ *
+ * @see module:engine/view/observer/compositionobserver~CompositionObserver
+ * @event module:engine/view/document~Document#event:compositionupdate
+ * @param {module:engine/view/observer/domeventdata~DomEventData} data Event data.
+ */
+
+/**
+ * Fired when composition ends inside one of the editables.
+ *
+ * Introduced by {@link module:engine/view/observer/compositionobserver~CompositionObserver}.
+ *
+ * Note that because {@link module:engine/view/observer/compositionobserver~CompositionObserver} is attached by the
+ * {@link module:engine/view/document~Document}
+ * this event is available by default.
+ *
+ * @see module:engine/view/observer/compositionobserver~CompositionObserver
+ * @event module:engine/view/document~Document#event:compositionend
+ * @param {module:engine/view/observer/domeventdata~DomEventData} data Event data.
+ */
--- a/tests/view/document/document.js
+++ b/tests/view/document/document.js
@@ -13,6 +13,7 @@ import SelectionObserver from '../../../src/view/observer/selectionobserver';
 import FocusObserver from '../../../src/view/observer/focusobserver';
 import KeyObserver from '../../../src/view/observer/keyobserver';
 import FakeSelectionObserver from '../../../src/view/observer/fakeselectionobserver';
+import CompositionObserver from '../../../src/view/observer/compositionobserver';
 import Renderer from '../../../src/view/renderer';
 import ViewRange from '../../../src/view/range';
 import DomConverter from '../../../src/view/domconverter';
@@ -23,7 +24,7 @@ import log from '@ckeditor/ckeditor5-utils/src/log';
 testUtils.createSinonSandbox();
 
 describe( 'Document', () => {
-	const DEFAULT_OBSERVERS_COUNT = 5;
+	const DEFAULT_OBSERVERS_COUNT = 6;
 	let ObserverMock, ObserverMockGlobalCount, instantiated, enabled, domRoot, viewDocument;
 
 	before( () => {
@@ -88,6 +89,7 @@ describe( 'Document', () => {
 			expect( viewDocument.getObserver( FocusObserver ) ).to.be.instanceof( FocusObserver );
 			expect( viewDocument.getObserver( KeyObserver ) ).to.be.instanceof( KeyObserver );
 			expect( viewDocument.getObserver( FakeSelectionObserver ) ).to.be.instanceof( FakeSelectionObserver );
+			expect( viewDocument.getObserver( CompositionObserver ) ).to.be.instanceof( CompositionObserver );
 		} );
 	} );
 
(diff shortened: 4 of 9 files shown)


src/view/renderer.js
```diff
@@ -177,6 +185,10 @@ export default class Renderer {
 	 * removed as long selection is in the text node which needed it at first.
 	 */
 	render() {
+		if ( this.isComposing ) {
+			return;
```
pjasiun: Looks like asking for serious troubles with collaboration. If you allow changing the view, but prevent changes in DOM, you may have a very different state of view and DOM if another user changed the same part of the document. Mutations calculates a diff between DOM and view and if there is something for the OT in the view it will be marked to be removed since it does not appear in the DOM.

pjasiun: You can say that it's an edge case, but it is exactly the same edge case you want to fix with this return (OT and composition in the same block). It means that to prevent breaking composition, you breaks OT.

pjasiun: For sure it need some tests with collaboration begore merging this PR because it can causes more bad than good.

scofalik:
Have you tested how it would work with OT / collaborative editing? I am worried it will not work properly. Consider following scenario:

1. You start composition in `<p>foo^bar</p>` -> `<p>foo[@]bar</p>`. `@` is composed char.
2. Rendering is blocked.
3. You get a change from server that inserts letter `a` at the beginning of `<p>`.
4.
Your model is:
`<paragraph>afoo[@]bar</paragraph>`
Your view is:
`<p>afoo[@]bar</p>`
But your DOM is:
`<p>foo[@]bar</p>`

5. Now, when you type another part of composed char, the mutation will be handled by typing and when DOM and view are diffed, it will appear that letter `a` should be removed from DOM. Amirite?


I think this has to be handled more cautiously. Some ideas (pretty wild so I don't know whether they are good):
* Maybe not only block rendering but also block typing? Then handle inserting composed character on `compositionend` and re-render view?
* Maybe block applying all operations when composing? But then we need to cache them and we would have to make some transformations.
* Maybe stop model->view conversion too and try to refresh things after composition has ended?

f1ames:
Thanks for pointing that out, I was wondering about issues connected to OT. What is the proper way to test OT for such cases, do we already have some integration/manual tests for similar cases?

scofalik:
You can test it directly by faking composition start (set `isComposing` by hand?) and apply correct `InsertOperation` (for example using `document.batch().insert()`).

f1ames:
Ok, first I will write some unit/integration test (using operations as @scofalik suggested) to check how it works with composition.
And then as @pjasiun suggested I will do some "real" collaboration testing.
Of course, having in mind that the current solution most likely will break things and needs to be implemented differently.

Reinmar:
A question (without checking much of the code) – does this PR blocks the entire rendering? Looking at some comments, I'd say that we should make it more precise from the day one by blocking changing only a certain container's content.

f1ames:
@Reinmar, yes it blocks the entire rendering. Blocking only the container (or part of it) in which the composition takes place may help to solve some cases with OT, however probably not all.
Do you suggest it only should block rendering of the node in which composition takes place?

Reinmar:
> Looking at some comments, I'd say that we should make it more precise from the day one by blocking changing only a certain container's content.

Although... I can imagine such situation:

1. One person is composing in `<p>a[bc]def</p>`.
2. The other person presses enter at this position: `<p>abcde[]f</p>`.

If we block rendering in that first `<p>`, the person who's composing will see something like this for some time:

```html
<p>a[bc]def</p>
<p>f</p>
```

Mind the duplicated "f".

After more thinking I guess that blocking rendering smaller bits of content may lead to really weird  situations ;|. So I'm not sure anymore if we shouldn't start with blocking everything.

Reinmar:
> @Reinmar, yes it blocks the entire rendering. Blocking only the container (or part of it) in which the composition takes place may help to solve some cases with OT, however probably not all.
> Do you suggest it only should block rendering of the node in which composition takes place?

I don't know. Right now I can see only these issues with weird things happening around the blocked container. @scofalik may be able to imagine how (and if) this would have any effect on OT. I think not. The model will actually be live. The problems may be `model+view <-> DOM` related – e.g. mapping. If we'll block rendering then view will diverge from the view so mapping the composition positions to the positions in the model may become tricky.

pjasiun:
I think that we should do as much as we can not to break anything, but when there is no other way, for changes we can not predict/handle, I think that it's better to break composition then OT. These will be edge cases, I hope, but if the composition will be broken, then it broken composition. If the OT is broken, it can be broken editor (or all connected editors in the worst scenario). 

I think, still, we can handle, most obvious scenarios. But blocking rendering should go in pair with handing observers well when the view totally changed. You can not predict what kind of changes occurred in the meantime. It mostly meant that you need to handle mutations, but I can also imagine some bugs with selection observer, hope other observers will be fine.

This is why I think that less we block, safer we are. Again, if part of the not blocked rendering will break composition, then only composition is broken, but if we will have a crash because of the not handled mutation/selection/whatever observer of the blocked element, then everything is broken.

I would start from the smallest possible part, so blocking changing content of the text node. Then we can handle inserting a new text node if composition creates it. Then breaking text block, moving it, etc. Or not, we can decide that if someone else breaks the block, it's too big change to prevent and the composition should be broken.

For handling typing in the same block, you still need to somehow save changes in the view, which are not rendered. I believe it's doable. For instance:
```
foo [locked]

foxxxo [in DOM, because of composition]

fyyyoo [in View, because of OT]
// yyy on position 1-4 are saved with the information that they are not rendered

DOM mutation:
[ change 'foo' to 'foxxxo' ]

Mutation after comparing it to view:
[ change 'fyyyoo' to 'foxxxo' ]

Mutation after handling saved letters:
[ change 'fyyyoo' to 'fyyyoxxxo' ] // This is the mutation which should be fired from the view.MutationObserver
```

pjasiun:
And, as I think about it, selection need to be handled as well. If I end composition by moving the selection to the end of `foxxxo{}`, then it should be moved to the `fyyyoxxxo{}` in the view.

pjasiun:
Also, note that renderer role is to compare view and DOM and not render nodes which does not need to be rendered (to not break composition). So elements which are the same in the view and DOM should not be touched. It, most probably, does not mean that we do not need renderer blocking, but as long as it can prevent some bugs, we should improve renderers nodes comparing, so render as few as possible.

Reinmar:
BTW, we can also improve how the text nodes are updated.

In https://github.com/w3c/editing/issues/149 I commented that:

> Touching the same text node (by using methods like CharacterData.replaceContent()) doesn't break the composition too: https://jsfiddle.net/gqnq4h5r/12/ (you can place caret inside "App*" and use IME there). This is super cool :).

I don't know if this applies to all browsers but if we want to go this way that OT is more important, then this would be the way to go.

**BUT** do note that composition can't have the lowest priority for us. E.g. if you want to render selections of other users in the text node in which someone is typing... you'll break the composition every time some other user moved the selection in that text node. This, in some situations, will not be an edge case, but a really popular scenario. It'd render composing text impossible which would be a huge issue.

Reinmar:
Final words – I guess we know a bit more about the scope and complexity of this problem now. It looks to me that it may be a killer issue so we should consider trying to find some quick wins and leave the rest for the future.

What I'd like to now first is:

1. Whether we have any problems with composition and how bad they are. I'm talking here about 1 user scenarios because we can be sure that collaboration breaks the composition.
2. What kind of quick wins we could talk about. E.g. can we block rendering/converting just the markers? That would at least prevent issues caused by other users' selections. Can we use `CharacterData.replaceContent()` today to prevent issues with people typing in one text node?
3. Do we have any viable plan for the future? We should have some solution designed now, so the temporal steps we'll do now will lead us closer to that designed solution. Otherwise, we may be wasting our time.

pjasiun:
> BUT do note that composition can't have the lowest priority for us. E.g. if you want to render selections of other users in the text node in which someone is typing... you'll break the composition every time some other user moved the selection in that text node. This, in some situations, will not be an edge case, but a really popular scenario. It'd render composing text impossible which would be a huge issue.

You did not get my point. I did not say that we should not prevent rendering anything, like the selection of other users. Or that it is not important. I said that we can not introduce changes which prevent breaking composition but crash editor instead. We need to be very careful and be sure that every introduced change is well tested and will not cause more harm than good.

f1ames:
Looks tough indeed. I start checking for failing 1 user scenarios while using composition. Checked MacOS accent balloon and Spanish accent composition for now.

### MacOS accent scenarios

1. Long press + number
1. Long press + mouse
1. Long press + arrow keys + enter
1. Long press + arrow keys + esc
1. Long press + click outside
1. Long press + Bold (edge case) + choose accent if possible
1. Long press + other letter

_Only failing scenarios are listed below_

**Chrome**
6. Inserts bolded accented letter but throws an error - `Uncaught CKEditorError: view-renderer-filler-was-lost: The inline filler node was lost.`. In other browsers the accent panel is just closed when applying bold.

**FF** - All OK

**Safari**
3. and 4. This is more general issue: It is impossible to navigate with arrow keys inside accent panel. After pressing left arrow last accent is inserted. After pressing right arrow first accent is inserted.

### Spanish-ISO scenarios

1. Accent (´) + normal letter (b)
2. Accent (´) + accentable letter (a)
3. Accent (´) + esc
4. Accent (´) + bold (edge case)

_Only failing scenarios are listed below_

**Chrome** - All OK

**FF**
4. After applying bold the second press inserts letter, first one does nothing.

**Safari**
The general issue is that accent (´) is removed when typed on non-collapsed selection.

3&period; Broken:
  1. `<h2>Headin[]g 1</h2>`
  2. Press accent
  3. Press esc
  4. Results in `<h2>Heading'[]g 1</h2>`

4&period; Broken:
  1. `<p>Hea[]ding 2</p>`
  2. Press accent
  3. Press bold
  4. Type `y`
  5. Results in `<p>He<strong>y</strong>´ding 2</p>` and Throws `TypeError: null is not an object (evaluating 'viewBlock.parent')` in `_defaultToModelPosition — mapper.js:222`

**To sum up, there is 1 issue in Chrome, 1 in FF and 4 in Safari.**
I am not sure how many of this issues are caused by the fact that DOM is rerendered during the composition (will check this later, after checking Hiragana) but seems like it is the cause for some of them.

Reinmar:
> 3. and 4. This is more general issue: It is impossible to navigate with arrow keys inside accent panel. After pressing left arrow last accent is inserted. After pressing right arrow first accent is inserted.

This is odd. Perhaps our handler for the filler kicks in and tries to handle the arrow keys. This is the only place where we listen to keys.

Reinmar:
> The general issue is that accent (´) is removed when typed on non-collapsed selection.

May be that our dirty unsafe key hack fails here :(

scofalik:
My 5 cents: of course blocking rendering does not crash OT directly. What PJ (and me) meant by crashing OT is more crashing collaborative editing, because of reasons that we already discussed.

f1ames:
Checked some scenarios with **Hiragana** too. Only a few but I think it is enough to give some general picture. The scenarios contains just a key sequences as it is the easiest way to describe it and the reproduce.

1. <kbd>a</kbd> + <kbd>a</kbd> + <kbd>a</kbd> + <kbd>enter</kbd>
2. <kbd>a</kbd> + <kbd>e</kbd> + <kbd>down</kbd> x3 + <kbd>enter</kbd> x2
3. <kbd>a</kbd> + <kbd>e</kbd> + <kbd>down</kbd> x3 + <kbd>a</kbd> + <kbd>e</kbd> + <kbd>down</kbd> x2 + <kbd>enter</kbd> x2
4. <kbd>a</kbd> + <kbd>e</kbd> + <kbd>a</kbd> + <kbd>left</kbd> + <kbd>right</kbd> + <kbd>left</kbd> + <kbd>up</kbd> x2
5. <kbd>a</kbd> + <kbd>e</kbd> + <kbd>a</kbd> + <kbd>up</kbd> hold
6. <kbd>a</kbd> + <kbd>e</kbd> + `Bold` + <kbd>a</kbd> + <kbd>e</kbd> + <kbd>down</kbd> x4
7. `Bold` + <kbd>a</kbd> + <kbd>e</kbd> + <kbd>a</kbd>

_Only failing scenarios are listed below_

**Chrome**
3. Broken - inserts two characters more than it should (`逢え和え[]和え` instead of `逢え和え[]`).
4. Broken - composition ends but composition dialog stays and cannot be closed.
5. Broken - spans a lot of additional characters when <kbd>up</kbd> is hold long enough.
6. Broken - seems like bold just ends the composition and text is inserted properly. However, next composition is broken.
7. Broken - first typed character is not a part of the composition.

**FF**
7. Broken - only first character is inserted.

**Safari**
_Navigating with arrow keys breaks the composition in most cases (breaks 2. - 6. scenario)._
7. Seems like composition instantly ends inside styled element

I think it will be good to check which of the failing cases could be solved by _blocking rendering during composition_ and or using `CharacterData.replaceContent()` (also I will take a quick look on https://github.com/ckeditor/ckeditor5-engine/pull/861#issuecomment-285290138 and https://github.com/ckeditor/ckeditor5-engine/pull/861#issuecomment-285290340) and after that we could probably conclude something based on the results and all previous comments.

pjasiun:
> I think it will be good to check which of the failing cases could be solved by blocking rendering during composition

I think that first, we should check why these cases are failing. It might be simple a bug in the renderer which renders more than it needs and may be fixed. `CharacterData.replaceContent()` might help too. But blocking rendering is the final workaround, when we are not able to find a real solution.

Reinmar:
I can see that the arrow keys break text input again. I think that this is the same issue as with what you described previously – that in e.g. Safari you can't use arrows to navigate the accent tooltip. I'd check this first. And it may not be related to rendering but to how we handle the filler.

f1ames:
@Reinmar if by `filler` you mean https://github.com/ckeditor/ckeditor5-engine/blob/master/src/view/filler.js#L156, it seems this code does not have any impact on this issue (both in MacOS accent panel and Hiragana), so the cause must be someweher else.

f1ames:
> May be that our dirty unsafe key hack fails here :(

Yep -> https://github.com/ckeditor/ckeditor5-typing/issues/82

f1ames:
As for https://github.com/ckeditor/ckeditor5-engine/pull/861#issuecomment-285319329 - while navigating with arrow keys inside balloon panel the selection changes so the renderer updates it (`_updateSelection`) which breaks composition.
Introducing e.g. blocking selection rendering during composition will prevent this issue but then, there is`_updateChildren` which is fired on accent change and replaces whole text node with the new one (which also breaks the composition).

I was also able to check how it works with `replaceData`. The `replaceData` does not break the composition only if the replaced part of text does not include the current selection. So using replace data might help in some cases but then it would require to use `replaceData` with diff to change only parts of text which were changes.

Reinmar:
> As for #861 (comment) - while navigating with arrow keys inside balloon panel the selection changes so the renderer updates it (_updateSelection) which breaks composition.
> Introducing e.g. blocking selection rendering during composition will prevent this issue but then, there is `_updateChildren` which is fired on accent change and replaces whole text node with the new one (which also breaks the composition).

We have two options here:

1. Make sure no rendering is needed because the selection we've read from native selectionchange events and converted through view=>model=>view is identical with the one with which we had initially. The same must apply to the content itself. That was @pjasiun's plan and in ideal world that would be the thing, but I personally find this approach to idealistic. Too many things can go wrong. But it will still be the first thing to check – why does selection converted from the model differs from the one we have in the view. Perhaps we have some bug.
2. Indeed block selection rendering when composition takes place. This should be easier than blocking content rendering, but it also assumes that for the composition to work, the renderer must decide that the DOM is identical with the view. That can still go wrong (because the view may diverge with the DOM and rendering will happen), but at least we'd be half way closer.

Reinmar:
> I was also able to check how it works with replaceData. The replaceData does not break the composition only if the replaced part of text does not include the current selection. So using replace data might help in some cases but then it would require to use replaceData with diff to change only parts of text which were changes.

Yep, we'd need to use the diff function and [diff results grouping](https://github.com/ckeditor/ckeditor5-utils/blob/master/src/difftochanges.js). I was worried about performance impact, but I guess we'll use it only when updating existing text node (which is rather rare action). When rendering for the first time we'll be of course creating new text nodes so it should be fine.

f1ames:
The two Safari issues (Spanish-ISO 3. and 4.) described in https://github.com/ckeditor/ckeditor5-engine/pull/861#issuecomment-285289224 are caused by the Safari auto correction mechanism so it is not releted to this issue. With auto correction off, it works fine.

pjasiun:
> We have two options here:
> 
> 1. Make sure no rendering is needed...
> 2. Indeed block selection rendering when composition takes place...

I agree with both. For content, we should do our best to improve renderer comparison, so the renderer will see that the view and the DOM are the same and do not need rendering, since, at the moment all other options seems to be very painful. For selection, we should try it too, but preventing selection rendering seems to be far less dangerous than contents.

pjasiun:
> I was worried about performance impact

Yep, we had performance problems with diff. It appears when you compare 2 totally different collections (strings). The time needed to calculate the diff is linear if collections are the same and grows exponentially with the length of the different texts. There was an idea to improve the diff function so if the difference is too big to calculate it quickly (something like 30 different characters), we overwrite the whole string.

Reinmar:
But do you think that it may be any performance issue if we use diffing only on strings which we know that got changed? This only affects typing and other kinds of text insertion which happen only a couple of times per second.

pjasiun:
If one paste a long text into the paragraph we may get a serious browser freeze (more than 10 sec). It might not be that bad, but the performance, in this case, needs to be tested in the first place.

I remember we experienced it already, and it was not around 500 ms, but more than 10 sec. But then we used a linear model and calculated a diff on the whole model, so in this case in might be not that bad.

f1ames:
https://github.com/ckeditor/ckeditor5-typing/issues/85

f1ames:
> I think that first, we should check why these cases are failing. It might be simple a bug in the renderer which renders more than it needs and may be fixed.

In case of selection during composition (MacOS accent panel) it is: https://github.com/ckeditor/ckeditor5-typing/issues/86

f1ames:
It looks like https://github.com/ckeditor/ckeditor5-typing/issues/86 also fixed most issues with **Hiragana** composition in Chrome (3. - 6.) mentioned above. And so for the above scenarios the remaining issues are:

**MacOS accent scenarios**

* Chrome -  _6._ - https://github.com/ckeditor/ckeditor5-typing/issues/90

**Spanish-ISO scenarios**

* FF - _4._ - https://github.com/ckeditor/ckeditor5-typing/issues/91

**Hiragana**

* Chrome, FF, Safari - _7._ - https://github.com/ckeditor/ckeditor5-typing/issues/89

All IME related found issues: https://github.com/ckeditor/ckeditor5-typing/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20IME

Reinmar:
@pjasiun, as I told you, we don't close unmerged PRs if they have some value. I've been looking for this PR now, knowing that it was somewhere there previously, and I couldn't find it. Hence, I'm reopening it.

f1ames:
Went through scenarios from https://github.com/ckeditor/ckeditor5-engine/pull/861#issuecomment-285289224 and https://github.com/ckeditor/ckeditor5-engine/pull/861#issuecomment-285312403 once again:

### MacOS accent

> 6. Inserts bolded accented letter but throws an error - Uncaught CKEditorError: view-renderer-filler-was-lost: The inline filler node was lost.. In other browsers the accent panel is just closed when applying bold.

Only this one is still reproducible on Chrome.

> 3. and 4. This is more general issue: It is impossible to navigate with arrow keys inside accent panel. After pressing left arrow last accent is inserted. After pressing right arrow first accent is inserted.

 The Safari one visible above related to key navigation inside accent panel was fixed.

### Spanish-ISO

Both:

> 4. After applying bold the second press inserts letter, first one does nothing.

> The general issue is that accent (´) is removed when typed on non-collapsed selection.

were fixed.

### Hiragana

> 6. Broken - seems like bold just ends the composition and text is inserted properly. However, next composition is broken.
> 7. Broken - first typed character is not a part of the composition.

Still reproducible in Chrome and Safari, looks like Bold is the issue here as described in https://github.com/ckeditor/ckeditor5-typing/issues/89. And on Firefox only **7.** case fails, seems to be related to the same issue. The rest of reported cases were fixed.

f1ames:
Went once again through scenarios which are still reproducible (https://github.com/ckeditor/ckeditor5-engine/pull/861#issuecomment-368943472) because of https://github.com/ckeditor/ckeditor5-engine/issues/1417.

### MacOS accent

> 6. Inserts bolded accented letter but throws an error - Uncaught CKEditorError: view-renderer-filler-was-lost: The inline filler node was lost.. In other browsers the accent panel is just closed when applying bold.

Only this one is still reproducible on Chrome.

### Hiragana

There is only problem with Safari when:

> Navigating with arrow keys when composition takes place inside empty inline element breaks it.

I suspect it is caused by the fact that when we navigate through composition panel, selection leaves text node (in which composition takes place) which causes inline filler to be removed and it breaks composition.

PR #864, opened 2017-03-09
T/795 Render document after focus

### LINKED ISSUE #795: Rendering view document after focusing
View document is rendered after each `blur` event but not after each `focus`. This means that if selection is placed in the exact same place (without selection change event) document will not be re-rendered:

output
_Nested ediable updates its CSS class after being focused but it is not re-rendered so the blue outline is not showing when selection is placed in the exact same place._

I've tried to test if adding `viewDocument.render()` to `FocusObserver` on `focus` event helps with this but unfortunately it is not working as expected. I think following things are happening when focusing nested editable:

* DOM element is focused, `FocusObserver` sets `viewDocument.isFocused` to `true`,
* `render()` method is called, view elements are rendered,
* old selection from view is rendered into DOM
* there is no selection change event with selection placed inside nested editable.

It looks like rendering on each focus event (which is fired when nested editable is focused) will not allow to put selection inside nested editable (because it will be rendered back to its position from view selection).

Any ideas how to solve it?

szymonkups:

Fix: View document is now re-rendered after focusing. Closes #795.

---

### Additional information
View document should be re-rendered when is focused. Unfortunately `selectionchange` event is fired asynchronously so immediate rendering after `focus` event will put selection from the view and overwrite new selection from DOM. First idea was to call `SelectionObserver#_handleSelectionChange` manually to process the DOM selection, but it is not there yet. Proposed solution is to add `selectionChangeHandling` event to `SelectionObserver` and listen to that event inside `FocusObserver` and render after that.


pjasiun:
For me the logic is fine. It's only the matter of the code structure (how the communication between focus and selection observers should look like). What you proposed, looks ok for me. @Reinmar suggested using a method/flag in `SelectionObserver`, what might be better (I hope he will write more about it). We could also merge these 2 observers into one class since they are strongly connected.

Reinmar:
Yes, I think that communication through event makes it more complex than it needs to be. I initially thought that the `FocusObserver` could just call on the `SelectionObserver` some `markDirty()` method, which the SO would remember and then use that knowledge to always fire `selectionChange` on the next native event.

However, @pjasiun's proposal makes even more sense. These two observers are strongly related. If we merge them, we wouldn't need communication between them to be protected – it could be private (it'd still need to be there :D).

OTOH, SO is pretty complex (and may be even more complex with time), so I'm a bit afraid that SO+FO will pass the boiling point and soon we'll split them back :D.

Reinmar:
tl;dr: Both options are fine for me (merging or method-based communication). But it seems that the latter will be quicker to code.

szymonkups:
During development we realized that if focus will be moved outside the document and then will be placed inside exact same place - `focus` will be fired but no `selectionchange` event. So the `FocusObserver` will wait for `SelectionObserver` which will never trigger it to render the document. Unfortunately `setTimeout` is the only solution for now.

src/view/observer/focusobserver.js
```diff
@@ -8,6 +8,10 @@
  */
 
 import DomEventObserver from './domeventobserver';
+import global from '@ckeditor/ckeditor5-utils/src/dom/global';
+
+const setTimeout = global.window.setTimeout;
+const clearTimeout = global.window.clearTimeout;
```
pjasiun: Can't you simple use Sinon fake timeout? http://sinonjs.org/releases/v2.0.0/fake-timers/

pjasiun: I mean: tests which ckecks if `setTimeout`/`clearTimeout` are called are too "unit" for me. They should check if `render` is called when it should be and is not called when it should not.

Reinmar: BTW, I haven't checked the code precisely, but IMO there's no point in using `global` if we're not stubbing something in the tests later. This module was initially created in order to allow stubbing globals.

Unless, of course, we decide that we should never use `window` and `document` directly. But it means that we'd need to change this in maaany places.

pjasiun: In this case, `global` is used because tests stub `setTimeout`.

Reinmar: In these particular tests I'd propose not to stub anything and have behaviour closest to the reality. The easier these tests will be the better because this is the most fragile part of our testing sute.

szymonkups: I was using `global` in a wrong way then - not for stubbing but to avoid direct access to globals. 
@pjasiun  - please look at the tests - they're checking if render was called or not. I will fix this to use sinon's fake timers.

szymonkups: Additionally I will add some integration test to make sure that it works as expected.

PR #865, opened 2017-03-09
Moved transferring markers from writer to batch

### LINKED ISSUE #862: Move transferring markers from from writer to batch
We decided that model `writer` will be a place where `DocumentFragment#markers` will be transferred to `Document#markers`. Unfortunately, this can be done here because:
1. it simply doesn't work
2. won't work for well for undo

I'll try to add markers as `MakerOperation` in `batch.insert`.

oskarwrobel:

Fix: Moved transferring markers from writer to batch. Closes #862.

---

### DIFF
--- a/src/model/delta/insertdelta.js
+++ b/src/model/delta/insertdelta.js
@@ -107,7 +107,14 @@ register( 'insert', function( position, nodes ) {
 	// When element is a DocumentFragment we need to move its markers to Document#markers.
 	if ( nodes instanceof DocumentFragment ) {
 		for ( const [ markerName, markerRange ] of nodes.markers ) {
-			this.setMarker( markerName, migrateRange( markerRange, position ) );
+			// We need to migrate marker range from DocumentFragment to Document.
+			const rangeRootPosition = Position.createAt( markerRange.root );
+			const range = new Range(
+				markerRange.start._getCombined( rangeRootPosition, position ),
+				markerRange.end._getCombined( rangeRootPosition, position )
+			);
+
+			this.setMarker( markerName, range );
 		}
 	}
 
@@ -115,26 +122,3 @@ register( 'insert', function( position, nodes ) {
 } );
 
 DeltaFactory.register( InsertDelta );
-
-// Migrates range to given position.
-//
-// @param {module:engine/model/range~Range} range
-// @param {module:engine/model/position~Position} position
-// @returns {module:engine/model/range~Range} Moved range.
-function migrateRange( range, position ) {
-	const positionRoot = position.parent.root;
-
-	// Clone paths.
-	let startPath = range.start.path.concat( [] );
-	let endPath = range.end.path.concat( [] );
-
-	// Move range out of DocumentFragment.
-	startPath.shift();
-	endPath.shift();
-
-	// Set Range on position.
-	startPath = position.path.concat( startPath );
-	endPath = position.path.concat( endPath );
-
-	return new Range( new Position( positionRoot, startPath ), new Position( positionRoot, endPath ) );
-}
--- a/src/model/delta/insertdelta.js
+++ b/src/model/delta/insertdelta.js
@@ -10,12 +10,12 @@
 import Delta from './delta';
 import DeltaFactory from './deltafactory';
 import RemoveDelta from './removedelta';
+import MarkerDelta from './markerdelta';
 import { register } from '../batch';
 import InsertOperation from '../operation/insertoperation';
+import MarkerOperation from '../operation/markeroperation';
 
 import DocumentFragment from '../documentfragment';
-import MarkerDelta from './markerdelta';
-import MarkerOperation from '../operation/markeroperation';
 import Range from '../../model/range.js';
 import Position from '../../model/position.js';
 
--- a/src/model/delta/insertdelta.js
+++ b/src/model/delta/insertdelta.js
@@ -106,15 +106,8 @@ register( 'insert', function( position, nodes ) {
 
 	// When element is a DocumentFragment we need to move its markers to Document#markers.
 	if ( nodes instanceof DocumentFragment ) {
-		for ( const marker of nodes.markers ) {
-			const doc = this.document;
-			const range = moveRange( marker[ 1 ], position );
-			const markerDelta = new MarkerDelta();
-			const markerOperation = new MarkerOperation( marker[ 0 ], null, range, doc.markers, doc.version );
-
-			this.addDelta( markerDelta );
-			markerDelta.addOperation( markerOperation );
-			doc.applyOperation( markerOperation );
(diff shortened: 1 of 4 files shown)


src/model/delta/insertdelta.js
```diff
@@ -95,7 +104,44 @@ register( 'insert', function( position, nodes ) {
 	delta.addOperation( insert );
 	this.document.applyOperation( insert );
 
+	// When element is a DocumentFragment we need to move its markers to Document#markers.
+	if ( nodes instanceof DocumentFragment ) {
+		for ( const marker of nodes.markers ) {
```
pjasiun: I this case destructuring might be fine (and nice) `for ( const [ markerName, markerRange ] of nodes.markers )`.

src/model/delta/insertdelta.js
```diff
@@ -95,7 +104,44 @@ register( 'insert', function( position, nodes ) {
...
 } );
 
 DeltaFactory.register( InsertDelta );
+
+// Moves range relative to given position.
+//
+// @param {module:engine/model/range~Range} range
+// @param {module:engine/model/position~Position} position
+// @returns {module:engine/model/range~Range} Moved range.
+function moveRange( range, position ) {
```
pjasiun: This function name is a little miss leading. You know, it sounds like you move range, but the range is the same. Maybe something like `migrateRanges` would be better.

src/model/delta/insertdelta.js
```diff
@@ -95,7 +104,44 @@ register( 'insert', function( position, nodes ) {
...
+	let startPath = range.start.path.concat( [] );
+	let endPath = range.end.path.concat( [] );
+
+	// Move range out of DocumentFragment.
+	startPath.shift();
+	endPath.shift();
+
+	// Set Range on position.
+	startPath = position.path.concat( startPath );
+	endPath = position.path.concat( endPath );
```
pjasiun: Isn't it done automatically by the position constructor?

pjasiun: https://github.com/ckeditor/ckeditor5-engine/blob/a7409be9830c7fa1fb9c921ee6b7eaaa0289e7c4/src/model/position.js#L66

pjasiun:
The change is fine for me. I commented some details.

src/model/delta/insertdelta.js
```diff
@@ -95,7 +104,44 @@ register( 'insert', function( position, nodes ) {
...
+	if ( nodes instanceof DocumentFragment ) {
+		for ( const marker of nodes.markers ) {
+			const doc = this.document;
+			const range = moveRange( marker[ 1 ], position );
+			const markerDelta = new MarkerDelta();
+			const markerOperation = new MarkerOperation( marker[ 0 ], null, range, doc.markers, doc.version );
+
+			this.addDelta( markerDelta );
+			markerDelta.addOperation( markerOperation );
+			doc.applyOperation( markerOperation );
```
pjasiun: You can call `this. setMarker` instead of doing it manually. It should be safer.

scofalik: Well, the funny thing about all of these is that the markers might be just temporary markers and you might not want to add them do the document, undo, etc.

scofalik: In fact, since we cannot (at the moment) use `Batch` API on `DocumentFragment` these markers are for sure not added to the document, so I hesitate whether we should add them to the document at all.

scofalik: On the other hand if you add markers with data that is added through `Batch` API it might be assumed that you want those markers synchronized (added to document, undo, etc.) too. I am fine with this.

scofalik: > You can call `this.setMarker` instead of doing it manually. It should be safer.

This is a good point I wanted to point the same (before I got the doubts I described above :P)

src/model/delta/insertdelta.js
scofalik: +1

scofalik: I also wonder whether one of many transformations methods in `Range` and `Position` may get handy, for example `Position#_getCombined`.

src/model/delta/insertdelta.js
oskarwrobel: I expect the same but:
- position where I want to insert element is `<root>abc[]</root>` -> path is `[ 3 ]`
- start position of my range in doc frag `<docfrag><p>d[]ef<p></docfrag>` `[ 0, 1 ]`

When I create position in this way `new Position( root, marker.start.path )` the result path will be `[ 0, 1 ]` but I need `[ 3, 0, 1 ]`.

src/model/delta/insertdelta.js
scofalik: If you use `_getTransformedByMove`, `sourcePosition` should be the first position in document fragment ( `Position.createAt( range.root )` ), `targetPosition` should be equal to passed `position` and `howMany` should be number of document fragment children.

Or you can go with `_getCombined`:
```javascript
return new Range(
	range.start._getCombined( Position.createAt( range.root ), position ),
	range.end._getCombined( Position.createAt( range.root ), position )
);
```

Try those and let me know if those worked.

src/model/delta/insertdelta.js
pjasiun: > On the other hand if you add markers with data that is added through Batch API it might be assumed that you want those markers synchronized (added to document, undo, etc.) too. I am fine with this.

Yea, I think that adding fragment with markers to the document is... adding markers to the document. One expects that they will be added. If I want to add only elements, I add elements and markers add to the `document.markers` directly, without `batch` API.

src/model/delta/insertdelta.js
pjasiun: Ach, I see. That's fine then.

src/model/delta/insertdelta.js
oskarwrobel: Looks like `Position#_getCombined` is exactly what I need 👌. Why is this method protected?

pjasiun:
I forget to copy a commit message, but it was an internal change which should be introduced with the feature at the first place.

oskarwrobel:
You can change message of last commit by `git commit --amend` and `git push origin master --force`

oskarwrobel:
But pushing with `--force` to the master is risky :)

pjasiun:
I know, but I don't think this message should go to the changelog. And I know @Reinmar will kill me for the wrong commit message, but I like living on the edge. :)

Reinmar:
You can't amend merge commit AFAIR. Just commit an empty commit right after the merge, this time with the proper message. That will be fine.

pjasiun:
You can, I just did it in `theme-lark`. But I agree that force push to the master is not a good idea in most cases.

PR #871, opened 2017-03-16
T/870 insertContent accept items

### LINKED ISSUE #870: `insertContent` should accept elements
Now `insertContent` accepts only a `DocumentFragment`. It could accept also `Element`s so the developer does not need to wrap a single `Element` into a `DocumentFragment` if he wants to insert it.

ma2ciek:

Feature: `DataController#insertContent()` now accepts also model items. Closes #870.


Reinmar:
It's a feature. Proposed message:

> Feature: `DataController#insertContent()` now accepts also model items. Closes #870.

src/controller/insertcontent.js
```diff
@@ -42,9 +42,17 @@ export default function insertContent( dataController, content, selection, batch
 
 	const insertion = new Insertion( dataController, batch, selection.anchor );
 
-	insertion.handleNodes( content.getChildren(), {
+	let nodesToInsert;
+
+	if ( content.is( 'documentFragment' ) ) {
 		// The set of children being inserted is the only set in this context
 		// so it's the first and last (it's a hack ;)).
```
pjasiun: The comment is in the wrong place now.

pjasiun:
But the change is fine, so as soon as you move the comment in the proper place the branch can be merged.

Reinmar:
Please add an explicit test for the new feature.

ma2ciek:
I removed the part of [util function](https://github.com/ckeditor/ckeditor5-engine/pull/871/files#diff-1c53d2450cb4e39a53a6bd3a8bc240caL615) that was wrapping elements in `DocumentFragment`, so except the first test which explicitly uses `DocumentFragment`, the rest is changed since the function should accept other elements. So for now 99% of the tests covers this change.

Reinmar:
The problem is that by reading the tests you can't easily check that. Tests are documentation too. Here, to know what kinds of arguments this function accept you need to scan through all the tests (because, as you said, only 1% uses DF) and also, you need to know what kind of results the `parse()` function returns. As a matter of fact – I didn't know that myself when writing these tests (I was assuming it's always returning DF) so I had to debug why the heck my tests aren't passing.

Finally, we may one day decide to change what `parse()` returns (so it always returns a DF) and we would silently lose a test for how `insertContent()` works with elements. These won't happen if we will have explicit tests for element and DF.

ma2ciek:
I agree that it's better to make tests more explicit. But it's hard for me to figure out how they should look like for now. There're 69 tests in `insertContent` and 68 uses of `insertHelper()` util. Should I write ~5 tests for DF and 5 tests for the Element and let the `insertHelper()` util to use DF?

Reinmar:
> There're 69 tests in insertContent and 68 uses of insertHelper() util. Should I write ~5 tests for DF and 5 tests for the Element and let the insertHelper() util to use DF?

No, you just need to write 2 tests – one which uses `insertContent()` with an element that you created manually. And one which uses `insertContent()` with a DF that you created. Only this is missing. The rest of the code and tests are fine now.

PR #872, opened 2017-03-16
T/857 View placeholder

### LINKED ISSUE #857: Create utility for creating placeholders
We need an utility that will help us to create placeholders in view elements. 
Placeholder should be created using CSS solution, for example, with `::before` or `::after` pseudo elements.

szymonkups:

Feature: Added placeholder utility that can be applied to view elements. Closes #857.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-engine/pull/864.


src/view/placeholder.js
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * @module engine/view/placeholder
+ */
+
+import extend from '@ckeditor/ckeditor5-utils/src/lib/lodash/extend';
+import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+
+const CSS_CLASS = 'ck-placeholder';
+const DATA_ATTRIBUTE = 'data-placeholder';
```
pjasiun: I remember I did something like this in my first PR to the CKEditor 4. And I remember that Fred gave me R- then telling it an unnecessary complication. Now, I see that we use it more and more. And to some point I like it.

But on the other hand, it's 2017 and all of us have IDEs which will easly find for us all `data-placeholder` and replace with anything we want. Also, I think that if one day we will want to replace it with something else, it probably will be a configurable variable so it will have to be changed anyway. At the same time, the fact that I have `data-placeholder` in the Chrome dev tools, and `DATA_ATTRIBUTE` in the code, and need to check and remember that it equals does not help to debug. It might not be hard with one variable, but we get more and more of these. I know, that it's a popular good practice, but, my question is: what do we get?

pjasiun: And I even think I know the answer. These kind of constants is useful to avoid magic strings and magic number. For instance:

```
const BLOCKS_SPACING = 15;

// ...

createBlock( BLOCK_SPACING );
```

is far more readable them:
```
createBlock( 15 );
```

But in this case, I'm not sure if these 2 are magic strings. I don't see anything hard to understand in:
```
element.setAttribute( 'data-placeholder', placeholderText );
```

I even like it more that the current version, because I immediately know what I need to look for in the browser's inspector in case of debugging.

So my proposal is to use this kind of constants only when the value or/and the context of its usage might be unclear and adding named const improve readability.

src/view/placeholder.js
```diff
@@ -0,0 +1,127 @@
...
+
+const listener = {};
+extend( listener, EmitterMixin );
+
+// Each document stores information about its placeholder elements and check functions.
+const documentPlaceholders = new Map();
```
pjasiun: Why not a `WeakMap`?

src/view/placeholder.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @param {module:engine/view/element~Element} element
+ */
+export function detachPlaceholder( element ) {
+	element.removeClass( CSS_CLASS );
+	element.removeAttribute( DATA_ATTRIBUTE );
+
+	for ( let placeholders of documentPlaceholders.values() ) {
+		placeholders.delete( element );
+	}
```
pjasiun: If you would change `documentPlaceholders` to the `WeakMap` you will have to change the loop to: `documentPlaceholders( element.document ).delete( element );`. But I think it worth to change it anyway.

pjasiun:
And one more thought: the `view/placeholder.js` file is. In fact, a singleton class. With public and private methods, properties. Don't get be wrong, I like how it is implemented, but I might spend too much time with OOP ;)

Reinmar:
> And one more thought: the view/placeholder.js file is, in fact, a singleton class.

It's a module. And indeed – a module can be like a singleton class. The exported stuff is the public API, the rest is private. It's exactly the same what we used to do with:

```js
const api = ( ( function() {
  var priv = [];

  return { 
    add: function( priv ) {
      priv.push( item );
    }
} )();
```

But a module is nicer :)

src/view/placeholder.js
szymonkups: You can't iterate over `WeakMap` items.

src/view/placeholder.js
szymonkups: I don't see many of this in our code. I've recently used this approach when I was working on CSS classes in image and widgets, but same as you - I have a feeling that we could get rid of this.

src/view/placeholder.js
pjasiun: In fact what you need is to add a `placeholders` property to the `document`, but in a nice way, without modifying `document` object. `WeakMap` do exactly this.

src/view/observer/focusobserver.js
```diff
@@ -8,6 +8,10 @@
  */
 
 import DomEventObserver from './domeventobserver';
+import global from '@ckeditor/ckeditor5-utils/src/dom/global';
+
+const setTimeout = global.window.setTimeout;
+const clearTimeout = global.window.clearTimeout;
```
pjasiun: Can we remove this?

tests/view/observer/focusobserver.js
```diff
@@ -3,11 +3,13 @@
...
 
-/* globals document */
-
 import FocusObserver from '../../../src/view/observer/focusobserver';
 import ViewDocument from '../../../src/view/document';
 import ViewRange from '../../../src/view/range';
+import global from '@ckeditor/ckeditor5-utils/src/dom/global';
+
+const setTimeout = global.window.setTimeout;
+const domDocument = global.document;
```
pjasiun: And this?

tests/view/observer/selectionobserver.js
```diff
@@ -3,31 +3,35 @@
...
+import FocusObserver from '../../../src/view/observer/focusobserver';
 import log from '@ckeditor/ckeditor5-utils/src/log';
-
+import global from '@ckeditor/ckeditor5-utils/src/dom/global';
 import { parse } from '../../../src/dev-utils/view';
 
+const setTimeout = global.window.setTimeout;
+const setInterval = global.window.setInterval;
+const clearInterval = global.window.clearInterval;
+const domDocument = global.document;
```
pjasiun: And this? ;)

src/controller/editingcontroller.js
```diff
@@ -23,6 +23,7 @@ import {
 } from '../conversion/model-selection-to-view-converters';
 
 import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
+import '../../theme/theme.scss';
```
pjasiun: I know that we used to have global `theme.scss` per repository, but since we use Webpack, we could load styles more locally. This is a perfect example of it. We could have `placeholder.scss` and import it from `placeholder.js`. Thanks to these styles will not be added to the build as long as someone will not include the `js` file (there might be no usage of the `placeholder.js` and, what is more important, it's very simple to find styles used in that JS fine. You just go to import like you would do with a function. //cc @oleq @Reinmar

oleq: If it really worked, we could get rid of nightmares like https://github.com/ckeditor/ckeditor5-link/blob/master/theme/theme.scss#L4. Waiting for confirmation from Reinmar, then I'm gonna create an umbrella issue in [ckeditor5](https://github.com/ckeditor/ckeditor5) because it applies to the entire codebase.

pjasiun: Yep, it works great. I use it a lot in another project.

oleq: I second that then.

szymonkups: I was thinking that this might be connected with a way how at the end we will build editor with proper theme. I am not sure if this is defined yet but let's wait for @Reinmar response.

Reinmar: Well, there are important things to remember here.

Importing an SCSS file many times may lead to code duplication and bugs. Even if `placeholder.scss` will be imported only once, it may be importing other SCSS files which are imported somewhere else already, making those rules appear twice. This is super sad and dangerous. I hit this hard when trying to start importing SCSS files in their proper JS components. It blew up.

The solution here is a refactoring of our SCSS. The code needs to be split into two layers – helpers and components. Helpers must not contain any rules – they should only contain mixins, functions, etc. This means that you will be able to import a helper multiple times without duplicating the code. The components will use helpers to define final CSS rules. However, a single component must not be imported more than once.

The second thing is that we must support multiple themes. No one so far made any research about that and the only related ticket so far is https://github.com/ckeditor/ckeditor5-ui/issues/144. But it clarifies that plugins (and engine) must not depend on a particular theme. This is a long topic and it doesn't make sense to spend time on it now, but how it applies to this particular case? It doesn't. The import can be anywhere. But the SCSS file should, at least for now, have no dependency on the theme-lark package.

oleq: That's a good point, actually. If we import things directly, switching to another theme will require changes in each single js file that imports SCSS vs. changing theme.scss file only. I wonder whether this could be possible.

pjasiun:
The change is fine for me. You can merge it as soon as you rebase the branch.

pjasiun:
Nvmd

PR #873, opened 2017-03-16
T/847: Move cannot be converted

### LINKED ISSUE #847: Move cannot be converted
Starting model:

```html
<messageQuote>
  <paragraph>abc</paragraph>
  <paragraph>xx</paragraph>
  <paragraph>def</paragraph>
</messageQuote>
```

Starting view:

```html
<div><blockquote><p>abc</p><p>xx</p><p>def</p></blockquote></div>
```

I want to unwrap that `<messageQuote>`. It first triggers move of its content before it:

```html
<paragraph>abc</paragraph>
<paragraph>xx</paragraph>
<paragraph>def</paragraph>
<messageQuote></messageQuote>
```

From what I can see, the `remove()` model-to-view-converter is called with the `sourcePostion`: `[0, 0]` and `offsetSize`: `1` which should map to:

```html
<div><blockquote>[<p>abc</p>]<p>xx</p><p>def</p></blockquote></div>
```

But apparently converts to:

```html
<div><blockquote><p>{a}bc</p><p>xx</p><p>def</p></blockquote></div>
```

;|

I can see that at that moment, the model looks like this already:

```html
<paragraph>abc</paragraph>
<paragraph>xx</paragraph>
<paragraph>def</paragraph>
<messageQuote></messageQuote>
```

So the `[ 0, 0 ]` position + 1 offset is clearly `<paragraph>[a]bc</paragraph>...`. Apparently, the `sourcePosition` needs to be transformed before move is converted.

Reinmar:

Fix: Fixed various issues with the move and unwrap deltas conversion. Closes #847.

---

### Additional information

Turns out that the plan we had in https://github.com/ckeditor/ckeditor5-engine/issues/847#issuecomment-284003395 wasn't enough. After changing the order in which remove and insert are fired I encountered a problem with a lost binding. It turned out that the remove operation was removing the wrong binding due to a duplicated view element. It was removing the freshly inserted view element's binding to the model element.

We've seen two options:

* Either checking where the mode element went in the meantime. If not to the graveyard then we decide to not remove any binding.
* However, the above solution leaves a binding from the old view element to the model element. So, a more correct solution could be if the mapper would check if (v->m; m->v2; v != v2) and if so, remove only the v->m binding. However, it means that `remove()` converter would need to iterate through elements in the view range to remove and pass all them to `unbindViewElement()`. This is a bit more work for a developer, so we decided to pick a shorter solution. In the future, we can think whether the mapper couldn't somehow do this by itself.

### DIFF
--- a/src/conversion/model-to-view-converters.js
+++ b/src/conversion/model-to-view-converters.js
@@ -440,7 +440,23 @@ export function remove() {
 		const viewRange = conversionApi.mapper.toViewRange( modelRange );
 
 		viewWriter.remove( viewRange.getTrimmed() );
-		conversionApi.mapper.unbindModelElement( data.item );
+
+		// Unbind this element only if it was moved to graveyard.
+		// The dispatcher#remove event will also be fired if the element was moved to another place (remove+insert are fired).
+		// Let's say that <b> is moved before <a>. The view will be changed like this:
+		//
+		// 1) start:    <a></a><b></b>
+		// 2) insert:   <b (new)></b><a></a><b></b>
+		// 3) remove:   <b (new)></b><a></a>
+		//
+		// If we'll unbind the <b> element in step 3 we'll also lose binding of the <b (new)> element in the view,
+		// because unbindModelElement() cancels both bindings – (model <b> => view <b (new)>) and (view <b (new)> => model <b>).
+		// We can't lose any of these.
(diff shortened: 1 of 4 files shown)


pjasiun:
Both the change and your proposal future solution are fine for me.

PR #879, opened 2017-03-20
Range#_getTransformedByDocumentChange will be "inclusive"

### LINKED ISSUE #877: LivePosition is still broken after move
Initial live range:

```html
<messageQuote><paragraph>abc</paragraph><paragraph>x[x</paragraph></messageQuote><paragraph>d]ef</paragraph>
```

The move (`<p>` out of `<messageQuote>`):

```html
<messageQuote><paragraph>abc</paragraph>[<paragraph>xx</paragraph>]</messageQuote>[]<paragraph>def</paragraph>
```

scofalik:

Other: `Range#_getTransformedByDocumentChange` will now include model items when range boundary is moved "out of" range. Closes #877.

BREAKING CHANGE: Significantly changed how `Range#_getTransformedByDocumentChange` works, which is a base for `LiveRange` and `LiveSelection` and markers transformations. See https://github.com/ckeditor/ckeditor5-engine/issues/877#issuecomment-287740021 for reasoning behind the change and examples.

### DIFF
--- a/src/model/delta/basic-transformations.js
+++ b/src/model/delta/basic-transformations.js
@@ -111,6 +111,8 @@ addTransformationCase( MarkerDelta, SplitDelta, transformMarkerDelta );
 addTransformationCase( MarkerDelta, MergeDelta, transformMarkerDelta );
 addTransformationCase( MarkerDelta, WrapDelta, transformMarkerDelta );
 addTransformationCase( MarkerDelta, UnwrapDelta, transformMarkerDelta );
+addTransformationCase( MarkerDelta, MoveDelta, transformMarkerDelta );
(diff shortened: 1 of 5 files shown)


Reinmar:
I like how the code actually got simpler. Looks fine to me and make MQ tests pass.

Reinmar:
An interesting thing about the changelog – it should be mostly focused on public API (see how I rewrote the message to focus first on what changed to the API user). Also, when doing any big fix we're actually doing some breaking change, so the fact that some behaviour has changed isn't necessarily a breaking change. Of course, some fixes may have such a wide effect that they may need to be marked as breaking – but this is something which doesn't have to concern us at this stage. It will be a problem for the future :D.

PR #885, opened 2017-03-21
T/884 Mapper: default position mapping algorithms may crash editor

### LINKED ISSUE #884: Mapper: default position mapping algorithms may crash editor.
At the moment, model-to-view and view-to-model position mapping use first the default algorithm and then additional callbacks may be fired and alter the position calculated using default algorithm.

This behaviour sounds okay, but is a problem in `List` feature because for some list structures the default algorithm throws.

Hence I propose that default algorithm should be used only if no custom callback has been used.

scofalik:

Other: Default `conversion.Mapper` position mapping algorithms are now added as callbacks with `low` priority and are fired only if earlier callbacks did not provide a result. Closes #884.

BREAKING CHANGES: Since default position mapping algorithms are attached with `low` priority, custom position mapping callbacks added with higher priority won't receive position calculated by default algorithms in `data`. To execute default position mapping algorithms and use their value, hook custom callback with lower priority.

### DIFF
--- a/src/conversion/mapper.js
+++ b/src/conversion/mapper.js
@@ -61,6 +61,36 @@ export default class Mapper {
 		 * @member {Map}
 		 */
 		this._viewToModelLengthCallbacks = new Map();
+
+		// Default mapper algorithm for mapping model position to view position.
+		this.on( 'modelToViewPosition', ( evt, data ) => {
+			if ( data.viewPosition ) {
+				return;
+			}
+
+			let viewContainer = this._modelToViewMapping.get( data.modelPosition.parent );
+
+			data.viewPosition = this._findPositionIn( viewContainer, data.modelPosition.offset );
+		}, { priority: 'low' } );
+
+		// Default mapper algorithm for mapping view position to model position.
+		this.on( 'viewToModelPosition', ( evt, data ) => {
+			if ( data.modelPosition ) {
+				return;
+			}
+
+			let viewBlock = data.viewPosition.parent;
+			let modelParent = this._viewToModelMapping.get( viewBlock );
+
+			while ( !modelParent ) {
+				viewBlock = viewBlock.parent;
+				modelParent = this._viewToModelMapping.get( viewBlock );
+			}
+
+			let modelOffset = this._toModelOffset( data.viewPosition.parent, data.viewPosition.offset, viewBlock );
+
+			data.modelPosition = ModelPosition.createFromParentAndOffset( modelParent, modelOffset );
+		}, { priority: 'low' } );
 	}
 
 	/**
@@ -164,20 +194,7 @@ export default class Mapper {
 
 		this.fire( 'viewToModelPosition', data );
 
-		return data.modelPosition ? data.modelPosition : this._defaultToModelPosition( viewPosition );
-	}
-
-	/**
-	 * Maps model position to view position using default mapper algorithm.
-	 *
-	 * @private
-	 * @param {module:engine/model/position~Position} modelPosition
-	 * @returns {module:engine/view/position~Position} View position mapped from model position.
-	 */
-	_defaultToViewPosition( modelPosition ) {
-		let viewContainer = this._modelToViewMapping.get( modelPosition.parent );
-
-		return this._findPositionIn( viewContainer, modelPosition.offset );
+		return data.modelPosition;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


src/conversion/mapper.js
```diff
@@ -159,13 +159,12 @@ export default class Mapper {
...
 		const data = {
 			viewPosition: viewPosition,
-			modelPosition: this._defaultToModelPosition( viewPosition ),
 			mapper: this
 		};
 
 		this.fire( 'viewToModelPosition', data );
 
-		return data.modelPosition;
+		return data.modelPosition ? data.modelPosition : this._defaultToModelPosition( viewPosition );
```
pjasiun: Can't we put `this._defaultToModelPosition( viewPosition )` as an event listener? Someone may need a mapped position do calculate his version.

pjasiun: In fact I was curious how such custom mapping may look like, how to create custom mapping without the result of the default mapping, and I realised there is no single test for this change.

src/conversion/model-to-view-converters.js
```diff
@@ -436,8 +436,12 @@ export function remove() {
...
 		}
 
-		const modelRange = ModelRange.createFromPositionAndShift( data.sourcePosition, data.item.offsetSize );
-		const viewRange = conversionApi.mapper.toViewRange( modelRange );
+		// Note: at this moment we should avoid mapping non existing model positions.
+		// TODO: try to simplify this if `Mapper` ever changes.
+		const viewPosition = conversionApi.mapper.toViewPosition( data.sourcePosition );
+		const viewRange = data.item.is( 'element' ) ?
+			ViewRange.createOn( viewPosition.nodeAfter ) :
+			_expandViewPositionOnText( viewPosition, data.item.offsetSize );
```
pjasiun: WTF?!

src/conversion/mapper.js
scofalik: > Can't we put this._defaultToModelPosition( viewPosition ) as an viewToModelPosition event listener?

We can, it was like this for some time some time ago.

> In fact I was curious how such custom mapping may look like, how to create custom mapping without the result of the default mapping, and I realised there is no single test for this change.

I don't understand what exactly you mean. There are test for adding a custom callback and there are tests for default mapping. What exactly do you need?

pjasiun: > We can, it was like this for some time some time ago.

I wanted to say in the ticket that I remember that it was already possible to add the custom listener before the default mapping and I don't know why it's not possible anymore, but I thought it was a dream ;)

pjasiun: > There are test for adding a custom callback and there are tests for default mapping. What exactly do you need?

Could you link where are these tests, because I may miss something.

scofalik: > I wanted to say in the ticket that I remember that it was already possible to add the custom listener before the default mapping and I don't know why it's not possible anymore, but I thought it was a dream ;)

Yes, it was like this for a while but then it got changed. I don't remember the exact reasons, though. I can change it so the default mapping callback is fired with `low` priority. Not a big deal.

> Could you link where are these tests, because I may miss something.

I meant these:
https://github.com/ckeditor/ckeditor5-engine/blob/master/tests/conversion/mapper.js#L209-L223
https://github.com/ckeditor/ckeditor5-engine/blob/master/tests/conversion/mapper.js#L282-L296

I haven't added new tests because, from API point of view, in fact nothing changed, other than the fact that you don't get "default" position in the callback.

pjasiun: I see that there were tests. As I said I was curious how such custom converter may look like if I do not have access to the default mapping. Tests is the place where I look for the sample use-case and I did not find any example how I could use this event.

scofalik: So, you mean something that resembles real case scenario. Well, the problem with such tests is that you need to think of some non-dumb case/feature and then code it. So you either go easy way (like the linked tests) or start re-writing other feature (i.e. List). Maybe it would be better to drop some reference in the event description (to List feature?).

scofalik: BTW. I just noticed that there's example in event's description.

pjasiun: > BTW. I just noticed that there's example in event's description.

I see. That's fine then.

scofalik:
Fixed.

PR #886, opened 2017-03-22
T/869 `insertContent` shouldn't clone given nodes

### LINKED ISSUE #869: `insertContent` should not clone nodes
Because of the clone in the `insertContent` method: https://github.com/ckeditor/ckeditor5-engine/blob/3e3839035c7d03af678b82eefea8fea83c66a625/src/controller/insertcontent.js#L130 the reference to the inserted node is broken. Inserting nodes should not be cloned.

ma2ciek:

Fix: Changed `insertContent` behaviour, so it doesn't clone given nodes. Closes #869.

### DIFF
--- a/tests/controller/insertcontent.js
+++ b/tests/controller/insertcontent.js
@@ -9,6 +9,7 @@ import insertContent from '../../src/controller/insertcontent';
 
 import DocumentFragment from '../../src/model/documentfragment';
 import Text from '../../src/model/text';
+import Element from '../../src/model/element';
 
 import { setData, getData, parse } from '../../src/dev-utils/model';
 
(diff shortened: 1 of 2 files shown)


tests/controller/insertcontent.js
```diff
@@ -62,6 +62,57 @@ describe( 'DataController', () => {
 			expect( getData( doc ) ).to.equal( 'xa[]x' );
 		} );
 
+		describe( 'saves the reference to the original object', () => {
+			it( 'inline-widget', () => {
```
pjasiun: It would be nice if the test name would create a sentence:

```
describe( 'insertContent' ...
it( 'should saves the reference to the original object' ...
```

It's not needed in all cases, for instance there is no reason to create strange sentences, if you test a function with a set of data, but here it's pretty nice and simple.

Also, I don't think we need a separate describe block here.

tests/controller/insertcontent.js
```diff
@@ -62,6 +62,57 @@ describe( 'DataController', () => {
...
+
+				dataController = new DataController( doc );
+
+				const batch = doc.batch();
+
+				setData( doc, '<paragraph>f[]oo</paragraph>' );
+
+				const content = parse( '<inlineWidget></inlineWidget>', doc.schema, {
+					context: [ '$clipboardHolder' ]
+				} );
```
pjasiun: That's a very strange way of creating an element. Why not `new Element(inlineWidget);`

tests/controller/insertcontent.js
```diff
@@ -62,6 +62,57 @@ describe( 'DataController', () => {
...
+
+				insertContent( dataController, content, doc.selection, batch );
+
+				expect( doc.getRoot().getChild( 0 ).getChild( 1 ) ).to.equal( content );
+			} );
+
+			it( 'image', () => {
```
pjasiun: Is the name of the element only difference between these 2 tests? If so, please remove one of them.

ma2ciek: No, in the first one I'm inserting inlineWidget and in second image, but I can remove the first one of course if only the second one is necessary.

tests/controller/insertcontent.js
ma2ciek: Ok, If this enhancement requires only one test, I can change it to your proposal. But take a look at the whole file. It uses a little bit different naming in descriptions.

tests/controller/insertcontent.js
ma2ciek: Hmm, I took this from the https://github.com/ckeditor/ckeditor5-engine/blob/a80abe3e2af8ca0703134b844ebf03441b5dbdd3/tests/controller/insertcontent.js#L693. But of course such a simple elements should be created in the normal way.

tests/controller/insertcontent.js
```diff
@@ -62,6 +62,57 @@ describe( 'DataController', () => {
...
+				doc.schema.objects.add( 'inlineWidget' );
+
+				dataController = new DataController( doc );
+
+				const batch = doc.batch();
+
+				setData( doc, '<paragraph>f[]oo</paragraph>' );
+
+				const content = parse( '<inlineWidget></inlineWidget>', doc.schema, {
+					context: [ '$clipboardHolder' ]
```
Reinmar: AFAICT, there's no sense for these particular tests to use a special context. I'm actually a bit unsure why I used it in the other tests (more precisely – in the ones using `insertHelper()`). It might've been to align them as close as possible to pasting scenarios and since I was using a helper I got lazy and used the same context everywhere. But since you're not using the helper in the new tests, don't use that context too.

ma2ciek:
I updated the tests regarding your comments.

PR #892, opened 2017-03-27
Reversed `ReinsertOperation` targets back to same graveyard holder

### LINKED ISSUE #891: Reversed ReinsertOperation should remove nodes back to the same graveyard holder
Whenever new `RemoveOperation` is created, `$graveyard` root is checked for how many `$graveyardHolder` nodes it has and the target of `RemoveOperation` is set to the new `$graveyardHolder` that will be inserted during operation execution at the end of `$graveyard`.

Simple example of situation when editor throws. Consider a batch that removes a character and removes bold from it afterwards. It's based on real scenario where post fixer callback fixes model and adds it's changes to the same batch (original issue: https://github.com/ckeditor/ckeditor5-list/issues/51):

1. `REMOVE` from `$root [ 0, 1 ]`, `1` node, to `$graveyard [ 0, 0 ]`.
2. `REMOVE ATTRIBUTE "bold"` from `$graveyard [ 0, 0 ] - [ 0, 1 ]`.

When it is undone, following batch is generated:
1. `SET ATTRIBUTE "bold"=true` on `$graveyard [ 0, 0 ] - [ 0, 1 ]`.
2. `REINSERT` from `$graveyard [ 0, 0 ]`, `1` node, to `$root [ 0, 1 ]`.

Now redo:
1. `REMOVE` from `$root [ 0, 1 ]`, `1` node, to `$graveyard [ 1, 0 ]` (new remove operation, creates new graveyard holder).
2. `REMOVE ATTRIBUTE "bold"` from `$graveyard [ 0, 0 ] - [ 0, 1 ]`. <-- this throws, as there are no nodes in `[ 0, 0 ]`.

It makes sense that `RemoveOperation` created from `ReinsertOperation` reversing should put nodes back in their original place. The only scenario where `ReinsertOperation` is reversed is redo.

scofalik:

Fix: Reversed `ReinsertOperation` targets back to same graveyard holder from which the nodes were re-inserted. Closes #891.

### DIFF
--- a/src/model/operation/reinsertoperation.js
+++ b/src/model/operation/reinsertoperation.js
@@ -45,7 +45,12 @@ export default class ReinsertOperation extends MoveOperation {
 	 * @returns {module:engine/model/operation/removeoperation~RemoveOperation}
 	 */
 	getReversed() {
-		return new RemoveOperation( this.targetPosition, this.howMany, this.baseVersion + 1 );
+		const removeOp = new RemoveOperation( this.targetPosition, this.howMany, this.baseVersion + 1 );
+
+		// Make sure that nodes are put back into the `$graveyardHolder` from which they got reinserted.
+		removeOp.targetPosition = this.sourcePosition;
+
+		return removeOp;
 	}
 
 	/**
--- a/src/model/operation/reinsertoperation.js
+++ b/src/model/operation/reinsertoperation.js
@@ -49,6 +49,7 @@ export default class ReinsertOperation extends MoveOperation {
 
 		// Make sure that nodes are put back into the `$graveyardHolder` from which they got reinserted.
 		removeOp.targetPosition = this.sourcePosition;
+		removeOp._needsHolderElement = false;
 
 		return removeOp;
 	}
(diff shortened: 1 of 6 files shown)


scofalik:
I pushed an integration test for this change to `ckeditor5-undo` repo: https://github.com/ckeditor/ckeditor5-undo/tree/ckeditor5-engine/t/891

Reinmar:
I can still reproduce issues with lists on this PR:

mar-27-2017 19-07-17

The following error is thrown:

image

Is it anyhow related?

scofalik:
Probably, I'll look into this...

Reinmar:
So I'll mark this as R- for now.

scofalik:
I've pushed a commit that should solve this issue. I wanted to propose that fix (from the last commit) anyway as a followup issue, but wasn't aware it was already needed.

Reinmar:
Will you also push a test to undo?

scofalik:
Yes

scofalik:
Done.

PR #893, opened 2017-03-27
T/889 Fixing tests in SelectionObserver

### LINKED ISSUE #889: SelectionObserver test is unstable
https://travis-ci.org/ckeditor/ckeditor5/builds/214216219#L1295

"SelectionObserver should warn and not enter infinite loop" started failing recently.

szymonkups:

Fix: Improved tests stability by changing how infinite loop detection works in SelectionObserver. Closes #889.

---

### Additional information

We decided to totally remove checking if selection is same as previous. Only counter is used now and I've manually tested its settings. I've decided that it will be cleared each second and warning will be showed if counter is greater than `200` before that time. I've managed to get the counter to `~150` by dragging selection around inline elements.

### DIFF
--- a/src/view/observer/selectionobserver.js
+++ b/src/view/observer/selectionobserver.js
@@ -148,7 +148,9 @@ export default class SelectionObserver extends Observer {
 		}
 
 		// Ensure we are not in the infinite loop (#400).
-		if ( this._isInfiniteLoop() ) {
+		// This counter is reset every 1 seconds. 200 selection changes in 1 seconds is enough high number
+		// to be very difficult (impossible) to achieve using just keyboard keys (during normal editor use).
+		if ( ++this._loopbackCounter > 200 ) {
 			/**
 			 * Selection change observer detected an infinite rendering loop.
 			 * Most probably you try to put the selection in the position which is not allowed
@@ -176,19 +178,6 @@ export default class SelectionObserver extends Observer {
 		// defined int the function time will elapse since the last time the function was called.
 		// So `selectionChangeDone` will be fired when selection will stop changing.
 		this._fireSelectionChangeDoneDebounced( data );
-	}
-
-	/**
-	 * Checks if selection rendering entered an infinite loop.
-	 *
-	 * See https://github.com/ckeditor/ckeditor5-engine/issues/400.
-	 *
-	 * @private
-	 * @param {module:engine/view/selection~Selection} newSelection DOM selection converted to view.
-	 * @returns {Boolean} True is the same selection repeat more then 10 times.
-	 */
-	_isInfiniteLoop() {
-		return ++this._loopbackCounter > 200;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Reinmar:
Hm... I'm afraid that this check stops making any sense when it has to hit 200 in 1s. Even if we'll have a real inf loop it may happen that it simply won't exceed this number.

Actually, this is exactly what happens now. The issue we have with `selectionChange` being fired up to 100 times per second when you drag the mouse is exactly an infinite loop (see #887). So, instead of avoiding catching it we should exactly catch those moments. I feel (after doing a very quick check ;)) that the right number is something around 50 events per second. I was able to trigger around ~30 normal selection change events by dragging (up to ~50 when I moved my mouse super quick) and the results were very similar on FF and Chrome. Any number higher than that has a high probability of being an inf loop. And it's better to log a warning and cancel `selectionChange` event (which has nearly invisible repercussions in such a case) than let the browser fall into inf loop.

Therefore, let's go lower with the number and see how often we'll get the warning. This will be a good way of checking how we do with #887. (cc @f1ames)

I'd propose 60 events per sec. This is around the frame rate at which I'd suppose the browser may be interested in checking the mouse position.

pjasiun:
Note that this check is not any real solution. It's here only to prevent browser freezing in case of the infinite loop. We should not use it to prevent or hide any real bug, only change freeze into console message.

Despite that we should not show a warning in cases we accept as normal behavior, I agree with @Reinmar that we could use this mechanism also to find buggy behavior, so it makes sense to set this limit low.

Reinmar:
> We should not use it to prevent or hide any real bug, only change freeze into console message.

Exactly. That's why we're working on #887. But exactly that bug should be caught by this check.

Reinmar:
There's one thing which worries me:
mar-28-2017 16-58-49

This happens on master and on this branch. We should get just one warning, not 60. I somehow understand why is this happening, but I have a feeling it's not good.

Reinmar:
BTW, it always happens exactly 63 times :D.

Reinmar:
I made such a change to catch any `render()` calls after the inf loop was caught:

image

Turns out that `render()` is never called after that. Which means that those native `selectionChange` events which we're handling are not triggered by us. Which means that with a high certainty they are triggered before we stop rendering and are in some kind of a buffer waiting to trigger our listener. And it even makes quite a lot of sense.

Reinmar:
I wanted to say – let's reset the counter after logging the warning... but nope – this is a very wrong idea. The stacked events will cause rendering on and on.

Reinmar:
This does the trick:

```js
this.stopListening( domDocument, 'selectionchange', this._callback );
setTimeout( () => {
	this.listenTo( domDocument, 'selectionchange', this._callback );
} );
```

But I guess it's unnecessary. Since render() isn't called after we kill the loop we're pretty safe. Only the warning gets duplicated.

PR #896, opened 2017-03-29
T/887 Renderer should not rerender similar non-collapsed selection.

### LINKED ISSUE #880: Infinite selectionChange event
### How to reproduce?

1. We need to enqueue some operation on the `ModelDocument` (can be empty function) on `selectionChangeDone` event.
```javascript
editor.editing.view.on( 'selectionChangeDone', () => {
	editor.document.enqueueChanges( () => {} );
} );
```

2. Place non-collapsed selection in the way that will end just after text with attributes.
```html
<paragraph>F[o<$text bold="true">o b</$text>]ar</paragraph>
```

mar-20-2017 13-19-13

### LINKED ISSUE #887: Renderer should not rerender similar non-collapsed selection.
See https://github.com/ckeditor/ckeditor5-typing/issues/87 for inital discussion.

There are some cases in which selection changes (the start/end position), but _visually_ it is the same. For example:
`<p><em>Text{1}</em></p>`
and
`<p><em>Text1[1</em>]</p>`

In such cases, we call it **similar selection** (basically, it is not the same, but very similar - only start/end container is different but offsets are somehow the same). In those cases the renderer should detect it and do not rerender the selection (because e.g. it breaks composition sometimes - https://github.com/ckeditor/ckeditor5-typing/issues/87).

This applies only to non-collapsed selection for now.

For collapsed selection it may break a few things, e.g. inserting empty `<strong></strong>` (with <kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>b</kbd>). The selection here is fixed and moved inside empty tag. It is a **similar selection** so it would not be rendered which is an invalid behavior.

f1ames:

Fix: Renderer does not rerender similar non-collapsed selection. Closes #887. Closes #880.

---

### Additional information

While working on this fix I have some doubts where logic checking if ranges are similar should be placed. It seems like a good place should be `view/selection` class (then it will have `isEqual` and `isSimilar` methods). But then it should also have method which trimms the selection (which is basically a selection with trimmed ranges).
Also the renderer is the one to decide how to check if selections are similar so any custom logic (some might be needed in the future) should be placed here. That was the reason the logic is placed in the `view/renderer` class (however, still not 100% sure about it).

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -582,13 +582,13 @@ export default class Renderer {
 			return;
 		}
 
-		if ( oldViewSelection && this._areSimilarSelections( oldViewSelection, this.selection ) ) {
+		if ( oldViewSelection && areSimilarSelections( oldViewSelection, this.selection ) ) {
 			const data = {
 				oldSelection: oldViewSelection,
 				currentSelection: this.selection
 			};
 
-			log.warn( 'renderer-selection-similar: Rendering selection was skipped due to its similarity to the previous selection.', data );
+			log.warn( 'renderer-skipped-selection-rendering: The selection was not rendered due to its similarity to the current one.', data );
 
 			return;
 		}
@@ -652,42 +652,38 @@ export default class Renderer {
 			}
 		}
 	}
-
-	/**
-	 * Checks if two given selections are similar. Selections are considered similar if they are non-collapsed
-	 * and their trimmed (see {@link #_trimSelection}) representations are equal.
-	 *
-	 * @private
-	 * @param {module:engine/view/selection~Selection} current
-	 * @param {module:engine/view/selection~Selection} previous
-	 * @returns {Boolean}
-	 */
-	_areSimilarSelections( current, previous ) {
-		return !current.isCollapsed && this._trimSelection( current ).isEqual( this._trimSelection( previous ) );
-	}
-
-	/**
-	 * Creates a copy of a given selection with all of its ranges
-	 * trimmed (see {@link module:engine/view/range~Range#getTrimmed getTrimmed}).
-	 *
-	 * @private
-	 * @param {module:engine/view/selection~Selection} selection
-	 * @returns {module:engine/view/selection~Selection} Selection copy with all ranges trimmed.
-	 */
-	_trimSelection( selection ) {
-		const newSelection = Selection.createFromSelection( selection );
-		const ranges = newSelection.getRanges();
-
-		let trimmedRanges = [];
-
-		for ( let range of ranges ) {
-			trimmedRanges.push( range.getTrimmed() );
-		}
-
-		newSelection.setRanges( trimmedRanges );
-
-		return newSelection;
-	}
 }
 
 mix( Renderer, ObservableMixin );
+
+// Checks if two given selections are similar. Selections are considered similar if they are non-collapsed
+// and their trimmed (see {@link #_trimSelection}) representations are equal.
+//
+// @private
+// @param {module:engine/view/selection~Selection} selection1
+// @param {module:engine/view/selection~Selection} selection2
+// @returns {Boolean}
+function areSimilarSelections( selection1, selection2 ) {
+	return !selection1.isCollapsed && trimSelection( selection1 ).isEqual( trimSelection( selection2 ) );
+}
+
+// Creates a copy of a given selection with all of its ranges
+// trimmed (see {@link module:engine/view/range~Range#getTrimmed getTrimmed}).
+//
+// @private
+// @param {module:engine/view/selection~Selection} selection
+// @returns {module:engine/view/selection~Selection} Selection copy with all ranges trimmed.
+function trimSelection( selection ) {
+	const newSelection = Selection.createFromSelection( selection );
+	const ranges = newSelection.getRanges();
+
+	let trimmedRanges = [];
+
+	for ( let range of ranges ) {
+		trimmedRanges.push( range.getTrimmed() );
+	}
+
+	newSelection.setRanges( trimmedRanges, newSelection.isBackward );
+
+	return newSelection;
+}
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -9,6 +9,7 @@
 
 import ViewText from './text';
 import ViewPosition from './position';
+import Selection from './selection';
 import { INLINE_FILLER, INLINE_FILLER_LENGTH, startsWithFiller, isInlineFiller, isBlockFiller } from './filler';
 
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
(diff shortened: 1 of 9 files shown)


Reinmar:
> That was the reason the logic is placed in the view/renderer class (however, still not 100% sure about it).

I think this is fine. It's better to start this way than trying to create "the wrong abstraction". There's a very high probability that the renderer would be the only user of this API, but when it'd be public we'd try to make it look reasonable which would certainly end badly :D

src/view/renderer.js
```diff
@@ -639,6 +652,42 @@ export default class Renderer {
...
+		const newSelection = Selection.createFromSelection( selection );
+		const ranges = newSelection.getRanges();
+
+		let trimmedRanges = [];
+
+		for ( let range of ranges ) {
+			trimmedRanges.push( range.getTrimmed() );
+		}
+
+		newSelection.setRanges( trimmedRanges );
```
Reinmar: You're losing the direction here.

src/view/renderer.js
```diff
@@ -639,6 +652,42 @@ export default class Renderer {
...
+	/**
+	 * Creates a copy of a given selection with all of its ranges
+	 * trimmed (see {@link module:engine/view/range~Range#getTrimmed getTrimmed}).
+	 *
+	 * @private
+	 * @param {module:engine/view/selection~Selection} selection
+	 * @returns {module:engine/view/selection~Selection} Selection copy with all ranges trimmed.
+	 */
+	_trimSelection( selection ) {
+		const newSelection = Selection.createFromSelection( selection );
```
Reinmar: `createFromSelection()` gives us `setTo()` call which e.g. copies fake selection's stuff so that's good. What I like less is the fact that we are copying the ranges here (`getRanges()` copies them) to copy them again when doing `setRanges()` to copy them again 3 times on another `getRanges()`, `getTrimmes()`, `setRanges()`. This means 6 times for every selection and we need to trim two selections...

Reinmar: Let me add that this is all due to our lack of trust that no one will touch the ranges and positions.

In a perfect immutable world, we'd only need to copy new instances on `getTrimmed()`. No other method would need to do that because no other method change ranges. And I think that we must go into this direction. Otherwise, we create a huge memory pressure.

Reinmar: I reported https://github.com/ckeditor/ckeditor5-engine/issues/897 to keep this issue in mind. Let's ignore it in this case.

src/view/renderer.js
```diff
@@ -639,6 +652,42 @@ export default class Renderer {
 			}
 		}
 	}
+
+	/**
+	 * Checks if two given selections are similar. Selections are considered similar if they are non-collapsed
+	 * and their trimmed (see {@link #_trimSelection}) representations are equal.
+	 *
+	 * @private
+	 * @param {module:engine/view/selection~Selection} current
```
Reinmar: There's no current and previous for `areSimilar()`. They are just two selections.

Reinmar:
The warning should not leak to the console: https://travis-ci.org/ckeditor/ckeditor5-engine/builds/216279012#L1484. Make sure to stub the `log.warn` function in these tests and verify that the error was logged. Of course, only if that was the expected behaviour there :D.

Reinmar:
I see that those logs are output exactly by the renderer tests, so it's very likely that you just need to mute them :).

Reinmar:
I've tested this solution on Firefox, Safari and Chrome. It seems to work very well. However, the general "feel" is destroyed by other issues with IMEs. I'll post all that I recorded in another issue so you can check them. None seemed to be related to this particular issue.

src/view/renderer.js
```diff
@@ -580,6 +582,17 @@ export default class Renderer {
 			return;
 		}
 
+		if ( oldViewSelection && this._areSimilarSelections( oldViewSelection, this.selection ) ) {
+			const data = {
+				oldSelection: oldViewSelection,
+				currentSelection: this.selection
+			};
+
+			log.warn( 'renderer-selection-similar: Rendering selection was skipped due to its similarity to the previous selection.', data );
```
Reinmar: It's super verbose now... but perhaps it's better to keep it logged like this for a time being so we know what's going on.

However, let's improve the text:

`renderer-skipped-selection-rendering: The selection was not rendered due to its similarity to the current one.`

src/view/renderer.js
```diff
@@ -639,6 +652,42 @@ export default class Renderer {
...
+
+	/**
+	 * Creates a copy of a given selection with all of its ranges
+	 * trimmed (see {@link module:engine/view/range~Range#getTrimmed getTrimmed}).
+	 *
+	 * @private
+	 * @param {module:engine/view/selection~Selection} selection
+	 * @returns {module:engine/view/selection~Selection} Selection copy with all ranges trimmed.
+	 */
+	_trimSelection( selection ) {
```
Reinmar: I wouldn't put methods which don't access `this` at least once in the class. They are better as private functions.

Reinmar:
The issues I recorded: https://github.com/ckeditor/ckeditor5-typing/issues/95

I feel that the next thing we should do is work on disabling filling char and element removal when composition takes place. Since things are transparent to the view (they are not visible there) I think that they could be kept longer if needed without much of a trouble. Am I right @pjasiun?

Also, I remember that the filling char is removed on keydown so it doesn't affect caret movement. This should also be prevented during composition.

Another thing is to start using character data's `replaceData()` method "just in case". We did a quick test with @f1ames on one of the issues whether it helps and it didn't, but I'm like 99% sure it must help with something. At least with collaboratively typed text.

Finally, there are some serious issues with indexes in Safari when using Hiragana or Spanish accents. This will be the next step too.

src/view/renderer.js
f1ames: Yes, it is... maybe we should think about something like e.g. `log.info` or introduce some config which turns off this log. But if we plan to remove it in the future anyway, maybe there is no sense in spending time on it.

src/view/renderer.js
f1ames: If we move `trimSelection` method to `view/selection` we could skip some cloning steps operating on `this._ranges`. However, this would be a partial solution only for this specific case.
It may need more general attention as you mentioned in #897.

pjasiun:
> I feel that the next thing we should do is work on disabling filling char and element removal 

What element removal, do you mean?


> when composition takes place. Since things are transparent to the view (they are not visible there) I think that they could be kept longer if needed without much of a trouble. Am I right @pjasiun?

Yep, both inline and block fillers should act the way they do not break the composition. If the do somthing else it's a bug and should be fixed.

f1ames:
Pushed requested changes.

Reinmar:
> What element removal, do you mean?

The block filler. It gets removed from the element you type in (e.g. `<p><strong>[]<br></strong></p>`).

> Yep, both inline and block fillers should act the way they do not break the composition. If the do somthing else it's a bug and should be fixed.

OK :)

PR #901, opened 2017-03-31
Fixing Renderer, part two.

### LINKED ISSUE #888: Renderer does not refresh nodes when they are reinserted.
Currently, when node is changed in view, for it to be re-rendered in DOM, it has to be "marked to sync". All nodes in view fire `change` event which is bubbled to the root. When `view.Document` creates `view.RootEditableElement` it attaches syncing callback to it:
```
viewRoot.on( 'change:children', ( evt, node ) => this.renderer.markToSync( 'children', node ) );
viewRoot.on( 'change:attributes', ( evt, node ) => this.renderer.markToSync( 'attributes', node ) );
viewRoot.on( 'change:text', ( evt, node ) => this.renderer.markToSync( 'text', node ) );
```
So all nodes in this root, when changed, bubble event to `RootEditableElement` and then are makred to sync. This is not true for `DocumentFragment`. It makes sense -- `view.DocumentFragment` are not rendered to DOM after all.

And here comes problematic scenario. Assume view: `<rootEditableElement><div><p>foo</p></div></rootEditableElement>`.

Now three actions follow:
1. Remove `<div>` -> `<rootEditableElement>` is marked to sync it's children.
2. Remove `<p>` -> `<div>` is NOT marked to sync.
3. Reinsert `<div>` -> it is reinserted with `<p>`  because it was not synced.

There are some possible fixes.

First what is coming to mind is "marking to sync" nodes that are in `view.DocumentFragment`. Seems simple but `DocumentFragment` does not have a reference to `view.Document` so it doesn't have a reference to `view.Renderer`. We can't implement it like this.

Digging deeper into issue I can see that interesting things happen when nodes are updated in `view.Renderer`. Synced view element's children are converted to DOM. New view elements are created but existing view elements use reference to their DOM counterparts. Elements that are no longer in view are removed from DOM.

We can solve the issue doing one of two:
1. When existing elements are converted, refresh their children.
2. Unbind elements that were removed, so they are re-created.

Since solution 1. would have to preform deep refreshing, it's actually less efficient and more difficult to implement than solution 2., so I'll go with solution 2. With solution 2. we know exactly which nodes were removed so we don't have to refresh deeply.

scofalik:

Fix: `view.Renderer` will deeply unbind DOM elements when they are removed from DOM. Closes #888.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -473,7 +473,7 @@ export default class Renderer {
 				i++;
 			} else if ( action === 'delete' ) {
 				// Whenever element is removed from DOM, unbind it and all of its children.
-				unbindDeep( actualDomChildren[ i ], this.domConverter );
+				this.domConverter.unbindDomElement( actualDomChildren[ i ] );
 				remove( actualDomChildren[ i ] );
 			} else { // 'equal'
 				i++;
@@ -686,17 +686,4 @@ function trimSelection( selection ) {
 	newSelection.setRanges( trimmedRanges, newSelection.isBackward );
 
 	return newSelection;
-}
-
-// Unbind given `domElement` and all of its children from view using given `domConverter`.
-//
-// @private
-// @param {HTMLElement} domElement
-// @param {module:engine/view/domconverter~DomConverter}
-function unbindDeep( domElement, domConverter ) {
-	domConverter.unbindDomElement( domElement );
-
-	for ( let child of domElement.childNodes ) {
-		unbindDeep( child, domConverter );
-	}
 }
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -472,8 +472,8 @@ export default class Renderer {
 				insertAt( domElement, i, expectedDomChildren[ i ] );
 				i++;
 			} else if ( action === 'delete' ) {
-				// Whenever element is removed from DOM, unbind it.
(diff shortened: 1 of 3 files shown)


src/view/renderer.js
```diff
@@ -440,10 +440,13 @@ export default class Renderer {
...
-		if ( !domElement ) {
-			// If there is no `domElement` it means that it was already removed from DOM.
-			// There is no need to update it. It will be updated when re-inserted.
-			return;
+		// Refresh children of this `viewElement` if they are already bound with DOM elements. Keep in mind if view
+		// element is bound with DOM element, the DOM element is not regenerated by `DomConverter`.
+		// Sometimes such elements are outdated. For that reason we have to refresh them.
+		for ( let viewChild of viewElement.getChildren() ) {
+			if ( viewChild.is( 'element' ) && domConverter.getCorrespondingDom( viewChild ) ) {
+				this._updateChildren( viewChild, options );
```
pjasiun: I would prefer to avoid recursion and do it straight: unbind all children on removing. This solution, even if more efficient, is a little bit hacky and we will have enough hacky code it there renderer anyway, so as long we can make something simpler we should do.

scofalik:
I can see that Travis errored even though the build passed.

Reinmar:
You know you can restart the build? ;)

PR #902, opened 2017-03-31
T/828 Introduce DeltaReplayer

### LINKED ISSUE #828: Create plugin that can replay operations
Few days ago @scofalik wrote a long text about engine complexity and debugging hell. He has already created a nice tool that enables logging a lot of useful informations to the console, but in many cases the amount of information is not enough to replay the bug. It's also hard to replay the bugs manually since they are complicated.

My proposal is to create a plugin for ckeditor5 that can filter and parse the logs and be able to replay logged operations on the initial state.

This plugin may expose following methods:
* `next( numberOfOperations = 1 )` - runs next n operations, breaks when hit a bug.
* `nextAll()` - runs all operations, breaks when hit a bug.
* `play( timeout )` and `stop()` -  plays operations until the error occurs.

It could be used to write complex integration / regression tests as well.

This plugin doesn't require so much work:
1. Improve logs, especially the logs from the insert operation.
1. Write a plugin for debugging.
   1. Create a method that gets text from the `history` (?) property from the ckeditor5 initialisation object.
   1. Create a method that filters the logs and returns only operations.
   1. Create a method that parses operation logs and returns array of operation instances.
   1. Create methods that can inject operations into editor.

It covers https://github.com/ckeditor/ckeditor5-engine/issues/821.

WDYT?

ma2ciek:

Feature: Introduced `dev-utils.DeltaReplayer`. Introduced new logging methods in `dev-utils.enableEngineDebug`. Closes #828.


Reinmar:
Missing tests. Coverage dropped below 100%.

Reinmar:
Still missing tests, plus, I noticed some warnings here: https://travis-ci.org/ckeditor/ckeditor5-engine/builds/218008934#L1462

ma2ciek:
Thanks for the attention. I'll look at this warnings. IMO the Promise wrapper could hide them.

Reinmar:
It depends on what logs them. Assuming that you expect that those things are logged, you could stub the proper `warn()` method and even validate that it was called.

ma2ciek:
Just realised that I misunderstood this warning.

src/dev-utils/deltareplayer.js
```diff
@@ -0,0 +1,122 @@
...
+/* global setTimeout, console */
+
+import DeltaFactory from '../model/delta/deltafactory';
+
+/**
+ * DeltaReplayer is a dev-tool created for easily replaying operations on the document from stringified deltas.
+ */
+export default class DeltaReplayer {
+	/**
+	 * @param {module:engine/model/document~Document} document.
```
scofalik: document Document to reply deltas on.

src/dev-utils/enableenginedebug.js
```diff
@@ -463,6 +476,22 @@ function enableDocumentTools() {
 		logDocument( this, version );
 	};
 
+	ModelDocument.prototype.initializeDebugging = function() {
```
scofalik: What is the reason for having this as separate function? Couldn't it be a part of `addAppliedDelta` (in `if` statement)?

scofalik:
I've added some comments.

Other than that, I am missing one thing: logging deltas should be automatic.

In real case scenario, developer will have to alter code with a lot of `addAppliedDelta`, basically everywhere. This is wasting time plus it's prone to errors, when you actually forget to put that somewhere. I understand that it's difficult, given that we don't have `model.Document.applyDelta`, but I think this is a must and we have to add this somehow.

One of ideas is to overwrite `Batch` methods, something like this:

```javascript
_batchInsert = Batch.prototype.insert;

Batch.prototype.insert = function( ...args ) {
	const deltasNumberBefore = this.deltas.length;

	_batchInsert.call( this, ...args );

	while ( deltasNumberBefore < this.deltas.length ) {
		const appliedDelta = this.deltas[ deltasNumberBefore++ ];
		this.document.addAppliedDelta( appliedDelta );
	}
}
```

scofalik:
Or, even better, overwrite `Batch.prototype.register`. This should save us some code.

scofalik:
Also - do we need logging operations if `DeltaReplayer` works on `Delta`s?

Aaanddd... It will be a pitty if delta separator would be contained in `InsertDelta` content, but since this is a dev tools, we can live with this.

scofalik:
> One of ideas is to overwrite Batch methods, something like this

Okay, the problem with this solution that a delta might not be applied through `Batch` API, for example in undo.

So another idea is to use `model.Document#applyOperation`. Since all operations from deltas have to be applied before another delta is created, we can sniff delta's "end" by comparing `operation.delta` property when `applyOperation` is called.

ma2ciek:
Ok. I added the hook to the `model.Document#applyOperation`, so now this replayer can be used independently.

scofalik:
> Ok. I added the hook to the model.Document#applyOperation, so now this replayer can be used independently.

This is fine now.

What about this:

> Also - do we need logging operations if `DeltaReplayer` works on `Delta`s?

?

ma2ciek:
Actually, logging operations could be useful if you don't want to use the DeltaReplayer.

scofalik:
Okay. I am also changing commit message. This is feature, not internal change. Internal change is a change that does not modify API and does not have a direct impact on how editor works. This PR introduces a new feature, which is `DeltaReplayer` and also new ways to log changes happening in editor. The fact that this is a dev tools doesn't change anything. After all, not only we (as core team) are supposed to use it, but external developers too.

ma2ciek:
ok, thanks @scofalik for the clarification.

src/dev-utils/deltareplayer.js
```diff
@@ -0,0 +1,135 @@
...
+
+	/**
+	 * Applies all deltas to replay at once.
+	 *
+	 * @returns {Promise}
+	 */
+	applyAllDeltas() {
+		return this.applyNextDelta()
+			.then( () => this.applyAllDeltas() )
+			.catch( () => {} );
```
Reinmar: THIS IS UBER BAD. This must never be done. Please open a ticket for it and remove this `.catch()`. This is a nightmare when debugging anything.

ma2ciek: Thanks. I just realizes how stupid approach it was.

PR #904, opened 2017-04-04
Changed: Fixed detaching model.LiveRange in model.LiveSelection.

### LINKED ISSUE #903: LiveSelection: LiveRanges are not detached correctly.
In https://github.com/ckeditor/ckeditor5-engine/issues/780#issuecomment-291171133 I've found a memory leak in `LiveSelection`. `LiveRange` callbacks are not correctly detached.

scofalik:

Fix: `model.LiveSelection` now correctly stops listening to removed `model.LiveRange`s. Closes #903.

---

### Additional information

image

After a minute of monkey clicking, there is just one instance of `LiveRange` and also `LiveSelection` retains just 4kb of data.

### DIFF
--- a/src/model/liveselection.js
+++ b/src/model/liveselection.js
@@ -116,8 +116,8 @@ export default class LiveSelection extends Selection {
 	 * Unbinds all events previously bound by document selection.
 	 */
 	destroy() {
-		for ( let i = 0; i < this._ranges.length; i++ ) {
-			this._ranges[ i ].detach();
+		while ( this._ranges.length ) {
+			this._popRange();
 		}
 
 		this.stopListening();
@@ -255,7 +255,19 @@ export default class LiveSelection extends Selection {
 	 * @inheritDoc
 	 */
 	_popRange() {
-		this._ranges.pop().detach();
+		this._removeRangeAtIndex( this._ranges.length - 1 );
+	}
+
+	/**
+	 * Removes a range from `LiveSelection` and detaches it.
+	 *
+	 * @private
+	 * @params {Number} index Index from which a range should be removed.
+	 */
+	_removeRangeAtIndex( index ) {
+		const range = this._ranges.splice( index, 1 )[ 0 ];
+
+		range.detach();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Reinmar:
I can see you used a different solution than we discussed initially. Was it due to those graveyard ranges or due to https://github.com/ckeditor/ckeditor5-utils/issues/144? If due to https://github.com/ckeditor/ckeditor5-utils/issues/144 then I would wait for that.

scofalik:
Mostly due to linked issue. Graveyard ranges are whatever.

Other solution would be to do: `this.stopListening(); this.off()` in `LiveRange#detach`. But I've gone with solution that is more logical given current implementation of events -- since it's `LiveSelection` who attaches callback, it's `LiveSelection` who is wholly responsible to take it back.

scofalik:
I've made a PR in `utils` so we can close this issue.

scofalik:
I've updated the solution, but keep in mind that it needs https://github.com/ckeditor/ckeditor5-utils/pull/145

jodator:
This fix should work with changes in `t/144-new`.

The main difference that prevents memory leaks is changing this:
```js
this.listenTo( liveRange, 'change:range' () => {} )
```
to:
```js
liveRange.on( 'change:range', () => {} );
```
Above change allows live ranges to be garbage collected because it is important where events are physically stored. 

As when using `this.listenTo` the `DocumentSelection` will hold references to `LiveRanges`. So one would have to on removing `LiveRange`:

```js
liveRange.deatch(); // this is just `liveRange.stopListening()`
// but it only stop listening event registered on this emitter.
// so this would be also needed:
this.stopListening( liveRange );
```

Reinmar:
This branch needs to be rebased in the first place :D

jodator:
I'd simplify this change to:
```
_prepareRange( range ) {
    // ... rest
    liveRange.on( 'change:range', ( evt, oldRange, data ) => {
    // ... rest
}
```
on line: https://github.com/ckeditor/ckeditor5-engine/blob/598bc78422d6da3dce7bae45c3d0e4c8569eb40f/src/model/documentselection.js#L328

As problematic here is the size of `DocumentSelection[ _listeingTo ]` size that keeps growing with each selection change.

Before:
selection_004

After:
selection_005

The reason is as previously stated the place where `listeningTo` are stored - it's better to keep it on shorter-living entity or remember to call `this.stopListening( emitter, 'event' )` every time emiter is not used any more. But doing `emitter.on( 'event' )` will "do it for us" as it will not store reference to `emitter` on listener.

scofalik:
AFAIR (and it was a long time ago) what you propose was my initial idea, but @Reinmar wanted a better solution that won't lead to problems in future (hence changes in `EmitterMixin`).

Reinmar:
Discareded by https://github.com/ckeditor/ckeditor5-engine/pull/1181.

PR #907, opened 2017-04-04
Changed the behaviour of deleteContent() in case of nested elements

### LINKED ISSUE #710: Blockquote implementation

Reinmar:

Other: Changed the behaviour of `DataController.deleteContent()` in a case of nested elements to better match situations like using <kbd>Backspace</kbd> after a block quotation. Closes #710.

---

### Additional information

Have fun analysing the code :).

### DIFF
--- a/src/controller/deletecontent.js
+++ b/src/controller/deletecontent.js
@@ -98,7 +98,9 @@ function mergeBranches( batch, startPos, endPos ) {
 	startPos = Position.createAfter( startParent );
 	endPos = Position.createBefore( endParent );
 
-	if ( endParent.childCount > 0 ) {
+	if ( endParent.isEmpty ) {
+		batch.remove( endParent );
+	} else {
 		// At the moment, next startPos is also the position to which the endParent
 		// needs to be moved:
 		// <a><b>x[]</b></a><c><d>{}y</d></c>
@@ -114,8 +116,19 @@ function mergeBranches( batch, startPos, endPos ) {
 		// To then become:
 		// <a><b>xy</b>[]</a><c>{}</c>
 		batch.merge( startPos );
-	} else {
-		batch.remove( endParent );
+	}
+
+	// Removes empty end ancestors:
+	// <a>fo[o</a><b><a><c>bar]</c></a></b>
+	// becomes:
+	// <a>fo[]</a><b><a>{}</a></b>
+	// So we can remove <a> and <b>.
+	while ( endPos.parent.isEmpty ) {
+		const parentToRemove = endPos.parent;
+
+		endPos = Position.createBefore( parentToRemove );
+
+		batch.remove( parentToRemove );
 	}
 
 	// Continue merging next level.
--- a/src/controller/deletecontent.js
+++ b/src/controller/deletecontent.js
@@ -78,7 +78,7 @@ function mergeBranches( batch, startPos, endPos ) {
 	}
 
 	// If one of the positions is a root, then there's nothing more to merge (at least in the current state of implementation).
-	// Theoretically in this case we could unwrap the <p>: <$root>x[]<p>[]y</p></$root>, but we don't need to support it yet
+	// Theoretically in this case we could unwrap the <p>: <$root>x[]<p>{}y</p></$root>, but we don't need to support it yet
 	// so let's just abort.
 	if ( !startParent.parent || !endParent.parent ) {
 		return;
@@ -92,18 +92,18 @@ function mergeBranches( batch, startPos, endPos ) {
 	}
 
 	// Remember next positions to merge. For example:
-	// <a><b>x[]</b></a><c><d>[]y</d></c>
+	// <a><b>x[]</b></a><c><d>{}y</d></c>
 	// will become:
-	// <a><b>xy</b>[]</a><c>[]</c>
+	// <a><b>xy</b>[]</a><c>{}</c>
 	startPos = Position.createAfter( startParent );
 	endPos = Position.createBefore( endParent );
 
 	if ( endParent.childCount > 0 ) {
 		// At the moment, next startPos is also the position to which the endParent
 		// needs to be moved:
-		// <a><b>x[]</b></a><c><d>[]y</d></c>
+		// <a><b>x[]</b></a><c><d>{}y</d></c>
 		// becomes:
-		// <a><b>x</b>[]<d>y</d></a><c>[]</c>
+		// <a><b>x</b>[]<d>y</d></a><c>{}</c>
 
 		// Move the end parent only if needed.
 		// E.g. not in this case: <p>ab</p>[]{}<p>cd</p>
@@ -112,7 +112,7 @@ function mergeBranches( batch, startPos, endPos ) {
 		}
 
 		// To then become:
-		// <a><b>xy</b>[]</a><c>[]</c>
+		// <a><b>xy</b>[]</a><c>{}</c>
 		batch.merge( startPos );
 	} else {
 		batch.remove( endParent );
--- a/src/controller/deletecontent.js
+++ b/src/controller/deletecontent.js
@@ -10,7 +10,6 @@
 import LivePosition from '../model/liveposition';
 import Position from '../model/position';
 import Element from '../model/element';
-import compareArrays from '@ckeditor/ckeditor5-utils/src/comparearrays';
 
 /**
  * Deletes content of the selection and merge siblings. The resulting selection is always collapsed.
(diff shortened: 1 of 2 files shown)


Reinmar:
Self R- unfortunately: https://github.com/ckeditor/ckeditor5-block-quote/pull/8#issuecomment-291613614

src/controller/deletecontent.js
```diff
@@ -99,8 +123,23 @@ function shouldAutoparagraph( doc, position ) {
 	return !isTextAllowed && isParagraphAllowed;
 }
 
-function checkCanBeMerged( element ) {
-	const schema = element.document.schema;
+function checkCanBeMerged( left, right ) {
+	const schema = left.document.schema;
```
scofalik: It's nice and all, but couldn't you just use `Treewalker` and see what objects you are stepping over?

scofalik: Or you thought it would be too inefficient?

src/controller/deletecontent.js
```diff
@@ -65,31 +64,56 @@ export default function deleteContent( selection, batch, options = {} ) {
...
+		return;
+	}
 
-			if ( !checkCanBeMerged( previousNode ) || !checkCanBeMerged( nextNode ) ) {
-				return;
-			}
+	// Check if operations we'll need to do won't need to cross object or limit boundaries.
+	// E.g., we can't merge endParent into startParent in this case:
+	// <limit><startParent>x</startParent></limit><endParent></endParent>
+	if ( !checkCanBeMerged( startParent, endParent ) ) {
```
scofalik: So I understand that if selection crosses limit elements, we don't do anything? (Instead of, for example, removing content from selection without merging)

src/controller/deletecontent.js
```diff
@@ -65,31 +64,56 @@ export default function deleteContent( selection, batch, options = {} ) {
...
-			if ( nextNode.childCount > 0 ) {
-				batch.merge( mergePos );
-			} else {
-				batch.remove( nextNode );
-			}
-		}
+	// Remember next positions to merge. For example:
+	// <a><b>x[]</b></a><c><d>[]y</d></c>
+	// will become:
+	// <a><b>xy</b>[]</a><c>[]</c>
```
scofalik: Are you sure this comment is correct? I don't understand it.

scofalik: Ah, okay, I get it know. But the comment doesn't say what is going on, and I thought that the example represents only what happens after two following lines of code are executed.

src/controller/deletecontent.js
```diff
@@ -65,31 +64,56 @@ export default function deleteContent( selection, batch, options = {} ) {
...
+		batch.move( endParent, startPos );
+
+		// To then become:
+		// <a><b>xy</b>[]</a><c>[]</c>
+		batch.merge( startPos );
+	} else {
+		batch.remove( endParent );
 	}
+
+	// Continue merging next level.
```
scofalik: Example of what/when should happen would be nice.

scofalik: Will this `<a><b>xy</b>[]</a><c>[]</c>` be changed to `<a><b>xy</b></a>`?

src/controller/deletecontent.js
scofalik: Reading commit diff, I haven't realised that this is already after removing content.

src/controller/deletecontent.js
```diff
@@ -65,31 +64,56 @@ export default function deleteContent( selection, batch, options = {} ) {
...
+	if ( !startParent.parent || !endParent.parent ) {
+		return;
+	}
 
-			if ( !checkCanBeMerged( previousNode ) || !checkCanBeMerged( nextNode ) ) {
-				return;
-			}
+	// Check if operations we'll need to do won't need to cross object or limit boundaries.
+	// E.g., we can't merge endParent into startParent in this case:
+	// <limit><startParent>x</startParent></limit><endParent></endParent>
```
scofalik: `[]` like in other examples would be nice.

src/controller/deletecontent.js
```diff
@@ -65,31 +64,56 @@ export default function deleteContent( selection, batch, options = {} ) {
...
+		// At the moment, next startPos is also the position to which the endParent
+		// needs to be moved:
+		// <a><b>x[]</b></a><c><d>[]y</d></c>
+		// becomes:
+		// <a><b>x</b>[]<d>y</d></a><c>[]</c>
+		batch.move( endParent, startPos );
+
+		// To then become:
+		// <a><b>xy</b>[]</a><c>[]</c>
+		batch.merge( startPos );
```
scofalik: I like this move+merge, because it is better from OT point of view. 👍

src/controller/deletecontent.js
scofalik: And what would happen if original scenario was:
`<a><b>x[]</b></a><c><d>[]y</d><e>z</e></c>`
Would it become finally:
`<a><b>xy</b><e>z</e></a>`
?

`<a><b>x[]</b></a><c><d>[]y</d><e>z</e></c>`
`<a><b>x</b>[]<d>y</d></a><c>[]<e>z</e></c>`
`<a><b>xy</b>[]</a><c>[]<e>z</e></c>`
`<a><b>xy</b>[]<c><e>z</e></c></a>[]`
`<a><b>xy</b>[]<e>z</e></a>[]`

How does it look like from the point of view of hypothetical features? Does it make sense? Have you thought about any feature that might use such deep structures?

scofalik: I guess this example shows this behaviour: https://github.com/ckeditor/ckeditor5-engine/pull/907/files#diff-b55e8246527f4a9a00e50fb4c1c88b10R269

It makes sense, I just wonder whether the function has to be that complicated.

src/controller/deletecontent.js
```diff
@@ -65,31 +64,56 @@ export default function deleteContent( selection, batch, options = {} ) {
...
+	startPos = Position.createAfter( startParent );
+	endPos = Position.createBefore( endParent );
+
+	if ( endParent.childCount > 0 ) {
+		// At the moment, next startPos is also the position to which the endParent
+		// needs to be moved:
+		// <a><b>x[]</b></a><c><d>[]y</d></c>
+		// becomes:
+		// <a><b>x</b>[]<d>y</d></a><c>[]</c>
+		batch.move( endParent, startPos );
```
scofalik: How about `<p>x[]</p><p>[]y</p>`? Maybe we could check whether the move is needed, to not create useless operations?

scofalik:
How about such scenario:

````
<bq>
  <p>a</p>
  <p>b</p>
</bq>
<bq>
  ^<img />
  <p>x</p>
</bq>
```

Will I get:
````
<bq>
  <p>a</p>
  <p>
    b
    
    <p>x</p>
  </p>
</bq>
```
(second `bq` is `endParent` that is moved after the parent of `startPos` and then is merged).

scofalik:
Okay I've analysed this solution and resolved each doubt I had. My only suggestion would be to try using `Treewalker` in `checkCanBeMerged` function, it might be easier and to understand and read, and would have at least same efficiency (there is no content between start and end position, so number of nodes is same as number of nodes to common ancestor).

src/controller/deletecontent.js
Reinmar: I haven't even thought about it. So, I'd need to iterate over all elements between those elements. That would be nice. I'll see how the code looks with it.

src/controller/deletecontent.js
Reinmar: Yep, we do remove the content, but we do touch the limit/object elements. So they can't be merged nor split.

Note: situations with selection crossing limit elements should not happen. This should be blocked on the model level. However, I wanted to have this covered at least partially here. I say partially because let's imagine this:

```html
<paragraph>x[x</paragraph>
<widget>
    noneditabletext
    <caption>x]x</caption>
</widget>
```

The expected result on delete is:


```html
<paragraph>x[]</paragraph>
<widget>
    noneditabletext
    <caption>x</caption>
</widget>
```

But you'll get:


```html
<paragraph>x[]</paragraph>
<widget>
    <caption>x</caption>
</widget>
```

So, the result is still bad, but we're at least a bit closer to the correct behaviour. The remaining obstacle is that being a limit element doesn't actually say that content within you isn't editable. So, we'll need some more element types in the schema to resolve this completely. That's why I thought that with the current types it'd be best to have the non-merging behaviour.

Reinmar:
> I guess this is a case for modifySelection though, as we could have startPos either in p or bq.

We can't quote images yet, so I'd need to hack it all a bit to test it. But I'll add a test in this repository. And yes, `modifySelection()` will first need to extend the selection to a point where the text is valid.

Reinmar:
> I guess this is a case for modifySelection though, as we could have startPos either in p or bq.

I've rethought that and you can't have such a selection at all:

```html
<bq>
  <p>a</p>
  <p>b</p>
</bq>
<bq>
  ^<img />
  <p>x</p>
</bq>
```

A more interesting test would be backspace in:

```html
<bq>
  <p>a</p>
  <p>b</p>
  <img />
</bq>
<bq>
  <p>^x</p>
</bq>
```

src/controller/deletecontent.js
Reinmar: We couldn't merge these two elements if not for move, so this is the only possible order :P

scofalik: What I mean is that merging is better than moving contents + deleting parent :).

Reinmar:
All done, except that odd test from https://github.com/ckeditor/ckeditor5-engine/pull/907#issuecomment-291823287 which I don't have energy now to hack.

scofalik:
Have you reported that case somewhere?

Reinmar:
No, of course not :D

Reinmar:
Done: https://github.com/ckeditor/ckeditor5-image/issues/54#issuecomment-292118301

Thanks.

PR #908, opened 2017-04-05
Remove invalid promise catches.

### LINKED ISSUE #906: Remove invaild catches from replayer
[These catches](https://github.com/ckeditor/ckeditor5-engine/pull/902/files/dd3015b82786b648e0c77f949c79d8a0cace4ca3#diff-891fb910d1bb2fa1a25f134274923658R103) should be changed in order to provide some correct errors.

ma2ciek:

Fix: Removed invalid promise catches from replayer tool. Closes #906.

### DIFF
--- a/src/dev-utils/deltareplayer.js
+++ b/src/dev-utils/deltareplayer.js
@@ -112,8 +112,8 @@ export default class DeltaReplayer {
 	}
 
 	/**
-	 * Applies the next delta to replay.
-	 * Returns promise with an `isFinished` parameter.
+	 * Applies the next delta to replay. Returns promise with `isFinished` parameter that is `true` if the last
+	 * delta in replayer has been applied, `false` otherwise.
 	 *
 	 * @returns {Promise.<Boolean>}
 	 */
--- a/src/dev-utils/deltareplayer.js
+++ b/src/dev-utils/deltareplayer.js
@@ -16,7 +16,7 @@ import DeltaFactory from '../model/delta/deltafactory';
  */
 export default class DeltaReplayer {
 	/**
-	 * @param {module:engine/model/document~Document} document Document to reply deltas on.
+	 * @param {module:engine/model/document~Document} document Document to replay deltas on.
 	 * @param {String} logSeparator Separator between deltas.
 	 * @param {String} stringifiedDeltas Deltas to replay.
 	 */
@@ -27,7 +27,7 @@ export default class DeltaReplayer {
 	}
 
 	/**
-	 * Parses given string containing stringified deltas and sets parsed deltas as deltas to reply.
+	 * Parses given string containing stringified deltas and sets parsed deltas as deltas to replay.
 	 *
 	 * @param {String} stringifiedDeltas Stringified deltas to replay.
 	 */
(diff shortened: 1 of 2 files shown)


src/dev-utils/deltareplayer.js
```diff
@@ -99,24 +104,28 @@ export default class DeltaReplayer {
...
+				}
+			} );
 	}
 
 	/**
 	 * Applies the next delta to replay.
+	 * Returns promise with a `isFinished` parameter.
 	 *
-	 * @returns {Promise}
+	 * @returns {Promise.<Boolean>}
```
scofalik: I know that you were doing documentation refactoring so you should know better but I have to ask - do we use such notation somewhere and does it/will it correctly show in built docs? I see it first time :P

src/dev-utils/deltareplayer.js
```diff
@@ -99,24 +104,28 @@ export default class DeltaReplayer {
...
+			.then( ( isFinished ) => {
+				if ( !isFinished ) {
+					return this.applyAllDeltas();
+				}
+			} );
 	}
 
 	/**
 	 * Applies the next delta to replay.
+	 * Returns promise with a `isFinished` parameter.
```
scofalik: an `isFinished`

src/dev-utils/deltareplayer.js
```diff
@@ -99,24 +104,28 @@ export default class DeltaReplayer {
...
 	 *
-	 * @returns {Promise}
+	 * @returns {Promise.<Boolean>}
 	 */
 	applyNextDelta() {
 		const document = this._document;
```
scofalik: I know that we've been discussing moving here this check:
```javascript
if ( !jsonDelta ) {
return Promise.resolve();
}
```

Was there a reason to abandon this idea?

ma2ciek: I think it's a safer way to do it. If we would want to change this class in the future and add some other `enqueuChanges()` modifying `this._deltasToReplay` then some strange errors could occur. (As far as I understand the idea behind `enqueueChanges` ;)

src/dev-utils/deltareplayer.js
ma2ciek: Yes, we do, e.g. here: https://github.com/ckeditor/ckeditor5-image/blob/0e8c6cab3759f5616f6d02a9ffc31aa099491adc/src/imagetextalternative.js#L108. The generic types are documented here: http://usejsdoc.org/tags-type.html, I didn't create the separate doc about them (maybe I should (?)). They're building correctly. And here is a list of currently supported generic types: https://github.com/ckeditor/ckeditor5-dev/blob/0aa523b441a052a441005041af9df4b7bd493c0e/packages/jsdoc-plugins/lib/validator/types.js

tests/dev-utils/deltareplayer.js
```diff
@@ -149,6 +131,26 @@ describe( 'DeltaReplayer', () => {
...
+				.map( d => JSON.stringify( d ) )
+				.join( '---' );
+
+			const deltaReplayer = new DeltaReplayer( doc, '---', stringifiedDeltas );
+
+			return deltaReplayer.play( 1 )
+				.then( () => {
+					throw new Error( 'It should throw an error' );
+				}, ( err ) => {
+					expect( err.name ).to.equal( 'CKEditorError' );
```
Reinmar: It's worth checking the message. We usually check it with `expect( err.message ).to.match( /^the-error-identifier:/ )` – this is precise enough, but you're also sure that you caught the right error. Not any CKEditorError.

PR #917, opened 2017-04-10
view.Element inline style attribute parser changed from RegExp-based to linear-loop based.

### LINKED ISSUE #881: Performance issues with complex content
I've been testing pasting some really complex websites (like https://gazeta.pl) and I noticed that it often causes browser to freeze.

I've been curious whether this is related to some problems with conversion (a potential inf loop) but I got something interesting once I asked Firefox to stop that heavy script execution:

image

I noticed that styles are mostly inlined on gazeta.pl and that there are images included there as base64 encoded strings (in `background-image`, I presume). This would nicely connect with the `parseInlineStyles()` being the last call in that stack trace.

There are two options:

* Either we use some poorly performing regexp or other kind of parser.
* Or we have even an inf loop in a regexp (this is possible).

scofalik:

Fix: Big property values in inline styles do not hang or crash editor anymore. Closes #881.

---

### Additional information

Inline styles parser was changed from RegExp based to loop based. New algorithm seeks for `"`, `'`, `:` and `;` characters and basing on them decides whether given part of inline style is a value or property name.


src/view/element.js
```diff
@@ -685,12 +685,70 @@ export default class Element extends Node {
...
+	// Fix inline styles that do not end with `;` so they are compatible with algorithm below.
+	if ( stylesString.charAt( stylesString.length - 1 ) != ';' ) {
+		stylesString = stylesString + ';';
+	}
+
+	// Seek the whole string for "special characters".
+	for ( let i = 0; i < stylesString.length; i++ ) {
+		const char = stylesString.charAt( i );
+
+		if ( quoteType === null ) {
```
Reinmar: You know you can use `case quoteType:`? :D

Reinmar: Turns out this logic can't be changed in the way I thought. So, this is actually the easiest implementation.

scofalik: 😎

Issue #600, opened 2016-09-19
Selection constructor should accept ranges array


Reinmar:
Cause, currently I have to do this:

``` js
const sel = new Selection();
sel.addRange( range );
```

Which could be simplified.

ma2ciek:
As I talked with @Reinmar, we'll create a new static function rather than change the constructor, so there'll be no need to rewrite existing code.

### Usage:
```js
const sel = Selection.createFromRanges( [ range1, range2 ] );
 // And with isLastBackward set to true:
const sel2 = Selection.createFromRanges( [ range1, range2 ], true );
```

The ` Selection.createFromRanges` uses the `Selection#setRanges` under the hood, so it's firing `change` event for now which is not needed I suppose.

Reinmar:
>  The Selection.createFromRanges uses the Selection#setRanges under the hood, so it's firing change event for now which is not needed I suppose.

It's not needed but it changed nothing because you can't listen on it :D

> As I talked with @Reinmar, we'll create a new static function rather than change the constructor, so there'll be no need to rewrite existing code.

TBH, I think I might've changed my mind. `Range()` does accept two positions and `Position()` does accept root and path. So, since range + direction is the most typical way of creating a selection I guess it could be the constructor's option.

Why I initially thought about moving it to a separate method was that I wasn't sure whether this is an identical case to `Range` and `Position` because the view selection can also be a fake one (so that's like another type of selection so why would the constructor handle just one). But... I've just completely overcomplicated this.

ma2ciek:
Ok, it makes sense, so I'll move this code to the constructors :D

ma2ciek:
I pushed the corrected version which still uses `setRanges()`. But now `setRanges` is reassigning `_ranges` and `_lastRangeBackward` and fires `change` event which is unnecessary. Maybe we can use the default parameters in the constructor function + simple `range instanceof` checks?

Reinmar:
Let's discuss this under #918.

PR #918, opened 2017-04-10
T/600 Introduce static method `Selection.createFromRanges`

ma2ciek:

Changed: Introduced static method `Selection.createFromRanges. Closes https://github.com/ckeditor/ckeditor5-engine/issues/600.

### DIFF
--- a/src/model/selection.js
+++ b/src/model/selection.js
@@ -621,7 +621,7 @@ export default class Selection {
 	}
 
 	/**
-	 * Creates and returns an instance of `Selection` that is created from the given ranges.
+	 * Creates and returns an instance of `Selection` from the given ranges.
 	 * Accepts a flag describing in which way the selection is made (see {@link #addRange addRange}).
 	 *
 	 * @param {Array.<module:engine/view/range~Range>} ranges Array of ranges to set.
--- a/src/model/selection.js
+++ b/src/model/selection.js
@@ -7,14 +7,14 @@
  * @module engine/model/selection
  */
 
-import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
-import Element from './element';
-import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
 import Position from './position';
+import Element from './element';
 import Range from './range';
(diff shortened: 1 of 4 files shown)


ma2ciek:
Please, read https://github.com/ckeditor/ckeditor5-engine/issues/600#issuecomment-292984632.

Reinmar:
https://github.com/ckeditor/ckeditor5-engine/issues/600#issuecomment-292988218 Sorry for that, but I'd move this to the constructor.

src/model/selection.js
```diff
@@ -621,6 +629,21 @@ export default class Selection {
...
 
 	/**
+	 * Creates and returns an instance of `Selection` from the given ranges.
+	 * Accepts a flag describing in which way the selection is made (see {@link #addRange addRange}).
+	 *
+	 * @param {Array.<module:engine/view/range~Range>} ranges Array of ranges to set.
+	 * @param {Boolean} [isLastBackward] Flag describing if last added range was selected forward - from start to end
+	 * (`false`) or backward - from end to start (`true`). Defaults to `false`.
+	 */
+	static createFromRanges( ranges, isLastBackward ) {
```
Reinmar: This method duplicates the constructor now so it can be removed.

Reinmar:
> I pushed the corrected version which still uses setRanges(). But now setRanges is reassigning _ranges and _lastRangeBackward and fires change event which is unnecessary. Maybe we can use the default parameters in the constructor function + simple range instanceof checks at the end?

IMO, the code is fine now. It could be more optimised, but then it'd be harder to read. Plus, it's a premature optimisation because this way of creating a selection isn't used even once yet and you'll never be able to listen to the selection's initial `change` event too.

So, KISS. Let's save our time for more important issues.

src/model/selection.js
ma2ciek: Fixed

src/model/selection.js
```diff
@@ -7,14 +7,14 @@
  * @module engine/model/selection
  */
 
-import Position from './position';
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
```
Reinmar: Why did you changed the imports? Please revert this change unless it's a must-have in this ticket.

PR #920, opened 2017-04-12
T/919 ViewConverterBuilder#fromAttribute() sets wrong matcher

### LINKED ISSUE #919: ViewConverterBuilder#fromAttribute() sets wrong matcher
`ViewConverterBuilder#fromAttribute()`sets wrong matcher when attribute is not `style` or `class`. Correct objects for matchers are, for example, `{ class: 'abc' }` or `{ style: 'font-weight:bold;' }`, but: `{ attribute: { foo: 'bar', abc: 'xyz' } }`. `fromAttribute` would set `{ foo: 'bar' }`.

scofalik:

Fix: `ViewConverterBuilder#fromAttribute()` was creating incorrect matcher object for `Matcher` if passed attribute was other than `class` or `style`. Closes #919.

Minor upgrades to `ViewConversionBuilder`:
* converters from `ViewConversionBuilder` will not convert if "creator function" returned `null`.
* simplified view converters building by making `ViewConversionBuilder#toAttribute()` `value` param optional. If not set, the attribute value is taken from converted view element.

### DIFF
--- a/src/conversion/buildviewconverter.js
+++ b/src/conversion/buildviewconverter.js
@@ -141,7 +141,7 @@ class ViewConverterBuilder {
 	 * @param {String|RegExp} [value] View attribute value.
 	 * @returns {module:engine/conversion/buildviewconverter~ViewConverterBuilder}
 	 */
-	fromAttribute( key, value = /.+/ ) {
+	fromAttribute( key, value = /.*/ ) {
 		let pattern = {};
 
 		if ( key === 'style' || key === 'class' ) {
--- a/src/conversion/buildviewconverter.js
+++ b/src/conversion/buildviewconverter.js
@@ -141,9 +141,15 @@ class ViewConverterBuilder {
 	 * @param {String|RegExp} [value] View attribute value.
 	 * @returns {module:engine/conversion/buildviewconverter~ViewConverterBuilder}
 	 */
-	fromAttribute( key, value = /.*/ ) {
+	fromAttribute( key, value = /.+/ ) {
 		let pattern = {};
-		pattern[ key ] = value;
+
+		if ( key === 'style' || key === 'class' ) {
+			pattern[ key ] = value;
+		} else {
+			pattern.attribute = {};
+			pattern.attribute[ key ] = value;
+		}
 
 		return this.from( pattern );
 	}
--- a/src/conversion/buildviewconverter.js
+++ b/src/conversion/buildviewconverter.js
@@ -151,9 +151,14 @@ class ViewConverterBuilder {
 			pattern.attribute[ key ] = value;
 		}
 
-		this.from( pattern );
+		const matcher = new Matcher( pattern );
 
-		this._from[ this._from.length - 1 ].attributeKey = key;
+		this._from.push( {
+			matcher: matcher,
+			consume: false,
+			priority: null,
+			attributeKey: key
+		} );
 
 		return this;
 	}
--- a/src/conversion/buildviewconverter.js
+++ b/src/conversion/buildviewconverter.js
@@ -151,7 +151,11 @@ class ViewConverterBuilder {
 			pattern.attribute[ key ] = value;
 		}
 
-		return this.from( pattern );
+		this.from( pattern );
+
+		this._from[ this._from.length - 1 ].attributeKey = key;
+
+		return this;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/conversion/buildviewconverter.js
```diff
@@ -314,13 +329,15 @@ class ViewConverterBuilder {
 	 * representing attribute key and attribute value or a function that returns an object with `key` and `value` properties.
 	 * If you provide creator function, it will be passed converted view element as first and only parameter.
 	 *
-	 *		buildViewConverter().for( dispatcher ).fromAttribute( 'style', { 'font-weight': 'bold' } ).toAttribute( 'bold', 'true' );
+	 *		buildViewConverter().for( dispatcher ).fromAttribute( 'alt' ).toAttribute( 'alt' );
+	 *		buildViewConverter().for( dispatcher ).fromAttribute( 'style', { 'font-weight': 'bold' } ).toAttribute( 'bold', true );
 	 *		buildViewConverter().for( dispatcher )
 	 *			.fromAttribute( 'class' )
-	 *			.toAttribute( ( viewElement ) => ( { key: 'class', value: viewElement.getAttribute( 'class' ) } ) );
+	 *			.toAttribute( ( viewElement ) => ( { key: 'class', value: 'class-' + viewElement.getAttribute( 'class' ) } ) );
```
Reinmar: What does this change mean?

src/view/matcher.js
```diff
@@ -130,7 +130,7 @@ export default class Matcher {
 	 *			pattern: <pattern used to match found element>,
 	 *			match: {
 	 *				name: true,
- 	 *				attributes: [ 'title', 'href' ],
+	 *				attribute: [ 'title', 'href' ],
 	 *				classes: [ 'foo' ],
```
Reinmar: AFAICS, it should be `class` and `style` too.

src/conversion/buildviewconverter.js
```diff
@@ -141,11 +141,21 @@ class ViewConverterBuilder {
...
+		if ( key === 'style' || key === 'class' ) {
+			pattern[ key ] = value;
+		} else {
+			pattern.attribute = {};
+			pattern.attribute[ key ] = value;
+		}
+
+		this.from( pattern );
+
+		this._from[ this._from.length - 1 ].attributeKey = key;
```
Reinmar: This is ugly. And where's this used anyway?

src/conversion/buildviewconverter.js
```diff
@@ -141,11 +141,21 @@ class ViewConverterBuilder {
 	 * @param {String|RegExp} [value] View attribute value.
 	 * @returns {module:engine/conversion/buildviewconverter~ViewConverterBuilder}
 	 */
-	fromAttribute( key, value = /.*/ ) {
+	fromAttribute( key, value = /.+/ ) {
```
Reinmar: Why this change? There are empty attributes which values can be empty but still they "are".

src/conversion/buildviewconverter.js
Reinmar: OK, I see where it's used (I must have made a typo looking for it before). Still, this is ugly, especially that you pass that key to `from()` anyway.

Reinmar:
Regarding the commit message – I don't think that our commit parser can handle multiple types. A change may be of one type – usually, it will be the most significant one – so "feature" comes before "fix". But in this case, we have some minor changes which accompany the fix so you can list them as separate points under the fix. Like:

```
Fix: What was fixed. Closes #123.

* Changed x.
* Changed also y.
```

src/conversion/buildviewconverter.js
scofalik: I've changed it to make the example relevant. If you want to map attribute to attribute without changing it, now you can do `.fromAttribute( 'class' ).toAttribute( 'class' )`.

src/view/matcher.js
scofalik: My bad, I haven't looked at other keys, I had a problem with `attribute(s)` only (and it made me really angry, BTW.)

src/conversion/buildviewconverter.js
scofalik: You pass it to `from()` but you can also use `from()` directly and pass whatever matcher/match object you want, so it is impossible to sniff whether `.fromAttribute( x ).toAttribute( y )` was used or not. I could check in `.toAttribute()` if `from` has just one value and it is `attribute` which has just one value, but I feel that that would be even uglier :P

src/conversion/buildviewconverter.js
scofalik: Hmmm, I haven't thought of this. This is a good point. I agree that this change was wrong.

scofalik:
I've pushed a couple of fixes and changed commit message.

src/conversion/buildviewconverter.js
Reinmar: It's not about that. It's about not using `from()` if you actually need to do more than that (and apparently, `from()` is not a sufficient generalisation for this case). `from()` is 5 LOC, just repeat that code here and all will be fine.

src/conversion/buildviewconverter.js
Reinmar: OK, I get it. I'll try to think about some better example.

PR #929, opened 2017-04-21
Fix docs and provide missing tests for the `DataController#insertContent`

### LINKED ISSUE #895: DataController.insertContent() and insertContent event have incorrect docs
We allowed passing an element to `insertContent()` function, but the docs of the method and event weren't updated and the change weren't tested on `DataController.insertContent()`.

ma2ciek:

Fix: Added missing test and updated docs in the `DataController#insertContent`. Closes #895.

### DIFF
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -251,7 +251,7 @@ export default class DataController {
 	 * See {@link module:engine/controller/insertcontent~insertContent}.
 	 *
 	 * @fires insertContent
-	 * @param {module:engine/model/documentfragment~DocumentFragment} content The content to insert.
+	 * @param {module:engine/model/documentfragment~DocumentFragment|module:engine/model/item~Item} content The content to insert.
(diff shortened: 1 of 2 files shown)


Reinmar:
What about tests?

ma2ciek:
I talked about the tests on the issue page (#895)

ma2ciek:
I added a test for this change.

Reinmar:
The `DataController.insertContent()` still doesn't have a test how it works with an item, but we can live with that. However, if you changed docs of two things (event and method) the same should ideally be reflected by the tests.

ma2ciek:
But the method `DataController.insertContent()` fires only an event (https://github.com/ckeditor/ckeditor5-engine/blob/3da3d625fc33a669c34c9f3ac857c2f90caa57de/src/controller/datacontroller.js#L259), so is this check necessary as we can pass everything to this method and the test will pass? (maybe I'm wrong, but for me one test is good enough).

PR #933, opened 2017-04-24
ViewConverterBuilder#toAttribute now passes model.DocumentFragment to data.output if was converter from view element.

### LINKED ISSUE #932: ViewConverterBuilder#toAttribute may return array as conversion result.
`ViewConverterBuilder#toAttribute()` may return array as conversion result. This may happen when view element is converted to model attribute: https://github.com/ckeditor/ckeditor5-engine/blob/master/src/conversion/buildviewconverter.js#L369

Here, `convertChildren` returns array which is then passed as `data.output`.

The solution is to wrap `convertChildren` result in `model.DocumentFragment`.

scofalik:

Fix: Unified values returned in `data.output` during view-to-model conversion. See breaking changes. Closes #932.

BREAKING CHANGE: `ViewConversionDispatcher.convert` now always returns `model.DocumentFragment` (which may be empty in various cases). `conversionApi#convertItem` can pass only `model.Node` or `model.DocumentFragment` or `null` to `data.output`. Warning will be logged if different value was passed. `conversionApi#convertChildren` now always returns `model.DocumentFragment`.

### DIFF
--- a/src/conversion/viewconversiondispatcher.js
+++ b/src/conversion/viewconversiondispatcher.js
@@ -193,7 +193,7 @@ export default class ViewConversionDispatcher {
 			 */
 			log.warn( 'view-conversion-dispatcher-incorrect-result: Dropped incorrect conversion result.', [ input, data.output ] );
 
-			data.output = null;
+			return null;
 		}
 
 		return data.output;
@@ -208,7 +208,7 @@ export default class ViewConversionDispatcher {
 		const viewChildren = Array.from( input.getChildren() );
 
 		// 1. Map those children to model.
-		// 2. Filter out wrong results.
+		// 2. Filter out items that has not been converted or for which conversion returned wrong result (for those warning is logged).
 		// 3. Extract children from document fragments to flatten results.
 		const convertedChildren = viewChildren
 			.map( ( viewChild ) => this._convertItem( viewChild, consumable, additionalData ) )
--- a/src/conversion/viewconversiondispatcher.js
+++ b/src/conversion/viewconversiondispatcher.js
@@ -8,15 +8,17 @@
  */
 
 import ViewConsumable from './viewconsumable';
-import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
-import mix from '@ckeditor/ckeditor5-utils/src/mix';
-import extend from '@ckeditor/ckeditor5-utils/src/lib/lodash/extend';
 import ModelRange from '../model/range';
 import ModelPosition from '../model/position';
 import ModelTreeWalker from '../model/treewalker';
 import ModelNode from '../model/node';
 import ModelDocumentFragment from '../model/documentfragment';
 import { remove } from '../model/writer';
+
+import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
+import mix from '@ckeditor/ckeditor5-utils/src/mix';
+import extend from '@ckeditor/ckeditor5-utils/src/lib/lodash/extend';
+import log from '@ckeditor/ckeditor5-utils/src/log';
 
 /**
  * `ViewConversionDispatcher` is a central point of {@link module:engine/view/view view} conversion, which is a process of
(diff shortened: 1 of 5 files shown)


Reinmar:
Yy... I'm unsure about this. Shouldn't we ensure that no method returns an array rather than post-convert this to DFs?

scofalik:
Well, it depends.

On one hand it feels safer.

On the other hand you imply that view conversion always is to model. We struggled to not do this, but as we can see now: it had no point (it was a nice idea) and we already lost (markers has to be handled there and this already makes `ViewConversionDispatcher` depend on model). Still, different methods use `conversionApi.convertChildren` for different purposes. However since you can do: `modelElement.appendChildren( modelFragment )` and `new ModelElement( 'xxx', null, modelFragment )` I can agree that it might be better to just return `model.DocumentFragment` in `ViewConversionDispatcher`.

scofalik:
Okay, I have changed the solution, more info in the first post.

src/conversion/viewconversiondispatcher.js
```diff
@@ -180,6 +179,30 @@ export default class ViewConversionDispatcher {
...
+		if ( data.output === undefined ) {
+			/**
+			 * Item has not been converted (output is undefined).
+			 *
+			 * @error view-conversion-dispatcher-item-not-converted
+			 */
+			log.warn( 'view-conversion-dispatcher-item-not-converted: Item has not been converted (output is undefined).', input );
+
+			data.output = null;
+		} else if ( !( data.output instanceof ModelNode || data.output instanceof ModelDocumentFragment ) ) {
```
Reinmar: I wanted to comment that it'd be better if we used duck typing with `is()` as it's safer. But I can see that we use a lot of instanceofs around our code still so we can change this at once in the future.

src/conversion/viewconversiondispatcher.js
```diff
@@ -180,6 +179,30 @@ export default class ViewConversionDispatcher {
 			this.fire( 'documentFragment', data, consumable, this.conversionApi );
 		}
 
+		// Handle incorrect `data.output`.
+		if ( data.output === undefined ) {
```
Reinmar: Is it really needed and correct? In what situations does it help?

scofalik: It seemed correct when I was working just on the dispatcher but "in real life" I can see that this is actually incorrect. There are situations where we expect that the view element will not be converted: for example when text is wrong with schema. On the other hand we have "default" converters which will take care of "non-intentionally not-converted elements". So yeah, this should be removed.

scofalik:
I fixed unneeded warning and fixed merge message.

scofalik:
The build failed but tests are fine:

```
No output has been received in the last 10m0s (...)
The build has been terminated
```

src/conversion/viewconversiondispatcher.js
```diff
@@ -188,11 +203,23 @@ export default class ViewConversionDispatcher {
...
-		const convertedChildren = viewChildren.map( ( viewChild ) => this._convertItem( viewChild, consumable, additionalData ) );
 
-		// Flatten and remove nulls.
-		return convertedChildren.reduce( ( a, b ) => b ? a.concat( b ) : a, [] );
+		// 1. Map those children to model.
+		// 2. Filter out wrong results.
+		// 3. Extract children from document fragments to flatten results.
+		const convertedChildren = viewChildren
+			.map( ( viewChild ) => this._convertItem( viewChild, consumable, additionalData ) )
+			.filter( ( converted ) => converted instanceof ModelNode || converted instanceof ModelDocumentFragment )
```
Reinmar: This is like a black hole. Can't we just throw here? Or at least warn the developer. But I think that throwing is better because this isn't even a runtime error but simply a serious implementation bug. The same applies to `_convertItem()` – I feel that we could throw there as well.

Reinmar:
I commented on one issue.

src/conversion/viewconversiondispatcher.js
```diff
@@ -188,11 +204,23 @@ export default class ViewConversionDispatcher {
...
 	 */
 	_convertChildren( input, consumable, additionalData = {} ) {
+		// Get all children of view input item.
 		const viewChildren = Array.from( input.getChildren() );
-		const convertedChildren = viewChildren.map( ( viewChild ) => this._convertItem( viewChild, consumable, additionalData ) );
 
-		// Flatten and remove nulls.
-		return convertedChildren.reduce( ( a, b ) => b ? a.concat( b ) : a, [] );
+		// 1. Map those children to model.
+		// 2. Filter out wrong results.
```
Reinmar: Bad @scofalik cause bad comment!

src/conversion/viewconversiondispatcher.js
```diff
@@ -180,6 +181,21 @@ export default class ViewConversionDispatcher {
...
+			 * Dropped incorrect conversion result.
+			 *
+			 * Item may be converted to either {@link module:engine/model/node~Node model node} or
+			 * {@link module:engine/model/documentfragment~DocumentFragment model document fragment}.
+			 *
+			 * @error view-conversion-dispatcher-incorrect-result
+			 */
+			log.warn( 'view-conversion-dispatcher-incorrect-result: Dropped incorrect conversion result.', [ input, data.output ] );
+
+			data.output = null;
```
Reinmar: Maybe `return null`?

src/conversion/viewconversiondispatcher.js
Reinmar: OK, `_convertChildren()` uses `_convertItem()` so the "wrong results" are not really wrong but just "non-convertable content". You can change the comment and it will be fine.

tests/conversion/viewconversiondispatcher.js
```diff
@@ -207,91 +195,169 @@ describe( 'ViewConversionDispatcher', () => {
...
+				expect( log.warn.calledOnce ).to.be.true;
+
+				log.warn.restore();
+			} );
 		} );
 	} );
+
+	// Silences warnings that pop up in tests. Use when the test checks a specific functionality and we are not interested in those logs.
+	// No need to restore `log.warn` - it is done in `afterEach()`.
+	function silenceWarnings() {
```
Reinmar: I'm not 100% sure but I'm afraid that this may silence too many warnings as `log` may be defined just once at the beginning of this file. And if we ever switch to bundling all test files into one module (to improve Webpack's performance) this will silence all logs forever.

scofalik: Why? `log.warn` is restored in `afterEach()`

scofalik:
I pushed one more commit. You, again, start being PicKy...

tests/conversion/viewconversiondispatcher.js
Reinmar: OK, I haven't noticed that. :*

PR #941, opened 2017-05-08
T/940

### LINKED ISSUE #940: Delta: enable history log when delta becomes transformed
It would make debugging easier, if delta instance kept a history of all deltas by which it was transformed, with all additional information that is needed to recreate that delta from it's original state to current state.

This is important to debug OT.

scofalik:

Feature: When engine debugging is on, deltas that are results of transformation will keep their history of changes in `#history` property. Closes #940.

---

### Additional information

This PR needs a change in `ckeditor5-undo`. Related PR https://github.com/ckeditor/ckeditor5-undo/pull/61

`Delta#history` is created and filled automatically when `engine/model/delta/transform~deltaTransform#transform` is used, if engine debugging is on (`engine/dev-utils/enableenginedebug~enableEngineDebug` was called).

`Delta#history` is an array. Each item is an object containing properties:
* `{String} before` - JSON string containing delta state before transformation,
* `{String} transformedBy` - JSON string containing delta which `before` was transformed by,
* `{Boolean} wasImportant` - whether `before` was more important than `transformedBy` when it was transformed,
* `{Number} resultsTotal` - how many deltas were returned as the transformation result,
* `{Number} resultIndex` - what was the index of that delta in the transformation result.

This info will let us recreate all steps which led to current delta state, even if it was transformed multiple times. First item's `before` property is the original delta. Then it has to be transformed by `transformBy` deltas of following items (using `wasImportant`). After each transformation, `resultIndex` delta has to be chosen from among returned deltas.

Non-first `before` values and `resultsTotal` values can be used for additional checking or when there is no need to perform transformation back from the original delta. Those could be omitted/removed if we don't want them.

Note: `before` and `transformedBy` JSON strings do not contain their own `history` properties. That would lead to enormous data duplication.

### DIFF
--- a/src/dev-utils/enableenginedebug.js
+++ b/src/dev-utils/enableenginedebug.js
@@ -268,38 +268,38 @@ function enableLoggingTools() {
 	};
 
 	AttributeOperation.prototype.toString = function() {
-		return getClassName( this ) + ': ' +
+		return getClassName( this ) + `( ${ this.baseVersion } ): ` +
 			`"${ this.key }": ${ JSON.stringify( this.oldValue ) } -> ${ JSON.stringify( this.newValue ) }, ${ this.range }`;
 	};
 
 	InsertOperation.prototype.toString = function() {
-		return getClassName( this ) + ': ' +
+		return getClassName( this ) + `( ${ this.baseVersion } ): ` +
 			`[ ${ this.nodes.length } ] -> ${ this.position }`;
 	};
 
 	MarkerOperation.prototype.toString = function() {
-		return getClassName( this ) + ': ' +
+		return getClassName( this ) + `( ${ this.baseVersion } ): ` +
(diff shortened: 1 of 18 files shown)


src/dev-utils/enableenginedebug.js
```diff
@@ -317,27 +320,57 @@ function enableLoggingTools() {
 		}
 	};
 
+	Delta.prototype.saveHistory = function( before, transformedBy, wasImportant, resultIndex, resultsTotal ) {
```
pjasiun: Shouldn't `saveHistory` be protected?

pjasiun: Also `before, transformedBy, wasImportant, resultIndex, resultsTotal` should be replaced by an object.

pjasiun:
After fixing `saveHistory` method, feel free to merge this PR and the one in the undo repository.

PR #942, opened 2017-05-10
Engine debug tools improvements

scofalik:

Feature: When engine debugging is on, additional logs will be provided when delta transformation causes editor to throw an error.


src/model/delta/attributedelta.js
```diff
@@ -15,6 +15,8 @@ import RootAttributeOperation from '../operation/rootattributeoperation';
 import Position from '../position';
 import Range from '../range';
 
+const range = Symbol( 'range' );
```
pjasiun: Overwrite `toJSON` instead of using symbol.

PR #944, opened 2017-05-10
MarkerDelta transformation should consider ranges equal to null.

### LINKED ISSUE #943: Editor crashes when transforming MarkerDelta that has null as one of its ranges.
Editor crashes when transforming `MarkerDelta` that has a `MarkerOperation` with `null` as one of its ranges.

`MarkerOperation` may have `null` values, for example when it creates or removes a marker. This was not taken into consideration when writing transformation in `engine/model/delta/basic-transformations~transformMarkerDelta`.

scofalik:

Fix: `MarkerDelta` transformation should no longer cause editor to crash, if a `MarkerOperation` had `null` as it's `oldRange` or `newRange`. Closes #943.

### DIFF
--- a/tests/model/delta/transform/markerdelta.js
+++ b/tests/model/delta/transform/markerdelta.js
@@ -233,7 +233,7 @@ describe( 'transform', () => {
 		} );
 	} );
 
-	it( 'transforming MarkerDelta with null ranges should not crash', () => {
+	it( 'null ranges of MarkerDelta should not be changed during transformation', () => {
 		const markerDelta = getMarkerDelta( 'name', null, null, baseVersion );
 
 		// Transform `markerDelta` by any other delta that has a special transformation case with `MarkerDelta`.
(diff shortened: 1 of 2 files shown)


tests/model/delta/transform/markerdelta.js
```diff
@@ -232,4 +232,17 @@ describe( 'transform', () => {
...
+	it( 'transforming MarkerDelta with null ranges should not crash', () => {
+		const markerDelta = getMarkerDelta( 'name', null, null, baseVersion );
+
+		// Transform `markerDelta` by any other delta that has a special transformation case with `MarkerDelta`.
+		const wrapElement = new Element( 'w' );
+		const wrapRange = new Range( new Position( root, [ 1 ] ), new Position( root, [ 2 ] ) );
+		const wrapDelta = getWrapDelta( wrapRange, wrapElement, baseVersion );
+
+		expect( () => {
+			transform( markerDelta, wrapDelta );
```
Reinmar: But what's the expected behaviour of such transformation?

Reinmar:
I left one comment. Wouldn't we like to test what's the expected output?

scofalik:
You are absolutely right.

scofalik:
Fixed

PR #952, opened 2017-05-11
t/951: DomConverter should actively prevent unwanted scrolling on focus

### LINKED ISSUE #707: Scrolled content of editable jumps to top on viewDocument.focus()
ezgif com-962e89e169

### LINKED ISSUE #951: Webpage scrolls when view's DomConverter focuses an element
It's a WebKit-specific behavior that makes the web page scroll to the beginning when editing view is focused. I see no other solution than post-focus `window.scrollTo` to fix the position of the scrollbar.

Note that Image text alternative balloon also suffers because of this.

A follow-up of https://github.com/ckeditor/ckeditor5-link/issues/100.

oleq:

Fix: `DomConverter` should actively prevent unwanted scrolling on focus. Closes #951. Closes #707.

---

### Additional information

~~Requires https://github.com/ckeditor/ckeditor5-utils/pull/156.~~

### DIFF
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -19,8 +19,6 @@ import ViewTreeWalker from './treewalker';
 import { BR_FILLER, INLINE_FILLER_LENGTH, isBlockFiller, isInlineFiller, startsWithFiller, getDataWithoutFiller } from './filler';
 
 import global from '@ckeditor/ckeditor5-utils/src/dom/global';
-import env from '@ckeditor/ckeditor5-utils/src/env';
-
 import indexOf from '@ckeditor/ckeditor5-utils/src/dom/indexof';
 import getAncestors from '@ckeditor/ckeditor5-utils/src/dom/getancestors';
 import getCommonAncestor from '@ckeditor/ckeditor5-utils/src/dom/getcommonancestor';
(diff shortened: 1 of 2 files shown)


Reinmar:
A more complete solution would be to look for all elements up in the tree (starting from the editable) which have `scrollTop` or something indicating that they are scrollable. But I guess that fixing the editable and viewport is OK for now.

Reinmar:
R- for a different reason, though. AFAIU this issue concerns Blink and Webkit. So:

* the issue and solution (`env` check) should not be about Webkit only,
* perhaps we could target the solution to all browsers in order to avoid code branching?

I'm not 100% sure that the latter is a good idea, but it's something worth checking. The Webkit thing should be resolved though.

oskarwrobel:
This PR also closes https://github.com/ckeditor/ckeditor5-engine/issues/707.

oleq:
@Reinmar 

> AFAIU this issue concerns Blink and Webkit

> the issue and solution (env check) should not be about Webkit only,

I'm a little bit confused what you meant by this.

Reinmar:
That this is not only about Webkit. This is about Webkit and Blink. Right? So why is there `if ( webkit )` in the code?

oleq:
Ah... so you're referring to https://github.com/ckeditor/ckeditor5-utils/pull/156#issuecomment-300880157, right?

Reinmar:
Yep. But it's not only about the code but also communication (issue, PR title, changelog, etc.).

oleq:
I made the fix generic for all web browsers in https://github.com/ckeditor/ckeditor5-engine/pull/952/commits/7431a142eee5921fad83a0d8f4cc8e6f99d471c9. Checked Firefox and Edge, and nothing looks wrong so I guess we may go with it this way.

Reinmar:
It seems that it also fixed https://github.com/ckeditor/ckeditor5-engine/issues/706! :)

Reinmar:
Would be green light but I miss a manual test. It's something that you need to test manually. OTOH... it's also an issue which we see quite well when using the editor. So, ok – I'm merging :)

PR #977, opened 2017-06-22
Introduced context for OT. Refactored: History, RemoveOperation, OT and other.

scofalik:

Feature: OT uses context information for better conflict resolving.

This includes refactoring of: `History`, `RemoveOperation`, operational transformation algorithms, delta transformation algorithms and more.

Context will be used instead of removing deltas from history, which caused bugs in more complicated scenarios. This mostly affects undo algorithms.

BREAKING CHANGE: `History` API for deleting undone deltas has been removed.
BREAKING CHANGE: `deltaTransform#transformDeltaSets` is now `protected`. Use `model.Document#transformDeltas` instead.

----------

Additional information:

* `RemoveOperation` no longer creates `$graveyardHolder` elements. We are not using `$graveyardHolder` anywhere anymore.
* `RemoveOperation` no longer sets its `targetPosition` on its own in constructor. `targetPosition` has to be passed to the constructor instead.
* `Batch#remove` now creates multiple `RemoveDelta`s each with one `RemoveOperation` instead of one delta with multiple operations.
* Introduced deltas normalization. Each delta has its set structure, for example `MoveDelta` has exactly one `MoveOperation`, `SplitDelta` has exactly `InsertOperation` and `MoveOperation` (or `NoOperation`) in this order, etc. After delta transformation using OT algorithms (default case), `MoveOperation` may get split into two, creating "not normal" deltas. For example `SplitDelta` with `InsertOperation` and two `MoveOperation`s. This is normalized to one `SplitDelta` (insert + first move) and one `MoveDelta` (second move).
* If `RemoveOperation` has been undone, it is not treated as "always more important" than the other operation, during `RemoveOperation` transformation.
* Removed deltas priority. They were used to change `isStrong` flag (now `context.isStrong`). In this new approach it might make too much mess. Deltas priority was supposed to fix some problems in delta transformations. After refactoring and removing this mechanisms no tests fail anyway. Maybe some of problems were fixed along the way, maybe they are more hidden. Anyway there will be a followup to revise all delta transformation special cases, so we will return to that problem later.


scofalik:
This needs PRs in other repos.

src/model/range.js
```diff
@@ -101,15 +101,24 @@ export default class Range {
...
 	 *
 	 * @param {module:engine/model/range~Range} otherRange Range to check.
+	 * @param {Boolean} [strict=true] Whether the check is strict or loose. If the check is strict (`true`), compared range cannot
+	 * start or end at the same position as this range boundaries. If the check is loose (`false`), compared range can start, end or
+	 * even be equal to this range. Note that collapsed ranges are always compared in strict mode.
 	 * @returns {Boolean} `true` if given {@link ~Range range} boundaries are contained by this range, `false` otherwise.
 	 */
-	containsRange( otherRange ) {
-		return this.containsPosition( otherRange.start ) && this.containsPosition( otherRange.end );
+	containsRange( otherRange, strict = true ) {
```
Reinmar: The default of a boolean flag should be `false`.

Reinmar: You shouldn't need to pass `false` as a param.

scofalik: Yes, I've seen you commented on it previously. I understand your point. My point was that this is creating double negation (`nonStrict == false`) which is sometimes confusing to read and understand when the code is longer.

Reinmar: I know. So maybe `weak`?

scofalik: I think that `loose` is a better term.

scofalik:
I've added some more commits but messed up a bit and rebased instead of merge with master, so sorry for this :).

scofalik:
I've created all needed PRs in other repos.

scofalik:
I found a bug in undo during manual testing, fixing it right now.

scofalik:
Okay, I've rewritten and fixed a lot of stuff in deltas transformations and operations transformations. For now it is okay, however I will create several followups.

scofalik:
I've updated related PRs and checked that all branches together work fine.

scofalik:
I've added follow up issues.

Reinmar:
I pushed `remove-refactor` to ckeditor5 with the branches setup.

Reinmar:
Please resolve conflicts in https://github.com/ckeditor/ckeditor5-undo/pull/63.

src/model/operation/transform.js
```diff
@@ -383,173 +406,206 @@ const ot = {
...
-					rangeB.containsPosition( a.targetPosition ) ||
-					rangeB.start.isEqual( a.targetPosition ) ||
-					rangeB.end.isEqual( a.targetPosition )
-				);
-
-			if ( common !== null && ( aCompB === 'extension' || ( aCompB === 'same' && isStrong ) || aIsInside ) && !bTargetsToA ) {
-				// Here we do not need to worry that newTargetPosition is inside moved range, because that
-				// would mean that the MoveOperation targets into itself, and that is incorrect operation.
-				// Instead, we calculate the new position of that part of original range.
+			// // If MoveOperations has common range it can be one of two:
```
Reinmar: Was it left like this intentionally?

src/model/operation/transform.js
```diff
@@ -52,12 +53,11 @@ import compareArrays from '@ckeditor/ckeditor5-utils/src/comparearrays';
  * `a` by `b` and also `b` by `a`. In both transformations the same operation has to be the important one. If we assume
  * that first or the second passed operation is always more important we won't be able to solve this case.
  *
- * @external module:engine/model/model.operation
+ * @protected
```
Reinmar: I'll remove this because it creates a precedence that we have a protected export. We should rather invest in marking internal modules all across the code.

src/model/operation/transform.js
```diff
@@ -383,173 +406,206 @@ const ot = {
...
+					// This means that the transformation will be "clean" and always return one result.
+					ranges[ i ] = ranges[ i ]._getTransformedByMove( ranges[ j ].start, newTargetPosition, howMany )[ 0 ];
 				}
+			}
 
-				return result;
-			} );
+			return makeMoveOperationsFromRanges( a, ranges, newTargetPosition );
+
+			// // Map transformed range(s) to operations and return them.
```
Reinmar: And here again – was it left intentionally?

src/model/range.js
```diff
@@ -101,15 +101,24 @@ export default class Range {
...
+		const containsStart = this.containsPosition( otherRange.start ) || ( loose && this.start.isEqual( otherRange.start ) );
+		const containsEnd = this.containsPosition( otherRange.end ) || ( loose && this.end.isEqual( otherRange.end ) );
+
+		return containsStart && containsEnd;
 	}
 
 	/**
-	 * Two ranges are equal if their {@link #start start} and
-	 * {@link #end end} positions are equal.
+	 * Two ranges are equal if their {@link #start start} and {@link #end end} positions are equal.
```
Reinmar: No need to type `#start start`. Just leave the property or class name unless you want to change the wording a bit.

Reinmar:
Please close tickets which should be closed now (and please cross-link them for the future reference).

Also, please check issues that you assigned to yourself, whether you're still working on them.

scofalik:
The commented out code was left there by mistake. I already pushed a commit to master fixing it.

PR #981, opened 2017-06-26
t/699: Live selection should properly set its properties in case of non-collapsed default range

### LINKED ISSUE #699: First object should be the default selection position
`Document.getNearestSelectionPosition()` needs to listen a... range containing the first object (see `Schema#objects`) if it's not preceded with another element which can contain text.

A TC for that is: load `<figure class=image></figure><p>x</p>` into the editor. Selection needs to be on that widget, not before it, nor in the paragraph.

Reinmar:

Fix: `LiveSelection` will correctly set its properties in case of a non-collapsed default range. This will fix loading data which starts with an object element. Closes #699.

---

### Additional information

There are a lot of failures in image, upload and widget packages which I need to fix now...

### DIFF
--- a/src/model/liveselection.js
+++ b/src/model/liveselection.js
@@ -88,7 +88,7 @@ export default class LiveSelection extends Selection {
 	get isCollapsed() {
 		const length = this._ranges.length;
 
-		return length === 0 ? true : super.isCollapsed;
+		return length === 0 ? this._document._getDefaultRange().isCollapsed : super.isCollapsed;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


Reinmar:
I pushed also:

* https://github.com/ckeditor/ckeditor5-image/tree/t/ckeditor5-engine/699
* https://github.com/ckeditor/ckeditor5-upload/tree/t/ckeditor5-engine/699
* https://github.com/ckeditor/ckeditor5-widget/tree/t/ckeditor5-engine/699

All ready to review now.

PR #990, opened 2017-06-29
T/986: Toggle attribute command is on when selecting an object with bolded text in a nested editable

### LINKED ISSUE #986: Selection should not inherit attributes from inside of object elements
Moved from https://github.com/ckeditor/ckeditor5-core/issues/73.

pomek:

Fix: `LiveSelection` will not read attributes from object element and its children. Closes #986.

### DIFF
--- a/src/model/liveselection.js
+++ b/src/model/liveselection.js
@@ -7,7 +7,6 @@
  * @module engine/model/liveselection
  */
 
-import Element from './element';
 import Position from './position';
 import Range from './range';
 import LiveRange from './liverange';
@@ -568,7 +567,7 @@ export default class LiveSelection extends Selection {
 			// ...look for a first character node in that range and take attributes from it.
 			for ( const item of range ) {
 				// If the item is an object, we don't want to get attributes from its children.
-				if ( item.item instanceof Element && schema.objects.has( item.item.name ) ) {
+				if ( item.item.is( 'element' ) && schema.objects.has( item.item.name ) ) {
 					break;
 				}
 
--- a/src/model/liveselection.js
+++ b/src/model/liveselection.js
@@ -7,6 +7,7 @@
  * @module engine/model/liveselection
  */
 
+import Element from './element';
 import Position from './position';
 import Range from './range';
 import LiveRange from './liverange';
@@ -556,6 +557,7 @@ export default class LiveSelection extends Selection {
 	 */
 	_getSurroundingAttributes() {
 		const position = this.getFirstPosition();
+		const schema = this._document.schema;
 
 		let attrs = null;
 
@@ -565,6 +567,11 @@ export default class LiveSelection extends Selection {
 
 			// ...look for a first character node in that range and take attributes from it.
 			for ( const item of range ) {
+				// If the item is an object, we don't want to get attributes from its children.
+				if ( item.item instanceof Element && schema.objects.has( item.item.name ) ) {
+					break;
+				}
+
 				// This is not an optimal solution because of https://github.com/ckeditor/ckeditor5-engine/issues/454.
 				// It can be done better by using `break;` instead of checking `attrs === null`.
 				if ( item.type == 'text' && attrs === null ) {
--- a/src/model/liveselection.js
+++ b/src/model/liveselection.js
@@ -565,16 +565,16 @@ export default class LiveSelection extends Selection {
 			const range = this.getFirstRange();
 
 			// ...look for a first character node in that range and take attributes from it.
-			for ( const item of range ) {
+			for ( const value of range ) {
 				// If the item is an object, we don't want to get attributes from its children.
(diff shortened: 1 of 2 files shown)


src/model/liveselection.js
```diff
@@ -565,6 +567,11 @@ export default class LiveSelection extends Selection {
 
 			// ...look for a first character node in that range and take attributes from it.
 			for ( const item of range ) {
+				// If the item is an object, we don't want to get attributes from its children.
+				if ( item.item instanceof Element && schema.objects.has( item.item.name ) ) {
+					break;
```
Reinmar: Is there a test which verifies that this must be "break" instead of "return"?

src/model/liveselection.js
```diff
@@ -565,6 +567,11 @@ export default class LiveSelection extends Selection {
 
 			// ...look for a first character node in that range and take attributes from it.
 			for ( const item of range ) {
+				// If the item is an object, we don't want to get attributes from its children.
+				if ( item.item instanceof Element && schema.objects.has( item.item.name ) ) {
```
Reinmar: I think it can be `item.item.is( 'element' )`

tests/model/liveselection.js
```diff
@@ -872,4 +872,37 @@ describe( 'LiveSelection', () => {
 			expect( values ).to.deep.equal( [] );
 		} );
 	} );
+
+	describe( '_updateAttributes()', () => {
```
Reinmar: Why a new describe? Aren't there any other tests which verify this mechanism?

Reinmar: Besides, there are many tests missing. E.g. `<p>x[<$text bold=true>bar</$text></img>foo]</p>` and similar where object is inside the selection (or in a place reachable by that algorithm) but should not affect the method's behaviour.

Reinmar: Always try think not only about the case you want to fix but also about the cases you don't want to break.

tests/model/liveselection.js
```diff
@@ -872,4 +872,37 @@ describe( 'LiveSelection', () => {
...
+			doc.schema.objects.add( 'image' );
+
+			doc.schema.registerItem( 'caption', '$block' );
+			doc.schema.allow( { name: '$inline', inside: 'caption' } );
+			doc.schema.allow( { name: 'caption', inside: 'image' } );
+			doc.schema.allow( { name: '$text', attributes: 'bold', inside: 'caption' } );
+
+			root.removeChildren( 0, root.childCount );
+
+			root.insertChildren( 0, [
```
Reinmar: Better to use `setData()` from dev utils.

src/model/liveselection.js
pomek: After adding this test - https://github.com/ckeditor/ckeditor5-engine/pull/990/commits/3fddb2a8f28af3c21a4944688feb043a696f3785#diff-2d9aba768c395558c529a50282531971R898 - the test will fail if we use `return` instead of `break`.

tests/model/liveselection.js
pomek: I've just added two new tests. But I don't have an idea where these tests should be located, so I left it as it is (in new describe).

pomek:
@Reinmar, ready to review once again.

src/model/liveselection.js
```diff
@@ -565,6 +566,11 @@ export default class LiveSelection extends Selection {
 
 			// ...look for a first character node in that range and take attributes from it.
 			for ( const item of range ) {
```
Reinmar: Let's also take the chance to fix this variable name – it should be `value`. So we don't have `item.item` later on.

pomek: Sure, `item.item` looks a little strange.

pomek: Done in e62fe83f.

tests/model/liveselection.js
```diff
@@ -872,4 +872,43 @@ describe( 'LiveSelection', () => {
...
+			doc.schema.registerItem( 'image', '$inline' );
+			doc.schema.disallow( { name: 'image', attributes: 'bold', inside: '$root' } );
+			doc.schema.objects.add( 'image' );
+
+			doc.schema.registerItem( 'caption', '$block' );
+			doc.schema.allow( { name: '$inline', inside: 'caption' } );
+			doc.schema.allow( { name: 'caption', inside: 'image' } );
+			doc.schema.allow( { name: '$text', attributes: 'bold', inside: 'caption' } );
+
+			root.removeChildren( 0, root.childCount );
```
Reinmar: This is unnecessary.

tests/model/liveselection.js
Reinmar: OK, I'll clean this up cause these tests are indeed spread around a few blocks.

PR #1003, opened 2017-07-06
T/992 Fixes in selection rendering

### LINKED ISSUE #701: Markers: Selection after creating marker.
1. Open manual test for markers integration http://localhost:8125/tests/engine/tickets/643/1.html
2. Make a selection intersecting with initial marker.
3. Create a marker using a button.
4. Selection after marker is created is in a wrong place.

This may be smaller or bigger issue. There might be more cases producing incorrect result.

### LINKED ISSUE #992: Selection is incorrectly rendered when performing certain actions
See:

* https://github.com/ckeditor/ckeditor5-engine/issues/711#issuecomment-309722442
* https://github.com/ckeditor/ckeditor5/issues/453
* https://github.com/ckeditor/ckeditor5-basic-styles/issues/46 (the reported bug is fixed, but you can observe the problem with selection and performing that steps)
* https://github.com/ckeditor/ckeditor5-engine/issues/701
* https://github.com/ckeditor/ckeditor5-engine/issues/741#issuecomment-319896039

---

italic_blockquote_1

---

italic_blockquote_2

szymonkups:

Fix: DOM selection is updated when child elements inside that selection are changed. Closes #992. Closes #701.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -9,6 +9,7 @@
 
 import ViewText from './text';
 import ViewPosition from './position';
+import ViewRange from './range';
 import Selection from './selection';
 import { INLINE_FILLER, INLINE_FILLER_LENGTH, startsWithFiller, isInlineFiller, isBlockFiller } from './filler';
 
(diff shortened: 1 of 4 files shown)


tests/manual/markers.js
```diff
@@ -40,12 +40,29 @@ ClassicEditor.create( document.querySelector( '#editor' ), {
...
 		} );
 
-	window.document.getElementById( 'add-yellow' ).addEventListener( 'click', () => addHighlight( 'yellow' ) );
-	window.document.getElementById( 'add-red' ).addEventListener( 'click', () => addHighlight( 'red' ) );
-	window.document.getElementById( 'remove-marker' ).addEventListener( 'click', () => removeHighlight() );
-	window.document.getElementById( 'move-to-start' ).addEventListener( 'click', () => moveSelectionToStart() );
-	window.document.getElementById( 'move-left' ).addEventListener( 'click', () => moveSelectionByOffset( -1 ) );
-	window.document.getElementById( 'move-right' ).addEventListener( 'click', () => moveSelectionByOffset( 1 ) );
+	const addYellowButton = window.document.getElementById( 'add-yellow' );
+	addYellowButton.addEventListener( 'click', () => addHighlight( 'yellow' ) );
```
Reinmar: Why listening to two events? Can we just listen to mousedown and perform both actions on it?

src/view/renderer.js
```diff
@@ -582,24 +583,30 @@ export default class Renderer {
 	 */
 	_updateDomSelection( domRoot ) {
 		const domSelection = domRoot.ownerDocument.defaultView.getSelection();
-		const oldViewSelection = domSelection && this.domConverter.domSelectionToView( domSelection );
 
-		if ( oldViewSelection && this.selection.isEqual( oldViewSelection ) ) {
-			return;
-		}
+		// Check selection is the same if children are not changed.
```
Reinmar: TODO If we'll keep this solution, fix this comment.

Reinmar:
I wonder whether this solution is ok.

Why do we need to wrap the "similar selection" check in another if? Why isn't the similar selection check working? If it worked fine, the selection would be re-rendered. So there's something wrong with this.

AFAIR, you told us that the similar selection check uses an old selection, before rendering element changes. Can't we just use a fresh one – one retrieved by changing the DOM?

tests/manual/markers.js
szymonkups: One event will be enough.

szymonkups:
It looks like my solution is only preventing browser's wrong behaviour, which is not the cleanest solution. We will report it and wait for the fix in the browser.

PR #1005, opened 2017-07-07
T/1002: Introduce Position#getCommonAncestor( otherPosition ) and Range#getCommonAncestor()

### LINKED ISSUE #1002: Introduce Position#getCommonAncestor( otherPosition ) and Range#getCommonAncestor()
In order to properly resolve this ticket – https://github.com/ckeditor/ckeditor5-typing/issues/61#issuecomment-313341258 – we need to introduce two methods:

- `Position#getCommonAncestor( otherPosition )` - should return the common ancestor for two positions,
- `Range#getCommonAncestor()` - should return the common ancestor for the range's positions.

pomek:

Feature: Introduced `Position#getCommonAncestor( position )` and `Range#getCommonAncestor()` methods for `view` and `model`. Closes #1002.

### DIFF
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -332,6 +332,7 @@ export default class Position {
 
 		const node = this.root.getNodeByPath( this.getCommonPath( position ) );
 
+		// Although paths can indicate a text node, text node is not an ancestor of a position.
 		if ( node.is( 'text' ) ) {
 			return node.parent;
 		}
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -319,27 +319,24 @@ export default class Position {
 	}
 
 	/**
-	 * Returns a {@link module:engine/model/node~Node} which is a common ancestor for both positions. The {@link #root roots}
-	 * of these two positions must be identical.
+	 * Returns an {@link module:engine/model/element~Element} or {@link module:engine/model/documentfragment~DocumentFragment}
+	 * which is a common ancestor for both positions. The {@link #root roots} of these two positions must be identical.
 	 *
 	 * @param {module:engine/model/position~Position} position The second position.
-	 * @returns {module:engine/model/node~Node|null} Node that contains both positions.
+	 * @returns {module:engine/model/element~Element|module:engine/model/documentfragment~DocumentFragment|null}
 	 */
 	getCommonAncestor( position ) {
 		if ( this.root !== position.root ) {
 			return null;
 		}
 
-		const ancestorsA = this.getAncestors();
-		const ancestorsB = position.getAncestors();
+		const node = this.root.getNodeByPath( this.getCommonPath( position ) );
 
-		let i = 0;
-
-		while ( ancestorsA[ i ] == ancestorsB[ i ] && ancestorsA[ i ] ) {
-			i++;
+		if ( node.is( 'text' ) ) {
+			return node.parent;
 		}
 
-		return i === 0 ? null : ancestorsA[ i - 1 ];
+		return node;
 	}
 
 	/**
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -319,6 +319,30 @@ export default class Position {
 	}
 
 	/**
+	 * Returns a {@link module:engine/model/node~Node} which is a common ancestor for both positions. The {@link #root roots}
+	 * of these two positions must be identical.
+	 *
+	 * @param {module:engine/model/position~Position} position The second position.
+	 * @returns {module:engine/model/node~Node|null} Node that contains both positions.
+	 */
+	getCommonAncestor( position ) {
+		if ( this.root !== position.root ) {
+			return null;
+		}
+
+		const ancestorsA = this.getAncestors();
+		const ancestorsB = position.getAncestors();
+
+		let i = 0;
+
+		while ( ancestorsA[ i ] == ancestorsB[ i ] && ancestorsA[ i ] ) {
+			i++;
+		}
+
+		return i === 0 ? null : ancestorsA[ i - 1 ];
+	}
+
+	/**
 	 * Returns a new instance of `Position`, that has same {@link #parent parent} but it's offset
 	 * is shifted by `shift` value (can be a negative value).
 	 *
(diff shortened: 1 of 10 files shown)


src/model/position.js
```diff
@@ -319,6 +319,30 @@ export default class Position {
 	}
 
 	/**
+	 * Returns a {@link module:engine/model/node~Node} which is a common ancestor for both positions. The {@link #root roots}
+	 * of these two positions must be identical.
+	 *
+	 * @param {module:engine/model/position~Position} position The second position.
+	 * @returns {module:engine/model/node~Node|null} Node that contains both positions.
```
Reinmar: DocumentFragment is not a node. So the correct notation is `Element|DF`.

src/model/position.js
```diff
@@ -319,6 +319,30 @@ export default class Position {
...
+	 */
+	getCommonAncestor( position ) {
+		if ( this.root !== position.root ) {
+			return null;
+		}
+
+		const ancestorsA = this.getAncestors();
```
Reinmar: I liked the previous implementation more. The one which reused `getCommonPath()`. It was shorter and more readable.

Let me guess – you decided to skip it because DF doesn't implement `getNodeByPath()`? ;) Why not borrowing it from `Element` for `DF`?

src/model/range.js
```diff
@@ -446,6 +446,15 @@ export default class Range {
 	}
 
 	/**
+	 * Returns a {@link module:engine/model/node~Node} which is a common ancestor for both positions.
+	 *
+	 * @returns {module:engine/model/node~Node|null}
```
Reinmar: Again, wrong type.

src/model/position.js
pomek: You are right :D 

If it does not a problem, I could do it as you said.

src/view/position.js
```diff
@@ -176,6 +176,25 @@ export default class Position {
 	}
 
 	/**
+	 * Returns a {@link module:engine/view/node~Node} which is a common ancestor for both positions.
+	 *
+	 * @param {module:engine/view/position~Position} position
+	 * @returns {module:engine/view/node~Node|null}
+	 */
+	getCommonAncestor( position ) {
+		const ancestorsA = this.getAncestors();
```
Reinmar: `getAncestors()` accept `options.includeNode` so will this work in such a case: `<p>x{y}z</p>`? For such positions the text node should be returned. And for such `<p>[x]</p>` the element.

Reinmar: The second issue – you didn't pass `options.parentFirst` too so how can you iterate from `i==0` (leafs)? What if one ancestors array is longer than the other?

src/view/position.js
```diff
@@ -176,6 +176,25 @@ export default class Position {
 	}
 
 	/**
+	 * Returns a {@link module:engine/view/node~Node} which is a common ancestor for both positions.
+	 *
+	 * @param {module:engine/view/position~Position} position
+	 * @returns {module:engine/view/node~Node|null}
```
Reinmar: In this case it should be `DF|Node` because here we can also get text nodes.

tests/model/position.js
```diff
@@ -844,4 +844,46 @@ describe( 'Position', () => {
 			).to.throw( CKEditorError, /model-position-fromjson-no-root/ );
 		} );
 	} );
+
+	describe( 'getCommonAncestor()', () => {
```
Reinmar: All tests use the same length of paths.

tests/view/position.js
```diff
@@ -519,4 +519,43 @@ describe( 'Position', () => {
 			document.destroy();
 		} );
 	} );
+
+	describe( 'getCommonAncestor()', () => {
```
Reinmar: Here you use positions on the same depts as well.

src/model/position.js
Reinmar: It will make a lot of sense to unify the APIs of Element and DF, so it's definitely good to add this method there.

src/model/position.js
```diff
@@ -319,6 +319,27 @@ export default class Position {
...
+	 * @returns {module:engine/model/element~Element|module:engine/model/documentfragment~DocumentFragment|null}
+	 */
+	getCommonAncestor( position ) {
+		if ( this.root !== position.root ) {
+			return null;
+		}
+
+		const node = this.root.getNodeByPath( this.getCommonPath( position ) );
+
+		if ( node.is( 'text' ) ) {
```
Reinmar: This is something I expected from this method's model implementation. But I wonder what other think. Is "text node" a position's ancestor?

Reinmar: OK, @scofalik and @pjasiun confirm that a text node is not an ancestor of position.

Reinmar: Let's add a comment here saying that "Although paths can indicate a text node, text node is not an ancestor of a position."

Reinmar:
@pjasiun and @scofalik pointed out that the tests miss these 2 tests:

```html
<p>{}<a></a></p>
<p>xxx{}xxx</p>
```

They will prove that using `getCommonPath()` will not work.

tests/view/position.js
```diff
@@ -519,4 +519,69 @@ describe( 'Position', () => {
...
+		} );
+
+		it( 'for two positions in the same element returns the element', () => {
+			const fPosition = new Position( li1, 0 );
+			const zPosition = new Position( li1, 2 );
+
+			test( fPosition, zPosition, li1 );
+		} );
+
+		it( 'works when one positions is nested deeper than the other', () => {
```
Reinmar: "when one *position*"

Reinmar: or rather "one of the positions"

tests/view/position.js
```diff
@@ -519,4 +519,69 @@ describe( 'Position', () => {
 			document.destroy();
 		} );
 	} );
+
+	describe( 'getCommonAncestor()', () => {
```
Reinmar: There's just a single test with positions in text nodes and it uses a `0` offset. Please add more,  with better offsets (ones which have a chance to break something).

PR #1007, opened 2017-07-10
T/1006 Expanded special case in `Range#_getTransformedByDelta`.

### LINKED ISSUE #1006: Range at the end of element is incorrectly transformed when split
Originally reported here: https://github.com/ckeditor/ckeditor5/issues/485

The issue is that in `Range#_getTransformedByDelta` the range is incorrectly transformed. Two ranges (that are touching) are returned through default algorithm. Then, one of those ranges is ignored. One range should be returned instead.

scofalik:

Fixed: Live ranges and markers, that are at the end of an element, are now correctly transformed when they are split. Closes #1006.

### DIFF
--- a/src/model/range.js
+++ b/src/model/range.js
@@ -462,13 +462,13 @@ export default class Range {
 		} else {
 			const sourceRange = Range.createFromPositionAndShift( sourcePosition, howMany );
 
-			// Edge case for merge detla.
+			// Edge case for merge delta.
 			if (
 				deltaType == 'merge' &&
 				this.isCollapsed &&
 				( this.start.isEqual( sourceRange.start ) || this.start.isEqual( sourceRange.end ) )
 			) {
-				// Collapsed range is in merged element.
+				// Collapsed range is in merged element, at the beginning or at the end of it.
 				// Without fix, the range would end up in the graveyard, together with removed element.
 				// <p>foo</p><p>[]bar</p> -> <p>foobar</p><p>[]</p> -> <p>foobar</p> -> <p>foo[]bar</p>
 				// <p>foo</p><p>bar[]</p>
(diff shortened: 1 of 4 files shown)


src/model/range.js
```diff
@@ -462,13 +462,13 @@ export default class Range {
...
 
-			// Edge case for merge detla.
+			// Edge case for merge delta.
 			if (
 				deltaType == 'merge' &&
 				this.isCollapsed &&
 				( this.start.isEqual( sourceRange.start ) || this.start.isEqual( sourceRange.end ) )
 			) {
-				// Collapsed range is in merged element.
+				// Collapsed range is in merged element, at the beginning or at the end of it.
```
Reinmar: Where's the "at the beginning" part checked in the code?

Reinmar: Actually, the whole condition got too complex and is hard to read. Could you move it to some helper fn and split into a couple of statements?

tests/model/range.js
```diff
@@ -918,8 +918,12 @@ describe( 'Range', () => {
 		} );
 
 		describe( 'by SplitDelta', () => {
-			it( 'split inside range', () => {
+			beforeEach( () => {
+				// Cleaning...
 				root.removeChildren( root.childCount );
```
Reinmar: removeChildren() accepts `index + howMany` so this looks like a mistake.

src/model/range.js
scofalik: > Where's the "at the beginning" part checked in the code?

`this.start.isEqual( sourceRange.start )`

Or do you mean tests?

> Actually, the whole condition got too complex and is hard to read.

I haven't changed it. Honestly there's not much to simplify here, except of the last part. I can make some kind of `isAtTheBeginningOrEndOfRange( this.start, sourceRange )` or `isRangeBoundary( this.start, sourceRange )` but IDK how much it helps.

tests/model/range.js
scofalik: Actually there's no need to add and remove any content from root in those tests...

PR #1011, opened 2017-07-12
T/1001: Remove selection attributes stored in element once it's not empty any more

### LINKED ISSUE #1001: Delete element attributes once it has any content
See https://github.com/ckeditor/ckeditor5-enter/issues/40:

> The style should be removed from a block once something is inserted in it. Who's job should it be? Can we actually understand which attrs applied to a block are inline attrs? I remeber we discussed this but I don't know how we implemented it.

And:

> We talked briefly with @scofalik and it seems that the listener which clears the element attributes (they can be recognised by the selection: prefix) will need to be placed in the Document.
> 
> It should be fairly simple – it needs to listen to insert and move operations and just clear attributes of an element to which insertion happens. It doesn't even need to check if it's empty or not – if, by any coincidence the attributes weren't cleared before they can be cleared now.

Reinmar:

Internal: Attributes which selection stores in elements should be removed once they are not empty any more. Closes #1001.

---

### Additional information

I marked this change as internal as it should invisible to the developers.

In this PR I also changed `LiveSelection` to `DocumentSelection` which is kinda internal change as well as this class is not meant to be used directly.

### DIFF
--- a/tests/model/documentselection.js
+++ b/tests/model/documentselection.js
@@ -997,14 +997,15 @@ describe( 'DocumentSelection', () => {
 				expect( emptyP.parent ).to.equal( root ); // Just to be sure we're checking the right element.
 			} );
 
-			it( 'are not removed when containing element is merged with another empty element', () => {
+			it( 'are not removed or merged when containing element is merged with another empty element', () => {
 				const emptyP2 = new Element( 'p', null );
 				root.appendChildren( emptyP2 );
 
 				emptyP.setAttribute( fooStoreAttrKey, 'bar' );
-				emptyP2.setAttribute( fooStoreAttrKey, 'bar' );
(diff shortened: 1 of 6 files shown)


tests/model/documentselection.js
```diff
@@ -893,49 +906,158 @@ describe( 'LiveSelection', () => {
...
+				const spy = sinon.spy( batch, 'removeAttribute' );
+
+				// <emptyP>{}<emptyP2>
+				batch.merge( Position.createAfter( emptyP ) );
+
+				expect( emptyP.hasAttribute( fooStoreAttrKey ) ).to.be.false;
+				expect( spy.calledOnce ).to.be.true;
+			} );
+
+			it( 'are not removed on transparent batches', () => {
```
scofalik: 👍

Issue #664, opened 2016-10-31
Renderer cannot set selection at the beginning of an attribute element (x<b>{}y</b>)


Reinmar:
It thinks that the filler isn't needed in this case, but unfortunately it is. The tests that we have are incorrect and used `domSelection.getRanges()` to check where the selection was rendered, but they should use the anchor and offset which are nicely normalised by Webkit/Blink.

I'll comment some tests out when working on #659, and we'll need to bring them back. It's possible that the only thing that needs to be done in this ticket is to fix `Renderer#_needsInlineFillerAtSelection()`.

scofalik:
This is connected with #629. I already made a PR fixing this issue but it may be overengineered. It all comes to the fact that the selection cannot be placed in `<b>^x</b>` and it is incorrectly (according to view) placed outside of it.

scofalik:
OTOH, this issue concerns only collapsed selection, while the one in #629 is connected with non-collapsed selections too.

We could rewrite `_needAddInlineFiller` so that it takes care of non-collapsed selections too. But that would mean possibiltiy of adding multiple inline fillers.

Reinmar:
I was confused by this ticket for two reasons:

* `anchorNode` is not normalised anymore by Blink and it's possible to place the selection at the beginning of the `<b>` element. In Webkit (Safari) it's still normalised so the selection apparently ends up outside the element. I think that it may have implications for composition when the engine would try to fix the text insertion which happened outside instead of inside the `<b>` element.
* However, besides that, I don't see any other TC where it would have some impact on how the editor works. And I think that it's a rather rare case for someone to press <kbd>Ctrl</kbd>+<kbd>B</kbd> when at `foo{}<b>bar</b>` to start typing with bold at that place.

Reinmar:
cc @f1ames sounds like part of your ZWS changes?

f1ames:
Yes, should be solved by a second inline filler improvement (first is #898 and second _Insert the inline filler when selection is at an inline element boundary_ mentioned in #1342).

scofalik:
"First inline improvement", "Second inline improvement"...

Sounds like some kind of physics laws, like the first law of thermodynamics.

PR #1016, opened 2017-07-16
T/1014 view.Renderer does not create inline filler in newly created DOM elements

### LINKED ISSUE #1014: view.Renderer does not create inline filler in newly created DOM elements
As the title says.

At the moment, inline filler is handled by `view.Renderer#_updateChildren`, but only if the filler is directly in the updated children. Howeer, newly created items are not processed directly by `_updateChildren` and do not receive inline filler.

scofalik:

Fix: Inline filler will be correctly inserted if it is deep in a newly created view structure that is just converted to DOM. Closes #1014.

### DIFF
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -252,7 +252,6 @@ export default class DomConverter {
 		let offset = 0;
 
 		for ( const childView of viewElement.getChildren() ) {
-			// debugger;
 			if ( blockFillerOffset === offset ) {
 				yield this.blockFiller( domDocument );
 			}
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -16,7 +16,10 @@ import ViewRange from './range';
 import ViewSelection from './selection';
 import ViewDocumentFragment from './documentfragment';
 import ViewTreeWalker from './treewalker';
-import { BR_FILLER, INLINE_FILLER_LENGTH, isBlockFiller, isInlineFiller, startsWithFiller, getDataWithoutFiller } from './filler';
+import {
+	BR_FILLER, INLINE_FILLER, INLINE_FILLER_LENGTH, isBlockFiller,
+	isInlineFiller, startsWithFiller, getDataWithoutFiller
+} from './filler';
 
 import global from '@ckeditor/ckeditor5-utils/src/dom/global';
 import indexOf from '@ckeditor/ckeditor5-utils/src/dom/indexof';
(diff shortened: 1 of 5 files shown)


src/view/domconverter.js
```diff
@@ -243,22 +247,41 @@ export default class DomConverter {
 	 * @returns {Iterable.<Node>} DOM nodes.
 	 */
 	* viewChildrenToDom( viewElement, domDocument, options = {} ) {
-		const fillerPositionOffset = viewElement.getFillerOffset && viewElement.getFillerOffset();
+		const blockFillerOffset = viewElement.getFillerOffset && viewElement.getFillerOffset();
+		const inlineFillerOffset = options.inlineFiller && options.inlineFiller.parent == viewElement ? options.inlineFiller.offset : null;
```
Reinmar: I never remember which operator has precedence over which. Could you wrap it with `()` to make it easier to read?

src/view/domconverter.js
```diff
@@ -243,22 +247,41 @@ export default class DomConverter {
...
-			if ( fillerPositionOffset === offset ) {
+			if ( blockFillerOffset === offset ) {
 				yield this.blockFiller( domDocument );
 			}
 
-			yield this.viewToDom( childView, domDocument, options );
+			const domNode = this.viewToDom( childView, domDocument, options );
+
+			// We need to insert inline filler before the dom node.
+			if ( inlineFillerOffset === offset ) {
```
Reinmar: And what if the offset indicates `text^<element>`? Will we create a new text node? Shouldn't we merge to the left then?

src/view/domconverter.js
```diff
@@ -243,22 +247,41 @@ export default class DomConverter {
...
+				// If dom node that was about to return is a text node, just add inline filler to that text node.
+				if ( this.isText( domNode ) ) {
+					domNode.data = INLINE_FILLER + domNode.data;
```
Reinmar: I see that until now DomConverter didn't create the inline filler. It's something new for it to start doing so. It did create block filler in the past, but I wonder what do you think about it, @pjasiun?

tests/view/renderer.js
```diff
@@ -1216,6 +1217,40 @@ describe( 'Renderer', () => {
 			} ).to.throw( CKEditorError, /^view-renderer-filler-was-lost/ );
 		} );
 
+		// #1014.
+		it( 'should create inline filler in newly created dom nodes', () => {
```
Reinmar: I wanted comment that perhaps this test should be split into two separate cases but its first part worked well before. As @scofalik commented in #1014 all these conditions need to be met.

src/view/domconverter.js
Reinmar: I'd be worried that the renderer may not expect to get the filler in some places. Especially that it needs to track it in `Renderer#_inlineFiller` – so how will it do this? I see that `_getInlineFillerNode()` should handle this but since we're leaking this logic to a separate class we'll risk that this will desync.

Reinmar:
I think that we need to answer the question whether this can be handled in the renderer itself. Either by calling `_updateChildren()` on the subtrees or handling it in a different way (e.g. when rendering the selection). In fact, I thought that we do add the filler once the selection changes when rendering the selection.

How does it happen that it's added in such a case:

```html
<p>[]foo<b>xx</b></p>
```

And you click after "foo"? The answer is: it does not happen. There's a bug. See also:

jul-18-2017 16-46-33

So I think that we should look wider – the filler needs more attention.

Reinmar:
There's also: https://github.com/ckeditor/ckeditor5-engine/issues/664

src/view/domconverter.js
scofalik: AFAI understand, this is incorrect scenario. Filler can be only at the beginning of the text node. Filler should not be needed in this case, and if it was already in the DOM, it will be just preversed (at the beginning of the text node).

scofalik:
After discussion we came to conclusion that we should solve this issue differently and we don't want creating inline filler in `DomConverter`.

PR #1019, opened 2017-07-18
Changed: Placeholder text should not be hidden if element has only ui elements.

### LINKED ISSUE #1018: Do not hide placeholder if ui element is only children
This is a step first to solve https://github.com/ckeditor/ckeditor5-engine/issues/1017.

We should not hide placeholder text if `view.UIElement`s are only children of placeholder-ed element.

scofalik:

Fix: Placeholder text now will not be hidden if the element has only ui elements. Closes #1018.

### DIFF
--- a/src/view/placeholder.js
+++ b/src/view/placeholder.js
@@ -110,15 +110,19 @@ function updateSinglePlaceholder( element, checkFunction ) {
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


scofalik:
There is no 100% CC but this is not caused by this PR and is already fixed in other PR.

PR #1023, opened 2017-07-18
Changed: Block filler should be rendered after all element's children.

### LINKED ISSUE #1021: Render block filler after children UI elements
Related: https://github.com/ckeditor/ckeditor5-engine/issues/1017

At the moment, block filler is always rendered at the beginning of an element. This is incorrect, since there is a possibility that such block filler will move UI element to a new line:

```html
<h2 class="ck-placeholder" data-placeholder="Type some header text...">
  <br data-cke-filler="true">
  <span class="myUi"></span>
</h2>
```

`view.ContainerElement#getFillerOffset()` should return an offset after all UI elements.

scofalik:

Fixed: Block filler was rendered before UI elements, interfering with their positioning. Now it will be properly rendered at the end of element. Closes #1021.

### DIFF
--- a/src/view/containerelement.js
+++ b/src/view/containerelement.js
@@ -78,5 +78,6 @@ export default class ContainerElement extends Element {
 //
 // @returns {Number|null} Block filler offset or `null` if block filler is not needed.
 function getFillerOffset() {
-	return Array.from( this.getChildren() ).some( element => !element.is( 'uiElement' ) ) ? null : 0;
+	// Render block filler at the end of element (after all ui elements).
+	return Array.from( this.getChildren() ).some( element => !element.is( 'uiElement' ) ) ? null : this.childCount;
 }
--- a/src/view/containerelement.js
+++ b/src/view/containerelement.js
@@ -78,6 +78,13 @@ export default class ContainerElement extends Element {
 //
 // @returns {Number|null} Block filler offset or `null` if block filler is not needed.
 function getFillerOffset() {
-	// Render block filler at the end of element (after all ui elements).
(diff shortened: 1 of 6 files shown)


scofalik:
I forgot about one thing, so auto -R for a while.

scofalik:
False alarm...

src/view/containerelement.js
```diff
@@ -78,5 +78,6 @@ export default class ContainerElement extends Element {
 //
 // @returns {Number|null} Block filler offset or `null` if block filler is not needed.
 function getFillerOffset() {
-	return Array.from( this.getChildren() ).some( element => !element.is( 'uiElement' ) ) ? null : 0;
+	// Render block filler at the end of element (after all ui elements).
```
Reinmar: This comment is incorrect. It's not after all UI elements - we don't do render the filler if there are any UI elements.

Reinmar: Or I don't get this condition.

Reinmar:
This change break the UI element manual test: http://localhost:8125/ckeditor5-engine/tests/view/manual/uielement.html

It either needs to be aligned to still makes sense or there's something wrong.

image

First thing is that using a div is a bad idea in general – p>div is invalid. 

Then, we should make it a "start paragraph" mark to work well with how the block filler is inserted now. 

But then... we still have the inline filler injected there and it will still break this test. So the selection would need to always end up after the UI element. I guess the test needs to be completely rewritten to better reflect how we're using markers.

scofalik:
Also the test sets the ui element to have `contenteditable=false` which brings some other complications (for example a caret can be placed after the ui element).

scofalik:
I don't know why this build does not pass...

Reinmar:
Most likely because it was based on a too old master.

PR #1025, opened 2017-07-18
Hide caret when an editor is read-only.

### LINKED ISSUE #1024: Hide caret when editor is read-only
See https://github.com/ckeditor/ckeditor5/issues/503.

oskarwrobel:

Feature: Hide caret when an editor is read-only. `EditingControler` is observable from now. Observable property `isReadOnly` is added to the `ViewDocument` and `EditingController`. Closes #1024. Closes ckeditor/ckeditor5#503.

---

### Related:
- https://github.com/ckeditor/ckeditor5-widget/pull/8
- https://github.com/ckeditor/ckeditor5-core/pull/99

### DIFF
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -23,11 +23,15 @@ import {
 } from '../conversion/model-selection-to-view-converters';
 
 import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
+import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
+import mix from '@ckeditor/ckeditor5-utils/src/mix';
 
 /**
  * Controller for the editing pipeline. The editing pipeline controls {@link ~EditingController#model model} rendering,
  * including selection handling. It also creates {@link ~EditingController#view view document} which build a
  * browser-independent virtualization over the DOM elements. Editing controller also attach default converters.
+ *
+ * @mixes module:utils/observablemixin~ObservableMixin
  */
 export default class EditingController {
 	/**
@@ -161,3 +165,5 @@ export default class EditingController {
 		this._listener.stopListening();
 	}
 }
+
+mix( EditingController, ObservableMixin );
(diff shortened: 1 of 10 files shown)


src/view/observer/selectionobserver.js
```diff
@@ -133,7 +133,7 @@ export default class SelectionObserver extends Observer {
 	 * @param {Document} domDocument DOM document.
 	 */
 	_handleSelectionChange( domDocument ) {
-		if ( !this.isEnabled || !this.document.isFocused ) {
+		if ( !this.isEnabled || ( !this.document.isFocused && !this.document.isReadOnly ) ) {
```
pjasiun: Comment needed.

pjasiun:
Also closes: https://github.com/ckeditor/ckeditor5/issues/503.

src/controller/editingcontroller.js
```diff
@@ -61,6 +64,19 @@ export default class EditingController {
 		this.mapper = new Mapper();
 
 		/**
+		 * Defines whether controller is in read-only mode.
```
Reinmar: The controller is not in a read only mode. It's this controller's view.

Reinmar: I mean – don't describe the code. Describe what it means.

And I've just thought that perhaps this property isn't needed at all in the EditingController and should be added to the view Document only? Then we'll avoid this strange situation which we can't even properly document.

Reinmar:
If you agree with my comments in https://github.com/ckeditor/ckeditor5-engine/pull/1025/files/81bbca10a46609b8659a9780fb89bb20ad06f875#diff-ab3352a1c25e7fa470c4c408009c6263 please open a ticket for that.

src/controller/editingcontroller.js
pjasiun: Yea, I was thinking about it too, we even talked about it with @oskarwrobel. On the one hand, we can say that editing controller is a facade, so it can have this property. On the other, we don't use it this way usually. In most cases, we use nested properties directly. So I agree that we could not make an exception here.

Reinmar: Good. Less code will be actually easier to comprehend. So I'll create a ticket.

PR #1027, opened 2017-07-19
T/2016: Introduced support for "keyup" event.

### LINKED ISSUE #1026: Support for `keyup` event on document
In order to detect whether the user keeps the <kbd>Backspace</kbd> key, we need to add support for the `keyup` event called on `Document`.

It's required in order to implement a test case **II** from here – https://github.com/ckeditor/ckeditor5-typing/issues/61#issuecomment-314705304.

pomek:

Feature: `KeyObserver` will handle the `keyup` event. Message. Closes #1026.

### DIFF
--- a/src/view/observer/keyobserver.js
+++ b/src/view/observer/keyobserver.js
@@ -50,20 +50,26 @@ export default class KeyObserver extends DomEventObserver {
  *
  * @see module:engine/view/observer/keyobserver~KeyObserver
  * @event module:engine/view/document~Document#event:keydown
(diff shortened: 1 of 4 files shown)


Reinmar:
Misses `keyup` event API docs.

Reinmar:
Did you check that it works and we have two events (keydown and keyup) with proper attributes?

pomek:
image

IMO works as it should.

Reinmar:
I meant documentation.

pomek:
After a fix added in bccf3f4 it works well:

image

PR #1032, opened 2017-07-20
T/1020 Fixes for placeholder with UI element and selection.

### LINKED ISSUE #613: It is posible to type at the begging of the 7ZWS
This one is tricky.

First of all because of: https://github.com/ckeditor/ckeditor5-engine/issues/462 it is not possible to reproduce it easily. On the other hand when https://github.com/ckeditor/ckeditor5-engine/issues/462 will be fixed this could be another reason of many issues.

Then problem is that:

We use 7 ZWS as an inline filling character. And we assume that user will always type at the end of it so only check if the string starts with 7 ZWS. But for some strange reason it is possible to put the caret at the begging of the 7 ZWS.

To reproduce the bug you can modify the classic editor manual test adding:
```js
editor.document.schema.registerItem( 't', '$block' );

buildModelConverter().for( editor.data.modelToView, editor.editing.modelToView )	
	.fromElement( 't' )
	.toElement( new ViewContainerElement( 'span', { contenteditable: false } ) );

buildViewConverter().for( editor.data.viewToModel )
	.fromElement( 'span' )
	.toElement( 't' );

editor.setData( '<p><span>foo</span></p><p>bar</p>' );
```

Then:
 - init editor,
 - put the caret at the beginning of the second line (before `bar`),
 - press arrow up,
 - start typing,
 - check console.

You will see that you type at the beginning of the inline filter what is unexpected. Then when you move the selection you get:
```
renderer.js:279 Uncaught CKEditorError: view-renderer-render-no-inline-filler: No inline filler on expected position.
```
Because renderer checks if the text starts with 7 ZWS to thread it as the filler and since you added text at the beginning of it, it will not work. Also, other algorithms may not work, like the one to jump over the filler.

The question is how it is possible to type before the 7 ZWS and if we can fix it to type after. If it is not possible some algorithms may need a fix to handle the situation where 7 ZWS are after the text.

### LINKED ISSUE #797: Selection in the UIElement should not be converter
...or should be moved before/after that `UIElement`, what could be a better solution in case of the non-collapsed selection.

### LINKED ISSUE #1020: Placeholder text interferes with UI elements
Related: #1017.

At the moment, placeholder text is created through a CSS pseudo-element:
```cs
.ck-placeholder::before {
	content: attr( data-placeholder );
	cursor: text;
	color: #c2c2c2;
	pointer-events: none; // See ckeditor/ckeditor5#469.
}
```

Unfortunately, this way, it collides with other elements that may be present in element with `.ck-placeholder` class. These are: fillers and `view.UIElement`s.

To properly display UI element, which is in _de facto_ empty element, we should display it at the beginning of placeholder element. On the other hand, placeholder cannot be after bogus BR (block filler) because it would end up in new line. This means that we can use neither `::before` nor `::after`.

My idea is to use `position: absolute` to make placeholder text independent of other possible content.

scofalik:

Fix: Placeholder text pseudo-element has now `position: absolute;` to not interfere with other elements that might be in "emptyish" element (for example UI elements). Closes #1020.
Fix: DOM Selection is now re-rendered and fixed when it is placed in DOM representation of `view.UIElement` or before/inside inline filler. Closes #797. Closes #613.

### DIFF
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -839,6 +839,50 @@ export default class DomConverter {
 	}
 
 	/**
+	 * Checks if given {Selection DOM Selection} boundaries are in correct places.
+	 *
+	 * Incorrect places for selection are:
+	 * * before or in the middle of inline filler sequence,
+	 * * inside DOM element that represents {@link module:engine/view/uielement~UIElement view ui element}.
+	 *
+	 * @param {Selection} domSelection DOM Selection object to be checked.
+	 * @returns {Boolean} `true` if given selection is at correct place, `false` otherwise.
+	 */
+	isCorrectDomSelection( domSelection ) {
+		return this._isCorrectDomSelectionPosition( domSelection.anchorNode, domSelection.anchorOffset ) &&
+			this._isCorrectDomSelectionPosition( domSelection.focusNode, domSelection.focusOffset );
+	}
+
+	/**
+	 * Checks if given DOM position is a correct place for selection boundary. See {@link ~isCorrectDomSelection}.
+	 *
+	 * @private
+	 * @param {Node} domParent Position parent.
+	 * @param {Number} offset Position offset.
+	 * @returns {Boolean} `true` if given position is correct place for selection boundary, `false` otherwise.
+	 */
+	_isCorrectDomSelectionPosition( domParent, offset ) {
+		// If selection is before or in the middle of inline filler string, it is incorrect.
+		if ( this.isText( domParent ) && startsWithFiller( domParent ) && offset < INLINE_FILLER_LENGTH ) {
+			// Selection in a text node, at wrong position (before or in the middle of filler).
+			return false;
+		} else if ( this.isElement( domParent ) && startsWithFiller( domParent.childNodes[ offset ] ) ) {
+			// Selection in an element node, before filler text node.
+			return false;
+		}
+
+		const viewParent = this.mapDomToView( domParent );
+
+		// If selection is in `view.UIElement`, it is incorrect. Note that `mapDomToView()` returns `view.UIElement`
+		// also for any dom element that is inside the view ui element (so we don't need to perform any additional checks).
+		if ( viewParent && viewParent.is( 'uiElement' ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
 	 * Takes text data from given {@link module:engine/view/text~Text#data} and processes it so it is correctly displayed in DOM.
 	 *
 	 * Following changes are done:
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -602,24 +602,27 @@ export default class Renderer {
 	 */
 	_updateDomSelection( domRoot ) {
 		const domSelection = domRoot.ownerDocument.defaultView.getSelection();
-		const oldViewSelection = domSelection && this.domConverter.domSelectionToView( domSelection );
 
-		if ( oldViewSelection && this.selection.isEqual( oldViewSelection ) ) {
-			return;
-		}
+		// Check whether dom selection needs fixing. If it does not need fixing maybe w don't have to update it at all.
+		if ( this.domConverter.isCorrectDomSelection( domSelection ) ) {
+			const viewSelectionFromDom = this.domConverter.domSelectionToView( domSelection );
 
-		if ( oldViewSelection && areSimilarSelections( oldViewSelection, this.selection ) ) {
-			const data = {
-				oldSelection: oldViewSelection,
-				currentSelection: this.selection
-			};
+			// Compare view selection assumed from dom with current view selection.
+			if ( this.selection.isCollapsed && this.selection.isEqual( viewSelectionFromDom ) ) {
+				return;
+			} else if ( areSimilarSelections( viewSelectionFromDom, this.selection ) ) {
+				const data = {
+					oldSelection: viewSelectionFromDom,
+					currentSelection: this.selection
+				};
 
-			log.warn(
-				'renderer-skipped-selection-rendering: The selection was not rendered due to its similarity to the current one.',
-				data
-			);
+				log.warn(
+					'renderer-skipped-selection-rendering: The selection was not rendered due to its similarity to the current one.',
+					data
+				);
 
-			return;
+				return;
+			}
 		}
 
 		// Multi-range selection is not available in most browsers, and, at least in Chrome, trying to
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -603,12 +603,17 @@ export default class Renderer {
 	_updateDomSelection( domRoot ) {
 		const domSelection = domRoot.ownerDocument.defaultView.getSelection();
 
-		// Check whether dom selection needs fixing. If it does not need fixing maybe w don't have to update it at all.
+		// Below we will check whether DOM Selection needs updating at all.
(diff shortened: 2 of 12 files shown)


scofalik:
If this is merged with master as is, there will be a new bug to fix -- it will be impossible to jump over UI element using right arrow key. When right arrow key is pressed, selection is placed inside UI element and then fixed back to where it was.

Reinmar:
Did you check how the placeholder behaves now in Firefox, Safari and Edge?

tests/manual/placeholder.html
```diff
@@ -1 +1,5 @@
+<style type="text/css">
+	.myUi { display: inline-block; width: 6px; height: 6px; position: relative; top: -16px; background: #AA0000; pointer-events: none; }
```
pjasiun: I don't think we should use `display: inline-block` since we know there are problems with it. Why not `display: inline;`?

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+	 * Incorrect places for selection are:
+	 * * before or in the middle of inline filler sequence,
+	 * * inside DOM element that represents {@link module:engine/view/uielement~UIElement view ui element}.
+	 *
+	 * @param {Selection} domSelection DOM Selection object to be checked.
+	 * @returns {Boolean} `true` if given selection is at correct place, `false` otherwise.
+	 */
+	isCorrectDomSelection( domSelection ) {
+		return this._isCorrectDomSelectionPosition( domSelection.anchorNode, domSelection.anchorOffset ) &&
+			this._isCorrectDomSelectionPosition( domSelection.focusNode, domSelection.focusOffset );
```
pjasiun: Wouldn't it be more correct to check all start and end positions of all ranges?

src/view/renderer.js
```diff
@@ -602,24 +602,27 @@ export default class Renderer {
 	 */
 	_updateDomSelection( domRoot ) {
 		const domSelection = domRoot.ownerDocument.defaultView.getSelection();
-		const oldViewSelection = domSelection && this.domConverter.domSelectionToView( domSelection );
 
-		if ( oldViewSelection && this.selection.isEqual( oldViewSelection ) ) {
-			return;
-		}
+		// Check whether dom selection needs fixing. If it does not need fixing maybe w don't have to update it at all.
```
pjasiun: w -> we

pjasiun: Actually, I don't get it. The comment says "If it does not need fixing maybe w don't have to update it at all." but the there is the condition which is executed only if the selection is correct.

pjasiun:
Also, I agree with @Reinmar that other browsers should be checked.

src/view/domconverter.js
scofalik: At this moment it doesn't matter, because AFAIR we only use DOM Selection's anchor and offset, because we don't support multi-range DOM selections. So theoretically you are correct, practically right now it doesn't matter.

pjasiun:
But in general the change looks good.

scofalik:
I've checked on other browsers. Have you found any errors?

Reinmar:
I still want to look into this. I'd like to understand why Chrome puts the selection inside an empty element. It never did that so it looks very suspicious but we have a theory that this is due to `display: inline-block` style.

tests/manual/placeholder.html
scofalik: I needed to give this marker a shape. With `display: inline` it would be empty. I could create a UI element that is inline and has `inline-block` or `position: absolute` element inside, though (but I didn't check if it will work).

src/view/renderer.js
scofalik: Yup.

What I meant was that selection "needs fixing" when [DOM] Selection is not correct. I will clarify this comment.

scofalik:
I can try to see what happens when there are two `display: inline` `<span>`s, one with `position: relative`, the other with `position: absolute`.

scofalik:
I've created UI Element composed of outer element with `position: relative` and inside element with `position: absolute`. None had `dispaly: inline-block`.

If there is a character after such UI Element, browser does not place selection in it. If this is the last element, the selection is placed in it.

jul-21-2017 15-47-29

scofalik:
I've removed R- label. Please tell me what are we doing with this issue. Not having `display: inline-block;` works better, but issue still exists. The fix solves issue both for `display: inline-block;` UI elements and for `position: absolute;` elements. But will introduce new issue and we will have to write `jumpOverUIElement` functionality. 

TBH with `display: inline-block;` it works better, because at least arrow down is working (with `position: absolute;` it does not).

theme/placeholder.scss
```diff
@@ -2,8 +2,15 @@
...
+	position: absolute;
+	top: 0;
+	left: 0;
 	content: attr( data-placeholder );
 	cursor: text;
 	color: #c2c2c2;
 	pointer-events: none; // See ckeditor/ckeditor5#469.
 }
+
+.ck-placeholder {
```
Reinmar: I'm afraid that we must not do that. This class is assigned to an element which, in the case of inline editors, may be a part of a website with which the editor is integrated. It will also be applied to e.g. nested editables which may also require a certain styling. So this is really unsafe.

Another unsafe thing here is the `pos:abs, top:0; left:0` styling. See e.g. https://jsfiddle.net/n2gzLqhz/2/:

image

Reinmar:
As [you might've seen](https://github.com/ckeditor/ckeditor5-engine/pull/1032/files#r128926648), part of this PR is based on CSS styles which I don't think we can use.

However, I don't understand one thing – is #1020 so important to introduce such changes? AFAIU this, we can simply hide those UIElements which want to appear in empty elements with placeholders in them. Can't we?

> Fix: DOM Selection is now re-rendered and fixed when it is placed in DOM representation of view.UIElement or before/inside inline filler. Closes #797. Closes #613.

This is a different story and I understand the importance of this part of the fix. However, I don't see how it's related to #1020 so I propose to focus on this mechanism only.

If you agree with me, please clean up this PR to contain only related changes. And we'd need some dedicated manual test on which we could verify selection fixing mechanisms.

theme/placeholder.scss
scofalik: There is no perfect solution.

First, let me address `position: relative` doubt. This might break something only if the element previously had `position: absolute` or `position: fixed`. Even then, most of the times, the element will have a "stronger path" if it is styled by external CSS and that will overwrite our `position: relative`.

`position: absolute; left: 0; top: 0;` is more dangerous, I agree, however it works with `margin`s. Maybe we could try just setting `position: absolute` without `top` and `left`.

Reinmar: If we'll do anything dangerous with these styles I'd like to have very strong reasons to do so. In fact, some bugs are easier to live with (or hack) than such CSS rules. Therefore, as I expressed in https://github.com/ckeditor/ckeditor5-engine/pull/1032#issuecomment-317274915, I'd like to understand if there are really so strong reasons to go this way.

scofalik:
The conclusion of this PR is this:
1. The PR touched two different problems: preventing placing selection in ui element and having ui elements and placeholders together.
2. The selection issue fix has been extracted to its own issue and PR: https://github.com/ckeditor/ckeditor5-engine/pull/1050
3. The solution for placeholder + ui elements issue presented in this PR is dangerous so we will abandon it and try to solve this issue differently.

This means that I am closing this PR.

PR #1037, opened 2017-07-21
T/1036: Introduced options.includeSelf to getCommonAncestor()

### LINKED ISSUE #1036: Node#getCommonAncestor() – add includeSelf option
Well, it turns out that the method we introduced in #1033 does not really solve the case that I have in the hands. And we realised that it's quite common that if you have a structure like:

```html
<a><b></b></a>
```

And elements "a" and "b" as some targets, then you may be interested that the entire target is still "a". So, it's more like "get common container".

PS. Turns out that `getAncestors()` already have `includeNode`. So, if in this case we can have non ancestors returned, then we can definitely do the same in `getCommonAncestor()`. But, I'd also like to rename `includeNode` to `includeSelf`.

Reinmar:

Internal: Introduced `options.includeSelf` to `getCommonAncestor()`. Closes #1036.

BREAKING CHANGE: The `includeNode` option of `Node#getAncestors()` methods (model and view) was renamed to `includeSelf`. See #1036.

---

### Additional information

Internal because it adds a feature to a method I've just added.

Additional branch: ckeditor5-image#t/ckeditor5-engine/1036. (Merge message: "Internal: Adjusted engine's method use to its changes.")

All clean on `npm t`: `Chrome 59.0.3071 (Mac OS X 10.12.4): Executed 6380 of 6386 (skipped 6) SUCCESS (1 min 2.733 secs / 16.285 secs)`

### DIFF
--- a/src/model/node.js
+++ b/src/model/node.js
@@ -274,11 +274,14 @@ export default class Node {
 	 * which is a common ancestor of both nodes.
 	 *
 	 * @param {module:engine/model/node~Node} node The second node.
+	 * @param {Object} options Options object.
+	 * @param {Boolean} [options.includeSelf=false] When set to `true` both nodes will be considered "ancestors" too.
+	 * Which means that if e.g. node A is inside B, then their common ancestor will be B.
 	 * @returns {module:engine/model/element~Element|module:engine/model/documentfragment~DocumentFragment|null}
 	 */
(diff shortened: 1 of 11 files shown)


pomek:
Merged with `Other` prefix because notes from `Internal` commits are ignored.

PR #1040, opened 2017-07-24
Virtual selection

### LINKED ISSUE #1015: Virtual selection
We have makers and marker to element converters which wraps text in the proper `<span>`. The problem is that it does not work for widgets, like images. This is why we need a move abstraction mechanism.

The idea is that instead returning specific `span` in the conversion builder:
```js
buildModelConverter().for( editor.editing.modelToView )
	.fromMarker( 'marker' )
	.toElement( data => {
		const elem = new AttributeElement( 'span', { 'class': 'marker-style', data-id: 42 } );
		element.priority = 1000;

		return element;
	} );
```

give more generic information:

```js
buildModelConverter().for( editor.editing.modelToView )
	.fromMarker( 'marker' )
	.toVirtualSelection( data => {
		return {
			'class': 'marker-style', // styling class
			'attributes': { data-id: 42 }, // additional attributes added to the span
			priority: 1000; 
		}
	} );
```

Event dispatcher should spit one big `markerAdd` to smaller pieces, for each text part or element in the range, like it does for attributes.

Then, the conversion builder should add listeners which will wrap text nodes into `span` with this attributes, but widget can be handled in a different way.

In ranges with an element which (mapped to the view element) has `setVirtualSelection` and `removeVirtualSelction`, it should call this method instead of wrapping it into `span`. Virtual selection parameters, defined in the converter, should be passed to these methods.

What the element will do will be different for every specific element (in the future tables will handle them differently), but for a standard widget (like an image), the styling class will be added to the element. However, if multiple virtual selections are applied only the one with the highest priority should add the styling class.

Additionally, if the element has `setVirtualSelection`, the converter added by conversion builder, should consume all element inside that element, so, for instance, the caption of the image will not get the selection if the whole image is selected.

Follow-up: since there will be no `fromMarker.toElement` converter anymore, we can rename `fromMarker.toStamp` to `fromMarker.toElement`, so the marker, instead of being convertable to "element" or "stamp" will be convertable to "virtual selection" or "element".

szymonkups:

Feature: Created virtual selection handling in model to view converters. Closes #1015. 

BREAKING CHANGE: `ModelConverterBuilder#toStamp()` functionality is moved to `ModelConverterBuilder#toElement`. Introduced `ModelConverterBuilder#toVirtualSelection` which replaces current marker to element conversion.
BREAKING CHANGE: Parameter change for `convertSelectionMarker()` function from `model-selection-to-view-converters.js`.
BREAKING CHANGE: Removed `wrapRange()` and `unwrapRange()` functions from `model-to-view-converters.js` as they're no longer used.
BREAKING CHANGE: Removed `breakViewRangePerContainer()` from view writer as it's no longer used.
BREAKING CHANGE: Renamed `StampCreatorData` to `MarkerViewElementCreatorData` and renamed its fields.

### DIFF
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -15,12 +15,11 @@ import {
 	removeUIElement,
 	wrapItem,
 	unwrapItem,
-	eventNameToConsumableType
+	virtualSelectionConverter
 } from './model-to-view-converters';
 
 import { convertSelectionAttribute, convertSelectionMarker } from './model-selection-to-view-converters';
 
-import ModelRange from '../model/range';
 import ViewAttributeElement from '../view/attributeelement';
 import ViewContainerElement from '../view/containerelement';
 import ViewUIElement from '../view/uielement';
@@ -247,93 +246,13 @@ class ModelConverterBuilder {
 		}
 	}
 
-	toVirtualSelection( selectionInfo ) {
+	toVirtualSelection( selectionDescriptor ) {
 		const priority = this._from.priority === null ? 'normal' : this._from.priority;
 
 		for ( const dispatcher of this._dispatchers ) {
-			dispatcher.on( 'selectionMarker:' + this._from.name, ( evt, data, consumable, api ) => {
-				const info = selectionInfo( data );
-				const wrapElement = new ViewAttributeElement( 'span', info.attributes );
-
-				if ( info.class ) {
-					wrapElement.addClass( info.class );
-				}
-
-				const wrap = convertSelectionMarker( wrapElement );
-
-				wrap( evt, data, consumable, api );
-			}, { priority } );
-
-			dispatcher.on( 'addMarker:' + this._from.name, ( evt, data, consumable, api ) => {
-				const modelItem = data.item;
-				const info = selectionInfo( data );
-
-				if ( modelItem.is( 'textProxy' ) ) {
-					const wrapElement = new ViewAttributeElement( 'span', info.attributes );
-
-					if ( info.class ) {
-						wrapElement.addClass( info.class );
-					}
-
-					const wrap = wrapItem( wrapElement );
-
-					wrap( evt, data, consumable, api );
-				}
-
-				if ( modelItem.is( 'element' ) ) {
-					const consumableType = eventNameToConsumableType( evt.name );
-					const viewElement = api.mapper.toViewElement( modelItem );
-
-					if ( !consumable.consume( modelItem, consumableType ) ) {
-						return;
-					}
-
-					if ( viewElement && viewElement.setVirtualSelection ) {
-						// Virtual selection will be handled by parent element - consume all children.
-						for ( const value of ModelRange.createIn( modelItem ) ) {
-							consumable.consume( value.item, consumableType );
-						}
-
-						viewElement.setVirtualSelection( info );
-					}
-				}
-			}, { priority } );
-
-			dispatcher.on( 'removeMarker:' + this._from.name, ( evt, data, consumable, api ) => {
-				const modelItem = data.item;
-				const info = selectionInfo( data );
-
-				if ( modelItem.is( 'textProxy' ) ) {
-					const info = selectionInfo( data );
-					const wrapElement = new ViewAttributeElement( 'span', info.attributes );
-
-					if ( info.class ) {
-						wrapElement.addClass( info.class );
-					}
-
-					const unwrap = unwrapItem( wrapElement );
-
-					unwrap( evt, data, consumable, api );
-				}
-
-				if ( modelItem.is( 'element' ) ) {
-					const consumableType = eventNameToConsumableType( evt.name );
-					const viewElement = api.mapper.toViewElement( modelItem );
-
-					if ( !consumable.consume( modelItem, consumableType ) ) {
-						return;
-					}
-
-					if ( viewElement && viewElement.setVirtualSelection ) {
-						// Virtual selection will be handled by parent element - consume all children.
-						for ( const value of ModelRange.createIn( modelItem ) ) {
-							consumable.consume( value.item, consumableType );
-						}
-
-						viewElement.removeVirtualSelection( info );
-					}
-				}
-			}, { priority } );
+			dispatcher.on( 'addMarker:' + this._from.name, virtualSelectionConverter( selectionDescriptor ), { priority } );
+			dispatcher.on( 'removeMarker:' + this._from.name, virtualSelectionConverter( selectionDescriptor, false ), { priority } );
+			dispatcher.on( 'selectionMarker:' + this._from.name, convertSelectionMarker( selectionDescriptor ), { priority } );
 		}
 	}
 
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -15,7 +15,7 @@ import {
 	removeUIElement,
 	wrapItem,
 	unwrapItem,
-	virtualSelectionConverter
+	markerToVirtualSelection
 } from './model-to-view-converters';
 
 import { convertSelectionAttribute, convertSelectionMarker } from './model-selection-to-view-converters';
@@ -250,8 +250,8 @@ class ModelConverterBuilder {
 		const priority = this._from.priority === null ? 'normal' : this._from.priority;
 
 		for ( const dispatcher of this._dispatchers ) {
-			dispatcher.on( 'addMarker:' + this._from.name, virtualSelectionConverter( selectionDescriptor ), { priority } );
-			dispatcher.on( 'removeMarker:' + this._from.name, virtualSelectionConverter( selectionDescriptor, false ), { priority } );
+			dispatcher.on( 'addMarker:' + this._from.name, markerToVirtualSelection( selectionDescriptor ), { priority } );
+			dispatcher.on( 'removeMarker:' + this._from.name, markerToVirtualSelection( selectionDescriptor ), { priority } );
 			dispatcher.on( 'selectionMarker:' + this._from.name, convertSelectionMarker( selectionDescriptor ), { priority } );
 		}
 	}
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -437,10 +437,10 @@ export default function buildModelConverter() {
  * can handle virtual selection separately by providing `setVirtualSelection` and `removeVirtualSelection` custom
  * properties.
  *
- * @property {Number} priority {@link module:engine/view/attributeelement~AttributeElement#priority} of the `span`
- * wrapping each text node in the virtual selection.
  * @property {String} class CSS class that will be added to `span`
  * {@link module:engine/view/attributeelement~AttributeElement} wrapping each text node in the virtual selection.
+ * @property {Number} [priority] {@link module:engine/view/attributeelement~AttributeElement#priority} of the `span`
+ * wrapping each text node in the virtual selection. If not provided, default 10 priority will be used.
  * @property {Object} [attributes] Attributes that will be added to `span`
  * {@link module:engine/view/attributeelement~AttributeElement} wrapping each text node it the virtual selection.
  */
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -297,7 +297,7 @@ class ModelConverterBuilder {
 	 * Also, descriptor creator function can be provided:
 	 *
 	 *		buildModelConverter.for( dispatcher ).fromMarker( 'search:blue' ).toVirtualSelection( data => {
-	 *			const color = data.name.split( ':' )[ 1 ];
+	 *			const color = data.markerName.split( ':' )[ 1 ];
 	 *
 	 *			return { class: 'search-' + color };
 	 *		} );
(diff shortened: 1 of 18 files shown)


pjasiun:
> Using this API it is possible to create four kinds of converters:
"five"

pjasiun:
We should stop using term "markers stamps" everywhere. It's still used in:
```
ckeditor5-engine/src/controller/datacontroller.js
ckeditor5-engine/src/conversion/viewconversiondispatcher.js
ckeditor5-engine/tests/conversion/viewconversiondispatcher.js
```

Fortunetly it's only comments and 1 variable name.

Reinmar:
To review this for real it would be good to have some prototype of integration with widgets.

AFAIR we planned to have a marker representing the selection and convert that marker in such a case. However, the thing which I don't see in this PR is the "toVirtualSelection()" mechanism using the `element.getCustomProperty( 'renderVirtualSelection' )` logic, so perhaps this PR is incomplete for this case.

src/conversion/model-to-view-converters.js
```diff
@@ -252,9 +254,11 @@ export function removeAttribute( attributeCreator ) {
  *
  * @param {module:engine/view/element~Element|Function} elementCreator View element, or function returning a view element, which will
  * be used for wrapping.
+ * @param {Function} [nameToConsumable] Optional function used to map conversion event name to consumable name. By default it
+ * uses {engine/conversion/model-to-view-converters~eventNameToConsumableType} function.
  * @returns {Function} Set/change attribute converter.
  */
-export function wrapItem( elementCreator ) {
+export function wrapItem( elementCreator, nameToConsumable = eventNameToConsumableType ) {
```
pjasiun: Wow! You pass a function as a parameter to the function which is used as a listener... I know that functional programming is trendy, but it's too much for me, this part of the code is complicated enough without such parameters.

Is it because we `eventNameToConsumableType` does not handle markers events well? Can't we make `eventNameToConsumableType` smarter, then? Or maybe we should just copy (or extract) part of the code instead of hacking it...(*)

src/conversion/model-to-view-converters.js
```diff
@@ -478,6 +385,63 @@ export function remove() {
...
+			return;
+		}
+
+		if ( modelItem.is( 'textProxy' ) ) {
+			const viewElement = virtualSelectionDescriptorToAttribute( descriptor );
+			const converter = addMarker ?
+				wrapItem( viewElement, eventName => eventName ) :
+				unwrapItem( viewElement, eventName => eventName );
+
+			converter( evt, data, consumable, conversionApi );
```
pjasiun: (*)... because these lines of code are unreadable after 17. They have to be simplified.

Reinmar: I got it on the second try :D So it's not that bad. The thing which confused me was the "...ToAttribute()" method because it doesn't return an attribute (key or value or something) but an element. The rest is fine (if you understand what wrapItem's second param does).

scofalik: Honestly I reallllllly dislike that we use "attribute" meaning `AttributeElement` in view. I don't even like `AttributeElement` name. It's called so only because those elements are mostly created through model attribute conversion.

pjasiun: I agree that we should not use "attribute" as a shortcut from "attribute element".

But my problem with this code is that when we use one converter functions factory in another it's very hard to understand how the code works. Especially when, like in this case, each of them wants to have this conversion working differently. Especially when, in fact, we don't need most of the code from the inside of `wrapItem` and `unwrapItem` in this case. We should use here the code which we need. In case, of having some complicated logic duplicated (like this condition https://github.com/ckeditor/ckeditor5-engine/blob/9f7e0a2c813a3a8f7c673c471d8caa3621200226/src/conversion/model-to-view-converters.js#L276-L279) we can export it to the separate function (what will improve readability anyway).

src/conversion/model-to-view-converters.js
Reinmar: I don't know what the code does and why but from used techniques it seems to be completely fine. You really looked too late at this, @pjasiun ;) It's just a callback, like many others all around our code. I don't see any particularly wrong with this one.

src/conversion/model-to-view-converters.js
```diff
@@ -552,3 +516,24 @@ export function eventNameToConsumableType( evtName ) {
...
 }
+
+/**
+ * Creates `span` {@link module:engine/view/attributeelement~AttributeElement view attribute element} from information
+ * provided by {@link module:engine/conversion/buildmodelconverter~VirtualSelectionDescriptor} object.
+ *
+ * @param {module:engine/conversion/buildmodelconverter~VirtualSelectionDescriptor }descriptor
+ * @return {module:engine/view/attributeelement~AttributeElement}
+ */
+export function virtualSelectionDescriptorToAttribute( descriptor ) {
```
pjasiun: We should not use `...ToAttribute` as a shortcut of `...ToAttributeElement`. It's very misleading.

scofalik: It's better to write `toElement` rather than `toAttribute` if you mean `AttributeElement`

pjasiun: It's better to write `toAttributeElement`. It's long but it's the only way to make it clear. `toElement` it for sure better, but still far from perfect.

src/conversion/model-to-view-converters.js
```diff
@@ -478,6 +385,63 @@ export function remove() {
...
+		if ( modelItem.is( 'textProxy' ) ) {
+			const viewElement = virtualSelectionDescriptorToAttribute( descriptor );
+			const converter = addMarker ?
+				wrapItem( viewElement, eventName => eventName ) :
+				unwrapItem( viewElement, eventName => eventName );
+
+			converter( evt, data, consumable, conversionApi );
+		}
+
+		if ( modelItem.is( 'element' ) ) {
```
pjasiun: The comment what section below is doing would be useful. In fact, we could consider splitting this function into two smaller and add both in the conversion builder. These are 2 separate logic: one to handle text and one to handle elements with the proper interface.

src/conversion/modelconversiondispatcher.js
```diff
@@ -399,10 +404,33 @@ export default class ModelConversionDispatcher {
...
 
-		const consumable = this._createMarkerConsumable( type, range );
-		const data = { name, range };
+		// In markers case, event name == consumable name.
+		const eventName = type + ':' + name;
 
-		this.fire( type + ':' + name, data, consumable, this.conversionApi );
+		// Fire event for whole marker. This can be used to convert whole marker to element.
+		const markerConsumable = this._createMarkerConsumable( eventName, range );
+		this.fire( eventName, { name, range }, markerConsumable, this.conversionApi );
```
pjasiun: Do we fire event on the same range twice from the dispatcher (once for the whole range and once for parts)? It's very strange.

src/conversion/model-to-view-converters.js
pjasiun: My problem is not the fact that we add a callback. My problem is that translating names to consumable was very very simple (see https://github.com/ckeditor/ckeditor5-engine/blob/2ccd95ff89e7d8522891a925a034056959f145b6/src/conversion/model-to-view-converters.js#L544-L554). You did not need to think much when you see `nameToConsumable` in the code. Now if it can be a callback we bring another level of abstraction to already very complex part of the code. We should think twice if we really need to do it.

src/conversion/model-to-view-converters.js
Reinmar: Agree with @pjasiun. That was the most confusing part for me when reading this code.

szymonkups:
Removing r- label. Please close together with https://github.com/ckeditor/ckeditor5-widget/pull/12.

src/conversion/model-to-view-converters.js
```diff
@@ -488,24 +402,86 @@ export function remove() {
...
+			selectionDescriptor;
 
-	for ( const value of walker ) {
-		if ( value.type == 'text' ) {
-			charactersFound++;
+		const modelItem = data.item;
 
-			if ( charactersFound == howMany ) {
-				return walker.position;
+		if ( !descriptor || data.markerRange.isCollapsed || !modelItem.is( 'textProxy' ) ) {
```
pjasiun: Will I be always a `textProxy`? Can't we have `text` as a model item?

szymonkups: `convertMarker` method from `ModelConversionDispatcher` iterates over nodes inside marker's range using `TreeWalker` which creates `TextProxies` for each text node.

pjasiun: Ach, that's true.

pjasiun: I did not remember that all text nodes are wrapped into text proxies.

src/conversion/buildmodelconverter.js
```diff
@@ -58,17 +58,22 @@ import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
...
- * Markers, basically, are {@link module:engine/model/liverange~LiveRange} instances, that are named. In this conversion, model range is
- * converted to view range, then that view range is wrapped (or unwrapped, if range is removed) in a view attribute element.
- * To learn more about markers, see {@link module:engine/model/markercollection~MarkerCollection}.
+ * 4. Model marker to virtual selection converter. This is a converter that converts model markers to virtual
+ * selection described by {@link module:engine/conversion/buildmodelconverter~VirtualSelectionDescriptor} object passed to
+ * {@link module:engine/conversion/buildmodelconverter~ModelConverterBuilder#toVirtualSelection} method.
  *
- *		const viewSpanSearchResult = new ViewAttributeElement( 'span', { class: 'search-result' } );
- *		buildModelConverter().for( dispatcher ).fromMarker( 'searchResult' ).toElement( viewSpanSearchResult );
+ *		buildModelConverter().for( dispatcher ).fromMarker( 'search' ).toVirtualSelection( descriptor );
```
pjasiun: Could you put here an example of a descriptor?

src/conversion/buildmodelconverter.js
```diff
@@ -198,7 +203,13 @@ class ModelConverterBuilder {
 	 * `string`, view element instance which will be cloned and used, or creator function which returns view element that
 	 * will be used. Keep in mind that when you view element instance or creator function, it has to be/return a
 	 * proper type of view element: {@link module:engine/view/containerelement~ContainerElement ViewContainerElement} if you convert
-	 * from element or {@link module:engine/view/attributeelement~AttributeElement ViewAttributeElement} if you convert from attribute.
+	 * from element, {@link module:engine/view/attributeelement~AttributeElement ViewAttributeElement} if you convert
+	 * from attribute and {@link module:engine/view/uielement~UIElement ViewUIElement} if you convert from marker.
```
pjasiun: Shouldn't `ViewAttributeElement` be used for virtual selection?

src/conversion/buildmodelconverter.js
szymonkups: Sure, I will add it.

src/conversion/buildmodelconverter.js
szymonkups: It's `toElement` method, used now instead of old `toStamp` method where `UIElements` are created.

src/conversion/model-to-view-converters.js
```diff
@@ -488,24 +404,86 @@ export function remove() {
...
+		} else {
+			viewWriter.unwrap( viewRange, viewElement );
+		}
+	};
+}
+
+/**
+ * Function factory, creates converter that converts all elements inside marker's range. Converter checks if element has
+ * functions stored under `setVirtualSelection` and `removeVirtualSelection` custom properties and calls them passing
+ * {@link module:engine/conversion/buildmodelconverter~VirtualSelectionDescriptor. In such case converter will consume
```
pjasiun: }

src/conversion/model-to-view-converters.js
```diff
@@ -488,24 +404,86 @@ export function remove() {
...
+ * all element's children, assuming that they were handled by element itself. If selection descriptor will not provide
+ * priority, priority 10 will be used as default, to be compliant with
+ * {@link module:engine/conversion/model-to-view-converters~convertTextsInsideMarker} method which uses default priority of
+ * {@link module:engine/view/attributeelement~AttributeElement}.
```
pjasiun: I miss here the `else` path ;) I mean what if element do not have `setVirtualSelection` or `removeVirtualSelection` properties.

pjasiun: And, since there is no check if the element has these methods you should write "And crash the element does not have this methods" ;) Or fix it ;)

src/conversion/modelconversiondispatcher.js
```diff
@@ -647,14 +668,23 @@ export default class ModelConversionDispatcher {
 	 *
 	 * @event addMarker
 	 * @param {Object} data Additional information about the change.
-	 * @param {String} data.name Marker name.
-	 * @param {module:engine/model/range~Range} data.range Marker range.
-	 * @param {module:engine/conversion/modelconsumable~ModelConsumable} consumable Values to consume.
+	 * @param {module:engine/model/item~Item} [data.item] Item contained in marker's range. Not present if collapsed range
+	 * is being converted.
+	 * @param {module:engine/model/range~Range} [data.range] Range spanning over item. Not present if collapsed range
+	 * is being converted.
```
pjasiun: "Not present if collapsed range is being converted." Is it correct?

src/conversion/model-to-view-converters.js
szymonkups: If element does not have this methods, it's child elements will be converted as the rest of the document :stuck_out_tongue:

src/conversion/buildmodelconverter.js
pjasiun: Ach, sorry. Then you're right.

src/conversion/model-to-view-converters.js
pjasiun: This is the comment I miss there :)

Reinmar:
@pjasiun, please copy source version of the commit message (click the "edit" button to see it) because you lost formatting.

PR #1043, opened 2017-07-25
T/1009: Element#getNodeByPath() should expect offsets not indexes

### LINKED ISSUE #1009: Element#getNodeByPath() should expect offsets not indexes
Path is an array of offsets (it's defined like this in `Position`), hence, whenever we accept path we should translate offsets to indexes (if we want to retrieve children).

pomek:

Fix: `model.Element#getNodeByPath()` and `model.DocumentFragment#getNodeByPath()` work with offsets instead of indexes. Closes #1009.

### DIFF
--- a/tests/model/documentfragment.js
+++ b/tests/model/documentfragment.js
@@ -309,16 +309,50 @@ describe( 'DocumentFragment', () => {
 		} );
 
 		it( 'should return a descendant of this node', () => {
+			const foo = new Text( 'foo' );
 			const image = new Element( 'image' );
 			const element = new Element( 'elem', [], [
 				new Element( 'elem', [], [
-					new Text( 'foo' ),
+					foo,
 					image
 				] )
 			] );
 			const frag = new DocumentFragment( element );
 
-			expect( frag.getNodeByPath( [ 0, 0, 1 ] ) ).to.equal( image );
+			expect( frag.getNodeByPath( [ 0, 0, 0 ] ) ).to.equal( foo );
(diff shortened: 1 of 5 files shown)


tests/model/documentfragment.js
```diff
@@ -309,16 +309,50 @@ describe( 'DocumentFragment', () => {
...
+			const paragraph = new Element( 'paragraph', [], [
+				foo,
+				bold,
+				bom
+			] );
+			const frag = new DocumentFragment( paragraph );
+
+			// <paragraph>foo<bold>bar</bold>bom</paragraph>
+
+			expect( frag.getNodeByPath( [ 0, 0 ] ) ).to.equal( foo );
```
Reinmar: It's not good when all tests start with `0` offset. E.g. if we'd implement this method recursively this would pass while there still might be a bug in DF itself.

Reinmar:
Minor R- for incomplete tests.

Besides that, have you checked how and why existing `getNodeByPath()` usage worked?

Reinmar:
The tests are fine now. But this question remains:

> Besides that, have you checked how and why existing getNodeByPath() usage worked?

pomek:
I found only one place in the whole project where the method is used: https://github.com/ckeditor/ckeditor5-engine/blob/e627f7e6f690b8fe45a58743201c28c56b4c97a8/src/controller/getselectedcontent.js#L40.

pomek:
Finally, I did it! :D

Reinmar:
Great. That's a really good test.

PR #1044, opened 2017-07-25
T/1042: Extract getLimitElement() from DataController#DeleteContent()

### LINKED ISSUE #1042: Extract getLimitElement() from DataController#DeleteContent()
`getLimitElement()` repeats twice in different places – [one in `DataController#DeleteContent()`](https://github.com/ckeditor/ckeditor5-engine/blob/master/src/controller/deletecontent.js#L171-L184) and the [second one in PR in ckeditor5-typing](https://github.com/ckeditor/ckeditor5-typing/pull/111/files#diff-723d5264f48546911fc24c4e03e1aeb7R169).

We need to figure out where it should land in order to avoid repeating the code.

Related comment – https://github.com/ckeditor/ckeditor5-typing/pull/111#issuecomment-317012503.

Proposed solution – https://github.com/ckeditor/ckeditor5-typing/pull/111#issuecomment-317653979.

pomek:

Feature: Moved a private function `getLimitElement()` from `DataController#DeleteContent()` to `Schema`. Closes #1042.

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -383,6 +383,26 @@ export default class Schema {
 	}
 
 	/**
+	 * Returns the lowest limit element defined in {@link module:engine/model/schema~Schema#limits} for passed selection.
+	 *
+	 * @param {module:engine/model/selection~Selection} selection Selection which will be checked.
+	 * @returns {module:engine/model/element~Element}
+	 */
+	getLimitElement( selection ) {
+		let element = selection.getFirstRange().getCommonAncestor();
+
+		while ( !this.limits.has( element.name ) ) {
+			if ( element.parent ) {
+				element = element.parent;
+			} else {
+				break;
+			}
+		}
+
+		return element;
+	}
+
+	/**
 	 * Returns {@link module:engine/model/schema~SchemaItem schema item} that was registered in the schema under given name.
 	 * If item has not been found, throws error.
 	 *
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -389,7 +389,15 @@ export default class Schema {
 	 * @returns {module:engine/model/element~Element}
 	 */
 	getLimitElement( selection ) {
-		let element = selection.getFirstRange().getCommonAncestor();
+		// Find the common ancestor for all selection's ranges.
+		let element = Array.from( selection.getRanges() )
+			.reduce( ( node, range ) => {
+				if ( !node ) {
+					return range.getCommonAncestor();
+				}
+
+				return node.getCommonAncestor( range.getCommonAncestor() );
+			}, null );
 
 		while ( !this.limits.has( element.name ) ) {
 			if ( element.parent ) {
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -385,7 +385,7 @@ export default class Schema {
 	/**
 	 * Returns the lowest limit element defined in {@link module:engine/model/schema~Schema#limits} for passed selection.
 	 *
-	 * @param {module:engine/model/selection~Selection} selection Selection which will be checked.
+	 * @param {module:engine/model/selection~Selection} selection Selection which returns the common ancestor.
 	 * @returns {module:engine/model/element~Element}
 	 */
 	getLimitElement( selection ) {
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -383,7 +383,8 @@ export default class Schema {
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


tests/model/schema/schema.js
```diff
@@ -641,4 +641,40 @@ describe( 'Schema', () => {
 			expect( result[ 1 ].end.path ).to.members( [ 1 ] );
 		} );
 	} );
+
+	describe( 'getLimitElement()', () => {
```
Reinmar: Missing tests for:

* non-collapsed selections
* multi-range selections

pomek:
@Reinmar, done.

tests/model/schema/schema.js
```diff
@@ -641,4 +641,87 @@ describe( 'Schema', () => {
...
+					'</section>' +
+					'<section>b[]ar</section>' +
+				'</div>'
+			);
+
+			const div = root.getNodeByPath( [ 0 ] );
+			expect( schema.getLimitElement( doc.selection ) ).to.equal( div );
+		} );
+
+		it( 'works fine with multi-range selections even if limit elements are not defined', () => {
```
Reinmar: If you cleared limits then this test checks nothing. The `reduce()` call may be completely broken and it will still pass. You must ensure that both ranges have some other limits and only when combined together they give some 3rd option **which still should not be the root** (so you know that it really works).

Reinmar: Oops. I can see the test above. So, extend the test above to match what I described. This test below is good then.

tests/model/schema/schema.js
```diff
@@ -641,4 +641,87 @@ describe( 'Schema', () => {
...
+			schema.limits.add( 'section' );
+
+			setData( doc, '<div><section><article><paragraph>foo[]bar</paragraph></article></section></div>' );
+
+			const article = root.getNodeByPath( [ 0, 0, 0 ] );
+
+			expect( schema.getLimitElement( doc.selection ) ).to.equal( article );
+		} );
+
+		it( 'returns the limit element which is the closest element to common ancestor for non-collapsed selection', () => {
```
Reinmar: This test checks pretty much nothing. The limit of both selection ends is the same, so the most important thing (so, knowing that both ends are taken into consideration) is not done here.

src/model/schema.js
```diff
@@ -383,6 +383,34 @@ export default class Schema {
 	}
 
 	/**
+	 * Returns the lowest limit element defined in {@link module:engine/model/schema~Schema#limits} for passed selection.
```
Reinmar: Returns the lowest {@link module:engine/model/schema~Schema#limits limit element} containing the entire selection or the root otherwise.

tests/model/schema/schema.js
Reinmar: Ha... there's also a case with selection spanning over multiple roots, but let's assume that this is invalid for now (cause it'd break lots of other things too).

Reinmar:
Few more issues in the tests and we're ready to  close it.

tests/model/schema/schema.js
pomek: I read the comment n times and every time I just don't get it. Should this comment apply to test "_works fine with multi-range selections_"?

Reinmar: No, it applies to a test with a single non-collapsed range selection. Both ends of this range should be placed in such elements that the test will check that both of them were taken into consideration when calculating the common ancestor:

```html
<root><limit><limit>[a</limit><limit>b]</limit></limit></root>
```

Reinmar: Tests should check as many things as possible. As many potential mistakes. And your current test would pass even if someone made a serious mistake and used only `selectionRange.start.parent` to calculate common ancestor of all selection ranges. From what I've seen, all test might pass. And we write them to prevent such situations.

pomek: Ok, thanks. I understand now. :D

pomek:
@Reinmar, I guess it's ready to review once again.

tests/model/schema/schema.js
```diff
@@ -641,4 +641,100 @@ describe( 'Schema', () => {
...
+			schema.limits.add( 'article' );
+			schema.limits.add( 'section' );
+
+			setData( doc, '<div><section><article>[foo></article><article>bar]></article></section></div>' );
```
Reinmar: `foo>` and `bar> are typos.

Reinmar:
I modified the suggested commit message because it made no sense for the reader to mention where this method came from. The only interesting fact is that it was added.

PR #1047, opened 2017-07-25
Fixed DocumentSelection#_fixGraveyardSelection

### LINKED ISSUE #1046: DocumentSelection: fixed range ends up in incorrect place
Whenever a range from `DocumentSelection` becomes moved to graveyard due to removing nodes around it, `DocumentSelection#_fixGraveyardSelection` is fired and the situation is fixed.

The algorithm works incorrectly when the whole content is removed though. If that happens, `root [ 0 ] - [ 0 ]` range is added to the selection. This wasn't well thought. Such range is probably incorrect. It may have even more consequences, like adding `inlineFiller` in wrong place, etc.

Instead of adding incorrect range, no range should be added. This would mean that if no other range was in the selection (which is true if all contents were removed), the "default range" would be returned.

The default range is better than `root [ 0 ] - [ 0 ]` `LiveRange` because it will become correct as soon as any content is added to the document.

scofalik:

Fixed: `DocumentSelection#_fixGraveyardSelection` now does not add incorrect range when all content is removed from document, instead it is empty. Closes #1046.

### DIFF
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -649,12 +649,7 @@ export default class DocumentSelection extends Selection {
 		const positionCandidate = Position.createFromPosition( removedRange.start );
 
 		// Find a range that is a correct selection range and is closest to the start of removed range.
-		let selectionRange = this._document.getNearestSelectionRange( positionCandidate );
-
-		// If nearest valid selection range cannot be found - use one created at the beginning of the root.
-		if ( !selectionRange ) {
-			selectionRange = new Range( new Position( positionCandidate.root, [ 0 ] ) );
-		}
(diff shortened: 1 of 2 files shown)


Reinmar:
Did you check all tests?

scofalik:
I've run `npm t`, though I might have out-dated other repos. Is something wrong? It looks like the build that failed had passed all the tests.

Reinmar:
> Is something wrong?

IDK, I just asked because this is the kind of change which might've broken something.

scofalik:
I've run `npm t` and clicked a bit (but not much) in manual test.

PR #1048, opened 2017-07-26
T/1045: Added a `preventReplacingWithParagraph` flag to `DataController#deleteContent()`

### LINKED ISSUE #1045: deleteContent replaces whole content with paragraph too often
Recently, we added a feature to `DataController#deleteContent`, which in some scenarios, instead of following default algorithm, just replaces the whole content with `<paragraph>` element.

Unfortunately, this feature kicks-in too often, for example in this scenario:
```html
<heading1></heading1>
<paragraph>[]</paragraph>
```
When user presses <kbd>Backspace</kbd>, both elements are removed and new `<paragraph>` is inserted.

This happens, because the selection is extended to:
```html
<heading1>[</heading1>
<paragraph>]</paragraph>
```
And then there is no difference between first scenario and "<kbd>Ctrl + A</kbd> <kbd>Delete</kbd>" scenario.

After talking with @Reinmar the conclusion is that probably it will be the best to solve this by adding another parameter in `deleteContent` function. As much as I don't like this solution, it is possibly the easiest to implement and maintain, so I propose `preventReplacingWithParagraph` flag.

When this is done, the flag should be set to `true` in `DeleteCommand` if original selection was collapsed.

pomek:

Fix: Added a flag "preventReplacingWithParagraph" to `DataController#deleteContent()` method. Closes #1045.

---

The rest of the ticket:

> After talking with @Reinmar the conclusion is that probably it will be the best to solve this by adding another parameter in `deleteContent` function. As much as I don't like this solution, it is possibly the easiest to implement and maintain, so I propose `preventReplacingWithParagraph` flag.
>
> When this is done, the flag should be set to `true` in `DeleteCommand` if original selection was collapsed.

will be solved in typing repository.

### DIFF
--- a/src/controller/deletecontent.js
+++ b/src/controller/deletecontent.js
@@ -19,6 +19,7 @@ import Element from '../model/element';
  * @param {module:engine/model/batch~Batch} batch Batch to which the deltas will be added.
  * @param {Object} [options]
  * @param {Boolean} [options.leaveUnmerged=false] Whether to merge elements after removing the content of the selection.
+ * @param {Boolean} [options.preventReplacingWithParagraph=false] Whether the entire content shouldn't be replaced with the paragraph.
  *
  * For example `<h>x[x</h><p>y]y</p>` will become:
  * * `<h>x^y</h>` with the option disabled (`leaveUnmerged == false`)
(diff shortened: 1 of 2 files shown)


Reinmar:
I need to still think on this – what kind of param do we need. Or maybe how to solve it completely differently.

Reinmar:
OK, I tried, but I failed to find a more elegant way of solving this issue. I can't make this option more general while still making sure that it's clear in context of `deleteContent()` itself.

I'll only try to figure out a more "entire content selected" related option name.

Reinmar:
OK, what I don't like in the current name is the "prevent" keyword, because it suggests that the option will do something additional. What it does, it skips some action.

Also, what it doesn't indicate is that it's about entire content selection (it's actually a simplification because it's "entire content of the closest limit element").

So, 🎉 `skipResettingEntireContene`. As bad as the previous proposal, but better.

Issue #796, opened 2017-01-31
Selection observer should call render


pjasiun:
The idea is that observers not only listen on DOM changes and fire view events but also take care about re-rendering DOM if it needs to be rendered. Especially, if the model did not fire `change` nor `changesDone` events it's observers role to bring DOM to the same state as view is, by calling `render()`.

Note that it's renderer who should compare DOM and view and do nothing if there is no good reason to render anything. Observers should be free to call render whenever the feel that something may need the update.

`render` was removed from the `SelectionObserver` by this commit: https://github.com/ckeditor/ckeditor5-engine/commit/f4c0d25398bf9a85376e2f24f32c26cff188e16e

Because of it 'noselection' manual test does not work: http://localhost:8125/ckeditor5-engine/tests/view/manual/noselection.html

It's also connected with other issues: https://github.com/ckeditor/ckeditor5-engine/issues/795

Reinmar:
I needed to change a code of http://localhost:8125/ckeditor5-engine/tests/view/manual/inline-filler.html a bit to make the test pass now.

Also, the http://localhost:8125/ckeditor5-engine/tests/view/manual/immutable.html test doesn't work now due to this.

szymonkups:
I am working on `UIElement` rendering now. All DOM selection that lands inside `UIElement` should be converted by `DOMConverter` to selection before/after `UIElement`. This means that each selection change inside `UIElement` is converted to the same selection. When `SelectionObserver` is not calling `render()`, we end up with "dirty" selection in DOM having correct selection in view and model.

ma2ciek:
In https://github.com/ckeditor/ckeditor5/issues/676 and https://github.com/ckeditor/ckeditor5-engine/issues/1157 we face a problem, that the `focus` and `selectionchange` events fire with some delay between them, which results in messing up the editor's selection (see https://github.com/ckeditor/ckeditor5/issues/676#issuecomment-349307800). We need to make the selection depend more on the `selectionchange` event, less on the `focus` event or make it somehow smarter.

> Observers should be free to call render whenever the feel that something may need the update.

For now, calling `render()` between `selectionchange` and `focus` events results in errors.

ma2ciek:
#1153 and #1155 are also connected with that topic.

ma2ciek:
[...]

PR #1050, opened 2017-07-27
T/1049 Selection will not be placed in UI elements

### LINKED ISSUE #1049: Prevent placing selection inside UI elements
Selection should not be able to be placed inside `view.UIElement`s. This concerns both DOM selection and `view.Selection`. Of course, ui elements do not exist in model.

At the moment, in certain cases, it is possible to place selection inside UI element and even write in it. Whenever a DOM selection ends up in `UIElement`, it should be fixed.

scofalik:

Fixed: Selection will not be placed in UI elements. Closes #1049.

### DIFF
--- a/src/conversion/view-selection-to-model-converters.js
+++ b/src/conversion/view-selection-to-model-converters.js
@@ -39,8 +39,10 @@ export function convertSelectionChange( modelDocument, mapper ) {
 
 		modelSelection.setRanges( ranges, viewSelection.isBackward );
 
-		modelDocument.enqueueChanges( () => {
-			modelDocument.selection.setTo( modelSelection );
-		} );
+		if ( !modelSelection.isEqual( modelDocument.selection ) ) {
+			modelDocument.enqueueChanges( () => {
+				modelDocument.selection.setTo( modelSelection );
+			} );
+		}
 	};
 }
--- a/src/conversion/view-selection-to-model-converters.js
+++ b/src/conversion/view-selection-to-model-converters.js
@@ -39,10 +39,8 @@ export function convertSelectionChange( modelDocument, mapper ) {
 
 		modelSelection.setRanges( ranges, viewSelection.isBackward );
 
-		if ( !modelSelection.isEqual( modelDocument.selection ) ) {
-			modelDocument.enqueueChanges( () => {
-				modelDocument.selection.setTo( modelSelection );
-			} );
-		}
+		modelDocument.enqueueChanges( () => {
+			modelDocument.selection.setTo( modelSelection );
+		} );
 	};
 }
--- a/src/view/document.js
+++ b/src/view/document.js
@@ -20,6 +20,7 @@ import KeyObserver from './observer/keyobserver';
 import FakeSelectionObserver from './observer/fakeselectionobserver';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
 import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
+import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
 
 /**
  * Document class creates an abstract layer over the content editable area.
@@ -126,6 +127,8 @@ export default class Document {
 		this.addObserver( FakeSelectionObserver );
 
 		injectQuirksHandling( this );
+
+		this.on( 'keydown', ( evt, data ) => _jumpOverUiElement( evt, data, this.domConverter ) );
 
 		this.decorate( 'render' );
 	}
@@ -347,3 +350,29 @@ mix( Document, ObservableMixin );
  *
  * @event render
  */
+
+// Selection cannot be placed in a `UIElement`. Whenever it is placed there, it is moved before it. This
+// causes a situation when it is impossible to jump over `UIElement` using right arrow key, because the selection
+// ends up in ui element (in DOM) and is moved back to the left. This handler fixes this situation.
+function _jumpOverUiElement( evt, data, domConverter ) {
+	if ( data.keyCode == keyCodes.arrowright ) {
+		const domSelection = data.domTarget.ownerDocument.defaultView.getSelection();
+
+		if ( domSelection.rangeCount == 1 && domSelection.getRangeAt( 0 ).collapsed ) {
+			const domParent = domSelection.getRangeAt( 0 ).startContainer;
+			const domOffset = domSelection.getRangeAt( 0 ).startOffset;
+
+			const viewPosition = domConverter.domPositionToView( domParent, domOffset );
+			// Skip all following ui elements.
+			const nextViewPosition = viewPosition.getLastMatchingPosition( value => value.item.is( 'uiElement' ) );
+
+			// If anything has been skipped, fix position.
+			// This `if` could be possibly omitted but maybe it is better not to mess with DOM selection if not needed.
+			if ( !viewPosition.isEqual( nextViewPosition ) ) {
+				const newDomPosition = domConverter.viewPositionToDom( nextViewPosition );
+
+				domSelection.collapse( newDomPosition.parent, newDomPosition.offset );
+			}
+		}
+	}
+}
--- a/src/view/document.js
+++ b/src/view/document.js
@@ -12,6 +12,7 @@ import Renderer from './renderer';
 import DomConverter from './domconverter';
 import RootEditableElement from './rooteditableelement';
 import { injectQuirksHandling } from './filler';
+import { injectUiElementHandling } from './uielement';
 import log from '@ckeditor/ckeditor5-utils/src/log';
 import MutationObserver from './observer/mutationobserver';
 import SelectionObserver from './observer/selectionobserver';
@@ -20,7 +21,6 @@ import KeyObserver from './observer/keyobserver';
 import FakeSelectionObserver from './observer/fakeselectionobserver';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
 import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
-import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
 
 /**
  * Document class creates an abstract layer over the content editable area.
@@ -127,8 +127,7 @@ export default class Document {
 		this.addObserver( FakeSelectionObserver );
 
 		injectQuirksHandling( this );
-
-		this.on( 'keydown', ( evt, data ) => _jumpOverUiElement( evt, data, this.domConverter ) );
+		injectUiElementHandling( this );
 
 		this.decorate( 'render' );
 	}
@@ -350,29 +349,3 @@ mix( Document, ObservableMixin );
  *
  * @event render
  */
-
-// Selection cannot be placed in a `UIElement`. Whenever it is placed there, it is moved before it. This
-// causes a situation when it is impossible to jump over `UIElement` using right arrow key, because the selection
-// ends up in ui element (in DOM) and is moved back to the left. This handler fixes this situation.
-function _jumpOverUiElement( evt, data, domConverter ) {
-	if ( data.keyCode == keyCodes.arrowright ) {
-		const domSelection = data.domTarget.ownerDocument.defaultView.getSelection();
-
-		if ( domSelection.rangeCount == 1 && domSelection.getRangeAt( 0 ).collapsed ) {
-			const domParent = domSelection.getRangeAt( 0 ).startContainer;
-			const domOffset = domSelection.getRangeAt( 0 ).startOffset;
-
-			const viewPosition = domConverter.domPositionToView( domParent, domOffset );
-			// Skip all following ui elements.
-			const nextViewPosition = viewPosition.getLastMatchingPosition( value => value.item.is( 'uiElement' ) );
-
-			// If anything has been skipped, fix position.
-			// This `if` could be possibly omitted but maybe it is better not to mess with DOM selection if not needed.
-			if ( !viewPosition.isEqual( nextViewPosition ) ) {
-				const newDomPosition = domConverter.viewPositionToDom( nextViewPosition );
-
-				domSelection.collapse( newDomPosition.parent, newDomPosition.offset );
-			}
-		}
-	}
-}
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -242,7 +242,7 @@ export default class DomConverter {
 	 * @param {Object} options See {@link module:engine/view/domconverter~DomConverter#viewToDom} options parameter.
 	 * @returns {Iterable.<Node>} DOM nodes.
 	 */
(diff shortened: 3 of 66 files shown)


src/conversion/view-selection-to-model-converters.js
```diff
@@ -39,10 +39,8 @@ export function convertSelectionChange( modelDocument, mapper ) {
 
 		modelSelection.setRanges( ranges, viewSelection.isBackward );
 
-		if ( !modelSelection.isEqual( modelDocument.selection ) ) {
```
Reinmar: Could you explain this change?

src/view/document.js
```diff
@@ -127,6 +128,8 @@ export default class Document {
 
 		injectQuirksHandling( this );
 
+		this.on( 'keydown', ( evt, data ) => _jumpOverUiElement( evt, data, this.domConverter ) );
```
Reinmar: Let's move this helper to `uielement.js` and expose it as `injectQuirksHandling()` just like for the filler.

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+	 * Checks if given {Selection DOM Selection} boundaries are in correct places.
+	 *
+	 * Incorrect places for selection are:
+	 * * before or in the middle of inline filler sequence,
+	 * * inside DOM element that represents {@link module:engine/view/uielement~UIElement view ui element}.
+	 *
+	 * @param {Selection} domSelection DOM Selection object to be checked.
+	 * @returns {Boolean} `true` if given selection is at correct place, `false` otherwise.
+	 */
+	isCorrectDomSelection( domSelection ) {
```
Reinmar: `isDomSelectionCorrect`?

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
 	}
 
 	/**
+	 * Checks if given {Selection DOM Selection} boundaries are in correct places.
```
Reinmar: missing `@link`?

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+	 *
+	 * Incorrect places for selection are:
+	 * * before or in the middle of inline filler sequence,
+	 * * inside DOM element that represents {@link module:engine/view/uielement~UIElement view ui element}.
```
Reinmar: that -> which I think

src/view/domconverter.js
Reinmar: "selection*'s* boundaries"

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
 	}
 
 	/**
+	 * Checks if given {Selection DOM Selection} boundaries are in correct places.
+	 *
+	 * Incorrect places for selection are:
```
Reinmar: "The following places are considered as incorrect for selection boundaries"?

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+	 * * before or in the middle of inline filler sequence,
```
Reinmar: the inline filler sequence?

src/view/domconverter.js
Reinmar: the DOM element
a view UI element

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+	 * * inside DOM element that represents {@link module:engine/view/uielement~UIElement view ui element}.
+	 *
+	 * @param {Selection} domSelection DOM Selection object to be checked.
+	 * @returns {Boolean} `true` if given selection is at correct place, `false` otherwise.
```
Reinmar: the given selection
in the correct place

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+	 */
+	isCorrectDomSelection( domSelection ) {
+		return this._isCorrectDomSelectionPosition( domSelection.anchorNode, domSelection.anchorOffset ) &&
+			this._isCorrectDomSelectionPosition( domSelection.focusNode, domSelection.focusOffset );
+	}
+
+	/**
+	 * Checks if given DOM position is a correct place for selection boundary. See {@link ~isCorrectDomSelection}.
```
Reinmar: the given DOM position

`#isCorrectDomSelection`

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+	 *
+	 * @private
+	 * @param {Node} domParent Position parent.
```
Reinmar: Is the type set to `Node` due to doc fragments?

Reinmar: I don't think that you can have a selection in a DF and this method is used for selection boundaries so this should be `Element`.

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+	 * @param {Number} offset Position offset.
+	 * @returns {Boolean} `true` if given position is correct place for selection boundary, `false` otherwise.
+	 */
+	_isCorrectDomSelectionPosition( domParent, offset ) {
+		// If selection is before or in the middle of inline filler string, it is incorrect.
+		if ( this.isText( domParent ) && startsWithFiller( domParent ) && offset < INLINE_FILLER_LENGTH ) {
+			// Selection in a text node, at wrong position (before or in the middle of filler).
+			return false;
+		} else if ( this.isElement( domParent ) && startsWithFiller( domParent.childNodes[ offset ] ) ) {
```
Reinmar: Unnecessary `else`?

src/view/domconverter.js
```diff
@@ -839,6 +839,50 @@ export default class DomConverter {
...
+			// Selection in an element node, before filler text node.
+			return false;
+		}
+
+		const viewParent = this.mapDomToView( domParent );
+
+		// If selection is in `view.UIElement`, it is incorrect. Note that `mapDomToView()` returns `view.UIElement`
+		// also for any dom element that is inside the view ui element (so we don't need to perform any additional checks).
+		if ( viewParent && viewParent.is( 'uiElement' ) ) {
```
Reinmar: Do you have a test that it works with deeper UIElement structures? AFAIR the mapper will anyway return the UI element itself even for its children, but it'd be good to have a test for it here too.

src/view/renderer.js
```diff
@@ -602,24 +602,33 @@ export default class Renderer {
...
-				currentSelection: this.selection
-			};
-
-			log.warn(
-				'renderer-skipped-selection-rendering: The selection was not rendered due to its similarity to the current one.',
-				data
-			);
-
-			return;
+		// Below we will check whether DOM Selection needs updating at all.
```
Reinmar: We rather use `DOM selection`, `DOM element`, etc. It's not a type here, but an instance. "DOM" is an adjective and `selection/engine` a noun.

tests/conversion/view-selection-to-model-converters.js
```diff
@@ -109,7 +109,8 @@ describe( 'convertSelectionChange', () => {
 		expect( model.selection.isBackward ).to.true;
 	} );
 
-	it( 'should not enqueue changes if selection has not changed', () => {
+	it( 'should re-convert selection even if it has not changed in model', () => {
+		// Selection might not have changed in model but it needs to be reconverted because it ended up in incorrect place in DOM.
```
Reinmar: in the model
the DOM
etc.

tests/view/domconverter/domconverter.js
```diff
@@ -190,4 +191,89 @@ describe( 'DomConverter', () => {
 			} );
 		} );
 	} );
+
+	describe( 'isCorrectDomSelection', () => {
```
Reinmar: Missing `()`

src/conversion/view-selection-to-model-converters.js
Reinmar: I saw the comment in the test. But... is setting the model selection to identical selection triggering the entire conversion (i.e. firing the change events)?  What's the implication of this change to potential consumers of these events? 

In fact, I'd say that the selection should not fire any events if selections are identical. I even think we worked on this, but maybe it was in a different class (maybe it was something about attributes). Anyway, this sounds wrong.

cc @pjasiun

pjasiun: Maybe we should try to render in the selection observer if there are no changes? I remember there was a ticket about it...

scofalik: What we really want to achieve is to re-render selection. So I think that what @pjasiun suggested will be fine. The problem is that we don't have a reference to `view.Document` there :(.

Anyway, as I said, what we want to achieve is re-rendering and we do so through `enqueueChanges`. Since nothing changes in `model.Selection` no event will be fired by it.

But we could also just fire `changesDone` event (but it's a bit hacky).

Or, pass the reference there somehow.

src/view/domconverter.js
scofalik: I don't agree. The function asks: is (passed argument) a correct dom selection? `isDomSelectionCorrect` would be fine if dom selection was a member of `DomConverter`.

Reinmar: We've just found `isDomSelectionBackward()` method in our code. Blame pointed to @pjasiun. Edit: Nope, it was @scofalik.

scofalik: 😢

src/view/domconverter.js
scofalik: http://dictionary.cambridge.org/grammar/british-grammar/at-on-and-in-place

Reinmar: GTK. Still, misses "the" :P

src/view/domconverter.js
scofalik: It should be `Element`.

src/conversion/view-selection-to-model-converters.js
pjasiun: https://github.com/ckeditor/ckeditor5-engine/issues/796

scofalik:
I've fixed all the raised concerns.

src/view/node.js
```diff
@@ -181,6 +182,15 @@ export default class Node {
 		}
 	}
 
+	toJSON() {
```
Reinmar: Missing API docs.

src/view/uielement.js
```diff
@@ -112,6 +112,12 @@ function jumpOverUiElement( evt, data, domConverter ) {
 			const domOffset = domSelection.getRangeAt( 0 ).startOffset;
 
 			const viewPosition = domConverter.domPositionToView( domParent, domOffset );
+
+			// In case if dom element is not converted to view or is not mapped or something. Happens for example in some tests.
+			if ( viewPosition === null ) {
```
Reinmar: Isn't this only a problem with these tests? How many tests were crashing on this?

src/view/uielement.js
```diff
@@ -82,9 +83,51 @@ export default class UIElement extends Element {
 	}
 }
 
+/**
+ * Assign key observer which will move cursor over a ui element if right arrow key is pressed and selection is before
```
Reinmar: "UI element"

src/view/uielement.js
```diff
@@ -82,9 +83,51 @@ export default class UIElement extends Element {
...
+ * ui element.
+ *
+ * @param {module:engine/view/document~Document} document Document instance we should inject quirks handling on.
```
Reinmar: Document to which the quirks handling will be injected.

src/view/uielement.js
Reinmar: Besides, this function's description should start with some general introduction. Then it can start describing particular quirks.

src/view/renderer.js
```diff
@@ -602,24 +601,33 @@ export default class Renderer {
...
-				'renderer-skipped-selection-rendering: The selection was not rendered due to its similarity to the current one.',
-				data
-			);
-
-			return;
+		// Below we will check whether DOM selection needs updating at all.
+		// We need to update DOM selection if either:
+		// * it is at incorrect position, or
+		// * it has changed (when compared to view selection).
+		if ( this.domConverter.isDomSelectionCorrect( domSelection ) ) {
```
Reinmar: Couldn't we move this whole "needs to be updated" check to a separate function? It'd make reading this simpler and that function can be simpler (early return if selection is not correct, etc.)

Reinmar:
npm t:

```
Chrome 59.0.3071 (Mac OS X 10.12.4): Executed 6474 of 6480 (1 FAILED) (skipped 6) (1 min 6.896 secs / 16.429 secs)
```

* 1 broken test
* 4 new "skipped-selection-rendering" warnings

src/view/renderer.js
scofalik: This is just a one-liner, one statement. And then you need those other checks below to steer function flow correctly.

scofalik: BTW. I really dislike that you cannot comment on a block of code instead of line on GitHub.

Reinmar: Of course the new function would have nearly the same conditions. But they wouldn't be nested any more. I had a problem with reading this code and understanding this change due to the nesting levels.

scofalik:
> 4 new "skipped-selection-rendering" warnings (they need to be checked and muted by stubbing the log function or changing tests a bit)

I think that we can do this in a followup ticket. Closing this PR is more important than some warnings. We can investigate new and old warnings in another issue.

Reinmar:
Only if you'll work on them right after closing this PR.

scofalik:
Added PR that fixes broken test in typing: https://github.com/ckeditor/ckeditor5-typing/pull/115

scofalik:
Travis fails because some weird problems with linting.

Reinmar:
Shift+arrow needs to be handled too:

aug-09-2017 14-27-43

Reinmar:
Besides:

* Now there are so many warnings that I cannot accept this PR like that. We need to clean them up.
* There's a test failing in `view/selection`:

   ```
  Selection
    isSimilar
      ✖ should return true if all ranges trimmed from both selections are equal
        Chrome 60.0.3112 (Mac OS X 10.12.4)
      AssertionError: expected false to be true
   ```

scofalik:
I've handled shift key, fixed failing tests and made changes do that warning is not logged.

Reinmar:
It doesn't work if there's an attribute element before the UI element.

scofalik:
Fixed. Try harder!

Reinmar:
OK, it seems to work fine now.

However, I miss several tests. Such as –  multiple UI elements, multiple attribute elements, root boundaries (the selection should not leave the last paragraph), etc. You may also be able to find a lot of false positives. I'd also like to see UI elements in the middle of the text in the manual test, because that can't be tested now.

I'll merge this PR to not block other things, but please add the missing tests.

PR #1052, opened 2017-07-28
Corrected RenameDelta x SplitDelta transformation special case

### LINKED ISSUE #1051: RenameDelta or RenameOperation transformation causes renaming incorrect nodes
Originally reported in comment: https://github.com/ckeditor/ckeditor5-undo/issues/65#issuecomment-313444518

What happens is that during undo rename is incorrectly applied to more than one element.

scofalik:

Fixed: Fixed a bug when renaming followed by merge or split, during undo, multiple elements where incorrectly renamed instead of one. Closes #1051.

### DIFF
--- a/src/model/delta/basic-transformations.js
+++ b/src/model/delta/basic-transformations.js
@@ -348,15 +348,16 @@ addTransformationCase( WrapDelta, SplitDelta, ( a, b, context ) => {
 
 // Add special case for RenameDelta x SplitDelta transformation.
 addTransformationCase( RenameDelta, SplitDelta, ( a, b, context ) => {
-	const splitPosition = new Position( b.position.root, b.position.path.slice( 0, -1 ) );
+	const undoMode = context.aWasUndone || context.bWasUndone;
+	const posBeforeSplitParent = new Position( b.position.root, b.position.path.slice( 0, -1 ) );
 
 	const deltas = defaultTransform( a, b, context );
 
-	if ( a.operations[ 0 ].position.isEqual( splitPosition ) ) {
+	if ( !undoMode && a.operations[ 0 ].position.isEqual( posBeforeSplitParent ) ) {
 		// If a node that has been split has it's name changed, we should also change name of
 		// the node created during splitting.
 		const additionalRenameDelta = a.clone();
-		additionalRenameDelta.operations[ 0 ].position = a.operations[ 0 ].position.getShiftedBy( 1 );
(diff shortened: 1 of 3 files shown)


src/model/delta/basic-transformations.js
```diff
@@ -365,20 +366,25 @@ addTransformationCase( RenameDelta, SplitDelta, ( a, b, context ) => {
...
-	const splitPosition = new Position( a.position.root, a.position.path.slice( 0, -1 ) );
-
-	if ( a._cloneOperation instanceof InsertOperation ) {
-		// If element to split had it's name changed, we have to reflect this change in an element
-		// that is in SplitDelta's InsertOperation.
-		if ( b.operations[ 0 ].position.isEqual( splitPosition ) ) {
-			a._cloneOperation.nodes.getNode( 0 ).name = b.operations[ 0 ].newName;
-		}
+addTransformationCase( SplitDelta, RenameDelta, ( a, b, context ) => {
+	const undoMode = context.aWasUndone || context.bWasUndone;
```
Reinmar: Since in both cases we use `undoMode = a || b`, couldn't we store just `undoMode` in the context.

scofalik: That was my idea but I expect that there might be cases where we just want one or two.

scofalik:
Additional integration tests are here: https://github.com/ckeditor/ckeditor5-undo/pull/68

src/model/delta/basic-transformations.js
Reinmar: OK, that's what I thought. Makes sense.

src/model/delta/transform.js
```diff
@@ -350,10 +354,21 @@ function padWithNoOps( deltas, howMany ) {
 // Using data given in `context` object, sets `context.insertBefore` and `context.forceNotSticky` flags.
 // Also updates `context.wasAffected`.
 function _setContext( a, b, context ) {
+	_setBWasUndone( b, context );
```
Reinmar: Do we only undo `b`s?

Reinmar:
I'm merging this PR, but please answer https://github.com/ckeditor/ckeditor5-engine/pull/1052#discussion_r130111839 for the future reference.

src/model/delta/transform.js
scofalik: Your question is almost correct ;).

We need only information about delta `b`, because delta `a` cannot be undone delta, because we are transforming it at the moment, so it is not yet applied. For delta `a` it would always be `false`.

PR #1054, opened 2017-07-31
T/1053 Incorrectly set context.forceWeakRemove for MergeDelta during OT

### LINKED ISSUE #1053: OT: incorrectly set context.forceWeakRemove for MergeDelta
This is how `context.forceWeakRemove` is set now:

```javascript
context.forceWeakRemove = !( originalB instanceof RemoveDelta && !history.isUndoneDelta( originalB ) );
```

This means that it is `true` whenever `originalB` is not `RemoveDelta`. This is wrong. Not only `RemoveDelta` has `RemoveOperation`. This causes a bug when transforming `MergeDelta`, described here: https://github.com/ckeditor/ckeditor5-undo/issues/65#issuecomment-313444977

Actually, since `context.forceWeakRemove` is used only for `RemoveOperation` x `MoveOperation` transformation, we can safely get rid of first part of this condition: `originalB instanceof RemoveDelta`. The flag doesn't have to have meaningful value if given delta has no `RemoveOperation`s.

scofalik:

Fixed: Fixed bug when additional list item has been created when undoing setting block quote on list followed by splitting list item in that list. Closes #1053.

---

### Additional information

1. When working on this issue I discovered another one. `RemoveOperation` x `MoveOperation` was changing `context.isStrong` value directly, which affected further transformations. Now `context.isStrong` is assigned to a different variable and that variable is changed if needed.

2. Tests for this issue will be written in `ckeditor5-undo`.

### DIFF
--- a/src/model/delta/transform.js
+++ b/src/model/delta/transform.js
@@ -444,21 +444,21 @@ function _setForceNotSticky( context ) {
 // always treats `RemoveOperation` as a stronger one, no matter how `context.isStrong` is set. It is like this
 // to provide better results when transformations happen.
 //
-// This is however works fine only when additional context is not used.
+// This, however, works fine only when additional context is not used.
 //
 // When additional context is used, we need a better way to decide whether `RemoveOperation` is "dominating" (or in other
 // words, whether nodes removed by given operation should stay in graveyard if other operation wants to move them).
 //
 // The answer to this is easy: if `RemoveOperation` has been already undone, we are not forcing given nodes to stay
 // in graveyard. In such scenario, we set `context.forceWeakRemove` to `true`. However, if the `RemoveOperation` has
-// not been undone, we set `context.forceWeakRemove` to `false` because we want remove to be "dominating".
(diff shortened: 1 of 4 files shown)


scofalik:
Tests: https://github.com/ckeditor/ckeditor5-undo/pull/69

Reinmar:
Tests for this change need to also be in this package. There can be additional integration tests in ckeditor5-undo but we need some basic verification here. Otherwise, someone (most likely you) will complain that he changed something in ckeditor5-engine, tests passed, but they failed in ckeditor5-undo ;).

scofalik:
I will write the test but you have to be aware that these kinds of tests are difficult to write in "non-integrational" way.

Reinmar:
But I didn't say that the test must be non-integrational. I just want it to be here. And there's a whole range of approaches and API levels between tests in the undo package and unit tests (which we don't want). In other words – I'd like to see as high-level test as you could implement in here.

scofalik:
I know I know :). I just wanted to explain why I haven't written any test in the first place.

scofalik:
I've added additional test (and also find another conceptual problem and fixed it).

PR #1056, opened 2017-08-01
T/1055 Selection attributes should be cleared in an enqueueChanges block

### LINKED ISSUE #1055: Selection attributes are cleared from elements outside enqueue changes block
Original issue: https://github.com/ckeditor/ckeditor5-undo/issues/65#issuecomment-314216007

It appears that the fixer which clears selection attributes is applied outside of enqueue changes block, which messes up and crashes editor.

scofalik:

Fixed: Selection attributes should be cleared in an enqueueChanges block. Closes #1055.

---

### Additional information

When fixing this bug, I found a bug in `AttributeDelta` x `SplitDelta` transformation. It was similar to a bug that I fixed very recently - concerning `RenameDelta` x `SplitDelta` transformation. I also reformatted some code in `delta/transform/basic-transformation.js` to make it more clear.

As with other undo related bugs, tests will come in `ckeditor5-undo`.

### DIFF
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -85,7 +85,7 @@ export default class DocumentSelection extends Selection {
 
 			// Whenever element which had selection's attributes stored in it stops being empty,
 			// the attributes need to be removed.
-			clearAttributesStoredInElement( changes, batch );
+			clearAttributesStoredInElement( changes, batch, this._document );
 		} );
 	}
 
@@ -691,7 +691,7 @@ function getAttrsIfCharacter( node ) {
 }
 
 // Removes selection attributes from element which is not empty anymore.
-function clearAttributesStoredInElement( changes, batch ) {
+function clearAttributesStoredInElement( changes, batch, document ) {
 	// Batch may not be passed to the document#change event in some tests.
 	// See https://github.com/ckeditor/ckeditor5-engine/issues/1001#issuecomment-314202352
 	// Ignore also transparent batches because they are... transparent.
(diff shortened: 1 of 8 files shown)


scofalik:
I've also added "undo mode" for `RemoveDelta` x `SplitDelta` transformation.

scofalik:
https://github.com/ckeditor/ckeditor5-undo/pull/70

scofalik:
Please do not review yet. I have to think about couple of things.

Reinmar:
I labelled it with R- so to know that I don't need to look on it yet.

scofalik:
Theoretically you can look at this. I just don't like what is happening to delta transformation special cases.

pjasiun:
Please solve conflicts with the master branch.

scofalik:
Done

src/model/documentselection.js
```diff
@@ -707,9 +707,15 @@ function clearAttributesStoredInElement( changes, batch ) {
 		return;
 	}
 
-	const storedAttributes = Array.from( changeParent.getAttributeKeys() ).filter( key => key.startsWith( storePrefix ) );
+	const anyStoredAttributes = Array.from( changeParent.getAttributeKeys() ).some( key => key.startsWith( storePrefix ) );
```
Reinmar: Why not having `storedAttributes`  and checking its length in the `if()` below?

scofalik: Maybe I am not right, but...

we need to "refresh" `storedAttributes` array anyway in the `enqueueChanges` because something might have changed.

But, if that's true, we should not have `if( anyStoredAttributes )` because there might be no stored attributes ATM but, at least theoretically, may be some later?

Argh, this `enqueueChanges` is too damn difficult!

scofalik: I think that we should make an agreement whether we are fine with `enqueueChanges` calls that do nothing. If we are fine with them, then we should not make conditional statements before `enqueueChanges`. The drawback is that we may fire more `changesDone` than needed, though.

If we are fine with empty `enqueueChanges` calls, we can remove that `if`.

Reinmar: Damn... I haven't even thought about enqChanges not being synchronous ;<

Regarding empty enqChanges – I can't think of any issue that would cause, so I guess we can remove that `if()`.

Reinmar:
I removed the `if()` which we discussed and unfortunately tests fail because they check how many times `enqChanges` was called (and it's now called all the time). So you need to update these tests somehow.

scofalik:
Done. I've removed failing checks from the tests and added one dedicated test that checks whether the change is done in enqueue changes block.

Reinmar:
And now this branch has conflicts :D

scofalik:
I resolved conflicts. I also had to push some changes to tests in `enter`, `typing` and `list` repo because they started to crash. Those tests either incorrectly checked whether `enqueueChanges` block was used, or didn't use `enqueueChanges` block when they needed.

PR #1057, opened 2017-08-01
t/660: Implemented ViewDocument#scrollToTheSelection method

### LINKED ISSUE #660: Various algorithms should scroll the viewport to the selection
1. Paste, press enter multiple time, etc.
2. The caret goes off screen and you stop seeing it.

The viewport should automatically be scrolled to show that position.

oleq:

Feature: Implemented ViewDocument#scrollToTheSelection method. Closes #660.

#### Commit message for integration branches:

Feature: View Document should scroll to the selection upon user input (see: ckeditor/ckeditor5-engine#660).

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-utils/pull/174.

Global mgit configuration (branch constellation): https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-engine/660.

Integration branches:
* https://github.com/ckeditor/ckeditor5-block-quote/tree/t/ckeditor5-engine/660
* https://github.com/ckeditor/ckeditor5-clipboard/tree/t/ckeditor5-engine/660
* https://github.com/ckeditor/ckeditor5-enter/tree/t/ckeditor5-engine/660
* https://github.com/ckeditor/ckeditor5-typing/tree/t/ckeditor5-engine/660

Issues:
* [Integration in the `ImageUploadEngine`](https://github.com/ckeditor/ckeditor5-upload/blob/master/src/imageuploadengine.js#L47-L54) – do we need to duplicate the code from the clipboard? Maybe we can use some other event which also fires on user input (paste) and it will handle scrolling once, in the clipboard?
* More integrations? List, link, etc. They're not critical but presumably, depending on the styling of the web page, they also may require scrolling after command execution.

### DIFF
--- a/src/view/document.js
+++ b/src/view/document.js
@@ -307,10 +307,12 @@ export default class Document {
 	scrollToTheSelection() {
 		const range = this.selection.getFirstRange();
 
-		scrollUtils.scrollViewportToShowTarget( {
-			target: this.domConverter.viewRangeToDom( range ),
-			viewportOffset: 20
-		} );
+		if ( range ) {
+			scrollUtils.scrollViewportToShowTarget( {
+				target: this.domConverter.viewRangeToDom( range ),
+				viewportOffset: 20
+			} );
+		}
 	}
 
 	/**
--- a/src/view/document.js
+++ b/src/view/document.js
@@ -20,6 +20,7 @@ import KeyObserver from './observer/keyobserver';
 import FakeSelectionObserver from './observer/fakeselectionobserver';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
 import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
+import scrollUtils from '@ckeditor/ckeditor5-utils/src/dom/scroll';
 
 /**
  * Document class creates an abstract layer over the content editable area.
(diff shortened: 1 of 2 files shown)


Reinmar:
I'm not sure whether we need to handle scrolling in case of blockquote (CKE4 doesn't). And I'm not sure whether we shouldn't handle scrolling for link instead.

aug-09-2017 10-36-42

In any case – scrolling for blockquote doesn't work.

Pasting, deleting, etc. works great.

oleq:
Blockquote didn't scroll because I enabled it for the enter key only. Now it scrolls always when executed.

oleq:
As for the link, we could enable scrolling there too. OTOH, link seldom makes the content look so different that it moves out of the viewport (unlike block quote which adds margins, paddings, etc.). It would only help if one selected text, scrolled somewhere else and then saved the value of the form to create a link, which is not a very common case really, isn't it?

Reinmar:
> Blockquote didn't scroll because I enabled it for the enter key only. Now it scrolls always when executed.

Oopsie. My bad. I haven't checked the context. So, the previous implementation made most sense (and was identical to CKE4's). Please revert https://github.com/ckeditor/ckeditor5-block-quote/commit/8c9ec706ae0106e71e73a645889cc4081d8fca51 especially that we didn't want to put scrolling in commands.

Reinmar:
> It would only help if one selected text, scrolled somewhere else and then saved the value of the form to create a link, which is not a very common case really, isn't it?

I agree that this is a rare case. That's a test case I thought immediately but I'm not sure how often you'd do that in real life. We can live without scrolling for now.

oleq:
> Please revert ckeditor/ckeditor5-block-quote@8c9ec70 especially that we didn't want to put scrolling in commands
Done.

oleq:
I updated all the branches to work with the latest ckeditor5-dev.

src/view/document.js
```diff
@@ -300,6 +301,21 @@ export default class Document {
...
 	/**
+	 * Scrolls the page viewport and {@link #domRoots} with their ancestors to reveal the
+	 * caret, if not already visible to the user.
+	 */
+	scrollToTheSelection() {
+		const range = this.selection.getFirstRange();
+
+		if ( range ) {
+			scrollViewportToShowTarget( {
+				target: this.domConverter.viewRangeToDom( range ),
```
Reinmar: I haven't noticed this before, but the `target` option isn't the most flexible one: https://github.com/ckeditor/ckeditor5-utils/issues/176.

PR #1059, opened 2017-08-04
Fixed: view.Range#getTrimmed() was returning incorrect ranges in some cases

### LINKED ISSUE #1058: view.Range#getTrimmed returns weird / incorrect results for ranges in empty elements
This is something that I spotted when writing some edge case-y tests. Assume such a range:

```html
<container:p>[<attribute:span></attribute:span>]</container:p>
```

When it gets `.getTrimmed()`, this is the result:

```html
<container:p>]<attribute:span></attribute:span>[</container:p>
```

Since not text and no container were found, positions were just moved to the edge of current container element and they got swapped. Actually, this is the result for any range that would be created in such structure.

I know that it's not common that there is container element with attribute element and no content but it happens. It might happen for ui element. I think we need to fix it. I would have expected that in case above the result will be a collapsed range in `<span>` but if that appears too difficult, I will take any collapsed range inside the `<p>`.

scofalik:

Fix: `view.Range#getTrimmed()` was returning incorrect ranges in some cases. Fixes #1058.

---

### Additional information

For now I've gone with the easy solution because it seems safer and was easy to implement.

Such range: `<p>[<span></span]</p>` will be trimmed to `<p><span></span>[]</p>`. However I am not sure it is really correct. I suggest @pjasiun to review this as he had introduced `getTrimmed` method so he might have a broader view on this.

### DIFF
--- a/src/view/range.js
+++ b/src/view/range.js
@@ -98,8 +98,8 @@ export default class Range {
 	 * @returns {module:engine/view/range~Range} Enlarged range.
 	 */
 	getEnlarged() {
-		let start = this.start.getLastMatchingPosition( enlargeShrinkSkip, { direction: 'backward' } );
-		let end = this.end.getLastMatchingPosition( enlargeShrinkSkip );
+		let start = this.start.getLastMatchingPosition( enlargeTrimSkip, { direction: 'backward' } );
+		let end = this.end.getLastMatchingPosition( enlargeTrimSkip );
 
 		// Fix positions, in case if they are in Text node.
 		if ( start.parent.is( 'text' ) && start.isAtStart ) {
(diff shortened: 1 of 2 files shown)


Reinmar:
I think that I'd expect that the range will end up inside the span: `<span>[]</span>`, but then, what should happen in such a case?

```html
<p>[<a></a><b></b>]</p>
```

In this case, you'd neither like it to end in `<a>` or `<b>`. Perhaps this would make sense: `<a>[</a><b>]</b>`.

Anyway, I'm just guessing here. We should have a real TC in hands before we change it otherwise we will all need to guess. So – can you think of any real scenario in which this case might be hit?

scofalik:
No, I don't. Also if we would like the algorithm to be more smart, it would have to prevent collapsing the selection inside UI element.

Reinmar:
> Edit: which two? I can see one.

One use case with two elements used in it :D

Reinmar:
> No, I don't.

And because of that I think that it doesn't make sense to spend on this time now. Let's just wait.

scofalik:
I've added additional test.

Reinmar:
That PR got R- from me because I don't see why we're fixing it now. We're doing it blindly and unless one of us can think of some real life cases (which will drive the solution) I wouldn't touch this. It's not worth our time otherwise.

scofalik:
I've added an explanation to tests that this is not meant to be the final solution. In future, if we have some real cases we may go back to this implementation and change it. Anyway, the fixed result is **without any doubt** more correct than current result, that's why we are fixing it.

PR #1061, opened 2017-08-09
Writer wrapping order.

### LINKED ISSUE #1060: Writer should keep order of wrapping for elements with same priority
Extracted from https://github.com/ckeditor/ckeditor5/issues/502.

Order of `AttributeElements` should be predictable for elements with the same priority when writer wraps them. Method `getIdentity()` should be introduced for view elements, and will return value that can be used to compare them. Writer compares elements using `element.isSimilar( otherElement )` method, so two "similar" elements should have same identity.

szymonkups:

Fix: Writer creates consistent hierarchy for attribute elements with same priorities. Introduced `viewElement.getIdentity()` method. Closes #1060.

### DIFF
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -701,7 +701,10 @@ export default class Element extends Node {
 		const styles = Array.from( this._styles ).map( i => `${ i[ 0 ] }:${ i[ 1 ] }` ).sort().join( ';' );
 		const attributes = Array.from( this._attrs ).map( i => `${ i[ 0 ] }="${ i[ 1 ] }"` ).sort().join( ' ' );
 
-		return `${ this.name } class="${ classes }" style="${ styles }"${ attributes == '' ? '' : ' ' + attributes }`;
+		return this.name +
+			( classes == '' ? '' : ` class="${ classes }"` ) +
+			( styles == '' ? '' : ` style="${ styles }"` ) +
+			( attributes == '' ? '' : ` ${ attributes }` );
 	}
 
 	/**
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -673,6 +673,13 @@ export default class Element extends Node {
 		yield* this._customProperties.entries();
 	}
 
+	getIdentity() {
+		const attributes = Array.from( this._attrs ).map( i => i[ 0 ] + '=' + i[ 1 ] ).sort();
+		const styles = Array.from( this._styles ).map( i => i[ 0 ] + '=' + i[ 1 ] ).sort();
+
+		return [ this.name, ...attributes, ...styles ].join( '|' );
+	}
+
 	/**
 	 * Returns block {@link module:engine/view/filler filler} offset or `null` if block filler is not needed.
 	 *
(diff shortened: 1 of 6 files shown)


src/view/writer.js
```diff
@@ -975,6 +975,29 @@ function wrapChildren( parent, startOffset, endOffset, attribute ) {
...
+// @param {module:engine/view/attributeelement~AttributeElement} b
+// @returns {Boolean}
+function shouldABeOutsideB( a, b ) {
+	if ( a.priority < b.priority ) {
+		return true;
+	} else if ( a.priority > b.priority ) {
+		return false;
+	}
+
+	if ( a.name == b.name ) {
```
Reinmar: Missing a comment. You wrote it to me on priv but it's missing here ;)

src/view/element.js
```diff
@@ -674,6 +674,37 @@ export default class Element extends Node {
 	}
 
 	/**
+	 * Returns identity string based on element's name, styles, classes and other attributes.
+	 * Two elements that {@link #isSimilar are similar} will have same identity string.
+	 * It has following format:
```
Reinmar: "the following format"

src/view/element.js
```diff
@@ -674,6 +674,37 @@ export default class Element extends Node {
 	}
 
 	/**
+	 * Returns identity string based on element's name, styles, classes and other attributes.
+	 * Two elements that {@link #isSimilar are similar} will have same identity string.
+	 * It has the following format:
+	 *
+	 *		"name|classes(class1,class2,class3)|styles(style1=val1,style2=val2)|attributes(attr1=val1,attr2=val2)"
```
pjasiun: I don't like the fact that we introduce one more custom format. Can't we return it in HTML-like format:
`name classes="class1,class2,class3" styles="style1:val1;style2:val2" attr1="val1" attr2="val2"`. It does not seams to be longer and people are used to parse it ealiy while debugging.

src/view/element.js
```diff
@@ -788,3 +819,14 @@ function normalize( nodes ) {
...
+
+// Returns string representation of povided map in following format:
+//
+//		"mapKey1=value1,mapKey2=value2,mapKey3=value3"
+//
+// NOTE: All map keys should be strings. Key-value pairs will be sorted.
+//
+// @returns {String}
+function mapToSortedString( map ) {
+	return Array.from( map ).map( i => i[ 0 ] + '=' + i[ 1 ] ).sort().join( ',' );
```
pjasiun: It would be great to have `sort` in `getStyleNames` and `getClassNames` so they always return values in the same order. We need to be sure that the order is the same in many cases (here, tests utils, `isSimillar` method). `getStyleNames` and `getClassNames` looks like a good place to ensure the same order, then `getIdentity` could use these methods instead of introducing another method which sort.

pjasiun: Or nevermind. In fact, it will be less efficient, and we use these methods very often. To make it in the propr way we should save styles, classes, and attributes in the ordered collection, but it's not a good ticket for it.

src/view/element.js
pjasiun: For instance:
`foo class="baz" style="border-color:white;color:red;" apple="20" banana="10"`
Is much easier for my brain to parse then:
`foo|classes(baz)|styles(border-color=white,color=red)|attributes(apple=20,banana=10)`

PR #1073, opened 2017-08-11
Fixed: AttributeElement with bogus BR should be pushed after all ui elements

### LINKED ISSUE #1072: Force rendering bogus BR after UI elements
At the moment, an attribute element that is in otherwise empty container element receives bogus BR:
```html
<p>
  <strong><br /></strong>
</p>
```
However it causes a problem when there is also a UI element in that container:
```html
<p>
  <strong><br /></strong>
  <span class="myUi"></span>
</p>
```
Because the UI element is rendered incorrectly.

We should force rendering such attribute element after all UI elements. The attribute element should be pushed if its parent has no non-UI elements.

scofalik:

Fix: `AttributeElement` with bogus `<br />` will now be placed after all UI elements which will fix how those elements are rendered. Closes #1072.

### DIFF
--- a/src/conversion/model-selection-to-view-converters.js
+++ b/src/conversion/model-selection-to-view-converters.js
@@ -200,10 +200,32 @@ function wrapCollapsedSelectionPosition( modelSelection, viewSelection, viewElem
 	}
 
 	let viewPosition = viewSelection.getFirstPosition();
+
+	// This hack is supposed to place attribute element *after* all ui elements if the attribute element would be
+	// the only non-ui child and thus receive a block filler.
+	if ( shouldPushAttributeElement( viewPosition.parent ) ) {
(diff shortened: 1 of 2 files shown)


src/conversion/model-selection-to-view-converters.js
```diff
@@ -200,12 +200,34 @@ function wrapCollapsedSelectionPosition( modelSelection, viewSelection, viewElem
 	}
 
 	let viewPosition = viewSelection.getFirstPosition();
+
+	// This hack is supposed to place attribute element *after* all ui elements if the attribute element would be
+	// the only non-ui child and thus receive a block filler.
```
pjasiun: This comment tells what you do, but not why you do it. At least link to the issue is needed.

scofalik:
I've added more explanation to the comment.

PR #1083, opened 2017-08-18
T/ckeditor5 typing/101 Multiple spaces fix

scofalik:

Fix: Multiple spaces in an empty paragraph are now allowed. Closes https://github.com/ckeditor/ckeditor5-typing/issues/101.

------

### Additional info

The whole algorithm got simplified and fixed and a few other edge cases has been also fixed.


Reinmar:
Why did you expose getTouchingTextNode as a util? Is it necessary?

scofalik:
It is used in `domconverter.js` and `renderer.js` and I couldn't find any good place where to fit it. I think that with time more functions might join there. For example I've used two times already a function that takes a view position and moves it forward after all ui elements.

If I were to put it in some existing place I'd probably do it as an exported function `treewalker.js` or maybe in `writer.js`, although the function does not change the view in any way...

src/view/renderer.js
```diff
@@ -420,6 +425,12 @@ export default class Renderer {
 
 		if ( actualText != expectedText ) {
 			domText.data = expectedText;
+
+			const nextTouchingTextNode = getTouchingTextNode( viewText, true );
```
Reinmar: Missing comment why we need to do that.

src/view/utils.js
```diff
@@ -0,0 +1,41 @@
...
+ * For licensing, see LICENSE.md.
+ */
+
+import ViewPosition from './position';
+import ViewTreeWalker from './treewalker';
+
+/**
+ * Contains utility functions for working on view.
+ *
+ * @module engine/view/utils
```
Reinmar: Let's rename this to `walker-utils` since the utils that come to our mind are Walker related. We also considered putting this to the walker's module but we don't want to make it too big (hence, a new module).

src/view/utils.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+/**
+ * For given {@link module:engine/view/text~Text view text node}, it finds previous or next sibling that is contained
+ * in the same container element. If there is no such sibling, `null` is returned.
+ *
+ * @param {module:engine/view/text~Text} node Reference node.
+ * @param {Boolean} getNext If `true` next touching sibling will be returned. If `false` previous touching sibling will be returned.
```
Reinmar: It'd be good to find it a better name... but I can't :D. `lookRight`?

Also, this is an optional param. Should be marked with `[]`. Maybe this will clarify its meaning (typically, boolean flags are optional).

Reinmar: As for this flag being optional – it's always passed in the tests: https://github.com/ckeditor/ckeditor5-engine/pull/1083/files#diff-41de3c28cf4c9630dfaec4e35850f0faR38. This should be changed also there.

src/view/renderer.js
```diff
@@ -407,6 +406,12 @@ export default class Renderer {
 	 */
 	_updateText( viewText, options ) {
 		const domText = this.domConverter.findCorrespondingDomText( viewText );
+
+		// If this is a new text node and it is not in DOM, it will be created and handled in `_updateChildren`.
+		if ( !domText ) {
```
Reinmar: This looks very suspicious. How did this function work without this condition so far? Why is it needed here now.

src/view/utils.js
```diff
@@ -0,0 +1,41 @@
...
+export function getTouchingTextNode( node, getNext ) {
+	const treeWalker = new ViewTreeWalker( {
+		startPosition: getNext ? ViewPosition.createAfter( node ) : ViewPosition.createBefore( node ),
+		direction: getNext ? 'forward' : 'backward'
+	} );
+
+	for ( const value of treeWalker ) {
+		if ( value.item.is( 'containerElement' ) ) {
+			// ViewContainerElement is found on a way to next ViewText node, so given `node` was first/last
+			// text node in it's container element.
```
Reinmar: its

tests/view/domconverter/view-to-dom.js
```diff
@@ -323,33 +331,80 @@ describe( 'DomConverter', () => {
...
 
 			test( [ 'x', '   y' ], 'x _ y' );
 			test( [ 'x ', '   y' ], 'x _ _y' );
-			test( [ 'x  ', '   y' ], 'x_ _ _y' );
+			test( [ 'x  ', '   y' ], 'x _ _ y' );
 			test( [ 'x   ', '   y' ], 'x _ _ _y' );
-			test( [ 'x    ', '   y' ], 'x_ _ _ _y' );
+			test( [ 'x    ', '   y' ], 'x _ _ _ y' );
+
+			// "Non-empty" + "empty" text nodes.
```
Reinmar: What about mirrored cases with "empty" + "non-empty"?

scofalik:
Unfortunately there is one more bug because text nodes are not refreshed when a node is removed or inserted. This is handled in `_updateChildren` not in `_updateText`.

For example, `<p><em>&nbsp;x</em>` becomes `<p>x<em>&nbsp;x</em></p>`.

Reinmar:
Let's not enlarge this PR with this case. Refreshing sibling text nodes was already something that we added here.

src/view/renderer.js
Reinmar: So why doesn't that condition save us from adding one here?

scofalik: I don't understand. Before there was a condition in `render` method` in the loop where `_updateText` is called. I removed this condition because I wanted `_updateText` to be fired more often. But this caused a situation where `_upadateText` was fired on text nodes that weren't yet converted to DOM. That's why I needed to add the statement in `_updateText`.

Reinmar: Aaand it turned out we can't make this change ;> We'll also need to make a step back and remove the fix for subsequent text nodes updating. It seems that we may not be able to finalise it now because of how the text nodes are mapped. 

We can neither update all text nodes before children are updated because changes in children might've changed the indexes of text nodes (so updating text nodes would fail). Changing the order to 1) update children, 2) update text nodes doesn't help too because in the following case, the whole paragraph is updated first and then text nodes are not:

1. `<p>x<strong> b</strong></p>`
2. Type space after "x".
3. Paragraph's children get updated (because for some reason the browser mutated text nodes too much when typing the space). 
4. The `"x "` text node is up to date when it comes to updating text nodes, so we don't mark `" b"` to be updated.

Reinmar: @scofalik will you report a ticket for this case and the one you mentioned in https://github.com/ckeditor/ckeditor5-engine/pull/1083#issuecomment-323760233? And of course, please revert changes which we can't do now.

scofalik: https://github.com/ckeditor/ckeditor5-engine/issues/1093

scofalik:
Ready for review.

PR #1091, opened 2017-08-21
Renamed VirtualSelection to Highlight.

### LINKED ISSUE #1085: Rename virtual selection to highlight
I was explaining @scofalik what virtual selection is and it takes a while to get that virtual selection and fake selection are different mechanisms. In fact, the selection has 2 meanings:

 - the place where the user is editing and
 - visual marker over some text/content.

While the fake selection is focused mostly on the first mechanism, the virtual selection is only about the second. And I think it's wrong. We used to associate selection with the document selection, with the first meaning. Virtual selection is misleading since it has nothing in common with the selection as the place where the user is editing.

So, since virtual selection is not a selection we should find a better name.

Maybe just 'marker'? It describes very well what it really is and model markers will be mostly converted to it. What I don't like in calling it "marker", is the "mostly" part. We can convert model markers to elements and it might be strange that model markers do not need to be a view marker.

But it's Friday, 17:40 and I have no better ideas.

szymonkups:

Other: Renamed virtual selection to highlight. Closes #1085.

---

### Additional information

Depending branch in widget repository: https://github.com/ckeditor/ckeditor5-widget/tree/t/ckeditor5-engine/1085


src/conversion/model-to-view-converters.js
```diff
@@ -441,23 +441,23 @@ export function convertTextsInsideMarker( selectionDescriptor ) {
...
- * When `setVirtualSelection` and `removeVirtualSelection` custom properties are not present, element is not converted
+ * When `setHighlight` and `removeHighlight` custom properties are not present, element is not converted
  * in any special way. This means that converters will proceed to convert element's child nodes.
  *
- * @param {module:engine/conversion/buildmodelconverter~VirtualSelectionDescriptor|Function} selectionDescriptor
+ * @param {module:engine/conversion/buildmodelconverter~HighlightDescriptor|Function} highlightDescriptor
  * @return {Function}
  */
-export function convertElementsInsideMarker( selectionDescriptor ) {
+export function convertElementsInsideMarker( highlightDescriptor ) {
```
pjasiun: Since most of the methods names here focus on what will be done on the view and `highlight` works well as a verb, maybe a better name would be `highlightElement` and `highlightText`?

src/conversion/model-to-view-converters.js
```diff
@@ -475,9 +475,9 @@ export function convertElementsInsideMarker( selectionDescriptor ) {
 
 		const viewElement = conversionApi.mapper.toViewElement( modelItem );
 		const addMarker = evt.name.split( ':' )[ 0 ] == 'addMarker';
-		const selectionHandlingMethod = addMarker ? 'setVirtualSelection' : 'removeVirtualSelection';
+		const highlightHandlingMethod = addMarker ? 'setHighlight' : 'removeHighlight';
```
pjasiun: I'm not sure about `setHighlight`. First sounds strange that it's `setHighlight` instead of just `highlight`. Second, if we have `remove` the better pair might be `add`, especially when you can add multiple highlights. So I would go with `addHighlight` instead of `setHighlight`.

pjasiun: Especially, when its default action is `stack.add` and `stack.remove`.

Reinmar:
Please do remember that these changes are internal, not breaking, because this whole API has appeared in this milestone.

szymonkups:
@Reinmar - true, I'll remove breaking changes from suggested merge commit message.

src/conversion/buildmodelconverter.js
```diff
@@ -15,8 +15,8 @@ import {
 	removeUIElement,
 	wrapItem,
 	unwrapItem,
-	convertTextsInsideMarker,
-	convertElementsInsideMarker
+	highlightTexts,
+	highlightElements
```
pjasiun: All other methods here are singular (`wrapItem`, `removeUIElement`). I think this should be too.

PR #1104, opened 2017-08-24
(insert|delete)Content strips disallowed attributes from nodes

### LINKED ISSUE #1088: (insert|delete)Content should strip disallowed attributes from nodes
See: https://github.com/ckeditor/ckeditor5-clipboard/issues/32.

oskarwrobel:

Fix: `DataController#insertContent()` and `DataController#deleteContent()` should strip disallowed attributes from text nodes. Closes #1088.

---


tests/controller/insertcontent.js
```diff
@@ -610,6 +612,18 @@ describe( 'DataController', () => {
 				insertHelper( '<disallowedWidget>xxx</disallowedWidget>' );
 				expect( getData( doc ) ).to.equal( '<paragraph>f[]oo</paragraph>' );
 			} );
+
+			it( 'filters out disallowed attributes', () => {
```
Reinmar: Just a single test? I can't believe that this ticket was so simple ;)

src/controller/insertcontent.js
```diff
@@ -226,6 +226,12 @@ class Insertion {
 		if ( node.is( 'element' ) ) {
 			this.handleNodes( node.getChildren(), context );
 		}
+		// When disallowed node is a text but text is allowed in current parent it means that our node
+		// contains disallowed attributes and we have to remove them.
+		else if ( node.is( 'text' ) && this.schema.check( { name: '$text', inside: [ this.position.parent ] } ) ) {
```
Reinmar: Use `inside: position` for a more precise check.

Reinmar: Uh, sorry. You passed the whole parent there. It's fine. I thought that you passed just a name.

Still, I think that `inside: this.position` might work.

oskarwrobel: I used `[ this.position.parent ]` to keep it consistent with `Insertion#_checkIsAllowed` calls.

oskarwrobel: But of course, you are right `this.position` will work.

oskarwrobel: Fixed.

tests/controller/insertcontent.js
oskarwrobel: I added one more test :)
If you have an idea for more tests (I'm sure you have)  I'm open :)

oskarwrobel: I'm also little surprised that it went so easy :) But looks like it really is. When disallowed element is a `Text` but in general `Text` is allowed in current position it means that this `Text` contains disallowed attributes and we have to remove them.

Reinmar: Hm... I can't think of anything else right now. I was worried that some falsy cases might not be covered, but I see that they are. So, it LGTM.

oskarwrobel:
There is one issue.
When block element is inserted into the other block element is not handled as a disallowed node and then stripped but is properly inserted and then merged with a parent without stripping.

```javascript
schema.disallow( { name: '$text', attributes: 'bold', inside: 'heading1' );

<heading1>Foo[] bar</heading1>

insertContent( '<paragraph><$text bold="true">biz</$text></paragraph>' );
```

Result:
```javascript
<heading1>Foo<$text bold="true">biz[]</$text> bar</heading1>
```

oskarwrobel:
I created a separate private function `removeDisallowedAttributes` for insert and delete. My first thought was to create one common util function and reuse it in insert and delete but we have 2 different flows here. `insertContent` filters attributes on detached nodes or before some batch operations so filtering is made using `Element` API. `deleteContent` filters attributes on nodes which are already in the document so I had to use `AttributeDelta`.

src/controller/deletecontent.js
```diff
@@ -132,6 +132,13 @@ function mergeBranches( batch, startPos, endPos ) {
 		// To then become:
 		// <a><b>xy</b>[]</a><c>{}</c>
 		batch.merge( startPos );
+
+		// We need to check and strip disallowed attributes in direct children because after merge
+		// some attributes could end up in a parent where are disallowed.
+		//
+		// e.g. bold is disallowed for <H1>
+		// <h1>Fo{o</h1><p>b}a<b>r</b><p> -> <h1>Fo{}a<b>r</b><h1> -> <h1>Fo{}ar<h1>.
+		removeDisallowedAttributes( batch, Array.from( startParent.getChildren() ), [ startParent ] );
```
scofalik: Theoretically, `schemaPath` equal to `[ startParent ]` might be incorrect, but this is a pretty edge case and I can't think of convincing real world scenario.

Anyway, it might return wrong result if, for example `attr` is allowed on element `A` and `B` but disallowed on `B` if `B` is in `X`.

Then: `<x><b>foo</b><a><$text attr=true>bar</$text></x>` upon merging `attr` should be cleared from text and it probably would not be now.

scofalik: You can pass a `Position` instance as `inside` parameter. Maybe that would be easier and even more clear in implementation. If that would work too, it might solve the issue I described.

src/controller/insertcontent.js
```diff
@@ -226,6 +226,12 @@ class Insertion {
 		if ( node.is( 'element' ) ) {
 			this.handleNodes( node.getChildren(), context );
 		}
+		// When disallowed node is a text but text is allowed in current position it means that our node
+		// contains disallowed attributes and we have to remove them.
+		else if ( node.is( 'text' ) && this.schema.check( { name: '$text', inside: this.position } ) ) {
```
scofalik: Is `node.is( 'text' )` needed here? We are in `else` after `node.is( 'element' )` so AFAIR the node can be only text node.

src/controller/deletecontent.js
scofalik: One more note - I am not sure about checking only direct children. What if one of children has an element or text with incorrect attribute?

src/controller/insertcontent.js
```diff
@@ -402,31 +423,45 @@ class Insertion {
...
+	return node.is( 'text' ) ? '$text' : node.name;
+}
 
-		return node.name;
+// Removes disallowed by schema attributes from given nodes.
+//
+// @param {module:engine/model/schema~Schema} schema
+// @param {module:engine/model/node~Node|Array<module:engine/model/node~Node>} nodes List of nodes or a single node to filter.
+// @param {module:engine/model/schema~SchemaPath} schemaPath
+export function removeDisallowedAttributes( schema, nodes, schemaPath ) {
```
scofalik: Do we want to export this?

src/controller/insertcontent.js
```diff
@@ -402,31 +423,45 @@ class Insertion {
...
+	if ( !Array.isArray( nodes ) ) {
```
scofalik: I'd skip this `if` and do it like in `deletecontent.js`. If you have just one node, wrap it in `[ ]` where the function is called. Especially if this is mostly a private helper function.

src/controller/deletecontent.js
scofalik: Those also apply to the other `removeDisallowedAttributes` function.

src/controller/insertcontent.js
oskarwrobel: I forgot to remove it.

src/controller/insertcontent.js
oskarwrobel: I was thinking about it.

scofalik:
> There is one issue. (...)

Was this fixed? If not, is it difficult to fix it?

oskarwrobel:
> Was this fixed? If not, is it difficult to fix it?

It is fixed.

scofalik:
Mostly looks fine. I've pushed some docs corrections.

The code that you had to deal with in this issue is strictly connected with pasting and autoparagraphing. Have you tested how autoparagraphing worked after your changes?

You could create an editor with just, let's say, basic styles and paragraph and check if pasting website works okay (for example links should be stripped but text should be maintained, list items should become paragraphs, like headings, etc. maybe you could even find or create sample content with bold in heading and try to paste it, etc.)

oskarwrobel:
> You could create an editor with just, let's say, basic styles and paragraph and check if pasting website works okay (for example links should be stripped but text should be maintained, list items should become paragraphs, like headings, etc. maybe you could even find or create sample content with bold in heading and try to paste it, etc.)

Ok, I'll create a manual test for this ticket.

oskarwrobel:
Ok, I fixed almost everything except checking a combination of attributes when filtering. Now when node with attributes is disallowed I remove attributes one by one. I'll create a follow up for this.

The rest should work fine. Here is a manual test: http://localhost:8125/ckeditor5-engine/tests/manual/tickets/1088/1.html

src/controller/deletecontent.js
```diff
@@ -208,9 +215,41 @@ function shouldEntireContentBeReplacedWithParagraph( schema, selection ) {
...
-	return true;
+// Gets a name under which we should check this node in the schema.
+//
+// @param {module:engine/model/node~Node} node The node.
+// @returns {String} node name.
+function getNodeSchemaName( node ) {
+	return node.is( 'text' ) ? '$text' : node.name;
+}
+
+// Adds deltas with `removeAttributes` operation for disallowed by schema attributes on given nodes and its children.
```
scofalik: Creates `AttributeDelta`s that removes attributes that are disallowed by schema on given node and its children.

scofalik:
I think you didn't understand me as far as passing arguments to `removeDisallowedAttributes` go. `nodes` need to be something iterable. It does not have to be an array. Don't do `Array.from()` in the function call if there's no need to.

oskarwrobel:
Yep, you are right. I don't have to do `Array.from()`. I'll fix it.

scofalik:
Also, change `schemaPath` parameter name to something more fitting and update docs.

scofalik:
In manual test I miss a case where content is autoparagraphed and attributes are kept. Also, autoparagraphing has "two faces". First is changing unrecognized nodes to paragraphs (for example, `<li>`). Then there is pasting bare text into a place where text is forbidden but text with paragraph is okay (`blockquote` as far as I remember?) It would be nice to enable both those scenarios and to have possibility to check whether parameters are stripped or not.

You could maybe better organise this sample. For example do not plug in bold, so all bolds should be stripped. Plug in italic so we will know that it has to be preserved. Do not plug in lists so we will know they have to be converted to paragraph. I also miss copying image (to see whether not-stripping attributes from image works).

oskarwrobel:
Lists are useful in this test because we can check filtering on child elements. List directly in root has different schema rules that list in blockquote.

scofalik:
Okay. Then I don't know which type of block we can dismiss, maybe one of heading types?

oskarwrobel:
Ok, I think everything is fixed. The manual test is improved but I still use `ArticlePresets` and disable `Bold` manually on `Paragraph` to prevent from importing many many plugins to run this sample.

PR #1109, opened 2017-08-28
Added HighlightDescriptor#id handling for a highlight on text nodes

### LINKED ISSUE #1108: Use HighlightDescriptor#id when marker over text nodes is converted
Highlight nicely solves some issues connected with creating converters for markers. However there is still one issue, which we might easily address tweaking a bit the mechanisms we created.

At the moment, when we remove a marker, or any attribute element, we use `view.writer#unwrap`. To know which attribute element should be unwrapped, we pass `AttributeElement` instance. The passed element is then compared with elements from range using `isSimilar` function.

Unfortunately, sometimes the elements are pretty complicated. By default, everything gets compared: `name`, attributes, classes and `priority`. Sometimes it is a bit tricky to write converter/element creator function to create proper similar element. Sometimes elements base not only on model but some other sources of data.

This should be easier. I propose to use `descriptor#id` to compare whether given `AttributeElement`s are similar, if they were created by `toHighlight` converter builder. This makes a lot of sense. By default `descriptor#id` is a marker name, which was converted. If later we remove the marker, we want to unwrap everything that is connected with that marker name.

scofalik:

Feature: Highlights on text nodes will be now unwrapped basing on descriptor id (which by default is marker name). Closes #1108.

---

### Additional information

I also moved docs a bit. I think that `model-to-view-converters` is a better place for `HighlightDescriptor` definition, let me know what you think about this.

### DIFF
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -59,7 +59,7 @@ import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
  *		buildModelConverter().for( dispatcher ).fromAttribute( 'bold' ).toElement( 'strong' );
  *
  * 4. Model marker to view highlight converter. This is a converter that converts model markers to view highlight
- * described by {@link module:engine/conversion/buildmodelconverter~HighlightDescriptor} object passed to
+ * described by {@link module:engine/conversion/model-to-view-converters~HighlightDescriptor} object passed to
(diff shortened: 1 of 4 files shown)


scofalik:
https://github.com/ckeditor/ckeditor5-widget/pull/18

szymonkups:
No problem with moving `HighlightDescriptor` to `model-to-view-converters`.

PR #1111, opened 2017-08-28
Schema.checkAttributeInSelection uses existing element's attributes when querying schema

### LINKED ISSUE #1110: Schema.checkAttributeInSelection should check with other attributes
Currently, `schema.checkAttributeInSelection` method queries schema only with attribute provided to that method. Sometimes this is not enough.
For example: we have an `image` element that has `src` attribute defined as required one and optional `alt` attribute. So when checking if `alt` attribute is allowed in model like this:
```
[<image src="image.png"></image>]
```
`checkAttributeInSelection` should return `true`.

szymonkups:

Fix: Schema.checkAttributeInSelection uses existing element's attributes when querying schema. Closes #1110.

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -332,7 +332,11 @@ export default class Schema {
 					// If returned item does not have name property, it is a TextFragment.
 					const name = value.item.name || '$text';
 
-					if ( this.check( { name, inside: value.previousPosition, attributes: attribute } ) ) {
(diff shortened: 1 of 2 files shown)


scofalik:
Fix: `Schema.checkAttributeInSelection` should use element's existing attributes when querying schema. Closes #1110.

PR #1123, opened 2017-09-03
Introduced `DataController#removeDisallowedAttributes`

### LINKED ISSUE #1120: A way to filter disallowed attributes on given nodes
We just introduced [filtering disallowed attributes](https://github.com/ckeditor/ckeditor5-engine/pull/1115) inside `insertContent` and `deleteContent`. This filtering is made by two different (but very similar) local private functions. This is because `InsertContent` has to:
- filter `Text` nodes which are not yet inserted into the document (using `Element` API)
- filer nodes and its children which are already in document (using `AttributeDelta`)

`DeleteContent` needs only to filter nodes including children which are already in the document so this function is less complex.

Creating two functions instead of one combo function was easier because we could avoid the problem "Where to put this function?" :) But... I turned out there is another place that needs filtering content after some operations. I'm sure there will be more such a places in the near future.

The idea is to create `DataController#removeDisallowedAttributes` method which handles filtering nodes and its children no matter if nodes are in doc or not (optional batch parameter).

Maybe a better solution would be to filter automatically changed document fragment after changes but it's definitely much more complex.

oskarwrobel:

Feature: Introduced `DataController#removeDisallowedAttributes` method to filter out disallowed by schema attributes from given nodes. Used this method to filter nodes by `insertContent` and `deleteContent`. Closes #1120.

### DIFF
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -29,6 +29,8 @@ import deleteContent from './deletecontent';
 import modifySelection from './modifyselection';
 import getSelectedContent from './getselectedcontent';
 
+import Schema from '../model/schema';
+
 /**
  * Controller for the data pipeline. The data pipeline controls how data is retrieved from the document
  * and set inside it. Hence, the controller features two methods which allow to {@link ~DataController#get get}
@@ -274,7 +276,7 @@ export default class DataController {
 	 * @param {Object} options See {@link module:engine/controller/deletecontent~deleteContent}'s options.
 	 */
 	deleteContent( selection, batch, options ) {
-		deleteContent( selection, batch, options );
+		deleteContent( this, selection, batch, options );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


Reinmar:
Why did you propose this method on `DataController`? Only the most generic, wide-range, high-level functions should end up there. This function doesn't match here.

oskarwrobel:
We talked about it with @scofalik and decided that it might be a good place for this method.

Reinmar:
I'd put this method to `Schema`. So far, `Schema` only contains methods for checking things, but I don't see why its responsibility couldn't also be to implement some simple helpers for actually reacting to the checks. There's no better place for them at the moment and I don't want to be introducing "schema tools" to have there just this one tool (and the other tools in the `Schema` itself).

scofalik:
I proposed it in `DataController` because it is used in `DataController`-related modules and, really, if `hasContent` fits `DataController`, then everything fits `DataController`.

oskarwrobel:
Here is a version with `Schema#removeDisallowedAttributes` https://github.com/ckeditor/ckeditor5-engine/pull/1124.

Reinmar:
Can we delete the branch?

PR #1124, opened 2017-09-04
Introduced `Schema#removeDisallowedAttribute`

### LINKED ISSUE #1120: A way to filter disallowed attributes on given nodes

oskarwrobel:

Feature: Introduced `Schema#removeDisallowedAttributes` method to filter out disallowed by schema attributes from given nodes. Closes #1120.

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -418,8 +418,8 @@ export default class Schema {
 	/**
 	 * Removes disallowed by {@link module:engine/model/schema~Schema schema} attributes from given nodes.
 	 * When {@link module:engine/model/batch~Batch batch} parameter is provided then attributes will be removed
-	 * by creating {@link module:engine/model/delta/attributedelta~AttributeDelta attributeDeltas} otherwise
-	 * attributes will be removed directly from provided nodes using {@link module:engine/model/node~Node node} API.
+	 * using that batch, by creating {@link module:engine/model/delta/attributedelta~AttributeDelta attribute deltas}.
+	 * Otherwise, attributes will be removed directly from provided nodes using {@link module:engine/model/node~Node node} API.
 	 *
 	 * @param {Iterable.<module:engine/model/node~Node>} nodes Nodes that will be filtered.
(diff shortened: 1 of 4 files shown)


src/model/schema.js
```diff
@@ -416,6 +416,43 @@ export default class Schema {
...
+			const attributes = Array.from( node.getAttributeKeys() );
+			const queryPath = Schema._normalizeQueryPath( inside );
+
+			// When node with attributes is not allowed in current position.
+			if ( !this.check( { name, attributes, inside: queryPath } ) ) {
+				// Let's remove attributes one by one.
+				// TODO: this should be improved to check all combination of attributes.
+				for ( const attribute of node.getAttributeKeys() ) {
+					if ( !this.check( { name, attributes: attribute, inside: queryPath } ) ) {
+						if ( batch ) {
```
Reinmar: This is... sad :D Do I understand correctly that this will be simplified once we got the unified API?

scofalik: Well, yes.

oskarwrobel: I hope so :)

PR #1128, opened 2017-09-06
Schema should ignore attributes stored in elements by DocumentSelection

### LINKED ISSUE #1127: Schema should skip checking attributes stored by DocumentSelection
As we all know, `Schema` is not perfect.

There's no use doing big changes or refactoring if we plan to write it from scratch and we haven't really make any decision how it will look.

However, there is one improvement needed right one. If `DocumentSelection` stores an attribute in an element, that element will not be allowed anywhere, unless we explicitly call `Schema.allow()`.

This is a problem, because a `paragraph` element with `selection:bold` attribute is suddenly not allowed in `blockquote`.

To make it all more managable, the skip would only be implemented in "allow" part of algorithm. That is, you won't have to write things like

`schema.allow( { name: '$block', attributes: [ 'selection:bold' ], inside: 'blockquote' } )`

but you will be able to do

`schema.disallow( { name: '$block', attributes: [ 'selection:bold' ], inside: 'blockquote'  } )`;

It is better and easier to disallow one specific thing than writing all allow rules that would grow bigger with every added feature.

scofalik:

Fix: Fixed bug when `blockquote` could not be applied on an empty `paragraph` with a basic style (bold, etc.) active. Closes #1127.

-------------

### Additional information

`model.Schema` now will ignore attributes stored by `DocumentSelection` when checking if given element with attributes is allowed in given path. It will be still possible to explicitly disallow such elements.

### DIFF
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -392,6 +392,17 @@ export default class DocumentSelection extends Selection {
 	}
 
 	/**
+	 * Checks whether given attribute key is an attribute stored on an element.
+	 *
+	 * @protected
+	 * @param {String} key
+	 * @returns {Boolean}
+	 */
+	static _isStoreAttributeKey( key ) {
+		return key.indexOf( storePrefix ) == 0;
+	}
+
+	/**
 	 * Internal method for setting `DocumentSelection` attribute. Supports attribute priorities (through `directChange`
 	 * parameter).
 	 *
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -392,14 +392,14 @@ export default class DocumentSelection extends Selection {
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/model/documentselection.js
```diff
@@ -392,6 +392,17 @@ export default class DocumentSelection extends Selection {
...
 
 	/**
+	 * Checks whether given attribute key is an attribute stored on an element.
+	 *
+	 * @protected
+	 * @param {String} key
+	 * @returns {Boolean}
+	 */
+	static _isStoreAttributeKey( key ) {
+		return key.indexOf( storePrefix ) == 0;
```
Reinmar: Like `startsWith()`?

scofalik: Oh to the hell with those new methods!

PR #1131, opened 2017-09-07
Tweaked SplitDelta transformation to act better when deltas are undone

### LINKED ISSUE #1130: SplitDelta x SplitDelta transformation leads to incorrect undo
Upon further research, it appeared that `SplitDelta` x `SplitDelta` transformation needs further improvements. This time, undo is incorrect when splitting is at the same position.

Right now, in such scenario, no matter whether the delta is strong or not, the same result is provided. It is fine on surface - but when undoing such deltas will result in wrong content.

scofalik:

Internal: Changed `SplitDelta` x `SplitDelta` transformation when split is at same position so it works better with undo. Closes #1130.


scofalik:
I'll also add a test in undo repo.

PR #1133, opened 2017-09-08
Fixed: Improved range by merge delta transformation case.

### LINKED ISSUE #1132: Range#getTransformedByDelta returns incorrect results for uncommon MergeDeltas
A special case transformation of range by `MergeDelta` assumes that `MoveOperation` in `MergeDelta` starts at offset `0`.

This is not true for all `MoveOperation`s in `MergeDelta`. Sometimes, due to transformations, `MergeDelta` may contain `MoveOperation` that does not start from offset `0`.

scofalik:

Fix: Fixed a bug in `Range#getTransformedByDelta()` that caused editor crash after some `MergeDelta`s was transformed. Closes #1132.

### DIFF
--- a/src/model/range.js
+++ b/src/model/range.js
@@ -492,8 +492,13 @@ export default class Range {
 				// Collapsed range is in merged element, at the beginning or at the end of it.
 				// Without fix, the range would end up in the graveyard, together with removed element.
 				// <p>foo</p><p>[]bar</p> -> <p>foobar</p><p>[]</p> -> <p>foobar</p> -> <p>foo[]bar</p>
-				// <p>foo</p><p>bar[]</p>
-				// return [ new Range( targetPosition.getShiftedBy( this.start.offset ) ) ];
+				// <p>foo</p><p>bar[]</p> -> <p>foobar</p><p>[]</p> -> <p>foobar</p> -> <p>foobar[]</p>
+				//
+				// In most cases, `sourceRange.start.offset` for merge delta's move operation would be 0,
+				// so this formula might look overcomplicated.
+				// However in some scenarios, after operational transformation, move operation might not
+				// in fact start from 0 and we need to properly count new offset.
+				// https://github.com/ckeditor/ckeditor5-engine/pull/1133#issuecomment-329080668.
 				const offset = this.start.offset - sourceRange.start.offset;
 
 				return [ new Range( targetPosition.getShiftedBy( offset ) ) ];
--- a/src/model/range.js
+++ b/src/model/range.js
@@ -493,7 +493,10 @@ export default class Range {
 				// Without fix, the range would end up in the graveyard, together with removed element.
 				// <p>foo</p><p>[]bar</p> -> <p>foobar</p><p>[]</p> -> <p>foobar</p> -> <p>foo[]bar</p>
 				// <p>foo</p><p>bar[]</p>
-				return [ new Range( targetPosition.getShiftedBy( this.start.offset ) ) ];
+				// return [ new Range( targetPosition.getShiftedBy( this.start.offset ) ) ];
+				const offset = this.start.offset - sourceRange.start.offset;
+
+				return [ new Range( targetPosition.getShiftedBy( offset ) ) ];
 			}
 			//
 			// Other edge cases:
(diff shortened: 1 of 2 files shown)


src/model/range.js
```diff
@@ -493,7 +493,10 @@ export default class Range {
 				// Without fix, the range would end up in the graveyard, together with removed element.
 				// <p>foo</p><p>[]bar</p> -> <p>foobar</p><p>[]</p> -> <p>foobar</p> -> <p>foo[]bar</p>
 				// <p>foo</p><p>bar[]</p>
-				return [ new Range( targetPosition.getShiftedBy( this.start.offset ) ) ];
+				// return [ new Range( targetPosition.getShiftedBy( this.start.offset ) ) ];
```
Reinmar: ⬆️

Reinmar:
I hope to understand at least this OT-related PR but you haven't left a comment why is this change needed. How can merge happen on a non-zero index?

src/model/range.js
```diff
@@ -493,7 +493,10 @@ export default class Range {
 				// Without fix, the range would end up in the graveyard, together with removed element.
 				// <p>foo</p><p>[]bar</p> -> <p>foobar</p><p>[]</p> -> <p>foobar</p> -> <p>foo[]bar</p>
 				// <p>foo</p><p>bar[]</p>
```
Reinmar: What does this comment mean? That last bit?

tests/model/range.js
```diff
@@ -1056,6 +1059,22 @@ describe( 'Range', () => {
 				expect( transformed[ 0 ].start.path ).to.deep.equal( [ 0, 0, 1 ] );
 				expect( transformed[ 0 ].end.path ).to.deep.equal( [ 0, 1, 1 ] );
 			} );
+
+			// #1132
+			it( 'merge delta has move operation that does not start from offset 0', () => {
```
Reinmar: Such tests without initial comments like these in lines 1048-1051 are completely unreadable for me (at least ;P)

src/model/range.js
scofalik: It is the same as the above, just the input data is same. I was hoping that the explanation above will be enough to understand that these are similar cases with different input.

scofalik: Well maybe output is not exactly the same because range ends up in a different spot. I'll write whole example.

scofalik:
> I hope to understand at least this OT-related PR but you haven't left a comment why is this change needed. How can merge happen on a non-zero index?

Yes it is because of OT but the example is complicated so I didn't want to leave long comment about it that would be not understood by anyone who does not know how transformations work.

Basically this happens:
1. `SplitDelta` A is transformed by `SplitDelta` B and they both split in same element, but on different positions. This is a collaborative editing scenario.
2. Both `SplitDelta`s are undone.
3. One of `MoveOperation` in reversed `SplitDelta` (`MergeDelta`) becomes something like:
`move [ a, x ] - [ a, y ] -> [ a, x ]` where `x` is not `0`. Basically, this move operation does not change anything as we move a range to the start of it. But because of special case in `Range` transformation method, it threw.

Now I understand that point 3. raises two concerns:
1. If this move operation does not do anything, why it is not a `NoOperation`?
2. If this move operation does not do anything, why we don't just skip such operations in `Range#getTransformedByDelta()`.

Here are some answers:
1. We already have problems in OT because of transforming operations to `NoOperation` when they seem to not do anything. This is a recurring theme that we already fixed on multiple occasions in multiple transformation cases. For sure we cannot transform to `NoOperation` during the transformation process. Maybe we could transform to `NoOperation` at the end when we already have results and are ready to apply them. But this is maybe.

2. We could just check if move operation does anything relevant and return if not. But at this point I cannot guarantee that there is no way to have such operation generated through OT:
`move [ a, x ] - [ a, y ] -> [ b, z ]` where `x` is not `0`.
So to prevent future errors I wanted to fix how `offset` is calculated because it seemed wrong. The new `offset` formula is same for `sourceRange.start.offset == 0` but is also correct for `sourceRange.start.offset != 0`.

scofalik:
What's up with this PR?

Reinmar:
W8ing for me having time to check it.

PR #1136, opened 2017-09-11
Allow model.Range to be passed to model.Document#getNearestSelectionRange

### LINKED ISSUE #1135: Allow model.Range to be passed to model.Document#getNearestSelectionRange
At the moment, there is no easy way to obtain a correct selection range basing on `model.Range`. There is `model.Document#getNearestSelectionRange` but it only takes `model.Position`.

A common scenario where we would like to get a correct selection range from given range is selection whole content: `[<paragraph>foo</paragraph><paragraph>bar</paragraph>]` -> `<paragraph>[foo</paragraph><paragraph>bar]</paragraph>`. In this case it would be insufficient to just pass the ranges start or end position because that would totally change the range.

This will be useful in undo where incorrect ranges may be created when restoring selection.

scofalik:

Feature: `model.Document#getNearestSelectionRange` now accepts also `model.Range`. Additionally, added `model.DocumentSelection#hasOwnRange` property. Closes #1135.

### DIFF
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -286,25 +286,49 @@ export default class Document {
 	}
 
 	/**
-	 * Basing on given `position`, finds and returns a {@link module:engine/model/range~Range Range} instance that is
-	 * nearest to that `position` and is a correct range for selection.
+	 * Basing on given {@link module:engine/model/position~Position} or {@link module:engine/model/range~Range} instance,
+	 * returns nearest {@link module:engine/model/range~Range Range} that is a correct range for selection.
 	 *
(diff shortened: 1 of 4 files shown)


scofalik:
After some testing it appears that `hasOwnRange` is enough to fix a bug in undo. I was under impression that we will have to fix the range (that was the first fix I've made and it worked) but it appears that it's enough to use `hasOwnRange` when storing batch in undo (and do not store selection ranges if they are default ranges).

Because of this I don't want to introduce unnecessary complication to code. We can leave this branch in case of we needed this kind of functionality in the future. I'll create a new issue about `hasOwnRange` and a proper PR.

PR #1138, opened 2017-09-11
Added model.DocumentSelection#hasOwnRange.

### LINKED ISSUE #1137: Introduce a way to check if DocumentSelection has only a default range
I propose introducing `DocumentSelection#hasOwnRange` as a way to check whether ranges set in selection are set directly (for example via `.addRange()` or `.setRanges()`) or is it just the default range.

This is needed to fix a bug in undo selection restoring.

scofalik:

Feature: Introduced `model.DocumentSelection#hasOwnRange`. Closes #1137.

### DIFF
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -120,6 +120,17 @@ export default class DocumentSelection extends Selection {
 	}
 
 	/**
+	 * Describes whether `DocumentSelection` has own range(s) set, or if it is defaulted to
+	 * {@link module:engine/model/document~Document#_getDefaultRange document's default range}.
+	 *
+	 * @readonly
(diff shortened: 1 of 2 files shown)


Reinmar:
OMG, like `hasOwnProperty()` :D. Do we have to use it in all for-of loops now? :trollface:

Reinmar:
OK, by my above comment you can clearly see that I did not understand what this property is all about. Which means that it might be worth rethinking its name.

Reinmar:
Nothing better came to my mind so let it be `hasOwnRanges`.

scofalik:
My other ideas were `isSet` or `isDefault`.

PR #1141, opened 2017-09-12
Nodes are removed from their old parent when they are appended to new parent

### LINKED ISSUE #1139: Children nodes are inserted to new parent without first removing them from their old parent
I spotted a bug that crashed editor during rendering. It appeared that `Renderer` wants to re-render a part of a tree that is detached from main tree. What was weird about that tree was that detached-root had a child but that child's `parent` property was set to `null`. Similarly, that child also had a child which also had `parent` set to `null`.

After debugging it appeared that in `view.Writer` some nodes were supposed to be moved from one node to another, however those nodes were not removed from their original parent before appending to new parent, which caused this bug.

scofalik:

Fix: `view` and `model` nodes will now be removed from their old parent when they are added to a new parent to prevent having same node on multiple elements' children lists. Closes #1139.

BREAKING CHANGE: If you will append `view` or `model` node that already has a parent, that node will be first removed from it's parent. This is important if you iterate through element's children and they are moved during that iteration. In that case it's safest to cache the element's children in an array.

### DIFF
--- a/src/dev-utils/view.js
+++ b/src/dev-utils/view.js
@@ -852,6 +852,8 @@ function _convertViewElements( rootNode ) {
 		const convertedElement = rootNode.is( 'documentFragment' ) ? new ViewDocumentFragment() : _convertElement( rootNode );
 
 		// Convert all child nodes.
+		// Cache the nodes in array. Otherwise, we would skip some nodes because during iteration we move nodes
+		// from `rootNode` to `convertedElement`. This would interfere with iteration.
 		for ( const child of [ ...rootNode.getChildren() ] ) {
 			if ( convertedElement.is( 'emptyElement' ) ) {
 				throw new Error( 'Parse error - cannot parse inside EmptyElement.' );
--- a/src/dev-utils/view.js
+++ b/src/dev-utils/view.js
@@ -852,7 +852,7 @@ function _convertViewElements( rootNode ) {
 		const convertedElement = rootNode.is( 'documentFragment' ) ? new ViewDocumentFragment() : _convertElement( rootNode );
 
 		// Convert all child nodes.
-		for ( const child of rootNode.getChildren() ) {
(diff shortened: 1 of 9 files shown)


scofalik:
Needed to fix a test in typing https://github.com/ckeditor/ckeditor5-typing/pull/121

src/dev-utils/view.js
```diff
@@ -852,7 +852,7 @@ function _convertViewElements( rootNode ) {
 		const convertedElement = rootNode.is( 'documentFragment' ) ? new ViewDocumentFragment() : _convertElement( rootNode );
 
 		// Convert all child nodes.
-		for ( const child of rootNode.getChildren() ) {
+		for ( const child of [ ...rootNode.getChildren() ] ) {
```
Reinmar: COMMMWENTR~!!#@

src/model/documentfragment.js
```diff
@@ -235,6 +235,11 @@ export default class DocumentFragment {
 		nodes = normalize( nodes );
 
 		for ( const node of nodes ) {
+			// If node that is being added to this element is already inside another element, first remove it from the old parent.
+			if ( node.parent !== null ) {
```
Reinmar: I'm hesitating whether I'd rather like `if ( node.parent )` or `!== null` ;| Dunno

src/dev-utils/view.js
scofalik: Why so serious?

scofalik: Fixed

src/model/documentfragment.js
scofalik: `!== null` is easier to find when searching codebase

scofalik:
What's up with this PR?

Reinmar:
W8ing for me having time to check it.

PR #1147, opened 2017-09-20
Make setting "undo mode" in transformation more explicit

### LINKED ISSUE #1146: Undo mode is not recognized properly during deltas transformations
We have `context.aWasUndone` and `context.bWasUndone` and they were used to check whether transformation is in "undo mode" and if special cases should be applied. However this is wrong in some situations (during selective undo). Transformation algorithms should recognize that they are in undo mode straight away and always when we are transforming for undo. This should not been based on history and whether some delta was undone or not.

scofalik:

Internal: Make setting "undo mode" in transformation more explicit. Closes #1146.

### DIFF
--- a/src/model/delta/basic-transformations.js
+++ b/src/model/delta/basic-transformations.js
@@ -56,7 +56,7 @@ addTransformationCase( AttributeDelta, SplitDelta, ( a, b, context ) => {
 		return defaultTransform( a, b, context );
 	}
 
-	const undoMode = context.aWasUndone || context.bWasUndone;
+	const undoMode = context.undoMode;
 	const splitPosition = new Position( b.position.root, b.position.path.slice( 0, -1 ) );
 
 	const deltas = defaultTransform( a, b, context );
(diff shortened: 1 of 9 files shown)


Reinmar:
If this is fixing the selective undo scenario (as I understand from the ticket, which has no labels, so I don't know if it's a bug or not), could we have integration tests for that?

Reinmar:
Bump?

scofalik:
An old issue ;). But I'll try to remind and come up with a test case.

scofalik:
So I've added an integration test here: https://github.com/ckeditor/ckeditor5-undo/pull/74.

To sum up: the transformation algorithm in deltas transformation case for `SplitDelta` x `SplitDelta` is incorrect for a special case, when both deltas try to return the same node from graveyard. This should happen only when undo is used in a specific scenario. Since the transformation should not be fired when undo is used I'd rather make sure that it is "turned off" rather than try to fix the algorithm.

Reinmar:
I guess we need to update the code added in https://github.com/ckeditor/ckeditor5-engine/pull/1150 to work with this changes. I'll check what I can do myself.

PR #1149, opened 2017-09-24
Fixed invalid error docs

pomek:

Docs: Fixed invalid error docs. Read more https://github.com/ckeditor/ckeditor5-dev/issues/288.


src/model/position.js
```diff
@@ -47,7 +47,7 @@ export default class Position {
 			/**
 			 * Position root invalid.
 			 *
-			 * @error position-root-invalid.
+			 * @error model-position-root-invalid.
```
Reinmar: There should be no period after the name.

PR #1150, opened 2017-09-29
MergeDelta x MoveDelta transformation case should not be applied in undo mode

scofalik:

Fix: Undo did no changes instead of merging elements, in a scenario when an element was split and then the "new" element was removed.

---

### Additional information

Originally reported here: https://github.com/ckeditor/ckeditor5-undo/issues/65#issuecomment-323682195

A PR in undo with a test will follow.

### DIFF
--- a/src/model/delta/basic-transformations.js
+++ b/src/model/delta/basic-transformations.js
@@ -141,8 +141,10 @@ addTransformationCase( MarkerDelta, RenameDelta, transformMarkerDelta );
 
 // Add special case for MoveDelta x MergeDelta transformation.
 addTransformationCase( MoveDelta, MergeDelta, ( a, b, context ) => {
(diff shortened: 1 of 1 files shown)


scofalik:
PR in undo: https://github.com/ckeditor/ckeditor5-undo/pull/75

Reinmar:
I'll pretend I didn't notice that you haven't written some test here :D But CC is ok and undo verifies this change, so ok.

scofalik:
Old tests verify that there's no regression ;)

Reinmar:
Regression testing is one thing. But why doing any change if all tests pass with the old code (which is true if no new tests were added :P). Which also means that there's no regression testing for the change in the future (you can break the feature and nothing's gonna tell you that).

scofalik:
I was just mocking you :). I understand what you meant with your earlier comment. But I created a test in undo because it makes more sense to test it in undo.

PR #1151, opened 2017-09-29
View selection correctness was not checked and fixed when view selection was converted to model

scofalik:

Fix: Incorrect view selection was converted to incorrect model selection, causing bugs when DOM selection was placed in a wrong place. Closes https://github.com/ckeditor/ckeditor5/issues/562.

### DIFF
--- a/src/conversion/view-selection-to-model-converters.js
+++ b/src/conversion/view-selection-to-model-converters.js
@@ -11,6 +11,7 @@
  */
 
 import ModelSelection from '../model/selection';
+import ModelRange from '../model/range';
 
 /**
  * Function factory, creates a callback function which converts a {@link module:engine/view/selection~Selection view selection} taken
@@ -34,7 +35,12 @@ export function convertSelectionChange( modelDocument, mapper ) {
 		const ranges = [];
 
 		for ( const viewRange of viewSelection.getRanges() ) {
-			ranges.push( mapper.toModelRange( viewRange ) );
+			const mappedRange = mapper.toModelRange( viewRange );
+
+			const correctModelRangeStart = modelDocument.getNearestSelectionRange( mappedRange.start ).start;
+			const correctModelRangeEnd = modelDocument.getNearestSelectionRange( mappedRange.end ).end;
+
+			ranges.push( new ModelRange( correctModelRangeStart, correctModelRangeEnd ) );
 		}
 
 		modelSelection.setRanges( ranges, viewSelection.isBackward );
--- a/src/conversion/view-selection-to-model-converters.js
+++ b/src/conversion/view-selection-to-model-converters.js
@@ -35,12 +35,27 @@ export function convertSelectionChange( modelDocument, mapper ) {
 		const ranges = [];
 
 		for ( const viewRange of viewSelection.getRanges() ) {
+			// Get a range from view selection and map it to model.
 			const mappedRange = mapper.toModelRange( viewRange );
 
+			// Check if model range mapped from view is correct.
+			// To do so, find nearest selection range near start and end of mapped model range.
+			//
+			// Note, that if `mappedRange.start` or `mappedRange.end` are in correct position,
+			// a collapsed range at those position will be returned.
+			//
+			// Example #1 - correct collapsed selection in the middle of text:
+			// <paragraph>Fo[]obar</paragraph> -> <paragraph>Fo{}[]obar</paragraph> -> <paragraph>Fo{]obar</paragraph>
+			//
+			// Example #2 - incorrect non-collapsed range:
+			// [<paragraph>Foobar</paragraph>] -> <paragraph>[]Foobar{}</paragraph> -> <paragraph>[Foobar}</paragraph>
+			//
+			// Example #3 - incorrect collapsed range near widget:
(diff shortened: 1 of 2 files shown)


Reinmar:
This doesn't look like the type of changes we'd like to merge just before a release. Let's get back to this after 1.0.0 alpha.

Reinmar:
Besides, I'd like such a change to be accompanied with proper tests. Automated and manual ones.

Reinmar:
cc @pjasiun – I guess you'd like to see this.

src/conversion/view-selection-to-model-converters.js
```diff
@@ -34,7 +35,12 @@ export function convertSelectionChange( modelDocument, mapper ) {
 		const ranges = [];
 
 		for ( const viewRange of viewSelection.getRanges() ) {
-			ranges.push( mapper.toModelRange( viewRange ) );
+			const mappedRange = mapper.toModelRange( viewRange );
+
+			const correctModelRangeStart = modelDocument.getNearestSelectionRange( mappedRange.start ).start;
+			const correctModelRangeEnd = modelDocument.getNearestSelectionRange( mappedRange.end ).end;
```
pjasiun: I am surprised that we did not have this before. We should use view-to-model conversion to fix wrong selection. However, we need to ensure that `getNearestSelectionRange` do a good job here and do no break anything. Also, this code is not very readable and need docs.

Reinmar:
This PR may fix https://github.com/ckeditor/ckeditor5-heading/issues/87 too. Let's verify this.

src/conversion/view-selection-to-model-converters.js
scofalik: What do you mean by "we need to ensure that `getNearestSelectionRange` do a good job here"? This is not a new method and we already use it in couple of places. Seems to work fine :). The method - in theory - is very simple. It takes given position and looks for the nearest position (left and right) where a model selection can be placed. As I said, we already use this method and it solves our problems.

pjasiun: I mean that now it will be called very often: whenever one type, move the selection, do... whatever. If this method will be called that ofter we should review it better than before ;) We need to ensure that it will not break any scenario. Also, we need to think about the efficiency.

Reinmar: `getNearestSelectionRange()` may be used in a couple of places, but here we use it in a bit special and new context. And critical for things like stability with various features and performance.

Reinmar:
I wonder – could we also consider fixing https://github.com/ckeditor/ckeditor5-engine/issues/1156 together with this to test it once and not 10 times?

src/conversion/view-selection-to-model-converters.js
scofalik: I understand your worries. Let's break it down.

Correctness. First of all, `getNearestSelectionRanges()` has quite a few tests written already. You can check them. I could have written a few tests for the conversion itself, but it would be mostly repeating those tests. In the PR I propose to use `getNearestSelectionRange()`  on a range instead of position. I've already wanted to introduce this to the method itself when solving another issue, but ultimately it was not needed. When you think about this solution it makes sense to take boundaries, expand them to ranges and then take appropriate boundaries. For example when selection is next to widget:
`foo<widget />[]` -> `foo{[<widget />]}` -> `foo{<widget />]` -> `foo[<widget />]`. This also works nicely for non-collapsed ranges. I haven't thought of all the cases, of course, but in theory it seems to be a good idea. I could write some test for non-collapsed ranges.

Efficiency. We have to do _any_ fixing. This is a fact. So we always have to check whether the selection is correct. I don't know of any correct and faster way than checking the schema. That's how we do it. As you can see, `getNearestSelectionRange()` has a check at the beginning which checks whether given position is correct in the first place. This is a minimal amount of processing that we have to do no matter what solution we choose. So we will do any extra processing only if the selection was in the wrong place. Which should not happen that often and should not have negative impact on editor smoothness.

Other worries. The only thing I am worried ATM is that we can get into some kind of stupid infinite loop. That deserves a second thought and a test. I am worried that this might happen if, for example there is no correct selection (editor is empty) or the browser tries to fix the selection in DOM all the time (but this is handled in another place).

scofalik:
Can we get back to this PR now after 1.0.0 alpha has been published? I added a lengthy comment, hopefully answering your worries.

scofalik:
I fixed a situation when there is no correct model position for selection.

scofalik:
I've updated comments in the converter.

Reinmar:
> Can we get back to this PR now after 1.0.0 alpha has been published? I added a lengthy comment, hopefully answering your worries.

I've read the comment – it clarifies the situation a lot. But I didn't manage to reply yet. Please, ping me again in a couple of days if I won't do that in the meantime :P.

Reinmar:
This doesn't look like the type of changes we'd like to merge just before a release. Let's get back to this after 1.0.0-alpha.2.

Reinmar:
Thinking about tables and images made me realise that the same algoritm perhaps will need to handle selections which cross editable element's boundaries. Those selections (which start in e.g. a caption and end outside) need to be fixed too.

Reinmar:
I'll just R- it so I know that it's not waiting for my review :D

Reinmar:
Replaced by https://github.com/ckeditor/ckeditor5-engine/pull/1431.

PR #1163, opened 2017-10-11
Fixed: Empty element should be merged instead of removed in deleteContents

### LINKED ISSUE #1161: Using RemoveDelta in deleteContent leads to wrong results in collaboration
In `deleteContent` we make an assumption, that if the merged element is empty, _maybe_ it is better to use `batch.remove()` instead of merging it.

It is all fine, until you realise that it is incorrect to take action basing on current model state. Here's a counter example:

1. Editor data is: `<p>foo</p><p></p><p>bar</p>`.
2. Second (empty) paragraph is "merged" with first paragraph.
3. Simultaneously, third paragraph ("bar") is merged with second paragraph.

If second paragraph is removed (instead of merged), then the data from third (merged) paragraph will be "moved" to the second paragraph and removed with it.

scofalik:

Fix: `deleteContent` algorithm will use merging to "remove" empty element which will ensure a better conflict resolution on collaborative editing. Closes #1161.


Reinmar:
Eeeem... as you wrote this is just a workaround and it's a workaround for a collab scenario where three blocks are merged simultaneously by two users. Additionally, this scenario leads to an incorrect data, not to a crash.

Taken that, I'm against doing this change. Avoiding using the remove op may help in the meantime but will need to be reverted later on. Unless, of course, you think that such algorithms should use merge instead of remove?

scofalik:
I proposed this PR anyway because this issue is:
- not difficult to reproduce,
- gives a very bad result,
- we won't fix it in transformations soon (requires refactor).

Also, it is not 100% clear that this should be a remove or merge. Both solutions are reasonable. I wanted to fix what can be easily fixed so we won't be stuck with a bug for a couple of months.

Reinmar:
OK, then let's change it. I can see that the code actually got simpler and after looking at the scenario again I realised that it makes sense to use merge there.

PR #1165, opened 2017-10-11
removeHighlight method now accepts highlight id

### LINKED ISSUE #1164: Removing highlight from element should only require highlight id
Elements can provide custom properties `setHighlight` and `removeHighlight` to handle highlighting by themselves. Currently, converters pass `HighlightDescriptor` object to both functions. This works well when converters are adding and removing highlights. When there is a separate logic to remove the highlight it must provide whole descriptor to `removeHighlight` method which is not so convenient. Maybe it would be better if:
 - `setHighlight` will accept full descriptor to set the highlight (as it is now),
 - `removeHighlight` accepts only the descriptor id to be removed.

szymonkups:

Other: The `removeHighlight()` function now accepts descriptor id instead of a `HighlightDescriptor` object. Closes #1164.

BREAKING CHANGE: The `removeHighlight()` function now accepts descriptor id instead of  a `HighlightDescriptor` object. See #1164.

---

### Additional information

Updates to handling highlighting in widgets: https://github.com/ckeditor/ckeditor5-widget/tree/t/ckeditor5-engine/1164.

### DIFF
--- a/src/conversion/model-to-view-converters.js
+++ b/src/conversion/model-to-view-converters.js
@@ -445,10 +445,9 @@ export function highlightText( highlightDescriptor ) {
 
 /**
  * Function factory, creates converter that converts all elements inside marker's range. Converter checks if element has
- * functions stored under `addHighlight` and `removeHighlight` custom properties and calls them passing
- * {@link module:engine/conversion/model-to-view-converters~HighlightDescriptor}. In such case converter will consume
- * all element's children, assuming that they were handled by element itself. If highlight descriptor will not provide
- * priority, priority `10` will be used as default, to be compliant with
+ * functions stored under `addHighlight` and `removeHighlight` custom properties and calls them when highlight should be added
+ * or removed. In such case converter will consume all element's children, assuming that they were handled by element itself.
+ * If highlight descriptor will not provide priority, priority `10` will be used as default, to be compliant with
  * {@link module:engine/conversion/model-to-view-converters~highlightText} method which uses default priority of
  * {@link module:engine/view/attributeelement~AttributeElement}.
  *
@@ -496,7 +495,7 @@ export function highlightElement( highlightDescriptor ) {
 				consumable.consume( value.item, evt.name );
 			}
 
-			viewElement.getCustomProperty( highlightHandlingMethod )( viewElement, descriptor );
+			viewElement.getCustomProperty( highlightHandlingMethod )( viewElement, addMarker ? descriptor : descriptor.id );
 		}
 	};
 }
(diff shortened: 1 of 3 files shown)


src/conversion/model-to-view-converters.js
```diff
@@ -445,10 +445,9 @@ export function highlightText( highlightDescriptor ) {
 
 /**
  * Function factory, creates converter that converts all elements inside marker's range. Converter checks if element has
```
Reinmar: > Converter function factory. Creates a function which applies the marker's highlight to all elements inside a marker's range. 

And then:

> The converter checks if an element has the `addHighlight` and `removeHighlight` functions stored as {@link TODO custom properties} and if so use them to apply the highlight...

Something like that.

src/conversion/model-to-view-converters.js
```diff
@@ -640,8 +639,11 @@ class HighlightAttributeElement extends ViewAttributeElement {
  * described by this object.
  *
  * Each element can handle displaying highlight separately by providing `addHighlight` and `removeHighlight` custom
- * properties. Those properties are passed `HighlightDescriptor` object upon conversion and should use it to
- * change the element.
+ * properties:
+ *  * `HighlightDescriptor` is passed to `addHighlight` function upon conversion and should be used to apply highlight to
```
Reinmar: > the `addHighlight()` function

src/conversion/model-to-view-converters.js
```diff
@@ -640,8 +639,11 @@ class HighlightAttributeElement extends ViewAttributeElement {
  * described by this object.
  *
  * Each element can handle displaying highlight separately by providing `addHighlight` and `removeHighlight` custom
```
Reinmar: > the highlight

src/conversion/model-to-view-converters.js
```diff
@@ -640,8 +639,11 @@ class HighlightAttributeElement extends ViewAttributeElement {
...
- * properties. Those properties are passed `HighlightDescriptor` object upon conversion and should use it to
- * change the element.
+ * properties:
```
Reinmar: In the docs above add prepositions before "highlight" and "content highlight" cause they are used as nouns.

Reinmar:
I don't know this code at all so I mainly focused on the docs. I leave the rest to you, @scofalik.

scofalik:
Please create a PR out of the changes in widgets repository so I can close this PR and those changes together.

szymonkups:
Done: https://github.com/ckeditor/ckeditor5-widget/pull/22

PR #1175, opened 2017-10-25
T/897: Make Position/Range immutable.

### LINKED ISSUE #897: Reduce memory pressure by stopping cloning all the time
Disclaimer: I didn't identify any performance issues yet. However, this is a low-level aspect of our engine, a code which is used everywhere millions of times, so we can be sure that it affects the performance. Also, since it touches archtectural decisions, it's a thing which will be hard to change later on. So let's talk about it now.

---

This topic started in https://github.com/ckeditor/ckeditor5-engine/pull/896/files#r108710032.

The code I commented on looked like this:

```js
const newSelection = Selection.createFromSelection( selection );
const ranges = newSelection.getRanges();

let trimmedRanges = [];

for ( let range of ranges ) {
	trimmedRanges.push( range.getTrimmed() );
}

newSelection.setRanges( trimmedRanges );
```

My comments:

> `createFromSelection()` gives us `setTo()` call which e.g. copies fake selection's stuff so that's good. What I like less is the fact that we are copying the ranges here (`getRanges()` copies them) to copy them again when doing `setRanges()` to copy them again 3 times on another `getRanges()`, `getTrimmed()`, `setRanges()`. This means 5 times for every selection and we need to trim two selections...

and:

> Let me add that this is all due to our lack of trust that no one will touch the ranges and positions.
> 
> In a perfect immutable world, we'd only need to copy new instances on `getTrimmed()`. No other method would need to do that because no other method change ranges. And I think that we must go into this direction. Otherwise, we create a huge memory pressure.

PS. Mind what happens with positions. They are cloned all the time too ;|

---

## Proposal

We introduced cloning on "input" like `setRanges()` and output like `getRanges()` because we were worried that someone might change one position somewhere and affect a couple of ranges and selection classes somewhere else. Those issues are usually hard to track.

While I'm still sure that we should worry about such issues I think that once we made ranges and positions nearly immutable (which happened later AFAIR and is still not enforced in the entire code) we secured the stability pretty well. The point is that we can secure the code by either breaking references everywhere or making things completely immutable. We former means loooots of cloning, so let's thing about the latter.

There are just 4 things which one can break:

* `range.start`, `range.end`,
* `position.root`, `position.path`.

If we made these 4 things totally read-only, we would be safe even without the cloning. Right?

So, how to make them really read-only? With the first 3 of them it's simple – just make them getters and keep the values under private properties. The path is a bigger issue because it can be modified after retrieving it. Therefore, I think that it can be a getter too but it'd need to clone the array to break the reference. Keep in mind that it's better to clone just this array (if one specifically accessed `position.path` – the other getters and methods can access the private `_path` property) rather than clone all the ranges and positions all the time (which clone the path too).

It seems that with this plan we should be safe. The only thing I'm not entirely sure about is whether making those 4 public properties getters which return private properties will not have negative performance impact too. Or rather – how strong it will be (if it exceeds cloning overhead, it'd be a pointless change :D). Unfortunately, without really making this move we won't learn that.

jodator:

Other: Make `Position` and `Range` immutable in model and view. Closes #897.

BREAKING CHANGE: `model.Position#offset` and `view.Position#offset` now have only a getter. To modify offset use `Postion#getShiftedBy()` or `Position#getShiftedTo()`.

BREAKING CHANGE: `model.Position#path` and `view.Position#path` now have only a getter and are immutable.

BREAKING CHANGE: `model.Range#start` and `model.Range#end` properties now have only a getter.

BREAKING CHANGE: `view.Range#start` and `view.Range#end` properties now have only a getter.

---

### Additional information

This PR required minor changes in other repos (centralized in: https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-engine/897)
- https://github.com/ckeditor/ckeditor5-list/tree/t/ckeditor5-engine/897
- https://github.com/ckeditor/ckeditor5-ui/tree/t/ckeditor5-engine/897
- https://github.com/ckeditor/ckeditor5-undo/tree/t/ckeditor5-engine/897

Merge commit message for related branches:

```
Internal: Aligned Positions and Ranges usage to the latest API of the framework (see ckeditor/ckeditor5-engine#897).
```

Other:
- Main idea behid this PR is to reduce the need for creating new `Position`/`Range` objects when using it in other methods. Now those object can be safely passed around without side effects.
- The `remove()` method in `view/writer` still does something with passed

### DIFF
--- a/src/dev-utils/view.js
+++ b/src/dev-utils/view.js
@@ -541,8 +541,7 @@ class RangeParser {
 				throw new Error( `Parse error - end of range was found '${ item.bracket }' but range was not started before.` );
 			}
 
-			// When second start of range is found when one is already opened - selection does not allow intersecting
-			// ranges.
+			// When second start of range is found when one is already opened - selection does not allow intersecting ranges.
 			if ( range && ( item.bracket == ELEMENT_RANGE_START_TOKEN || item.bracket == TEXT_RANGE_START_TOKEN ) ) {
 				throw new Error( `Parse error - start of range was found '${ item.bracket }' but one range is already started.` );
 			}
--- a/src/dev-utils/view.js
+++ b/src/dev-utils/view.js
@@ -550,7 +550,7 @@ class RangeParser {
 			if ( item.bracket == ELEMENT_RANGE_START_TOKEN || item.bracket == TEXT_RANGE_START_TOKEN ) {
 				range = new Range( item.position, item.position );
 			} else {
-				range.end = item.position;
+				range = new Range( range.start, item.position );
 				ranges.push( range );
 				range = null;
 			}
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -179,25 +179,29 @@ const ot = {
 				// Take the start and the end of the range and transform them by deletion of moved nodes.
 				// Note that if rangeB was inside AttributeOperation range, only difference.end will be transformed.
 				// This nicely covers the joining simplification we did in the previous step.
-				difference.start = difference.start._getTransformedByDeletion( b.sourcePosition, b.howMany );
-				difference.end = difference.end._getTransformedByDeletion( b.sourcePosition, b.howMany );
+				const range = new Range(
+					difference.start._getTransformedByDeletion( b.sourcePosition, b.howMany ),
+					difference.end._getTransformedByDeletion( b.sourcePosition, b.howMany )
+				);
 
 				// MoveOperation pastes nodes into target position. We acknowledge this by proper transformation.
 				// Note that since we operate on transformed difference range, we should transform by
 				// previously transformed target position.
 				// Note that we do not use Position._getTransformedByMove on range boundaries because we need to
 				// transform by insertion a range as a whole, since newTargetPosition might be inside that range.
-				ranges = difference._getTransformedByInsertion( b.getMovedRangeStart(), b.howMany, true, false ).reverse();
+				ranges = range._getTransformedByInsertion( b.getMovedRangeStart(), b.howMany, true, false ).reverse();
 			}
 
 			if ( common !== null ) {
 				// Here we do not need to worry that newTargetPosition is inside moved range, because that
 				// would mean that the MoveOperation targets into itself, and that is incorrect operation.
 				// Instead, we calculate the new position of that part of original range.
-				common.start = common.start._getCombined( b.sourcePosition, b.getMovedRangeStart() );
-				common.end = common.end._getCombined( b.sourcePosition, b.getMovedRangeStart() );
+				const range = new Range(
+					common.start._getCombined( b.sourcePosition, b.getMovedRangeStart() ),
+					common.end._getCombined( b.sourcePosition, b.getMovedRangeStart() )
+				);
 
-				ranges.push( common );
+				ranges.push( range );
 			}
 
 			// Map transformed range(s) to operations and return them.
@@ -376,7 +380,7 @@ const ot = {
 			// Setting and evaluating some variables that will be used in special cases and default algorithm.
 			//
 			// Create ranges from `MoveOperations` properties.
-			const rangeA = Range.createFromPositionAndShift( a.sourcePosition, a.howMany );
+			let rangeA = Range.createFromPositionAndShift( a.sourcePosition, a.howMany );
 			const rangeB = Range.createFromPositionAndShift( b.sourcePosition, b.howMany );
 
 			// Assign `context.isStrong` to a different variable, because the value may change during execution of
@@ -428,8 +432,10 @@ const ot = {
 			if ( bTargetsToA && rangeA.containsRange( rangeB, true ) ) {
 				// There is a mini-special case here, where `rangeB` is on other level than `rangeA`. That's why
 				// we need to transform `a` operation anyway.
-				rangeA.start = rangeA.start._getTransformedByMove( b.sourcePosition, b.targetPosition, b.howMany, !includeB );
-				rangeA.end = rangeA.end._getTransformedByMove( b.sourcePosition, b.targetPosition, b.howMany, includeB );
+				rangeA = new Range(
+					rangeA.start._getTransformedByMove( b.sourcePosition, b.targetPosition, b.howMany, !includeB ),
+					rangeA.end._getTransformedByMove( b.sourcePosition, b.targetPosition, b.howMany, includeB )
+				);
 
 				return makeMoveOperationsFromRanges( [ rangeA ], newTargetPosition, a );
 			}
@@ -444,8 +450,10 @@ const ot = {
 			if ( aTargetsToB && rangeB.containsRange( rangeA, true ) ) {
 				// `a` operation is "moved together" with `b` operation.
 				// Here, just move `rangeA` "inside" `rangeB`.
-				rangeA.start = rangeA.start._getCombined( b.sourcePosition, b.getMovedRangeStart() );
-				rangeA.end = rangeA.end._getCombined( b.sourcePosition, b.getMovedRangeStart() );
+				rangeA = new Range(
+					rangeA.start._getCombined( b.sourcePosition, b.getMovedRangeStart() ),
+					rangeA.end._getCombined( b.sourcePosition, b.getMovedRangeStart() )
+				);
 
 				return makeMoveOperationsFromRanges( [ rangeA ], newTargetPosition, a );
 			}
@@ -466,8 +474,10 @@ const ot = {
 				// Transform `rangeA` by `b` operation and make operation out of it, and that's all.
 				// Note that this is a simplified version of default case, but here we treat the common part (whole `rangeA`)
 				// like a one difference part.
-				rangeA.start = rangeA.start._getTransformedByMove( b.sourcePosition, b.targetPosition, b.howMany, !includeB );
-				rangeA.end = rangeA.end._getTransformedByMove( b.sourcePosition, b.targetPosition, b.howMany, includeB );
+				rangeA = new Range(
+					rangeA.start._getTransformedByMove( b.sourcePosition, b.targetPosition, b.howMany, !includeB ),
+					rangeA.end._getTransformedByMove( b.sourcePosition, b.targetPosition, b.howMany, includeB )
+				);
 
 				return makeMoveOperationsFromRanges( [ rangeA ], newTargetPosition, a );
 			}
@@ -500,10 +510,12 @@ const ot = {
 			// This is an array with one or two ranges. Two ranges if `rangeB` is inside `rangeA`.
 			const difference = rangeA.getDifference( rangeB );
 
-			for ( const range of difference ) {
+			for ( const rangeInDiff of difference ) {
 				// Transform those ranges by `b` operation. For example if `b` moved range from before those ranges, fix those ranges.
-				range.start = range.start._getTransformedByDeletion( b.sourcePosition, b.howMany );
-				range.end = range.end._getTransformedByDeletion( b.sourcePosition, b.howMany );
+				const range = new Range(
+					rangeInDiff.start._getTransformedByDeletion( b.sourcePosition, b.howMany ),
+					rangeInDiff.end._getTransformedByDeletion( b.sourcePosition, b.howMany )
+				);
 
 				// If `b` operation targets into `rangeA` on the same level, spread `rangeA` into two ranges.
 				const shouldSpread = compareArrays( range.start.getParentPath(), b.getMovedRangeStart().getParentPath() ) == 'same';
@@ -513,12 +525,14 @@ const ot = {
 			}
 
 			// Then, we have to manage the "common part" of both move ranges.
-			const common = rangeA.getIntersection( rangeB );
+			const intersectionRange = rangeA.getIntersection( rangeB );
 
-			if ( common !== null && isStrong && !bTargetsToA ) {
+			if ( intersectionRange !== null && isStrong && !bTargetsToA ) {
 				// Calculate the new position of that part of original range.
-				common.start = common.start._getCombined( b.sourcePosition, b.getMovedRangeStart() );
-				common.end = common.end._getCombined( b.sourcePosition, b.getMovedRangeStart() );
+				const common = new Range(
+					intersectionRange.start._getCombined( b.sourcePosition, b.getMovedRangeStart() ),
+					intersectionRange.end._getCombined( b.sourcePosition, b.getMovedRangeStart() )
+				);
 
 				// Take care of proper range order.
 				//
@@ -627,9 +641,10 @@ function joinRanges( ranges ) {
 	} else if ( ranges.length == 1 ) {
 		return ranges[ 0 ];
 	} else {
-		ranges[ 0 ].end = ranges[ ranges.length - 1 ].end;
-
-		return ranges[ 0 ];
+		return new Range(
+			ranges[ 0 ].start,
+			ranges[ ranges.length - 1 ].end
+		);
 	}
 }
 
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -179,7 +179,7 @@ const ot = {
 				// Take the start and the end of the range and transform them by deletion of moved nodes.
 				// Note that if rangeB was inside AttributeOperation range, only difference.end will be transformed.
 				// This nicely covers the joining simplification we did in the previous step.
-				const range = new Range(
+				const differenceTransformed = new Range(
 					difference.start._getTransformedByDeletion( b.sourcePosition, b.howMany ),
 					difference.end._getTransformedByDeletion( b.sourcePosition, b.howMany )
 				);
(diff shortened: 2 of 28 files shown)


src/dev-utils/view.js
```diff
@@ -550,7 +550,7 @@ class RangeParser {
 			if ( item.bracket == ELEMENT_RANGE_START_TOKEN || item.bracket == TEXT_RANGE_START_TOKEN ) {
```
scofalik: Maybe this part of the code can be refactored to store range start and range end, instead of creating a range only to modify it later.

scofalik:
I just started to review this, but I wonder if it will be a positive change at all. AFAICS now positions will be created in some scenarios when they weren't, pretty much a lot in OT.

src/model/operation/transform.js
```diff
@@ -179,25 +179,29 @@ const ot = {
 				// Take the start and the end of the range and transform them by deletion of moved nodes.
 				// Note that if rangeB was inside AttributeOperation range, only difference.end will be transformed.
 				// This nicely covers the joining simplification we did in the previous step.
-				difference.start = difference.start._getTransformedByDeletion( b.sourcePosition, b.howMany );
-				difference.end = difference.end._getTransformedByDeletion( b.sourcePosition, b.howMany );
+				const range = new Range(
```
scofalik: Please change this const name to `differenceTransformed`.

src/model/operation/transform.js
```diff
@@ -179,25 +179,29 @@ const ot = {
...
+				ranges = range._getTransformedByInsertion( b.getMovedRangeStart(), b.howMany, true, false ).reverse();
 			}
 
 			if ( common !== null ) {
 				// Here we do not need to worry that newTargetPosition is inside moved range, because that
 				// would mean that the MoveOperation targets into itself, and that is incorrect operation.
 				// Instead, we calculate the new position of that part of original range.
-				common.start = common.start._getCombined( b.sourcePosition, b.getMovedRangeStart() );
-				common.end = common.end._getCombined( b.sourcePosition, b.getMovedRangeStart() );
+				const range = new Range(
```
scofalik: `commonTransformed`.

src/model/position.js
```diff
@@ -482,26 +513,25 @@ export default class Position {
 	 * @returns {module:engine/model/position~Position|null} Transformed position or `null`.
 	 */
 	_getTransformedByDeletion( deletePosition, howMany ) {
-		const transformed = Position.createFromPosition( this );
-
 		// This position can't be affected if deletion was in a different root.
 		if ( this.root != deletePosition.root ) {
-			return transformed;
+			return this;
```
scofalik: Are we sure this is not harmful in any way? I know that positions are now immutable but this method, by name, suggests that a new instance is returned. I think we don't have (and maybe won't have) such case in code, but one thing that comes to my mind is storing values in `Map` with `Position`s as keys.

src/model/position.js
```diff
@@ -533,19 +567,17 @@ export default class Position {
 	 * @returns {module:engine/model/position~Position} Transformed position.
 	 */
 	_getTransformedByInsertion( insertPosition, howMany, insertBefore ) {
-		const transformed = Position.createFromPosition( this );
-
 		// This position can't be affected if insertion was in a different root.
 		if ( this.root != insertPosition.root ) {
-			return transformed;
+			return this;
```
scofalik: Same as above.

src/model/position.js
```diff
@@ -459,20 +476,34 @@ export default class Position {
...
+	 */
+	toJSON() {
+		return {
+			root: this.root.toJSON(),
+			path: this.path
+		};
+	}
+
+	/**
 	 * Returns a copy of this position that is updated by removing `howMany` nodes starting from `deletePosition`.
```
scofalik: Docs might need an update if we won't be returning a copy in every scenario (see comment in code below).

src/model/position.js
```diff
@@ -782,6 +818,41 @@ export default class Position {
...
+ */
+export function setPath( position, path ) {
+	position[ _path ] = path;
+}
+
+// Helper for setting offset on give path array.
+// @private
+// @param {Array.<Number>} path Position path. See {@link module:engine/model/position~Position#path}.
+function getOffset( path ) {
+	return last( path ) || 0;
```
scofalik: Why `|| 0`? Wouldn't `undefined` be better?

Anyway, it is an incorrect situation if `path` has no items.

jodator: I didn't want to change logic behind this AFAIR. It was:
```js
get offset() {
    return last( path ) || 0;
}
```
I don't know if it would have any side effects or not.

src/model/position.js
jodator: Yeah.. maybe I've gone too deep into removing new instances. Here might be better to always create a copy.

src/model/range.js
```diff
@@ -804,13 +829,13 @@ export default class Range {
 		// 4. At this moment we don't need the original range.
 		// We are going to modify the result and we need to return a new instance of Range.
 		// We have to create a copy of the reference range.
-		const result = new this( ref.start, ref.end );
+		let result = new this( ref.start, ref.end );
```
scofalik: AFAICS, range `result` is not used directly, just its `start` and `end` properties. Maybe we could just go with storing `Position`s and it would be better?

src/model/treewalker.js
```diff
@@ -204,33 +204,37 @@ export default class TreeWalker {
 	 */
 	_next() {
 		const previousPosition = this.position;
```
scofalik: Maybe we can drop `position` variable in this function? It is confusing and I am not really sure it is needed.

src/model/treewalker.js
```diff
@@ -278,27 +285,31 @@ export default class TreeWalker {
 	 */
 	_previous() {
```
scofalik: Same with `position` here.

src/view/treewalker.js
```diff
@@ -187,7 +187,7 @@ export default class TreeWalker {
 	 * @returns {module:engine/view/treewalker~TreeWalkerValue} return.value Information about taken step.
 	 */
 	_next() {
-		let position = Position.createFromPosition( this.position );
+		let position = this.position;
```
scofalik: Same with `position`.

src/view/treewalker.js
```diff
@@ -292,7 +293,7 @@ export default class TreeWalker {
 	 * @returns {module:engine/view/treewalker~TreeWalkerValue} return.value Information about taken step.
 	 */
 	_previous() {
-		let position = Position.createFromPosition( this.position );
+		let position = this.position;
```
scofalik: Same with `position`.

scofalik:
All in all, good job! @Reinmar can you check what impact does it have on the original issue?

scofalik:
Could you make PRs out of commits in other repos?

scofalik:
@Reinmar @jodator Original issue described problem with copying selection. Unfortunately, this PR only partially fixes it. `Selection#_pushRange` is still the same. One of the reasons is that we don't want people to push `LiveRange`s to `Selection`. So there is still `Range#createFromRange()` there. What do we do with it?

src/model/position.js
scofalik: Weird. Position always has to have an offset :S. Try if anything breaks without it ;).

src/model/position.js
```diff
@@ -782,6 +818,41 @@ export default class Position {
...
+	position[ _root ] = root;
+}
+
+/**
+ * Method used to expose path setter to child classes.
+ * @protected
+ * @param {module:engine/model/position~Position} position Position of which path should be modified.
+ * @param {Array.<Number>} path Position path. See {@link module:engine/model/position~Position#path}.
+ */
+export function setPath( position, path ) {
```
pjasiun: I would make this a protected class method.

scofalik: I have mixed feelings about this.

From one point we don't want to mess around with position path. On the other hand, though, we need this or `LivePosition` won't work (there is an exactly same problem with `LiveRange#start` and `LiveRange#end`).

If these are protected methods then we don't have to really refactor some parts of the code. Just use the new protected methods! But is this really the way to go?

So IDK, I am torn. I liked this as a utility function because it is more discouraging to use it. But IDK. Maybe `LiveRange` should be in the same file as `Range`. Then we would just use the symbol defined in `range.js`.

pjasiun: For me, it the matter of convention. We have classes with protected methods exposed for other related classes in modules. For instance: https://github.com/ckeditor/ckeditor5-engine/blob/79da6fb1c0782cb6442527d3f2a50497b89b0985/src/model/operation/insertoperation.js#L84 We already had a talk about exposing protected methods for child classes at some point of designing editor class. We never did it as a separate export.

src/model/position.js
```diff
@@ -13,6 +13,9 @@ import compareArrays from '@ckeditor/ckeditor5-utils/src/comparearrays';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 import Text from './text';
 
+const _path = Symbol( 'path' );
+const _root = Symbol( 'root' );
```
pjasiun: We don't use symbols for private properties. `@private` is enough.

src/model/position.js
scofalik: I know what is the purpose of `@protected` tag and how we use it. The question is whether we want to have such `@protected` method at all. Because if we do, we can revert (or rather re-do) many of the changes proposed in this PR (at least in OT and tree walkers as these are used multiple time consecutively).

src/model/position.js
scofalik: It is not really `@private`, it is read-only. Also, we do use `Symbol`s in a couple of places. I think these are fine here.

src/model/position.js
pjasiun: Now I see that also there is a symbol used for the property. I would go with the `@protected` property (without a symbol) and public getters. We have such pair already in some places in the code. We can even write that the property is exposed only for `LiveRange` and should not be used anywhere.

src/model/position.js
ma2ciek: But symbols make them private too.

jodator: > But symbols make them private too.

A bit better hidden but also accessible ([more work required](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertySymbols) though...) also not enumerable.

But I get that in CKEditor 5 there is already established convention and using straightforward `this._path` in child classes would be simpler.

I only don like "duplication" on object as the getter `this.path` would just return `this._path`.

pjasiun: Feel free to create a separate ticket to change it, but since we already do it this way, we should follow the convention:

https://github.com/ckeditor/ckeditor5-engine/blob/79da6fb1c0782cb6442527d3f2a50497b89b0985/src/model/rootelement.js#L53-L55

https://github.com/ckeditor/ckeditor5-engine/blob/79da6fb1c0782cb6442527d3f2a50497b89b0985/src/view/selection.js#L107-L119

https://github.com/ckeditor/ckeditor5-engine/blob/79da6fb1c0782cb6442527d3f2a50497b89b0985/src/view/text.js#L58-L60

scofalik: I never liked `this._data` and was wondering why it is this way. I guess I will like to change it after more important subjects are dealt with.

Reinmar:
Please don't merge this PR before we'll release alpha.2. It's a bit risky change.

Reinmar:
Since alpha.2 is out, please conclude this PR and let me know when we could have a final check on it.

src/model/position.js
jodator: @scofalik: I was able to fail a test after removing `|| 0`:
```js
expect( new Position( root, [ false ] ) ).to.have.property( 'offset' ).that.equals( 0 );
expect( new Position( root, [ undefined ] ) ).to.have.property( 'offset' ).that.equals( 0 );
```

But this looks like a bad API use - so we might as well drop `|| 0` check as `position` accepts only non-empty array as path.

jodator:
> * getRanges() {
> 	for ( const range of this._ranges ) {
> 		yield Range.createFromRange( range );
> 	}
> }
> 
> Both in model.Selection and view.Selection?
> 
> Same with getFirstRange and getLastRange.

So probably directe getters are to be created as it would make no sense for `getBlabla` methods?

scofalik:
No matter what `getRanges()` will return, it should not let modify selection ranges without using selection API. So:
```javascript
get ranges() {
    return this._ranges;
}
```
Is an incorrect solution if you ask me.

src/model/position.js
scofalik: Yeah, I'd remove `|| 0` and the test.

jodator:
@scofalik 

```js
return [ ...this._ranges ];
```
?

scofalik:
Although it was a generator function before, and some people on this project love their generators...

jodator:
OK so what to do with those iterators?

I'll grab 🍿 and let you @pjasiun ⚔️  @scofalik fight over removing or not iterators :)

ps.: I'm for returning new array:

```js
get ranges() {
    return [ ...this._ranges ];
```
to not mess with internal array.

scofalik:
I am for returning array, because in this case you always want to handle all of them. If not, there are already `getFirstRange()` and `getLastRange()` methods.

pjasiun:
I would go with:
```js
get ranges() {
    return this. _ranges[ Symbol.iterator ]();
}
```

This way you will not create any new object and it will be still not mutable.

jodator:
@scofalik I've done with all the suggested changes.

I've also removed `Range.createFromRange()` from `getFirst/LastRange()` and updated docs for those methods.

src/model/range.js
```diff
@@ -291,8 +311,11 @@ export default class Range {
 				ranges.push( new Range( pos, pos.getShiftedBy( howMany ) ) );
 			}
 
-			pos.path = pos.path.slice( 0, -1 );
-			pos.offset++;
+			const path = pos.getParentPath();
```
scofalik: I think this is same as `Position.createAfter( pos.parent )`?

src/model/range.js
```diff
@@ -755,9 +791,8 @@ export default class Range {
 	 */
 	static createCollapsedAt( itemOrPosition, offset ) {
 		const start = Position.createAt( itemOrPosition, offset );
-		const end = Position.createFromPosition( start );
 
-		return new Range( start, end );
+		return new Range( start, start );
```
scofalik: Even `new Range( start )` should be fine

src/model/range.js
```diff
@@ -804,13 +839,14 @@ export default class Range {
 		// 4. At this moment we don't need the original range.
 		// We are going to modify the result and we need to return a new instance of Range.
 		// We have to create a copy of the reference range.
-		const result = new this( ref.start, ref.end );
+		let start = ref.start;
+		let end = ref.end;
 
 		// 5. Ranges should be checked and glued starting from the range that is closest to the reference range.
 		// Since ranges are sorted, start with the range with index that is closest to reference range index.
 		for ( let i = refIndex - 1; i >= 0; i++ ) {
```
scofalik: I wonder how the hell this works if the condition is `i >= 0` and `i` is increment in each loop. Seems wrong.

src/model/selection.js
```diff
@@ -255,7 +253,7 @@ export default class Selection {
 	getLastPosition() {
 		const lastRange = this.getLastRange();
```
scofalik: -> `last`

src/view/position.js
```diff
@@ -24,20 +27,28 @@ export default class Position {
...
-		 */
-		this.parent = parent;
-
-		/**
-		 * Position offset.
-		 *
-		 * @member {Number} module:engine/view/position~Position#offset
-		 */
-		this.offset = offset;
+		this[ _parent ] = parent;
```
scofalik: Didn't we agreed that this should be `this._parent` and `this._offset`?

src/view/range.js
```diff
@@ -23,19 +26,8 @@ export default class Range {
...
-		 */
-		this.start = Position.createFromPosition( start );
-
-		/**
-		 * End position.
-		 *
-		 * @member {module:engine/view/position~Position}
-		 */
-		this.end = end ? Position.createFromPosition( end ) : Position.createFromPosition( start );
+		this[ _start ] = start;
```
scofalik: `this._parent` and `this._offset`

src/view/writer.js
```diff
@@ -353,9 +352,7 @@ export function remove( range ) {
 	const removed = parentContainer.removeChildren( breakStart.offset, count );
 
 	// Merge after removing.
-	const mergePosition = mergeAttributes( breakStart );
-	range.start = mergePosition;
-	range.end = Position.createFromPosition( mergePosition );
+	mergeAttributes( breakStart );
```
scofalik: Note to self: check how `mergeAttributres` works now.

scofalik:
I've pushed some minor changes.

scofalik:
Hm, for some reason it broke: https://travis-ci.org/ckeditor/ckeditor5-engine/builds/306753449?utm_source=email&utm_medium=notification although tests pass for engine for me locally.

Reinmar:
I had to revert this change because it completely blows up editor tests.

Also, see https://github.com/ckeditor/ckeditor5-engine/issues/897#issuecomment-346862873.

scofalik:
What do you mean exactly by "blows up editor tests"? Performance issues or something else?

Reinmar:
> What do you mean exactly by "blows up editor tests"? Performance issues or something else?

Many failing tests in lists, images and undo.

PR #1191, opened 2017-11-24
Conversion refactor

### LINKED ISSUE #1172: Conversion refactor
I am opening this ticket to have a place for discussion about upcoming conversion refactor.

## Goals

There are a few goals that we want to achieve with this.

The first goal is to stop converting an incorrect model state to the view. This happens now because we fire model-to-view conversion after each atomic change on the model.

There is also a problem with mapping during conversion. Sometimes we need to map non-existing model position/range to the view (a case for remove). Sometimes we want to map an element from the model to the view, but the element was already mapped to another element (a case for rename and move AFAIR). It would be perfect if those problems would vanish away.

The second goal is to make it easier to write converters. At the moment, converters are the most complicated part of developing a feature. For a new developer, it may be not clear what exactly they have to do to write a correct converter.

This is also connected with providing more / different / better / easier to use conversion helpers. Now we have more knowledge about how the converters actually look, so we know what kind of helpers we can introduce.

Additionally, markers conversion isn't pretty right now. It'd be nice to revisit it.

The bonus goal is to prepare an environment where we can easily hook callbacks / fixers that should react to changes in the model.

Keep in mind, that developers have their own experiences and are accustomed to some patterns. If the solution is easier for the user, it is probably better.

Think if there is anything that can be improved in view-to-model conversion. Since this topic all began from talking about "converting model changes to view on `changesDone`" it was natural, that we mostly focused on model-to-view conversion. In my opinion, view-to-model conversion is fine, but I might be missing something.

## Related tickets

https://github.com/ckeditor/ckeditor5-engine/issues/829 - Simplify model to view conversion
https://github.com/ckeditor/ckeditor5-engine/issues/736 - Add more conversion helpers
https://github.com/ckeditor/ckeditor5-engine/issues/778 - Change-based conversion can be really confusing
https://github.com/ckeditor/ckeditor5-engine/issues/779 - It shouldn't be needed to listen on 3 events to convert an attribute
https://github.com/ckeditor/ckeditor5-engine/issues/945 - Consider simplifying attribute conversion

### Thinking out of the box

I'd like to make it clear, that we should try to forget about how conversion looks now. Of course, we have experience, we know what worked and what did not work. However, maybe we could come up with a totally different idea for conversion mechanism.

The only assumptions to make are:
1. There is a model element which is composed of name and attributes. We need to convert that model element to a view element. 
2. The model element might change. We need to reflect that change in the view.
3. The model might be in incorrect state after some changes are applied. We need a good solution how to fix the model if that happens.

### Limit amount of change types

This already has been tweaked a bit earlier. Anyway, This is a minimal set of changes that we need to support:
 - insert node,
 - remove node,
 - attribute changes.

Insert and remove changes handle, of course, removing and inserting elements, but also moving or renaming them.

Attribute changes are a bit more complicated. I am not sure what we need here. Unfortunately "add attribute" + "remove attribute" does not seem to be the correct way to change an attribute. Mostly because some elements require some attributes so removing attribute is forbidden. Take `listItem` for example. Changing it's `type` attribute should be converted "in one step", especially because it doesn't make sense to have a `listItem` without `type` attribute.

Maybe the solution is to have just one "attribute change" change type, and the converter should handle it all: inserting, removing and changing the attribute. The question is what we gain here - other than having to attach only one callback (instead of three, like now).

Another idea is to remove the element which attributes got changed and then insert it with the changed attribute. I am not sure how this will impact view tree and then DOM rendering. I am not sure how inefficient such solution would be. Thankfully, the biggest efficiency bottleneck is typing and attribute conversion is not connected with it.

### Conversion on `changesDone`

This is the original idea from which it all started. It is simple - do not convert model changes right away. Instead, buffer them. When all changes are applied, all callbacks and fixers did their changes, convert the model to the view. This should happen on `changesDone` event. At this moment, all fixers should already have fixed the model.

What do we gain from this solution? Mostly, the converters will be easier to write because they won't need to convert incorrect model states.

How to enable conversion on `changesDone`? There are a couple of ideas here:

1. Mark changed elements. Have a data structure that will keep changes done to the model: which elements were removed, which were added and which attributes changed. This is a bit difficult for `Text` instances. Text changes might need to be kept as `LivePosition`s, especially when whole text node was removed. On `changesDone` iterate through this data structure and apply changes to the view one by one. This is the simplest and original proposal. I'd treat this as a "fallback solution". It does not change a lot in the conversion itself, it just postpones the whole conversion to a later stage. It solves some of our problems. If we won't come up with something better, this should be okay.
2. Copy model or parts of the model to diff them later. Instead of tracking down each change, we could copy the whole model when first change is applied and then diff old (saved) model with the new one. Unfortunately, I fear that this would be too inefficient for typing.
3. Save which model elements have changed. Then convert them to the view and diff them with the actual view. This idea is similar to what we have right now in view->DOM rendering. We only change what really changed. I am not sure, but AFAICS this solution could also resolve some problems with mapping.

Do not forget about markers! Markers are also a part of the model and need to be converted on `changesDone`. We would have to store which markers have changed, and how. The exact implementation will probably depend on the chosen solution for model tree conversion.

Those ideas do not touch `change` callbacks, though. Those callbacks / fixers, similarly to converters, are now a bit too complicated. At the end of this post I've included a list of all callbacks listening to `change` event.

### Decouple building and inserting elements

We came from an idea that we are converting changes, not model state. This is why we'd rather say "convert image insertion" rather than "convert image element". On the other hand, it is much easier to think that "`listItem` model element should be represented as `<li></li>` in view". People are familiar with this kind of thinking. We should take it into consideration.

Hence, I propose to think about decouple inserting elements and building elements. Inserting is same anyway for most features. For now, only lists have their own needs for inserting -- when you insert an element between `listItem`s, in view you need to break `<ul>`/`<ol>` element. Similarly, when you remove an element, it may happen that two lists need to be merged.

The reason for this proposal is that maybe if we think about "building" elements, it will be easier to refactor conversion mechanisms and propose new conversion utilities.

This idea is strongly connected with rebuilding items a lot of time (for example, rebuild the whole item when its attribute changed). We need to think how elements change and what impact does rebuilding have on creating a view structure and then DOM.

### Firing event before the change is applied to the model

This is a wild idea and I don't know what it would achieve expect of giving us more possibilities but I just wanted to throw it here and see if maybe you will find it useful.

The idea is simple: instead of applying a change to the model right away, fire an event. Listeners would be able to modify or cancel the change.

### Do we need consumables?

This was brought up during talks and is not originally my idea. The question was whether we need consumables and how do we use them at the moment. The doubt was that consumables have the same role as event canceling. The bottom line was that we maybe need them in view-to-model conversion but probably are able to remove them from model-to-view conversion.

Of course, if we will go with "building" items, the model-to-view conversion will get a bit similar to the view-to-model conversion. Then we might need consumables to build elements from the model in a right way.

### Use cases for `change` event other than `ModelConversionDispatcher`

CKE5 uses `change` event for following callbacks / fixing:

 - autoformatting (typing of certain characters are detected on `change` event),
 - image caption (adding `image` element is detected and if the `image` does not have `caption` element, the element is added),
 - lists fixer (a lot of incorrect scenarios related with `type` and `indent` attributes are detected and fixed),
 - auto paragraphing (when an empty root is detected, `paragraph` element is inserted),
 - change buffer is updated after each `change` event (typing),
 - undo (undo stack is refreshed after each change on the model),
 - image upload.

Additionally, these are Letters use cases:

 - document structure is fixed to always have title and body,
 - comments management heavily depends on detecting marker changes,
 - saving changes to send to other clients,
 - comments position in the sidebar is refreshed after each change.

I think that in overall, there are two types of callbacks:
 - ones that care about model state or how it was changed and then fix the model or refresh some other states,
 - others that have to react to every change, mostly to track it.

### What helpers do we need?

This is to bootstrap a work that has to be done with conversion helpers. There was an idea to drop converter builders and instead provide a set of utility functions that will be simpler to use but will produce more focused converters.

We need to study how do we use converter builders. Which cases need to be ported to utility functions. Then, we need to see what was impossible to do using builders.

A short list of helpers that we need. Was not researched much yet:
 - model element <-> view element (by element name),
 - model attribute <-> view element (by names),
 - model attribute <-> view attribute (by names),
 - markers conversion,
 - collapsed selection conversion (it breaks attributes and inserts empty attribute elements),
 - **new** view style converter (make it easy to add and remove a part of style attribute),
 - **new** view class converter (make it easy to add and remove a part of class attribute),
 - **new** wrapper converters (case where generated view structure is wrapped in an element, for example `<figure>` element in image feature)

## Tasks

A list of tasks that we need to do no matter how we proceed with this refactoring.

1. Clean `view.writer`. For example, it has `rename` method and I am not sure if it is used at all.
2. Check how `UIElement`s are created and added to the view tree. Try to come up with an API to manage `UIElement`s better.

scofalik:

Other: Conversion refactor. Closes #1172.

---

### Additional information

This PR is not ready yet. I am creating it so anyone who is interested can take a look at proposed changes.

### DIFF
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -7,6 +7,7 @@
  * @module engine/controller/editingcontroller
  */
 
+import ModelDiffer from '../model/differ';
 import ViewDocument from '../view/document';
 import Mapper from '../conversion/mapper';
 import ModelConversionDispatcher from '../conversion/modelconversiondispatcher';
@@ -64,10 +65,9 @@ export default class EditingController {
 		this.mapper = new Mapper();
 
 		/**
-		 * Model to view conversion dispatcher, which converts changes from the model to
-		 * {@link #view editing view}.
+		 * Model to view conversion dispatcher, which converts changes from the model to {@link #view the editing view}.
 		 *
-		 * To attach model to view converter to the editing pipeline you need to add lister to this property:
+		 * To attach model-to-view converter to the editing pipeline you need to add a listener to this dispatcher:
 		 *
 		 *		editing.modelToView( 'insert:$element', customInsertConverter );
 		 *
@@ -83,36 +83,62 @@ export default class EditingController {
 			viewSelection: this.view.selection
 		} );
 
-		// Convert changes in model to view.
-		this.listenTo( this.model, 'change', ( evt, type, changes ) => {
-			this.modelToView.convertChange( type, changes );
-		}, { priority: 'low' } );
+		// Model differ object. It's role is to buffer changes done on model and then calculates a diff of those changes.
+		// The diff is then passed to model conversion dispatcher which generates proper events and kicks-off conversion.
+		const modelDiffer = new ModelDiffer();
 
-		// Convert model selection to view.
-		this.listenTo( this.model, 'changesDone', () => {
-			const selection = this.model.selection;
+		// Before an operation is applied on model, buffer the change in differ.
+		this.listenTo( this.model, 'operation', ( evt, operation ) => {
+			modelDiffer.bufferOperation( operation );
+		} );
 
-			this.modelToView.convertSelection( selection );
-			this.view.render();
-		}, { priority: 'low' } );
-
-		// Convert model markers changes.
+		// Buffer marker changes.
+		// This is not covered in buffering operations because markers may change outside of them (when they
+		// are modified using `document.markers` collection, not through `MarkerOperation`).
 		this.listenTo( this.model.markers, 'add', ( evt, marker ) => {
-			this.modelToView.convertMarker( 'addMarker', marker.name, marker.getRange() );
+			// Whenever a new marker is added, buffer that change.
+			modelDiffer.bufferMarkerChange( marker.name, null, marker.getRange() );
+
+			// Whenever marker changes, buffer that.
+			marker.on( 'change', ( evt, oldRange ) => {
+				modelDiffer.bufferMarkerChange( marker.name, oldRange, marker.getRange() );
+			} );
 		} );
 
 		this.listenTo( this.model.markers, 'remove', ( evt, marker ) => {
-			this.modelToView.convertMarker( 'removeMarker', marker.name, marker.getRange() );
+			// Whenever marker is removed, buffer that change.
+			modelDiffer.bufferMarkerChange( marker.name, marker.getRange(), null );
 		} );
 
-		// Convert view selection to model.
+		// When all changes are done, get the model diff containing all the changes and convert them to view and then render to DOM.
+		this.listenTo( this.model, 'changesDone', () => {
+			// First, before the view is changed, remove from the view markers which has changed.
+			this.modelToView.removeMarkers( modelDiffer.getMarkersToRemove() );
+
+			// Calculate a diff between old model and new model. Then convert those changes to the view.
+			this.modelToView.convertBufferedChanges( modelDiffer.getChanges() );
+
+			// After the view is updated, convert markers which has changed.
+			this.modelToView.addMarkers( modelDiffer.getMarkersToAdd() );
+
+			// Reset model diff object. When next operation is applied, new diff will be created.
+			modelDiffer.reset();
+
+			// After the view is ready, convert selection from model to view.
+			this.modelToView.convertSelection( this.model.selection );
+
+			// When everything is converted to the view, render it to DOM.
+			this.view.render();
+		}, { priority: 'low' } );
+
+		// Convert selection from view to model.
 		this.listenTo( this.view, 'selectionChange', convertSelectionChange( this.model, this.mapper ) );
 
-		// Attach default content converters.
+		// Attach default model converters.
 		this.modelToView.on( 'insert:$text', insertText(), { priority: 'lowest' } );
 		this.modelToView.on( 'remove', remove(), { priority: 'low' } );
 
-		// Attach default selection converters.
+		// Attach default model selection converters.
 		this.modelToView.on( 'selection', clearAttributes(), { priority: 'low' } );
 		this.modelToView.on( 'selection', clearFakeSelection(), { priority: 'low' } );
 		this.modelToView.on( 'selection', convertRangeSelection(), { priority: 'low' } );
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -89,7 +89,9 @@ export default class EditingController {
 
 		// Before an operation is applied on model, buffer the change in differ.
 		this.listenTo( this.model, 'operation', ( evt, operation ) => {
-			modelDiffer.bufferOperation( operation );
+			if ( operation.isDocumentOperation ) {
+				modelDiffer.bufferOperation( operation );
+			}
 		} );
 
 		// Buffer marker changes.
--- a/src/conversion/buildmodelconverter.js
+++ b/src/conversion/buildmodelconverter.js
@@ -10,13 +10,10 @@
 import {
 	insertElement,
 	insertUIElement,
-	setAttribute,
-	removeAttribute,
 	removeUIElement,
-	wrapItem,
-	unwrapItem,
-	highlightText,
-	highlightElement
+	setAttribute,
+	wrap,
+	highlight
 } from './model-to-view-converters';
 
 import { convertSelectionAttribute, convertSelectionMarker } from './model-selection-to-view-converters';
@@ -254,15 +251,13 @@ class ModelConverterBuilder {
 
 				dispatcher.on( 'insert:' + this._from.name, insertElement( element ), { priority } );
 			} else if ( this._from.type == 'attribute' ) {
-				// From model attribute to view element -> wrap and unwrap.
+				// From model attribute to view element -> wrap.
 				element = typeof element == 'string' ? new ViewAttributeElement( element ) : element;
 
-				dispatcher.on( 'addAttribute:' + this._from.key, wrapItem( element ), { priority } );
-				dispatcher.on( 'changeAttribute:' + this._from.key, wrapItem( element ), { priority } );
-				dispatcher.on( 'removeAttribute:' + this._from.key, unwrapItem( element ), { priority } );
-
+				dispatcher.on( 'attribute:' + this._from.key, wrap( element ), { priority } );
 				dispatcher.on( 'selectionAttribute:' + this._from.key, convertSelectionAttribute( element ), { priority } );
-			} else { // From marker to element.
+			} else {
+				// From marker to element.
 				const priority = this._from.priority === null ? 'normal' : this._from.priority;
 
 				element = typeof element == 'string' ? new ViewUIElement( element ) : element;
@@ -326,12 +321,8 @@ class ModelConverterBuilder {
 		}
 
 		for ( const dispatcher of this._dispatchers ) {
-			// Separate converters for converting texts and elements inside marker's range.
-			dispatcher.on( 'addMarker:' + this._from.name, highlightText( highlightDescriptor ), { priority } );
-			dispatcher.on( 'addMarker:' + this._from.name, highlightElement( highlightDescriptor ), { priority } );
-
-			dispatcher.on( 'removeMarker:' + this._from.name, highlightText( highlightDescriptor ), { priority } );
-			dispatcher.on( 'removeMarker:' + this._from.name, highlightElement( highlightDescriptor ), { priority } );
+			dispatcher.on( 'addMarker:' + this._from.name, highlight( highlightDescriptor ), { priority } );
+			dispatcher.on( 'removeMarker:' + this._from.name, highlight( highlightDescriptor ), { priority } );
 
 			dispatcher.on( 'selectionMarker:' + this._from.name, convertSelectionMarker( highlightDescriptor ), { priority } );
 		}
@@ -407,9 +398,7 @@ class ModelConverterBuilder {
 		for ( const dispatcher of this._dispatchers ) {
 			const options = { priority: this._from.priority || 'normal' };
 
-			dispatcher.on( 'addAttribute:' + this._from.key, setAttribute( attributeCreator ), options );
-			dispatcher.on( 'changeAttribute:' + this._from.key, setAttribute( attributeCreator ), options );
-			dispatcher.on( 'removeAttribute:' + this._from.key, removeAttribute( attributeCreator ), options );
+			dispatcher.on( 'attribute:' + this._from.key, setAttribute( attributeCreator ), options );
 		}
 	}
 }
--- a/src/conversion/mapper.js
+++ b/src/conversion/mapper.js
@@ -110,23 +110,43 @@ export default class Mapper {
 	/**
 	 * Unbinds given {@link module:engine/view/element~Element view element} from the map.
 	 *
+	 * **Note:** view-to-model binding will be removed, if it existed. However, corresponding model-to-view binding
+	 * will be removed only if model element is still bound to passed `viewElement`.
+	 *
+	 * This behavior lets for re-binding model element to another view element without fear of losing the new binding
+	 * when the previously bound view element is unbound.
+	 *
 	 * @param {module:engine/view/element~Element} viewElement View element to unbind.
 	 */
 	unbindViewElement( viewElement ) {
 		const modelElement = this.toModelElement( viewElement );
 
-		this._unbindElements( modelElement, viewElement );
+		this._viewToModelMapping.delete( viewElement );
+
+		if ( this._modelToViewMapping.get( modelElement ) == viewElement ) {
+			this._modelToViewMapping.delete( modelElement );
+		}
 	}
 
 	/**
 	 * Unbinds given {@link module:engine/model/element~Element model element} from the map.
+	 *
+	 * **Note:** model-to-view binding will be removed, if it existed. However, corresponding view-to-model binding
+	 * will be removed only if view element is still bound to passed `modelElement`.
+	 *
+	 * This behavior lets for re-binding view element to another model element without fear of losing the new binding
+	 * when the previously bound model element is unbound.
 	 *
 	 * @param {module:engine/model/element~Element} modelElement Model element to unbind.
 	 */
 	unbindModelElement( modelElement ) {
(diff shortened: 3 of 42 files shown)


scofalik:
Things left to do:
1. Simplify code in `engine.model.Differ`.
2. Provide docs for `engine.model.Differ`.
3. Write tests for new code.
4. Split all the changes into sane commits.

src/conversion/buildmodelconverter.js
```diff
@@ -10,13 +10,10 @@
...
 	insertElement,
 	insertUIElement,
-	setAttribute,
-	removeAttribute,
 	removeUIElement,
-	wrapItem,
-	unwrapItem,
-	highlightText,
-	highlightElement
+	setAttribute,
```
pjasiun: changeAttribute?

src/conversion/mapper.js
```diff
@@ -110,23 +110,43 @@ export default class Mapper {
...
 	 */
 	unbindViewElement( viewElement ) {
 		const modelElement = this.toModelElement( viewElement );
 
-		this._unbindElements( modelElement, viewElement );
+		this._viewToModelMapping.delete( viewElement );
+
+		if ( this._modelToViewMapping.get( modelElement ) == viewElement ) {
+			this._modelToViewMapping.delete( modelElement );
+		}
```
pjasiun: Not sure isn't it simpler to just call both methods.

src/conversion/model-to-view-converters.js
```diff
@@ -91,6 +90,35 @@ export function insertText() {
...
+ */
+export function remove() {
+	return ( evt, data, conversionApi ) => {
+		// Find view range start position by mapping model position at which the remove happened.
+		const viewStart = conversionApi.mapper.toViewPosition( data.position );
+
+		// Find view range end by providing to mapper view container and expected model offset.
+		// Get the offset from passed model position and passed length of removed content.
+		const viewContainer = conversionApi.mapper.toViewElement( data.position.parent );
+		const viewEnd = conversionApi.mapper._findPositionIn( viewContainer, data.position.offset + data.length );
```
pjasiun: Instead of using private method, we should provide mapper API. Also it we skip custom callbacks we can get wrong position if this callbacks change offset.

src/conversion/model-to-view-converters.js
```diff
@@ -105,110 +133,98 @@ export function insertText() {
  * @returns {Function} Insert element event converter.
  */
 export function insertUIElement( elementCreator ) {
-	return ( evt, data, consumable, conversionApi ) => {
+	return ( evt, data, conversionApi ) => {
```
pjasiun: Why there are no consumables here? That's strange that if you wants to prevent child consuming here you need to do something different that in other converteres.

src/conversion/model-to-view-converters.js
```diff
@@ -105,110 +133,98 @@ export function insertText() {
...
- * {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher#event:addAttribute}
- * or {@link module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher#event:changeAttribute} event.
- * @returns {Function} Set/change attribute converter.
+ * @param {module:engine/view/uielement~UIElement|Function} elementCreator View ui element, or function returning
+ * a view ui element, which will be used as a pattern when look for element to remove at the marker start position.
+ * @returns {Function} Remove ui element converter.
  */
-export function setAttribute( attributeCreator ) {
-	attributeCreator = attributeCreator || ( ( value, key ) => ( { value, key } ) );
+export function removeUIElement( elementCreator ) {
```
pjasiun: I just realised that conversion of markers removal is now problematic, because we convert markers on the document which is different in view and model.

src/conversion/model-to-view-converters.js
```diff
@@ -264,19 +285,26 @@ export function removeAttribute( attributeCreator ) {
...
  *
- *		modelDispatcher.on( 'addAttribute:bold', wrapItem( new ViewAttributeElement( 'strong' ) ) );
+ *		modelDispatcher.on( 'attribute:bold', wrapItem( new ViewAttributeElement( 'strong' ) ) );
  *
  * @param {module:engine/view/element~Element|Function} elementCreator View element, or function returning a view element, which will
  * be used for wrapping.
  * @returns {Function} Set/change attribute converter.
  */
-export function wrapItem( elementCreator ) {
+export function wrap( elementCreator ) {
```
pjasiun: `export function wrap&unwrap` (can we have `&` in the function name? :P) )

src/conversion/model-to-view-converters.js
```diff
@@ -286,273 +314,163 @@ export function wrapItem( elementCreator ) {
...
-
-		const modelItem = data.item;
-
-		if ( !descriptor || data.markerRange.isCollapsed || !modelItem.is( 'textProxy' ) ) {
-			return;
-		}
-
-		if ( !consumable.consume( modelItem, evt.name ) ) {
+export function highlight( highlightDescriptor ) {
+	return ( evt, data, conversionApi ) => {
```
pjasiun: I miss `consumable` here.

pjasiun: We should consinder splitting listeners for add marker and remove marker.

src/conversion/model-to-view-converters.js
```diff
@@ -286,273 +314,163 @@ export function wrapItem( elementCreator ) {
...
+			}
+			// Skip or step inside other elements (mostly attribute elements).
+
+			// If new position got specified, apply it to the walker.
+			if ( newPosition ) {
+				if ( newPosition.isBefore( viewRange.start ) ) {
+					// If the new position is beyond the range, stop processing.
+					break;
+				} else {
+					treeWalker.position = newPosition;
```
pjasiun: `treeWalker.position` is at the moment readonly. We should check why.

src/controller/editingcontroller.js
```diff
@@ -83,36 +83,62 @@ export default class EditingController {
...
+		// When all changes are done, get the model diff containing all the changes and convert them to view and then render to DOM.
+		this.listenTo( this.model, 'changesDone', () => {
+			// First, before the view is changed, remove from the view markers which has changed.
+			this.modelToView.removeMarkers( modelDiffer.getMarkersToRemove() );
+
+			// Calculate a diff between old model and new model. Then convert those changes to the view.
+			this.modelToView.convertBufferedChanges( modelDiffer.getChanges() );
+
+			// After the view is updated, convert markers which has changed.
+			this.modelToView.addMarkers( modelDiffer.getMarkersToAdd() );
```
pjasiun: I like how this code looks like, but we could consider calling `removeMarkers` and `addMarkers ` inside `convertBufferedChanges` since you should not call `convertBufferedChanges` whitout them.

pjasiun: However I'm not sure. This code tell very well what is going on.

src/model/liveposition.js
```diff
@@ -35,7 +35,7 @@ export default class LivePosition extends Position {
 	 * @see module:engine/model/position~Position
 	 * @param {module:engine/model/rootelement~RootElement} root
 	 * @param {Array.<Number>} path
-	 * @param {module:engine/model/position~PositionStickiness} [stickiness] Defaults to `'sticksToNext'`.
+	 * @param {module:engine/model/position~PositionStickiness} [stickiness] Defaults to `'notSticky'`.
```
pjasiun: @scofalik said we don't need this change anymore.

src/view/writer.js
```diff
@@ -480,25 +480,31 @@ export function wrap( range, attribute ) {
...
+	if ( breakEnd.isEqual( breakStart.getShiftedBy( 1 ) ) ) {
+		const node = breakStart.nodeAfter;
+
+		if ( node instanceof AttributeElement && wrapAttributeElement( attribute, node ) ) {
+			const start = mergeAttributes( breakStart );
+			const end = mergeAttributes( breakEnd );
+
+			return new Range( start, end );
+		}
+	}
```
pjasiun: There are missing tests for this change.

src/model/differ.js
```diff
@@ -0,0 +1,703 @@
...
+				newRange
+			} );
+		} else {
+			if ( buffered.oldRange == null && newRange == null ) {
+				// The marker is going to be removed (`newRange == null`) but it did not exist before the change set
+				// (`buffered.oldRange == null`). In this case, do not keep the marker in buffer at all.
+				this._changedMarkers.delete( markerName );
+			} else {
+				buffered.newRange = newRange;
+			}
```
pjasiun: Rafactor?
```js
} else {
	buffered.newRange = newRange;
}

if ( buffered.oldRange == null && buffered.newRange == null ) {
	// The marker is going to be removed (`newRange == null`) but it did not exist before the change set
	// (`buffered.oldRange == null`). In this case, do not keep the marker in buffer at all.
	this._changedMarkers.delete( markerName );
}
```

src/model/differ.js
```diff
@@ -0,0 +1,703 @@
...
+
+			// Process every action in the diff.
+			for ( const action of actions ) {
+				if ( action === 'insert' ) {
+					if ( elementChildren[ i ] instanceof Element ) {
+						// New element got inserted. We should not buffer the same change twice.
+						// If that element's children has changed -- remove that from the buffer.
+						this._changedElements.delete( elementChildren[ i ] );
+						// If that element has changed -- remove that change the buffer.
+						this._changedChildren.delete( elementChildren[ i ] );
```
pjasiun: All descendants can have changes, not only direct children.

src/model/differ.js
```diff
@@ -0,0 +1,703 @@
...
+			} else {
+				// Element name has changed. Since this requires removing old element and inserting new element,
+				// we don't care about attributes. If those have changed, they will be refreshed with the newly inserted element.
+				// Generate diff items for element "refreshing".
+				diffSet.push( this._getRemoveDiff( element.parent, element.startOffset, snapshot.name ) );
+				diffSet.push( this._getInsertDiff( element.parent, element.startOffset, element.name ) );
+			}
+		}
+
+		// At this point, all buffered changes has been processed and the diff set is calculated. Now, it will be further processed.
```
pjasiun: Why this have coverage?

src/model/differ.js
```diff
@@ -0,0 +1,703 @@
...
+			if ( a.position.isEqual( b.position ) ) {
+				// Keep chronological order of operations.
+				return a.changeCount < b.changeCount ? -1 : 1;
+			}
+
+			// If positions differ, position "on the left" should be earlier in the result.
+			return a.position.isBefore( b.position ) ? -1 : 1;
+		} );
+
+		// Glue together multiple changes (mostly on text nodes).
```
pjasiun: It would be nice to handle text with multiple changed attributes. It can be done as a follow-up.

src/model/differ.js
```diff
@@ -0,0 +1,703 @@
...
+	}
+
+	/**
+	 * Resets `Differ`. Removes all buffered changes.
+	 */
+	reset() {
+		this._changedChildren.clear();
+		this._changedElements.clear();
+		this._changedMarkers.clear();
+		this._changeCount = 0;
```
pjasiun: You should reset change could at the end of `getChanges` not in reset.

pjasiun:
We need to check performance after these changes.

pjasiun:
We should check if it's possible to call `getChanges` multiple times and if it returns the same data.

src/conversion/mapper.js
scofalik: It looks that it's not that easy after all. For example, take the default `remove` converter. You'd have to move the same logic there:

```javascript
for ( const child of ViewRange.createIn( removed ).getItems() ) {
	conversionApi.mapper.unbindViewElement( child );
}
```

The converter does not receive the removed item, just the remove position, remove length and a name of the removed item. So you can't just call `unbindModelElement` on it.

src/conversion/model-to-view-converters.js
scofalik: Let's keep it as `wrap()` :)

src/controller/editingcontroller.js
scofalik: I am for moving `removeMarkers` and `addMarkers` to `ModelConversionDispatcher`. You are right that there is only one way to process this conversion, and you never want to fire one of those function without firing other ones, and exactly in this order.

src/conversion/model-to-view-converters.js
scofalik: On the second look, I am not sure if we would achieve that much with this change.

src/conversion/model-to-view-converters.js
scofalik: Consumables are pretty awkward here after refactoring. Earlier, every item in the marker range was put in consumables (so it was multiple values unless the range was collapsed, then it was just the range).

It made sense earlier to have consumables. Right now, because the model at the moment of conversion is already changed, we cannot look at it and generate consumables as earlier. I could pass just the range there but then... what's really the difference compared to stopping the event? It would be there just to be there.

Of course, the other way would be to make it possible to consume only a part of a range and keep this information in `consumable`. But this would make the default converter more complicated and it would make writing new converters much more complicated.

If you have any good example where consumables would actually be useful, please comment. If not, I propose keeping it the way it is.

src/model/differ.js
scofalik: Not relevant anymore.

src/conversion/model-to-view-converters.js
scofalik: > Also it we skip custom callbacks we can get wrong position if this callbacks change offset.

This is a bit true but a bit not.

`remove` event is fired for every item one by one. This means that if given kind of element (like `listItem`) has custom mapping, it should also provide custom removing converter. It works like this now and I think we don't experience any bugs when it comes to removing lists?

So, I'd just remember about it and deal with it if it will prove to be troublesome.

pjasiun: Still, using the private API looks bad for me.

src/conversion/model-to-view-converters.js
scofalik: It seems that there is no real reason to have it read-only, other than it is read-only in `model.TreeWalker`. There, however, it makes more sense because we cache `this._visitedParent` -- `parent` of current `position` (it's for efficiency reasons).

However, I can propose an alternative solution for this function (not using `TreeWalker` at all).

PR #1192, opened 2017-11-28
Fixed T/897: Make Position/Range immutable.

jodator:

---

### Additional information

* This is updated https://github.com/ckeditor/ckeditor5-engine/pull/1175 pull with fixes in:
    - [CKEditor5-undo](https://github.com/ckeditor/ckeditor5-undo/commits/t/ckeditor5-engine/897)
    - [CKEditor5-typing](https://github.com/ckeditor/ckeditor5-typing/commits/t/ckeditor5-engine/897)

This fix was probably due to how `Selection.getRanges()` work. It was necessary to create a copy of a range in undo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,20 @@
 Changelog
 =========
 
+## [1.0.0-alpha.2](https://github.com/ckeditor/ckeditor5-engine/compare/v1.0.0-alpha.1...v1.0.0-alpha.2) (2017-11-14)
+
+### Bug fixes
+
+* `model.Range` will now be extended if it was collapsed and it was transformed by insertion. Closes [#1159](https://github.com/ckeditor/ckeditor5-engine/issues/1159). ([5f020b0](https://github.com/ckeditor/ckeditor5-engine/commit/5f020b0))
+* Prevent adding inline filler to non-editable content. Closes [#1170](https://github.com/ckeditor/ckeditor5-engine/issues/1170). ([07a01b1](https://github.com/ckeditor/ckeditor5-engine/commit/07a01b1))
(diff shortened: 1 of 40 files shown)


Reinmar:
> This fix was probably due to how Selection.getRanges() work. It was necessary to create a copy of a range in undo.

Why was it necessary? Since no other code may change a range, it should not be necessary.

Reinmar:
BTW, there was a branch `t/897` and it's still available. I think that it would be good to continue that branch instead of reverting a revert. This seems to be cleaner.

jodator:
> BTW, there was a branch t/897 and it's still available.

Well I hadn't changed anything in this branch and since `t/897` is merged into master:
https://github.com/ckeditor/ckeditor5-engine/compare/t/897-test?expand=1

> master is up to date with all commits from t/897-test. Try switching the base for your comparison.

Reinmar:
You'd need to rebase it.

Reinmar:
We decided to postpone this work, unfortunately. It turned out that there's more to do in the OT algorithms in order to really fix the problem. We need to figure a way to avoid moving so many positions there. We can either introduce something like internal, fast, mutable positions or simply making the API to modify existing positions protected so the OT code (which is part of the engine) will be able to use that API. It will, of course, need to initially clone all positions that it gets from the "world" but that happens now anyway.

jodator:
@Reinmar & @mlewand: I think that we could close this PR as it is probably so outdated that if we ever start to work on this we have to start working from scratch anyway.

If we'd like to retain history we could create a branch category for such cases, `reference/t/897` maybe?

Reinmar:
We don't clean old branches anyway (at least as of now), so I think we can just remove the PR. Plus, I agree that after such time there's less and less value in such a PR.

PR #1194, opened 2017-11-28
T/858 Refactoring: improve batch API

pjasiun:
Refactoring: improve batch API. Make it possible to use batch API on detached nodes. Use batch API instead of elements API in view to model conversion.


src/conversion/viewconversiondispatcher.js
```diff
@@ -203,24 +208,23 @@ export default class ViewConversionDispatcher {
...
+
+		for ( const viewChild of Array.from( input.getChildren() ) ) {
+			const modelChild = this._convertItem( viewChild, consumable, additionalData );
+
+			if ( modelChild instanceof ModelNode ) {
+				batch.append( modelChild, documentFragment );
+			} else if ( modelChild instanceof ModelDocumentFragment ) {
+				for ( const child of Array.from( modelChild ) ) {
+					batch.append( child, documentFragment );
+				}
```
pjasiun: Can't we just appen document fragment to another document fragment?

src/model/documentselection.js
```diff
@@ -723,15 +723,15 @@ function clearAttributesStoredInElement( changes, batch, document ) {
 
 	// `changes.range` is not set in case of rename, root and marker operations.
 	// None of them may lead to the element becoming non-empty.
-	if ( !changeParent || changeParent.isEmpty ) {
+	if ( !changeParent || changeParent.is( 'documentFragment' ) || changeParent.isEmpty ) {
```
pjasiun: Is it still needed? I see now change is fired only for docment changes.

oskarwrobel: I think you are right.

pjasiun:
In: https://github.com/ckeditor/ckeditor5-typing/compare/master...t/ckeditor5-engine/858

```
this._buffer.batch.insert( paragraph, limitElement );
```
Is not correct API usage. Should be:
```
this._buffer.batch.insert( paragraph, limitElement, 0 );
```
Or:
```
this._buffer.batch.append( paragraph, limitElement );
```
(does not matter if you insert at the beginning or at the end, the element is empty)

pjasiun:
Same in `ckeditor5-image` (`conversionApi.batch.insert( modelChildren, modelImage );`).

pjasiun:
https://github.com/ckeditor/ckeditor5-image/compare/master...t/ckeditor5-engine/858#diff-f91836fb3901004a578ed9233b54df72R177

Shouldn't be:
```
batch.appendElement( 'caption', item, 'end' );
```
(attributes parameter is optional, but the position was at the end)

oskarwrobel:
This url doesn't work but append is always at the end.

pjasiun:
Nevermid, my bad.

PR #1195, opened 2017-11-29
T/1193: Added "disableEngineDebug" function

### LINKED ISSUE #1193: How to disable debug tools?
When tests are executed as a single entry point, two tests fail:

https://github.com/ckeditor/ckeditor5-engine/blob/82335aee6426a7418416927fbe0db5d471c71465/tests/model/document/document.js#L113-L141

and

https://github.com/ckeditor/ckeditor5-engine/blob/82335aee6426a7418416927fbe0db5d471c71465/tests/model/document/document.js#L143-L153

image

In order to fix these tests, I can add a method `toJSON` into `operation` object (`toJSON: sinon.spy())`. But I'm not sure whether this solution is proper. I would like to disable engine debug tools but I guess it's tricky.

This ticket is a last blocker for https://github.com/ckeditor/ckeditor5-dev/issues/290.

pomek:

Feature: Engine debug tools can be easily disabled using `disableEngineDebug()` function. Closes #1193.


scofalik:
@ma2ciek could you check if all test passes for `npm t` and there are no unneeded logs and if so merge the PR?

ma2ciek:
Sure, I'll check it.

pomek:
Checked with single entry point:

image

ma2ciek:
I checked it with the current CKE5-dev setup. Was green too.

PR #1203, opened 2017-12-14
engine/model reorganization.

oskarwrobel:
`engine/model` reorganization.

List of related changes is here: https://github.com/ckeditor/ckeditor5/blob/t/engine/1186/mgit.json


src/model/documentselection.js
```diff
@@ -52,15 +52,24 @@ export default class DocumentSelection extends Selection {
 	 * Creates an empty live selection for given {@link module:engine/model/document~Document}.
 	 *
 	 * @param {module:engine/model/document~Document} document Document which owns this selection.
+	 * @param {module:engine/model/model~Model} model Data model.
 	 */
-	constructor( document ) {
+	constructor( document, model ) {
```
pjasiun: `model` should be enough. There is no need to have both model and document.

src/model/model.js
```diff
@@ -0,0 +1,223 @@
...
+	/**
+	 * Fires when all queued model changes are done.
+	 *
+	 * @see #change
+	 * @see #enqueueChange
+	 * @event changesDone
+	 */
+
+	/**
+	 * @event applyOperation
```
pjasiun: `applyOperation` event docs are missing. I will fix it.

PR #1204, opened 2017-12-14
Add fix for typing space on the FF on the edge of inline elements inside the paragraph

ma2ciek:

Fix: Added fix for typing space on the FF on the edge of inline elements inside the paragraph. Closes ckeditor/ckeditor5#692.

### DIFF
--- a/src/view/observer/mutationobserver.js
+++ b/src/view/observer/mutationobserver.js
@@ -204,8 +204,8 @@ export default class MutationObserver extends Observer {
 		}
 
 		for ( const viewElement of mutatedElements ) {
-			const viewChildren = Array.from( viewElement.getChildren() );
 			const domElement = domConverter.mapViewToDom( viewElement );
+			const viewChildren = Array.from( viewElement.getChildren() );
 			const newViewChildren = Array.from( domConverter.domChildrenToView( domElement ) );
 
 			// It may happen that as a result of many changes (sth was inserted and then removed),
--- a/src/view/observer/mutationobserver.js
+++ b/src/view/observer/mutationobserver.js
@@ -204,8 +204,8 @@ export default class MutationObserver extends Observer {
 		}
 
 		for ( const viewElement of mutatedElements ) {
-			const domElement = domConverter.mapViewToDom( viewElement );
 			const viewChildren = Array.from( viewElement.getChildren() );
+			const domElement = domConverter.mapViewToDom( viewElement );
 			const newViewChildren = Array.from( domConverter.domChildrenToView( domElement ) );
 
 			// It may happen that as a result of many changes (sth was inserted and then removed),
--- a/tests/view/observer/mutationobserver.js
+++ b/tests/view/observer/mutationobserver.js
@@ -271,14 +271,21 @@ describe( 'MutationObserver', () => {
 
 	// https://github.com/ckeditor/ckeditor5/issues/692 Scenario 1.
 	it( 'should handle space after inline filler at the end of container', () => {
-		const { view, selection } = parse( '<container:p>foo<attribute:b>[]</attribute:b></container:p>' );
+		const { view, selection } = parse(
+			'<container:p>' +
+				'foo' +
+				'<attribute:b>[]</attribute:b>' +
+			'</container:p>'
+		);
 
 		viewRoot.appendChildren( view );
 		viewDocument.selection.setTo( selection );
 
 		viewDocument.render();
 
-		const inlineFiller = domEditor.childNodes[ 2 ].childNodes[ 1 ].childNodes[ 0 ];
+		// Appended container is third in the tree.
+		const container = domEditor.childNodes[ 2 ];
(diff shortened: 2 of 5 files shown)


ma2ciek:
Could you test it too @Mgsy?

Mgsy:
> Could you test it too, @Mgsy?

I've checked it and your solution fixes this problem 👌

Reinmar:
Missing higher-level integration test for typing this space. With the current test the only thing we'll know is that `DomConverter` does what it seems to should do.

Reinmar:
Another thing is that I miss an explanation somewhere (in the ticket for instance) what exactly happens and why this change is needed. Looking at the code I'm not able to easily figure this out.

szymonkups:
We are investigating this issue further. Me or @ma2ciek will post more detailed information about this issue soon.

ma2ciek:
I've added 3 higher-level integration test for 3 scenarios from ckeditor/ckeditor5#692. They work on all browsers.

tests/view/observer/mutationobserver.js
```diff
@@ -267,6 +267,66 @@ describe( 'MutationObserver', () => {
...
+
+		const inlineFiller = domEditor.childNodes[ 2 ].childNodes[ 1 ].childNodes[ 0 ];
+
+		inlineFiller.data += ' ';
+
+		mutationObserver.flush();
+
+		expect( lastMutations.length ).to.equal( 1 );
+		expect( lastMutations[ 0 ].type ).to.equal( 'children' );
+		expect( lastMutations[ 0 ].node ).to.equal( selection.getFirstPosition().parent );
```
scofalik: Shouldn't we check what's in the mutation? Would be more complete. Here and below.

src/view/observer/mutationobserver.js
```diff
@@ -204,8 +204,8 @@ export default class MutationObserver extends Observer {
 		}
 
 		for ( const viewElement of mutatedElements ) {
-			const domElement = domConverter.mapViewToDom( viewElement );
 			const viewChildren = Array.from( viewElement.getChildren() );
+			const domElement = domConverter.mapViewToDom( viewElement );
```
scofalik: Why changing the order?

tests/view/observer/mutationobserver.js
```diff
@@ -267,6 +267,66 @@ describe( 'MutationObserver', () => {
...
+	// https://github.com/ckeditor/ckeditor5/issues/692 Scenario 1.
+	it( 'should handle space after inline filler at the end of container', () => {
+		const { view, selection } = parse( '<container:p>foo<attribute:b>[]</attribute:b></container:p>' );
+
+		viewRoot.appendChildren( view );
+		viewDocument.selection.setTo( selection );
+
+		viewDocument.render();
+
+		const inlineFiller = domEditor.childNodes[ 2 ].childNodes[ 1 ].childNodes[ 0 ];
```
scofalik: It'd be nice to comment what is in `domEditor.childNode[ 2 ]`. Here and below.

scofalik: Or how DOM looks like.

ma2ciek: Two children are appended in the `beforeEach()` to root. In these three tests, I'm only adding new elements to prevent some weird actions happening in the `MutationObserver`. Few tests for the inline filler does it. I'll add a note to these tests.

src/view/observer/mutationobserver.js
ma2ciek: At some point, I've mistakenly restored old code. I'll revert the original sequence.

tests/view/observer/mutationobserver.js
ma2ciek: Unfortunately, FF will have bogus `br` element in children, while other browsers won't have it at all. I'm not sure whether it safe to check it. But I'll try to improve that part.

ma2ciek:
> I've pushed a commit with updated comment in docs, because for me it wasn't easy to grasp what is going on. Let me know if I understood the problem and if the comment after my change is correct.

Your docs improvement is good and accurate.

> Other than that, I am not sure if those "integration tests" would satisfy @Reinmar :). I think they are lacking the ultimate result. I'd check if the editor data is okay at the end of them. I am not sure if mutationobserver.js is a proper place for such tests, though.

Well, I'll try to write a more integration test in `tickets/` directory. For me, it's the best place for such tests.

tests/view/observer/mutationobserver.js
scofalik: I see. Don't try too hard then. If you won't come up with something in a couple of minutes just add a comment like your post above.

scofalik:
`tickets/` will be fine. The test needs to be convincing, so it should check if the final result after inserting a space is valid.

ma2ciek:
I've just added 2 more integrational test.

Reinmar:
I've done additional manual testing and improved the tests.

PR #1205, opened 2017-12-14
T/1198: ViewElementConfig with helper converters.

### LINKED ISSUE #1198: Introduce ViewElementConfigDefinition interface and utility classes
This gathers requirements from both [heading](https://github.com/ckeditor/ckeditor5/issues/651) and https://github.com/ckeditor/ckeditor5-font/issues/2#issuecomment-347838197 features.

It was proposed by @Reinmar [in here](https://github.com/ckeditor/ckeditor5/issues/638#issuecomment-347271402).

The `ViewElementConfigDefinition` would server as an `view` configuration for some features that requires flexibility in parsing and should allow to greater customization of editor features like Heading and FontSize.

Full structure (whether values makes sense or not):
```js
{
    // Defines name of rendered element
    name: 'h1',
    // Optional classes
    class: 'ck-heading-one',
    // Optional styles:
    style: 'font-family:serif',
    alternatives: [
        { name: 'h2', /* styles, class, etc */ }
    ]
}
```

This would allow to:
- parse one or more HTML elements into model Element/Attribute,
- enable extensibility of existing features beyond what we can imagine right now.
- create util methods for conversions for features using those interfaces (right now Headings and Font features AFAIK).

Things to consider:
1. I'd go with `class` and `style` rather then `styles` and `classes` - such forms are already used in other places AFAICS.
2. How to define multiple classes/styles?
3. Are other attributes (`data-header=true`) needed?

---

Ad. 2. Examples:
```js
const singleClass = {
    name: 'span',
    class: 'foo'
}

const multipleClass = {
    name: 'span',
    class: [ 'foo', 'bar', 'baz' ]
}

const multipleStyles = {
    name: 'span',
    style: {
        'font-size': '12px',
        'font-weight': 'bolder'
    }
}
```
This might grow heavy so alternative configuration would by just strings with defined format:
```js
const multipleClass = {
    name: 'span',
    class: 'foo,bar,baz' // or as in HTML: 'foo bar baz'
}

const multipleStyles = {
    name: 'span',
    style: 'font-size:12px;font-weight: bolder'
}
```

As for second flavour we already have utility methods for parsing those:
https://github.com/ckeditor/ckeditor5-engine/blob/8eda4e5308f192cee58dd3d4b425f42769f66f17/src/view/element.js#L728
https://github.com/ckeditor/ckeditor5-engine/blob/8eda4e5308f192cee58dd3d4b425f42769f66f17/src/view/element.js#L804

jodator:

Enhancement: Introduce ViewElementDefinition. Closes #1198.

---

### Additional information

I'd love to hear about this implementation as I'm not sure if it's the right way to do this. Basically the most interesting file would be `configurationdefinedconverters.js` which has 4 helpers for defining V->M and M->V converters for both `ContainerElement` and `AttributeElmeent`. All methods have docs so please check the usage first.

What I'm not sure about is mostly naming of that file. The other thing is that I've used other helpers that already existed `buildModelConverter` and `buildViewConverter` to build upon this helper methods.

What I'm not sure right now is how to pass `ConversionDefinition` object. Right now I pass config directly as it should be compatible with `ConversionDefinition` object as I've used option 2 [from this](https://github.com/ckeditor/ckeditor5-engine/issues/1198#issuecomment-350707799) as @pjasiun  suggested. Maybe it would be nicer to group all those required properties on one object?

The other thins is the `model` part and how those util method accepts them as for `Element` I've used `model` string directly whereas for `Attribute` conversion I've used it as `attributeValue` as you pass `attributeName` as method parameter. The rationale behind this was how those two features were structured. In Heading there were many possible elements (hence many `model` attributes) while in Font I had only one attribute name (`fontSize`) with many size options as different values.

To see both in action check (or checkout `t/ckeditor5-font/2` branch on ckeditor5 repo):
* `Element` conversion: heading [branch](https://github.com/ckeditor/ckeditor5-heading/tree/t/ckeditor5-engine/1198),  [example test](https://github.com/ckeditor/ckeditor5-heading/compare/t/ckeditor5-engine/1198?expand=1#diff-a5a0a633207c3c796583fc5f0faaf0baR78)
* `Attribute` conversion: font [branch](https://github.com/ckeditor/ckeditor5-font/tree/t/2) and example [configuration tests](https://github.com/ckeditor/ckeditor5-font/compare/t/2?expand=1#diff-ac3f059d35f0909413ecc8f5dfdfc064R195).


jodator:
So the `model` part might be also like this:
```js
{
    model: {
        attributeName: 'foo',
        attributeValue: 'bar'
    }
}
// or
{
    model: {
        element: 'baz'
    }
}
```
...but it could lead to mis configuration for feature that expects attribute to have the same name.

So after second thought I'm for leaving `model`'s attribute meaning as is.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+  *		}, [ dispatcher ] );
+ *
+ * or with `acceptAlso` property to match many elements:
+ *
+ *		viewToModelElement( {
+ *			model: 'heading1',
+ *			view: {
+ *				name: 'h1'
+ *			},
+ *			acceptAlso: [
```
oskarwrobel: Maybe `view` could be an `array` in case of multiple definitions?

jodator: The `view` option defines default model->view conversion and it should always be one as it would be impossible to convert one model value to many (random?) view ;).

The `acceptAlso` is used as a way to convert existing markup to model in view->model conversion. As in this example you always want to convert to `h1` element but when loading content you want to map other elements to `heading1` also.

So TL;DR: I don't think so ;)

oskarwrobel: I mean view as array only for view -> model conversion. I wouldn't be afraid that someone will try to use it for converting to multiple random element in case of model -> view conversion.

oskarwrobel: Or I don't understand something. Maybe exactly the same definition will be used for both conversions.

jodator: Bu the configuration (like in heading) would be one: 

```js
return VirtualTestEditor
	.create( {
		plugins: [ HeadingEngine ],
		heading: {
			options: [
				{ model: 'paragraph', title: 'paragraph' },
				{
					model: 'heading1',
					view: 'h1',
					acceptsAlso: [
						{ name: 'p', attributes: { 'data-heading': 'h1' }, priority: 'high' }
					],
					title: 'User H1'
				}
			]
		}
	} )
`

Comment (oskarwrobel) [reply src/conversion/configurationdefinedconverters.js]: Ok, makes sense now :)

Comment (jodator) [reply src/conversion/configurationdefinedconverters.js]: And also the feature would use it as:

https://github.com/ckeditor/ckeditor5-heading/blob/b8ef80f86121f8a1ec20c66314bf09eb837cb818/src/headingengine.js#L85-L100

File: src/conversion/configurationdefinedconverters.js [hunk 2]
```diff
@@ -0,0 +1,310 @@
+/**
```
Reinmar: This file is named incorrectly. If it doesn't export a variable called `configurationDefinedConverters` then these are 3 separate words and hence the file should be called `configuration-defined-converters.js`.

I'm also unsure about naming this file after "configuration". What if I just want to use these helpers in my feature (which has no configuration)? Can we think about a different name?

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *			view: {
+ *				name: 'h1',
+ *				class: [ 'header', 'article-header' ],
+ *				attributes: {
+ *					data-header: 'level-1',
+ *				}
+ *			},
+ *		}, [ dispatcher ] );
+ *
+ * Above will generate an HTML tag:
```
Reinmar: "the following DOM element:"

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+import ViewContainerElement from '../view/containerelement';
+
+import buildModelConverter from './buildmodelconverter';
+import buildViewConverter from './buildviewconverter';
+
+/**
+ * Helper for creating model to view converter from model's element
+ * to {@link module:engine/view/containerelement~ContainerElement ViewContainerElement}. The `acceptAlso` property is ignored.
+ *
+ * You can define conversion as simple model element to view element conversion using simplified definition:
```
Reinmar: I think it's always "by using". "using" means that the definition is an unrelated condition. https://english.stackexchange.com/questions/217815/what-is-difference-between-using-and-by-using

The same applies below – "by defining".

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+
+import AttributeElement from '../view/attributeelement';
+import ViewContainerElement from '../view/containerelement';
+
+import buildModelConverter from './buildmodelconverter';
+import buildViewConverter from './buildviewconverter';
+
+/**
+ * Helper for creating model to view converter from model's element
+ * to {@link module:engine/view/containerelement~ContainerElement ViewContainerElement}. The `acceptAlso` property is ignored.
```
Reinmar: What does the last sentence do here? You're introducing the function, so don't go into details like what kind of property is ignored because you haven't introduced this property yet.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+	const { model: modelElement, viewDefinition } = parseConverterDefinition( definition );
+
+	buildModelConverter()
+		.for( ...dispatchers )
+		.fromElement( modelElement )
+		.toElement( () => ViewContainerElement.fromViewDefinition( viewDefinition ) );
+}
+
+/**
+ * Helper for creating view to model converter from view to model element. It will convert also all matched view elements defined in
```
Reinmar: The first sentence makes no sense. "view to model converter" and then "from view to model"?

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ * As a full-flavored definition:
+ *
+ *		viewToModelElement( {
+ *			model: 'heading1',
+ *			view: {
+ *				name: 'p',
+ *				attributes: {
+ *					'data-heading': 'true'
+ *				},
+ *				// it might require to define higher priority for elements matched by other features
```
Reinmar: Upper case and period.

> You may need to use high-priority listener to catch elements which are handled by other (usually – more generic) converters too.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *				priority: 'high'
+ *			}
+  *		}, [ dispatcher ] );
+ *
+ * or with `acceptAlso` property to match many elements:
+ *
+ *		viewToModelElement( {
+ *			model: 'heading1',
+ *			view: {
+ *				name: 'h1'
```
Reinmar: Why using the object syntax in this case?

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *			},
+ *			acceptAlso: [
+ *				{ name: 'p', attributes: { 'data-heading': 'level1' }, priority: 'high' },
+ *				{ name: 'h2', class: 'heading-main' },
+ *				{ name: 'div', style: { 'font-weight': 'bold', font-size: '24px' } }
+ *			]
+ *		}, [ dispatcher ] );
+ *
+ *    Above example will convert such existing HTML content:
```
Reinmar: The above
an existing view elements

src/conversion/configurationdefinedconverters.js
Reinmar: actually, isn't it a view element?

src/view/element.js
```diff
@@ -712,6 +712,48 @@ export default class Element extends Node {
 			( attributes == '' ? '' : ` ${ attributes }` );
 	}
 
+	/**
+	 * Creates element instance from provided viewElementDefinition.
+	 *
+	 * @param {module:engine/view/viewelementdefinition~ViewElementDefinition} viewElementDefinition
+	 * @returns {Element}
```
Reinmar: Nope

src/view/element.js
```diff
@@ -712,6 +712,48 @@ export default class Element extends Node {
...
+		const classes = viewElementDefinition.classes;
+
+		if ( classes ) {
+			attributes.class = Array.isArray( classes ) ? classes.join( ' ' ) : classes;
+		}
+
+		const stylesObject = viewElementDefinition.styles;
+
+		if ( stylesObject ) {
+			attributes.style = toStylesString( stylesObject );
```
Reinmar: Do you guys think that we could avoid this stringification? E.g. by setting styles after creating this element (then, view `Element` should then have `setStyles()` method)?

Uncareful stringification is unsafe.

src/view/element.js
```diff
@@ -712,6 +712,48 @@ export default class Element extends Node {
...
+	 * @param {module:engine/view/viewelementdefinition~ViewElementDefinition} viewElementDefinition
+	 * @returns {Element}
+	 */
+	static fromViewDefinition( viewElementDefinition ) {
+		const attributes = {};
+
+		const classes = viewElementDefinition.classes;
+
+		if ( classes ) {
+			attributes.class = Array.isArray( classes ) ? classes.join( ' ' ) : classes;
```
Reinmar: Same as with styles below.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *
+ *		<h1>A heading</h1>
+ *		<h2 class="heading-main">Another heading</h2>
+ *		<p data-heading="level1">Paragraph-like heading</p>
+ *		<div style="font-size:24px; font-weigh:bold;">Another non-semantic header</div>
+ *
+ * into `heading1` model element so after rendering it the output HTML will be cleaned up:
```
Reinmar: Again - avoid using "HTML".

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+	const { model: modelElement, viewDefinitions } = parseConverterDefinition( definition );
+
+	const converter = prepareViewConverter( dispatchers, viewDefinitions );
+
+	converter.toElement( modelElement );
+}
+
+/**
+ * Helper for creating model to view converter from model's attribute
+ * to {@link module:engine/view/attributeelement~AttributeElement AttributeElement}. The `acceptAlso` property is ignored.
```
Reinmar: Don't specify the second param of `{@link}` because what you did here is actually breaking the link. It should look like: [`AttributeElement`](x) but you made it [AttributeElement](x). If you won't specify the second param it will be rendered correctly. Specify the second param only if you want to write something custom like `config.foo.bar` (which would be better than the default `bar`). Also, make sure to always wrap code names in backticks.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ * You can define conversion as simple model element to view element conversion using simplified definition:
+ *
+ *		modelAttributeToViewAttributeElement( 'bold', {
+ *			model: 'true',
+ *			view: 'strong',
+ *		}, [ dispatcher ] );
+ *
+ * Or defining full-flavored view object:
+ *
+ *		modelAttributeToViewAttributeElement( 'fontSize' {
```
Reinmar: Missing comma.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *					'font-size': '1.2em'
+ *				}
+ *			},
+ *		}, [ dispatcher ] );
+ *
+ * Above will generate an HTML tag for model's attribute `fontSize` with a `big` value set:
+ *
+ *		<span style="font-size:1.2em;">...</span>
+ *
+ * @param {String} attributeName Attribute name from which convert.
```
Reinmar: "to convert" or something even more understandable "The name of the model attribute which should be converted."

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *
+ *		viewToModelAttribute( 'fontSize', {
+ *			model: 'big',
+ *			view: {
+ *				name: 'span',
+ *				style: {
+ *					'font-size': '1.2em'
+ *				}
+ *			}
+  *		}, [ dispatcher ] );
```
Reinmar: Something's wrong with indentation here.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *		<p>An example text with some big elements:
+ *			<span class="text-big>one</span>,
+ *			<span data-size="big>two</span>,
+ *			<span class="font font-huge>three</span>,
+ *			<span style="font-size:18px>four</span>
+ *		<p>
+ *
+ * into `fontSize` model attribute with 'big' value set so after rendering it the output HTML will be cleaned up:
+ *
+ *		<p>An example text with some big elements:
```
Reinmar: By convention, "An example..." should be in a new line.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *			<span class="text-big>one</span>,
+ *			<span class="text-big>two</span>,
+ *			<span class="text-big>three</span>,
+ *			<span class="text-big>four</span>
+ *		<p>
```
Reinmar: Missing `/`.

src/conversion/configurationdefinedconverters.js
Reinmar: `definition-based-converters.js`? :D

Reinmar: Or we need to find some proper name for these definitions. Something distinctive that we could use all around the code. Conversion specs, conversion maps, conversion rules, conversion... dunno. None of these is better than conversion definitions.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *		}, [ dispatcher ] );
+ *
+ *    Above example will convert such existing HTML content:
+ *
+ *		<p>An example text with some big elements:
+ *			<span class="text-big>one</span>,
+ *			<span data-size="big>two</span>,
+ *			<span class="font font-huge>three</span>,
+ *			<span style="font-size:18px>four</span>
+ *		<p>
```
Reinmar: Missing `/` (and below too).

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ * @param {String} attributeName Attribute name to which convert.
+ * @param {module:engine/conversion/configurationdefinedconverters~ConverterDefinition} definition A conversion configuration.
+ * @param {Array.<module:engine/conversion/viewconversiondispatcher~ViewConversionDispatcher>} dispatchers
+ */
+export function viewToModelAttribute( attributeName, definition, dispatchers ) {
+	const { model: attributeValue, viewDefinitions } = parseConverterDefinition( definition );
+
+	const converter = prepareViewConverter( dispatchers, viewDefinitions );
+
+	converter.toAttribute( () => ( {
```
Reinmar: Are we fine with this notation or we prefer a `return` statement?

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+
+	if ( attributes ) {
+		pattern.attribute = attributes;
+	}
+
+	return pattern;
+}
+
+/**
+ * Defines conversion details.
```
Reinmar: Missing links to where is all this used.

src/view/viewelementdefinition.jsdoc
```diff
@@ -0,0 +1,22 @@
...
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module engine/view/viewelementdefinition
+ */
+
+/**
+ * An object defining view element used for defining elements for conversion.
```
Reinmar: Some examples and links to places where this is used will help.

src/view/viewelementdefinition.jsdoc
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @typedef {Object} module:engine/view/viewelementdefinition~ViewElementDefinition
```
Reinmar: I don't know if it wouldn't be easier to define here that an element may be defined by a single string. Otherwise, everywhere this is used you'd need to write:

```
@property {String|module:engine/view/viewelementdefinition~ViewElementDefinition}
```

And you'll need to check the type of this property before deciding whether to use `ViewElement.createFromDefinition()` or the constructor (if a string was passed).

src/view/element.js
```diff
@@ -712,6 +712,48 @@ export default class Element extends Node {
 			( attributes == '' ? '' : ` ${ attributes }` );
 	}
 
+	/**
+	 * Creates element instance from provided viewElementDefinition.
+	 *
+	 * @param {module:engine/view/viewelementdefinition~ViewElementDefinition} viewElementDefinition
+	 * @returns {Element}
+	 */
+	static fromViewDefinition( viewElementDefinition ) {
```
Reinmar: Misses a verb and repeating "view" isn't necessary IMO. `ViewElement.createFromDefinition()`. should be fine.

src/conversion/configurationdefinedconverters.js
jodator: I might saw that in code so any advice on team-preferred code style is welcome :)

Reinmar:
I really like these converter definitions. They have a potential to handle 90% of use cases. I even think that with these converters being used by us and available for 3rd party use we can work on the rest of the utils without blocking people out there. Of course, all the lower-level conversion utils still require our attention, but we'll finally be able to write any documentation for conversion :)

Reinmar:
> What I'm not sure about is mostly naming of that file. The other thing is that I've used other helpers that already existed buildModelConverter and buildViewConverter to build upon this helper methods.

I commented on that in https://github.com/ckeditor/ckeditor5-engine/pull/1205#pullrequestreview-83801311.

Reinmar:
> What I'm not sure right now is how to pass ConversionDefinition object. Right now I pass config directly as it should be compatible with ConversionDefinition object as I've used option 2 from this as @pjasiun suggested. Maybe it would be nicer to group all those required properties on one object?

I don't understand what you mean.

BTW, I think that it's a `ConverterDefinition`. Conversion is a process for me and we define here a specific converter. cc @pjasiun

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *			model: 'heading1',
+ *			view: {
+ *				name: 'h1'
+ *			},
+ *			acceptAlso: [
+ *				{ name: 'p', attributes: { 'data-heading': 'level1' }, priority: 'high' },
+ *				{ name: 'h2', class: 'heading-main' },
+ *				{ name: 'div', style: { 'font-weight': 'bold', font-size: '24px' } }
+ *			]
+ *		}, [ dispatcher ] );
```
Reinmar: Please use a real dispatchers here – I mean, something like `editor.editing.fromView`. Otherwise it will be super confusing what this dispatcher is and where to take it from. In general, code samples should be as realistic as possible.

src/conversion/configurationdefinedconverters.js
Reinmar: Ooops. I haven't noticed that you mention here retrieving the data back. So it makes sense to call it HTML. But at the same time, maybe it's better to only mention here what ends up in the model. It will be clearer where and when the unification happen.

Reinmar:
> The other thins is the model part and how those util method accepts them as for Element I've used model string directly whereas for Attribute conversion I've used it as attributeValue as you pass attributeName as method parameter. The rationale behind this was how those two features were structured. In Heading there were many possible elements (hence many model attributes) while in Font I had only one attribute name (fontSize) with many size options as different values.

So, you did this:

```
+ *		modelAttributeToViewAttributeElement( 'bold', {
 + *			model: 'true',
 + *			view: 'strong',
 + *		}, [ dispatcher ] );
```

And this:

```
 + *		modelAttributeToViewAttributeElement( 'fontSize', {
 + *			model: 'big',
 + *			view: {
 + *				name: 'span',
 + *				styles: {
 + *					'font-size': '1.2em'
 + *				}
 + *			},
 + *		}, [ dispatcher ] );
```

I think that it's pretty ok. The name of the model attr should not be a part of the config so it's ok if you specify it as a separate param of this function. We can even simplify it because the `model: true` (it's a boolean IIRC) isn't necessary in the first case, so it could be ommited. So if we'll make the bold feature configurable you'd be defining two things:

```
config.bold = {
    view: ...
    acceptAlso: ...
}
```

This feature doesn't need to make the model part configurable cause it's just too simple.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	return converter;
+}
+
+// Converts viewDefinition to a matcher pattern.
+//
+// @param {module:engine/view/viewelementdefinition~ViewElementDefinition} viewDefinition
+// @returns {module:engine/view/matcher~Pattern}
+function definitionToPattern( viewDefinition ) {
```
pjasiun: Instead if translating syntaxes, we should make them the same (change matcher pattern to fit definition pattern).

pjasiun:
I spent 3 hours reading this code, usecases of the conversion builders, and custom conversions. All I'm sure at this point is that this is only a tip of the iceberg. We need a general clean-up of the conversion utils, on all level. For sure there is no good reasone to keep 2 declarative APIs (difinition and builder), I'm not sure if the view to model and model to view should be defined by 2 separate methods, I'm not sure if we these are all utils we need, etc.

I think that at this point we sould focus on unlocking features, and we will come back to it in January, together with the general conversion refatoring.

src/conversion/configurationdefinedconverters.js
```diff
@@ -0,0 +1,310 @@
...
+ *				style: {
+ *					'font-size': '1.2em'
+ *				}
+ *			}
+  *		}, [ dispatcher ] );
+ *
+ * or with `acceptAlso` property to match many elements:
+ *
+ *		viewToModelAttribute( 'fontSize', {
+ *			model: 'big',
```
pjasiun: I don't like the fact that attirbute name is passed as a separate parameter (what also @jodator notice in the PR description). It is also strange that in one case `model` is the model element name and in the other it is attibute value. 

At, the same time I undrestand that for feature configuration it is the best, because feature knows what is the name of the attribute.

However, I would go here with:

```js
model: {
   attributes: { 'fontSize': 'big' }
}
```

This is configuration of the conversion building tool. The plugin configuration does not need to be the same. Plugin can do some transformation adding obvious data to it, to make its configuration as simple as possible. However, this config will not fit all features need and we should not try to make it 1-1 with the config.

src/conversion/configurationdefinedconverters.js
jodator: Yeah... I go with that `definition-based-converts` and other cleanups. We can review it once again after I make all the requested changes.

src/view/element.js
pjasiun: Note that soon view writer will be the only proper way of changing model, so such methods will have to be moved.

jodator:
Most things were cleaned up. Remainig are:
1. extract `matcher~Pattern` definition (in docs) and add full support of it in `acceptAlso` (RegExp & callback remaining)
2. what to do with conflicting solutions for `ConverterDefinition#model`?

Ad. 1. It might be a follow up - but I can do it rather quickly now.
Ad. 2. It's either [we're OK with](https://github.com/ckeditor/ckeditor5-engine/pull/1205#issuecomment-352023681) it or [it is adding a full flavor configuration for](https://github.com/ckeditor/ckeditor5-engine/pull/1205#discussion_r157533878). (I'm for OK option here):

Full:
```js
const elementConfig = {
    view: 'h1',
    model: {
        element: 'heading1'
    }
};
const attributeConfig = {
    view: { style: { 'font-size': 'big' } },
    model: {
        attribute: 'fontSize',
        value: 'big'
    }
};
```
Current:
```js
const elementConfig = {
    view: 'h1',
    model: 'heading1'
};

const attributeConfig = {
    view: { style: { 'font-size': 'big' } },
    model: 'big' // A feature knows on what attribute it works ('fontSize')
};
```

And after looking how features will be configured it would be a lot easier to change only `model` value instead of reading which values of 'model' object to define. What can be more frustrating is that users could redefine attribute name in such (full) configuration which is also wrong.

I'm still for using `model` as either element name or attribute value - especially if we gonna let that be configurable.

/cc @pjasiun @Reinmar

Reinmar:
> Ad. 1. It might be a follow up - but I can do it rather quickly now.

If we can postpone it I'd rather do it later. Let's have features unblocked asap.

> And after looking how features will be configured it would be a lot easier to change only model value instead of reading which values of 'model' object to define. What can be more frustrating is that users could redefine attribute name in such (full) configuration which is also wrong.

Exactly. It's enough to have `model` exposed in the **config**. The API of those helper methods can be different if @pjasiun doesn't like the current API that much. Those are independent things.

src/conversion/definition-based-converters.js
```diff
@@ -0,0 +1,288 @@
...
+// @param {module:engine/conversion/definition-based-converters~ConverterDefinition} definition An object that defines view to model
+// and model to view conversion.
+// @returns {Object}
+function parseConverterDefinition( definition ) {
+	const model = definition.model;
+	const view = definition.view;
+
+	const viewDefinition = typeof view == 'string' ? { name: view } : view;
+
+	const viewDefinitions = Array.from( definition.acceptsAlso ? definition.acceptsAlso : [] );
```
pjasiun: Calling 2 variables in so similar way is asking for troubles. Why not leave "view" or "acceptsAlso"?

src/conversion/definition-based-converters.js
```diff
@@ -0,0 +1,288 @@
...
+		value: attributeValue
+	} ) );
+}
+
+// Prepares a {@link module:engine/conversion/definition-based-converters~ConverterDefinition definition object} for building converters.
+//
+// @param {module:engine/conversion/definition-based-converters~ConverterDefinition} definition An object that defines view to model
+// and model to view conversion.
+// @returns {Object}
+function parseConverterDefinition( definition ) {
```
pjasiun: This function gets JSON and return JSON, it parse nothing. It should be called "normaliseConverterDefinition".

src/conversion/definition-based-converters.js
pjasiun: As, I see that you adds `viewDefinition` to `viewDefinitions`. Still, I would call it better: `fromView` and `toView` (or `sourceView`/`targetView`).

src/view/element.js
```diff
@@ -713,6 +713,28 @@ export default class Element extends Node {
 			( attributes == '' ? '' : ` ${ attributes }` );
 	}
 
+	/**
+	 * Creates element instance from provided viewElementDefinition.
+	 *
+	 * @param {module:engine/view/viewelementdefinition~ViewElementDefinition} viewElementDefinition
+	 * @returns {module:engine/view/element~Element}
+	 */
+	static createFromDefinition( viewElementDefinition ) {
```
pjasiun: Since this method will be removed from the view element I would put it in `definition-based-converters.js`.

src/conversion/definition-based-converters.js
jodator: That's the word I was seeking then :)

pjasiun:
A agree with @Reinmar that there is no sense to do a deep review at this point. I only left a few comments to the code.

In fact, the only thing I don't like in the current definition for features is the `model` property for attribute values and the fact that the same property in the very similar definition means something very different (in one case it is node name, in the other, it is attribute value). It might be very misleading for people writing these plugins, or if we will want to support mixed configuration at some point. Maybe we could call it just `modelAttributeValue`?

jodator:
> Maybe we could call it just modelAttributeValue?

If so we should then be consequent and ad two: `modelAttribueValue` and `modelElementName` and make those in configuration of features as in previous comment.

Also I don't see how those configuration would be mixed in one feature.

pjasiun:
I talked with @jodator and realized that in fact, users do not need to know if they are defining element names or attribute values. For headings, which is the only case for element names so far, `heading1` might be element name or attribute value and it's only a matter of the implementation which should not be important for the end user.

TL;DR: `model` is enough.

jodator:
TL;DR: I win 😜

src/conversion/definition-based-converters.js
Reinmar: `normalizeConverterDefinition()` in AmE.

jodator:
Unfortunatelly I've found one bug. So please hold back this PR :(

jodator:
So the but was when changing attribute value from 'a' to 'b' in one step. I've caught it when dealing with `fontSize`. Probably defining `modelAttributeToview()` converters for each `attributeValue` was not a good way to do this.

I've changed `modelAttributeToViewAttributeElement()` to accept array of all configured attribute values so there will be only one converter for given `attribute` instead of `n` as previously.

PR #1206, opened 2017-12-15
T/1172-b Conversion refactor

### LINKED ISSUE #1172: Conversion refactor

scofalik:

Other: Conversion refactoring. Introduced `model.Differ`. Changes now will be converted after all changes in a change block are done. Closes #1172.

### DIFF
--- a/src/conversion/mapper.js
+++ b/src/conversion/mapper.js
@@ -110,23 +110,43 @@ export default class Mapper {
 	/**
 	 * Unbinds given {@link module:engine/view/element~Element view element} from the map.
 	 *
+	 * **Note:** view-to-model binding will be removed, if it existed. However, corresponding model-to-view binding
+	 * will be removed only if model element is still bound to passed `viewElement`.
+	 *
+	 * This behavior lets for re-binding model element to another view element without fear of losing the new binding
+	 * when the previously bound view element is unbound.
+	 *
 	 * @param {module:engine/view/element~Element} viewElement View element to unbind.
 	 */
 	unbindViewElement( viewElement ) {
 		const modelElement = this.toModelElement( viewElement );
 
-		this._unbindElements( modelElement, viewElement );
+		this._viewToModelMapping.delete( viewElement );
+
+		if ( this._modelToViewMapping.get( modelElement ) == viewElement ) {
+			this._modelToViewMapping.delete( modelElement );
+		}
 	}
 
 	/**
 	 * Unbinds given {@link module:engine/model/element~Element model element} from the map.
+	 *
+	 * **Note:** model-to-view binding will be removed, if it existed. However, corresponding view-to-model binding
+	 * will be removed only if view element is still bound to passed `modelElement`.
+	 *
+	 * This behavior lets for re-binding view element to another model element without fear of losing the new binding
+	 * when the previously bound model element is unbound.
 	 *
 	 * @param {module:engine/model/element~Element} modelElement Model element to unbind.
 	 */
 	unbindModelElement( modelElement ) {
 		const viewElement = this.toViewElement( modelElement );
 
-		this._unbindElements( modelElement, viewElement );
+		this._modelToViewMapping.delete( modelElement );
+
+		if ( this._viewToModelMapping.get( viewElement ) == modelElement ) {
+			this._viewToModelMapping.delete( viewElement );
+		}
 	}
 
 	/**
@@ -202,12 +222,15 @@ export default class Mapper {
 	 *
 	 * @fires modelToViewPosition
 	 * @param {module:engine/model/position~Position} modelPosition Model position.
+	 * @param {Boolean} [isPhantom=false] Should be set to `true` if the model position to map is pointing to a place
+	 * in model tree which no longer exists. For example, it could be an end of a removed model range.
 	 * @returns {module:engine/view/position~Position} Corresponding view position.
 	 */
-	toViewPosition( modelPosition ) {
+	toViewPosition( modelPosition, isPhantom = false ) {
 		const data = {
 			modelPosition,
-			mapper: this
+			mapper: this,
+			isPhantom
 		};
 
 		this.fire( 'modelToViewPosition', data );
@@ -293,18 +316,6 @@ export default class Mapper {
 	}
 
 	/**
-	 * Removes binding between given elements.
-	 *
-	 * @private
-	 * @param {module:engine/model/element~Element} modelElement Model element to unbind.
-	 * @param {module:engine/view/element~Element} viewElement View element to unbind.
-	 */
-	_unbindElements( modelElement, viewElement ) {
-		this._viewToModelMapping.delete( viewElement );
-		this._modelToViewMapping.delete( modelElement );
-	}
-
-	/**
 	 * Gets the length of the view element in the model.
 	 *
 	 * The length is calculated as follows:
@@ -368,7 +379,7 @@ export default class Mapper {
 	 *		We are in the text node so we can simple find the offset.
 	 *		<p>fo<b>ba|r</b>bom</p> -> expected offset: 2, actual offset: 2 -> position found
 	 *
-	 * @private
+	 * @protected
 	 * @param {module:engine/view/element~Element} viewParent Tree view element in which we are looking for the position.
 	 * @param {Number} expectedOffset Expected offset.
 	 * @returns {module:engine/view/position~Position} Found position.
@@ -460,19 +471,32 @@ export default class Mapper {
 	 *			}
 	 *		} );
 	 *
-	 * **Note:** keep in mind that custom callback provided for this event should use provided `data.modelPosition` only to check
-	 * what is before the position (or position's parent). This is important when model-to-view position mapping is used in
-	 * remove change conversion. Model after the removed position (that is being mapped) does not correspond to view, so it cannot be used:
+	 * **Note:** keep in mind that sometimes a "phantom" model position is being converted. "Phantom" model position is
+	 * a position that points to a non-existing place in model. Such position might still be valid for conversion, though
+	 * (it would point to a correct place in view when converted). One example of such situation is when a range is
+	 * removed from model, there may be a need to map the range's end (which is no longer valid model position). To
+	 * handle such situation, check `data.isPhantom` flag:
 	 *
-	 *		// Incorrect:
-	 *		const modelElement = data.modelPosition.nodeAfter;
-	 *		const viewElement = data.mapper.toViewElement( modelElement );
-	 *		// ... Do something with `viewElement` and set `data.viewPosition`.
+	 * 		// Assume that there is "customElement" model element and whenever position is before it, we want to move it
+	 * 		// to the inside of the view element bound to "customElement".
+	 *		mapper.on( 'modelToViewPosition', ( evt, data ) => {
+	 *			if ( data.isPhantom ) {
+	 *				return;
+	 *			}
 	 *
-	 *		// Correct:
-	 *		const prevModelElement = data.modelPosition.nodeBefore;
-	 *		const prevViewElement = data.mapper.toViewElement( prevModelElement );
-	 *		// ... Use `prevViewElement` to find correct `data.viewPosition`.
+	 *			const sibling = modelPosition.nodeBefore; // Might crash for phantom position that does not exist in model.
+	 *
+	 *			// Check if this is the element we are interested in.
+	 *			if ( !sibling.is( 'customElement' ) ) {
+	 *				return;
+	 *			}
+	 *
+	 *			const viewElement = data.mapper.toViewElement( sibling );
+	 *
+	 *			data.viewPosition = new ViewPosition( sibling, 0 );
+	 *
+	 *			evt.stop();
+	 *		} );
 	 *
 	 * **Note:** default mapping callback is provided with `low` priority setting and does not cancel the event, so it is possible to
 	 * attach a custom callback after default callback and also use `data.viewPosition` calculated by default callback
--- a/src/conversion/mapper.js
+++ b/src/conversion/mapper.js
@@ -222,15 +222,16 @@ export default class Mapper {
 	 *
 	 * @fires modelToViewPosition
 	 * @param {module:engine/model/position~Position} modelPosition Model position.
-	 * @param {Boolean} [isPhantom=false] Should be set to `true` if the model position to map is pointing to a place
+	 * @param {Object} [options] Additional options for position mapping process.
+	 * @param {Boolean} [options.isPhantom=false] Should be set to `true` if the model position to map is pointing to a place
 	 * in model tree which no longer exists. For example, it could be an end of a removed model range.
 	 * @returns {module:engine/view/position~Position} Corresponding view position.
 	 */
-	toViewPosition( modelPosition, isPhantom = false ) {
+	toViewPosition( modelPosition, options = { isPhantom: false } ) {
 		const data = {
 			modelPosition,
 			mapper: this,
-			isPhantom
+			isPhantom: options.isPhantom
 		};
 
 		this.fire( 'modelToViewPosition', data );
@@ -379,7 +380,7 @@ export default class Mapper {
 	 *		We are in the text node so we can simple find the offset.
 	 *		<p>fo<b>ba|r</b>bom</p> -> expected offset: 2, actual offset: 2 -> position found
 	 *
-	 * @protected
+	 * @private
 	 * @param {module:engine/view/element~Element} viewParent Tree view element in which we are looking for the position.
 	 * @param {Number} expectedOffset Expected offset.
 	 * @returns {module:engine/view/position~Position} Found position.
@@ -484,7 +485,8 @@ export default class Mapper {
 	 *				return;
 	 *			}
 	 *
-	 *			const sibling = modelPosition.nodeBefore; // Might crash for phantom position that does not exist in model.
+	 *			// Below line might crash for phantom position that does not exist in model.
+	 *			const sibling = data.modelPosition.nodeBefore;
 	 *
 	 *			// Check if this is the element we are interested in.
 	 *			if ( !sibling.is( 'customElement' ) ) {
--- a/src/conversion/model-to-view-converters.js
+++ b/src/conversion/model-to-view-converters.js
@@ -3,17 +3,16 @@
  * For licensing, see LICENSE.md.
  */
 
+import ModelRange from '../model/range';
+
 import ViewElement from '../view/element';
 import ViewAttributeElement from '../view/attributeelement';
 import ViewText from '../view/text';
 import ViewRange from '../view/range';
-import ViewPosition from '../view/position';
-import ViewTreeWalker from '../view/treewalker';
 import viewWriter from '../view/writer';
-import ModelRange from '../model/range';
 
 /**
- * Contains {@link module:engine/model/model model} to {@link module:engine/view/view view} converters for
+ * Contains model to view converters for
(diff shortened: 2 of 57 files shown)


src/conversion/mapper.js
```diff
@@ -368,7 +379,7 @@ export default class Mapper {
 	 *		We are in the text node so we can simple find the offset.
 	 *		<p>fo<b>ba|r</b>bom</p> -> expected offset: 2, actual offset: 2 -> position found
 	 *
-	 * @private
+	 * @protected
```
pjasiun: It should be still private.

src/conversion/mapper.js
```diff
@@ -202,12 +222,15 @@ export default class Mapper {
 	 *
 	 * @fires modelToViewPosition
 	 * @param {module:engine/model/position~Position} modelPosition Model position.
+	 * @param {Boolean} [isPhantom=false] Should be set to `true` if the model position to map is pointing to a place
+	 * in model tree which no longer exists. For example, it could be an end of a removed model range.
 	 * @returns {module:engine/view/position~Position} Corresponding view position.
 	 */
-	toViewPosition( modelPosition ) {
+	toViewPosition( modelPosition, isPhantom = false ) {
```
pjasiun: Instead of `isPhantom` -> `options.isPhantom`.

src/conversion/model-to-view-converters.js
```diff
@@ -91,6 +90,34 @@ export function insertText() {
...
+ *
+ * @returns {Function} Remove event converter.
+ */
+export function remove() {
+	return ( evt, data, conversionApi ) => {
+		// Find view range start position by mapping model position at which the remove happened.
+		const viewStart = conversionApi.mapper.toViewPosition( data.position );
+
+		const modelEnd = data.position.getShiftedBy( data.length );
+		const viewEnd = conversionApi.mapper.toViewPosition( modelEnd, true );
```
pjasiun: `const viewEnd = conversionApi.mapper.toViewPosition( modelEnd, { isPhantom: true } );` would be more readable.

src/conversion/model-to-view-converters.js
```diff
@@ -108,107 +135,95 @@ export function insertUIElement( elementCreator ) {
...
-		// Marker that is collapsed has consumable build differently that non-collapsed one.
-		// For more information see `addMarker` and `removeMarker` events description.
-		// If marker's range is collapsed - check if it can be consumed.
-		if ( markerRange.isCollapsed && !consumable.consume( markerRange, eventName ) ) {
-			return;
-		}
-
-		// If marker's range is not collapsed - consume all items inside.
-		for ( const value of markerRange ) {
-			if ( !consumable.consume( value.item, eventName ) ) {
```
pjasiun: Why consume is removed?

src/conversion/modelconversiondispatcher.js
```diff
@@ -131,64 +146,49 @@ export default class ModelConversionDispatcher {
...
+				continue;
+			}
 
-		if ( type == 'rename' && data.element.root.rootName == '$graveyard' ) {
-			return;
+			if ( entry.type == 'insert' ) {
+				this.convertInsert( Range.createFromPositionAndShift( entry.position, entry.length ) );
+			} else if ( entry.type == 'remove' ) {
+				this.convertRemove( entry.position, entry.length, entry.name );
+			} else {
```
pjasiun: Comment.

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+	 * @returns {Array.<Object>} Diff between old and new model tree state.
+	 */
+	getChanges() {
+		// Will contain returned results.
+		const diffSet = [];
+
+		// Check all changed elements.
+		for ( const element of this._changesInElement.keys() ) {
+			// If the element is inside other changed element, skip changes in this element.
+			if ( this._isInsertedOrRemoved( element ) ) {
```
pjasiun: Add comment that it's all about children.

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+				continue;
+			}
+
+			// Get changes for this element and sort them.
+			const changes = this._changesInElement.get( element ).sort( ( a, b ) => {
+				if ( a.offset === b.offset ) {
+					if ( a.type != b.type ) {
+						return a.type == 'remove' ? -1 : 1;
```
pjasiun: Comment needed.

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+// - 'a' for 'attribute' - when item at that position has it attributes changed.
+function _generateActionsFromChanges( oldChildrenLength, changes ) {
+	const actions = [];
+
+	let offset = 0;
+	let oldChildrenHandled = 0;
+
+	for ( const change of changes ) {
+		if ( change.offset > offset ) {
+			actions.push( ...'e'.repeat( change.offset - offset ).split( '' ) );
```
pjasiun: Comment "fill holes between end of one change and beginning of another with 'e'".

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+	return snapshot;
+}
+
+// Generates array of actions for given changes set.
+// It simulates what `diff` function does.
+// Generated actions are:
+// - 'e' for 'equal' - when item at that position did not change,
+// - 'i' for 'insert' - when item at that position was inserted,
+// - 'r' for 'remove' - when item at that position was removed,
+// - 'a' for 'attribute' - when item at that position has it attributes changed.
```
pjasiun: A single sample would be great.

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+		for ( const [ key, oldValue ] of oldAttributes ) {
+			// Check what is the new value of the attribute (or if it was removed).
+			const newValue = newAttributes.has( key ) ? newAttributes.get( key ) : null;
+
+			// If values are different (or attribute was removed)...
+			if ( newValue !== oldValue ) {
+				// Add diff item.
+				diffs.push( {
+					type: 'attribute',
+					position: range.start,
```
pjasiun: It's strange that I get both position and range.

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+			if ( old.type == 'attribute' ) {
+				if ( inc.offset <= old.offset ) {
+					old.offset += inc.howMany;
+				} else if ( inc.offset < oldEnd ) {
+					const howMany = old.howMany;
+
+					old.howMany = inc.offset - old.offset;
+
+					// Unshift to prevent further processing of this change.
+					changes.unshift( { type: 'attribute', offset: incEnd, howMany: howMany - old.howMany, count: this._changeCount++ } );
```
pjasiun: Comment why you add change in the first place is needed.

pjasiun: And since you unshift because you know that this function is called in the loop, the loop should be moved inside the function.

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+				} else if ( inc.offset < oldEnd ) {
+					if ( incEnd <= oldEnd ) {
+						const howMany = old.howMany;
+
+						old.howMany = inc.offset - old.offset;
+
+						const howManyAfter = howMany - old.howMany - inc.howMany;
+
+						// Unshift to prevent further processing of this change.
+						changes.unshift( { type: 'attribute', offset: inc.offset, howMany: howManyAfter, count: this._changeCount++ } );
```
pjasiun: Why?

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+		}
+
+		if ( inc.type == 'attribute' ) {
+			if ( old.type == 'insert' ) {
+				if ( inc.offset < old.offset && incEnd > old.offset ) {
+					if ( incEnd > oldEnd ) {
+						const changeItem = { type: 'attribute', offset: oldEnd, howMany: incEnd - oldEnd, count: this._changeCount++ };
+
+						for ( const oldChange of changes ) {
+							this._handleChange( changeItem, oldChange, changes );
```
pjasiun: Why?

src/model/differ.js
```diff
@@ -0,0 +1,774 @@
...
+			if ( change.newRange != null ) {
+				result.push( { name, range: change.newRange } );
+			}
+		}
+
+		return result;
+	}
+
+	/**
+	 * Calculates diff between old model tree state (before all the buffered operations) and the new model tree state
```
pjasiun: `(before all the buffered operations)` -> `(since the last {@link #reset})`?

src/conversion/model-to-view-converters.js
scofalik: For some time there were no consumables and I forgot to bring back this part of code.

src/model/differ.js
scofalik: `position` is useful for further sorting (as other changes has `position` too), but I will remove it from the final results. BTW. attribute change item also has `length` which is not used at all - I will remove it.

src/conversion/mapper.js
```diff
@@ -460,19 +471,32 @@ export default class Mapper {
...
-	 *		const viewElement = data.mapper.toViewElement( modelElement );
-	 *		// ... Do something with `viewElement` and set `data.viewPosition`.
+	 * 		// Assume that there is "customElement" model element and whenever position is before it, we want to move it
+	 * 		// to the inside of the view element bound to "customElement".
+	 *		mapper.on( 'modelToViewPosition', ( evt, data ) => {
+	 *			if ( data.isPhantom ) {
+	 *				return;
+	 *			}
+	 *
+	 *			const sibling = modelPosition.nodeBefore; // Might crash for phantom position that does not exist in model.
```
pjasiun: `modelPosition.nodeBefore` -> `data.modelPosition.nodeBefore`?

src/conversion/model-to-view-converters.js
```diff
@@ -496,65 +412,98 @@ export function highlightElement( highlightDescriptor ) {
...
-			viewElement.getCustomProperty( highlightHandlingMethod )( viewElement, addMarker ? descriptor : descriptor.id );
+			viewElement.getCustomProperty( 'addHighlight' )( viewElement, descriptor );
 		}
 	};
 }
 
 /**
- * Function factory, creates a default model-to-view converter for removing {@link module:engine/view/uielement~UIElement ui element}
- * basing on marker remove change.
+ * Function factory, creates a converter that converts model marker add/change/remove to the view.
```
pjasiun: Add, change?

src/conversion/modelconversiondispatcher.js
```diff
@@ -68,11 +83,11 @@ import extend from '@ckeditor/ckeditor5-utils/src/lib/lodash/extend';
  *			// Add the newly created view element to the view.
  *			viewWriter.insert( viewPosition, viewElement );
  *
- *			// Remember to stop the event propagation if the data.item was consumed.
+ *			// Remember to stop the event propagation.
  *			evt.stop();
  *		} );
  *
- * Callback that "overrides" other callback:
+ * Callback that overrides other callback:
```
pjasiun: I would remove this sample, it is too complicated, after reading it I understand less than before. Instead you could add a regular convresion for attribute->wrap (like bold).

src/model/differ.js
scofalik: It appears that I was wrong, both `length` and `position` are used in further processing but also both are not really needed later so they can be removed before changes are returned by `getChanges()`

PR #1216, opened 2017-12-27
T/1207 Refactored events fired by model classes

### LINKED ISSUE #1207: Clean-up model change events
After [introducing differ](https://github.com/ckeditor/ckeditor5-engine/issues/1172) we need to clean-up events model and document fire.

1. Current `change` and `changesDone` events should be removed.
2. API to add postfixers should be added ([see prototype here](https://gist.github.com/pjasiun/74b5c4e22b2c60ec17d82918a9663fd0#file-engine-js-L55-L73)) and all change listeners which are post fixers should be changed to post-fixers.
3. New `document.change` event should be introduced, which should be fired after all changes related to the single batch are done. Changes which are executed on the separate batch (i.e. autoformatter) should listen on this event and do `enqueueChange()`.
4. Listeners which really need to listen on each operation apply (live positions, differ) should now listen on `applyOperation`. 
5. We should consider removing data object which is returned by the operation execute, or change it. It made sense when the change event was fired base on the operation applied, but now it might not be needed anymore.
6. `model.change` event should be private (`model._change`), is should be used only by the document to fire `document.change`.

scofalik:

Other: Refactored events fired by model classes. Closes #1207.


scofalik:
PRs in multiple other repos are needed. I am on it.

scofalik:
Related PRs:
https://github.com/ckeditor/ckeditor5-autoformat/pull/44
https://github.com/ckeditor/ckeditor5-basic-styles/pull/62
https://github.com/ckeditor/ckeditor5-core/pull/111
https://github.com/ckeditor/ckeditor5-image/pull/163
https://github.com/ckeditor/ckeditor5-list/pull/86
https://github.com/ckeditor/ckeditor5-paragraph/pull/28
https://github.com/ckeditor/ckeditor5-typing/pull/127
https://github.com/ckeditor/ckeditor5-undo/pull/76
https://github.com/ckeditor/ckeditor5-upload/pull/76

scofalik:
I am not sure why Travis build fails. Maybe some dependencies in the package itself are outdated?

Reinmar:
You can push a branch to ckeditor/ckeditor5 and you'll see whether everything works there.

scofalik:
Looks like it passes although I am not 100% sure I did everything correctly :D https://travis-ci.org/ckeditor/ckeditor5/builds/322315528

Reinmar:
We usually modify `mgit.json` and I'm not sure whether modifying `package.json` had any effect.

scofalik:
I've pushed modified `mgit.json` and checked whether tests count is same as local and it is. It wasn't previously, without `mgit.json`.

Still, passing.

src/model/differ.js
```diff
@@ -168,9 +191,23 @@ export default class Differ {
...
 	 * another one, it will be on an earlier index in the diff set.
 	 *
+	 * Because calculating diff is a costly operation, the result is cached. If no new operation was buffered since the
+	 * previous {@link #getChanges} call, the next call with return the cached value.
+	 *
+	 * @params {Boolean} [includeChangesInGraveyard=false] If set to `true`, also changes that happened in graveyard root will be returned.
 	 * @returns {Array.<Object>} Diff between old and new model tree state.
 	 */
-	getChanges() {
+	getChanges( includeChangesInGraveyard = false ) {
```
pjasiun: I would use object `option`  with `includeChangesInGraveyard` as property instead of boolean   `includeChangesInGraveyard`. It would be safer for the future if we will introduce more properties and `getChanges( { includeChangesInGraveyard: true } )` is more readable than `getChanges( true )`.

src/model/differ.js
```diff
@@ -168,9 +191,23 @@ export default class Differ {
...
+		if ( this._cachedChanges ) {
+			// If there are cached changes, just return them instead of calculating changes again.
+			if ( !includeChangesInGraveyard ) {
+				// Filter out graveyard changes.
+				return this._cachedChanges.slice().filter( _changesInGraveyardFilter );
```
pjasiun: You check the root of every position in the cached changes. It's not very effective, especially if in most cases changes in the graveyard will not be used. Can't we just keep the information if the cashed changes include changes in the graveyard? Or have 2 caches (in one store change with and in the other without these in graveyard)?

src/model/document.js
```diff
@@ -80,28 +81,26 @@ export default class Document {
...
+		/**
+		 * Model differ object. Its role is to buffer changes done on model document and then calculate a diff of those changes.
+		 *
+		 * @readonly
+		 * @member {module:engine/model/differ~Differ}
+		 */
+		this.differ = new Differ();
+
+		/**
+		 * Post-fixer callbacks registered to the model.
```
pjasiun: "registered to the model document"?

src/model/document.js
```diff
@@ -278,6 +322,46 @@ export default class Document {
...
+	 *
+	 *		document.registerPostFixer( writer => {
+	 *			const changes = document.differ.getChanges();
+	 *
+	 *			// Check if the changes lead to an empty root in an editor.
+	 *			let applied = false;
+	 *
+	 *			for ( const entry of changes ) {
+	 *				if ( entry.type == 'remove' && entry.position.root.isEmpty ) {
+	 *					writer.insertElement( 'paragraph', ModelPosition.createAt( entry.position.root, 0 ) );
```
pjasiun: You can writer `writer.insertElement( 'paragraph', entry.position.root, 0 );`.

Also since, multi root editor is not very common, you can use `document.getRoot()`

src/model/document.js
```diff
@@ -341,48 +425,30 @@ export default class Document {
...
+					break;
+				}
+			}
+		} while ( wasFixed );
+	}
 
 	/**
-	 * Fired when all queued document changes are done. See {@link module:engine/model/model~Model#change}.
+	 * Fired after outermost {@link module:engine/model/model~Model#change change} or
+	 * {@link module:engine/model/model~Model#enqueueChange enqueueChange} block has been executed and
```
pjasiun: That's not true about the outermost `enqueueChange`. The change event will be fired after each enqueueChange block.

src/model/model.js
```diff
@@ -144,10 +163,11 @@ export default class Model {
 	 *			writer.insertText( 'foo', paragraph, 'end' );
 	 *		} );
 	 *
-	 * @fires event:change
-	 * @fires event:changesDone
+	 * `Batch` instance can be obtained from {@link module:engine/model/writer~Writer#batch the writer}.
+	 *
+	 * @fires event:_change
```
pjasiun: I'm not sure if we need to document protected event.

pjasiun:
I see that all postfixers define `let applied = false;` and return it at the end. Can we change it the way that postfixers return nothing (`undefined`) when they do nothing and return `true` when they do something? See https://github.com/ckeditor/ckeditor5-paragraph/pull/28/files#r159883141

pjasiun:
I see that `ChangeBuffer` listens on the `applyOpperation`, instead of the `change` event (see https://github.com/ckeditor/ckeditor5-typing/pull/127/files#r159888698). Is it because the change event has no information about the current batch? If so, maybe it should have `batch` argument.

src/model/differ.js
scofalik: I don't think that's that ineffective. It's just one check for each `entry`. `root` is just a simple property of `Range` or `Position` (it is not evaluated). I can make two caches if you really want to but I don't think this is that crucial.

src/model/document.js
scofalik: I'll leave `entry.position.root` as it is more correct.

src/model/model.js
scofalik: Do we need to document protected methods or private methods?

Yeah, that's what I thought :)

scofalik:
> I see that all postfixers define let applied = false; and return it at the end. Can we change it the way that postfixers return nothing (undefined) when they do nothing and return true when they do something? See https://github.com/ckeditor/ckeditor5-paragraph/pull/28/files#r159883141 and https://github.com/ckeditor/ckeditor5-image/pull/163/files#r159883461

Well, this depends on how you want the postfixer to work. I didn't want to make an early return because we mostly loop through a set of changes from differ. There can be multiple changes to be fixed. OTOH, when a change is fixed, all postfixers are fired again, so that other change would have another chance to be postfixed again. So I think that both solutions have their merits. If you don't like `let applied` (I don't), we can change it to `return true`.

> I see that ChangeBuffer listens on the applyOpperation, instead of the change event

Well, `applyOperation` is a direct substitute for `change` event, so it was the easier way to port this callback.

src/model/differ.js
pjasiun: We talked F2F, and @scofalik realized what is the problem here.

src/model/model.js
pjasiun: Documentation to the event is fine. What I don't like is the documentation to the method that it fires event. There is no way to mention that this event is protected and it will be displayed in the public documentation too. This is why I prefer to have no information that this event is fired here.

pjasiun:
> > I see that all postfixers define let applied = false; and return it at the end. Can we change it the way that postfixers return nothing (undefined) when they do nothing and return true when they do something? See https://github.com/ckeditor/ckeditor5-paragraph/pull/28/files#r159883141 and https://github.com/ckeditor/ckeditor5-image/pull/163/files#r159883461
>
> Well, this depends on how you want the postfixer to work. I didn't want to make an early return because we mostly loop through a set of changes from differ. There can be multiple changes to be fixed. OTOH, when a change is fixed, all postfixers are fired again, so that other change would have another chance to be postfixed again. So I think that both solutions have their merits. If you don't like `let applied` (I don't), we can change it to `return true`.

It depends what kind of post-fixer it is. If it fixes 30 attributes on one run it makes sense to improve performance and fix everything before restarting post-fixers. In other cases, it does not worth. As you notice all post-fixers will be called again anyway, so if you missed something because of the early return it will be fixed on the second round anyway.

> > I see that ChangeBuffer listens on the applyOpperation, instead of the change event
>
> Well, `applyOperation` is a direct substitute for `change` event, so it was the easier way to port this callback.
  
I think we should try to avoid using `applyOperation` as much as we can. Let's try to use `change` here.

scofalik:
I've left `let applied` in lists post fixer cause the callback is more complicated there. Other than that I fixed the rest.

src/model/document.js
```diff
@@ -278,6 +322,46 @@ export default class Document {
...
+	 *
+	 *			for ( const entry of changes ) {
+	 *				if ( entry.type == 'remove' && entry.position.root.isEmpty ) {
+	 *					writer.insertElement( 'paragraph', entry.position.root, 0 );
+	 *
+	 *					applied = true;
+	 *				}
+	 *			}
+	 *
+	 *			return applied;
```
pjasiun: I think we should remove this property from the sample to make it simpler.

src/model/document.js
```diff
@@ -341,48 +425,30 @@ export default class Document {
...
+	}
 
 	/**
-	 * Fired when all queued document changes are done. See {@link module:engine/model/model~Model#change}.
+	 * Fired after an {@link module:engine/model/model~Model#enqueueChange enqueueChange block} or the outermost
+	 * {@link module:engine/model/model~Model#change change block} has been executed and the document model tree was changed
+	 * during that block execution.
 	 *
-	 * @event changesDone
+	 * @event change
```
pjasiun: Change event has batch parameter which is not documented.

pjasiun:
There is also unsolve note in the upload repo that `change` event, not post-fixer, should be used there.

scofalik:
All done.

PR #1220, opened 2017-12-28
T/1211 Operations that do not operate on document should have baseVersion set to null

### LINKED ISSUE #1211: Operations which are not on the document should have base version null
Operations which does not change the document (are applied on detached trees) does not increment document version. It means that multiple operations may have the same base version, what is not correct.

The proper way would be to set `null` as the base version for operations which works on detached trees.

scofalik:

Other: Operations that do not operate on a document should have `baseVersion` set to `null`. Closes #1211.
Fixed: Markers again are properly converted in `engine.controller.DataController`.
Fixed: Markers are cleared now before an operation is applied to `model.Document` tree to fix scenarios where marker range could not be converted to the view after the model changed.

---

### Additional information

This PR bases on #1216. Please close that PR first.

### DIFF
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -135,10 +135,10 @@ export default class DataController {
 	 * @returns {String} Output data.
 	 */
 	stringify( modelElementOrFragment ) {
-		// model -> view
+		// Model -> view.
 		const viewDocumentFragment = this.toView( modelElementOrFragment );
 
-		// view -> data
+		// View -> data.
 		return this.processor.toData( viewDocumentFragment );
 	}
 
@@ -153,6 +153,7 @@ export default class DataController {
 	 * @returns {module:engine/view/documentfragment~DocumentFragment} Output view DocumentFragment.
 	 */
 	toView( modelElementOrFragment ) {
+		// First, convert elements.
 		const modelRange = ModelRange.createIn( modelElementOrFragment );
 
 		const viewDocumentFragment = new ViewDocumentFragment();
@@ -160,6 +161,17 @@ export default class DataController {
 
 		this.modelToView.convertInsert( modelRange );
 
+		if ( !modelElementOrFragment.is( 'documentFragment' ) ) {
+			// Then, if document element is converted, convert markers.
+			// Get only those markers that contain or are contained by the element.
+			const markers = _getMarkersRelativeToElement( modelElementOrFragment );
+
+			for ( const [ name, range ] of markers ) {
+				this.modelToView.convertMarkerAdd( name, range );
+			}
+		}
+
+		// Clear bindings so the next call to this method gives correct results.
 		this.mapper.clearBindings();
 
 		return viewDocumentFragment;
@@ -234,3 +246,28 @@ export default class DataController {
 }
 
 mix( DataController, ObservableMixin );
+
+// Helper function for converting part of a model to view.
+//
+// Takes a document element (element that is added to a model document) and checks which markers are inside it
+// and which markers are containing it. If the marker is intersecting with element, the intersection is returned.
+function _getMarkersRelativeToElement( element ) {
+	const result = [];
+	const doc = element.root.document;
+
+	if ( !doc ) {
+		return [];
+	}
+
+	const elementRange = ModelRange.createIn( element );
+
+	for ( const marker of doc.model.markers ) {
+		const intersection = elementRange.getIntersection( marker.getRange() );
+
+		if ( intersection ) {
+			result.push( [ marker.name, intersection ] );
+		}
+	}
+
+	return result;
+}
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -162,8 +162,8 @@ export default class DataController {
 		this.modelToView.convertInsert( modelRange );
 
 		if ( !modelElementOrFragment.is( 'documentFragment' ) ) {
-			// Then, if document element is converted, convert markers.
-			// Get only those markers that contain or are contained by the element.
+			// Then, if a document element is converted, convert markers.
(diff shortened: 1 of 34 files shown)


pjasiun:
I will wait with the review to close #1216.

src/controller/datacontroller.js
```diff
@@ -153,13 +153,25 @@ export default class DataController {
...
 		const modelRange = ModelRange.createIn( modelElementOrFragment );
 
 		const viewDocumentFragment = new ViewDocumentFragment();
 		this.mapper.bindElements( modelElementOrFragment, viewDocumentFragment );
 
 		this.modelToView.convertInsert( modelRange );
 
+		if ( !modelElementOrFragment.is( 'documentFragment' ) ) {
+			// Then, if document element is converted, convert markers.
+			// Get only those markers that contain or are contained by the element.
```
pjasiun: What other elements do we have?

src/controller/editingcontroller.js
```diff
@@ -110,6 +107,58 @@ export default class EditingController {
 		this.modelToView.on( 'selection', convertRangeSelection(), { priority: 'low' } );
 		this.modelToView.on( 'selection', convertCollapsedSelection(), { priority: 'low' } );
 
+		// Convert markers removal.
+		//
+		// Markers should be removed from the view before changes to model are applied. This is because otherwise
+		// it would be impossible to map some markers to view (if, for example, marker boundary parent got removed).
+		//
+		// `removedMarkers` keeps information which markers already has been removed to prevent removing them twice.
+		// (The second remove would not be at "the soonest moment" and it could crash.)
```
pjasiun: > (The second remove would not be at "the soonest moment" and it could crash.)
I don't get it.

src/controller/editingcontroller.js
```diff
@@ -110,6 +107,58 @@ export default class EditingController {
...
+				}
+
+				const markerRange = marker.getRange();
+
+				if ( _operationAffectsMarker( operation, marker ) ) {
+					// And if the operation in any way modifies the marker, remove the marker from the view.
+					removedMarkers.add( marker.name );
+					this.modelToView.convertMarkerRemove( marker.name, markerRange );
+
+					// TODO: This stinks but this is the safest place to have this code.
```
pjasiun: You mean all of these listeners? ;)

src/controller/editingcontroller.js
```diff
@@ -110,6 +107,58 @@ export default class EditingController {
...
+					this.model.document.differ.bufferMarkerChange( marker.name, markerRange, markerRange );
+				}
+			}
+		}, { priority: 'high' } );
+
+		// If a marker with given name was added again, remove it from `removedMarkers`. This is to prevent a bug in
+		// a situation when marker is removed, then added, then removed again (would not got removed if it is saved in `removedMarkers`).
```
pjasiun: Are you sure we need this? If a marker is added, we add it on "changes done", so it should not be removed immediately, only removed from differ.

src/model/writer.js
```diff
@@ -574,7 +564,9 @@ export default class Writer {
 		const delta = new RenameDelta();
 		this.batch.addDelta( delta );
 
-		const renameOperation = new RenameOperation( Position.createBefore( element ), element.name, newName, this.model.document.version );
+		const version = element.root.document ? this.model.document.version : null;
```
pjasiun: More natural is: `element.root.document ? element.root.document.version : null;`

src/model/writer.js
```diff
@@ -605,21 +597,24 @@ export default class Writer {
 		}
 
 		const copy = new Element( splitElement.name, splitElement.getAttributes() );
+		const insertVersion = splitElement.root.document ? this.model.document.version : null;
```
pjasiun: `splitElement.root.document ? splitElement.root.document.version : null;`

src/model/writer.js
```diff
@@ -670,16 +665,20 @@ export default class Writer {
 		const delta = new WrapDelta();
 		this.batch.addDelta( delta );
 
-		const insert = new InsertOperation( range.end, element, this.model.document.version );
+		const insertVersion = range.root.document ? this.model.document.version : null;
```
pjasiun: `range.root.document ? range.root.document.version : null;`

src/model/writer.js
```diff
@@ -707,26 +706,20 @@ export default class Writer {
 		this.batch.addDelta( delta );
 
 		const sourcePosition = Position.createFromParentAndOffset( element, 0 );
+		const moveVersion = sourcePosition.root.document ? this.model.document.version : null;
```
pjasiun: `sourcePosition.root.document ? sourcePosition.root.document : null;`

src/model/writer.js
```diff
@@ -939,6 +939,30 @@ function addMarkerOperation( writer, name, oldRange, newRange ) {
...
+// @param {module:engine/model/position~Position} position Position from which nodes are removed.
+// @param {Number} howMany Number of nodes to remove.
+// @param {module:engine/model/delta~Delta} delta Delta to add new operation to.
+// @param {module:engine/model/model~Model} model Model instance on which operation will be applied.
+function addRemoveOperation( position, howMany, delta, model ) {
+	let operation;
+
+	if ( position.root.document ) {
+		const doc = model.document;
+		const gyPosition = new Position( doc.graveyard, [ 0 ] );
```
pjasiun: `graveyardPosition`

src/controller/editingcontroller.js
scofalik: Yes.

src/controller/editingcontroller.js
scofalik: Wow, good catch! I've added a test and it is true that it is not only unnecessary but it's even harmful.

PR #1223, opened 2018-01-04
Prevent Writer from usage outside of the change block

### LINKED ISSUE #1212: Check if writer is not used outside a change block
The model writer should be used only inside `enqueueChange`, `change` or post fixer block. It can be passed to the inner function, but should not be stored in a variable and used outside of the proper block. Fortunately, we should be able to check it easily. 

Each writer method, when called, should check if `this.model._currentWritter == this` and throw exception if it is not true.

ma2ciek:

Other: Prevented `Writer` from usage outside of the `change` block. Closes #1212.


ma2ciek:
Few tests for errors miss. I'm on it.

src/model/writer.js
```diff
@@ -150,6 +150,10 @@ export default class Writer {
 	 * second parameter is a {@link module:engine/model/item~Item model item}.
 	 */
 	insert( item, itemOrPosition, offset ) {
+		if ( this.model._currentWriter !== this ) {
+			throw new CKEditorError( 'writer-detached-writer-tries-to-modify-model: Detached writer tries to modify the model' );
+		}
```
pjasiun: Wouldn't it be better to have it in a private method?

ma2ciek: Probably yes. I moved it to the separate private method.

ma2ciek:
Ready for the review.

PR #1242, opened 2018-01-16
T/1239 Allowed passing true as view.Matcher's attribute match to check if that attribute is set

### LINKED ISSUE #1239: Matcher atribute value wildcard
Right now, when we want to define the match that any value of an attribute is acceptable we use `/./`: https://github.com/ckeditor/ckeditor5-image/blob/2141a495f11413a1dfd610d1a1d5a5004ed766db/src/image/imageengine.js#L69

Or `/.?/`: https://github.com/ckeditor/ckeditor5-link/blob/e3d17d466139d038f4e37c9e55238caa7eacc08d/src/linkengine.js#L51

It is not very readable nor correct (`/./` will not accept an empty value).

Since the values of the view element attributes should be always strings (including `"true"`), we should improve Matcher so boolean attribute value `true` should mean that all values are accepted.

Note: check if view element attributes really use always strings.

scofalik:

Feature: Allowed passing `true` as `view.Matcher`'s attribute match to check if that attribute is set. Closes #1239.

---

### Additional information

Close only after PR #1241.

### DIFF
--- a/src/view/matcher.js
+++ b/src/view/matcher.js
@@ -41,8 +41,9 @@ export default class Matcher {
 	 *
 	 *		matcher.add( {
 	 *			attribute: {
-	 *				title: 'foobar',
-	 *				foo: /^\w+/
+	 *				title: 'foobar',	// Attribute title should equal 'foobar'.
+	 *				foo: /^\w+/,		// Attribute foo should match /^\w+/ regexp.
+	 *				bar: true			// Attribute bar should be set (can be empty).
 	 *			}
 	 *		} );
 	 *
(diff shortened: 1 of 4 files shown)


src/view/matcher.js
```diff
@@ -95,8 +96,10 @@ export default class Matcher {
 	 * {@link module:engine/view/matcher~Matcher#match match} or {@link module:engine/view/matcher~Matcher#matchAll matchAll} methods.
 	 * @param {String|RegExp} [pattern.name] Name or regular expression to match element's name.
 	 * @param {Object} [pattern.attribute] Object with key-value pairs representing attributes to match. Each object key
-	 * represents attribute name. Value under that key can be either a string or a regular expression and it will be
-	 * used to match attribute value.
+	 * represents attribute name. Value under that key can be either:
+	 * * `true` - then attribute is just required (can be empty),
+	 * * a string - then attribute has to be equal, or
+	 * * a regular expression - then attribute has to match the expression.
 	 * @param {String|RegExp|Array} [pattern.class] Class name or array of class names to match. Each name can be
```
pjasiun: @param {true|String|RegExp|Array}

PR #1244, opened 2018-01-18
[WIP] Focus the new editing host to workaround FF's errors

Reinmar:

Fix: Fixed a bug where Firefox would throw an `NS_ERROR_FAILURE` error when moving selection from a nested editable to the root editable. Closes ckeditor/ckeditor5#721.

---

### Additional information

This is WIP. @Mgsy could you check whether this PR doesn't cause some regressions? Unfortunately, you'd need to check all browsers and various scenarios because such a change has a huge spectrum of potential regressions.


Mgsy:
I've checked your fix and it works fine. I haven't noticed any regression.

Reinmar:
OK, then let's add a manual test and automated test (if possible). I leave that to you, @ma2ciek

ma2ciek:
Should I add the manual test to the engine or to the image package?

Reinmar:
Here, in the engine.

Reinmar:
If it was a normal manual ticket test putting it into the image might make more sense. But in this case, when this is engine's feature (nested editables), let's have it here.

ma2ciek:
Ok, so I'll also add `ckeditor5-image` to `ckeditor5-engine`'s `devDeps`.

Reinmar:
That can be a strictly engine's test. With a pseudo widget.

ma2ciek:
I've added a manual test to that bugfix.

PR #1247, opened 2018-01-19
Simplified conversion by dropping `selectionAttribute` event.

### LINKED ISSUE #1238: Try to automate selection attribute converstion
Right now when you define an attribute to element converter you need to define also selection attribute converter. It is something you need to remember about and it is not very intuitive.

https://github.com/ckeditor/ckeditor5-engine/blob/47291ebfee56cb20a110f6c38407476159dfbd69/src/conversion/buildmodelconverter.js#L259-L260

Since these converters are very similar, we should try to merge them to handle everything using a single event.

scofalik:

Other: Simplified model to view selection conversion. Closes #1238.

---

### Additional information

Changed: `view.Writer#wrap` and `view.Writer#wrapPosition` got unified.
Changed: `view.Writer#wrap` has optional, third parameter, a view selection.
Changed: `view.Writer#wrap` will modify view selection if it is collapsed and wrap is at the selection position.


src/conversion/model-to-view-converters.js
```diff
@@ -334,7 +338,11 @@ export function wrap( elementCreator ) {
 
 		// Then wrap with the new wrapper.
 		if ( data.attributeNewValue !== null ) {
-			viewWriter.wrap( viewRange, newViewElement );
+			if ( data.item instanceof ModelSelection ) {
+				viewWriter.wrap( conversionApi.viewSelection.getFirstRange(), newViewElement, conversionApi.viewSelection );
+			} else {
+				viewWriter.wrap( viewRange, newViewElement );
+			}
```
pjasiun: You remember that we agreed that this condition should be earlier, don't you?

src/conversion/modelconversiondispatcher.js
```diff
@@ -534,10 +510,15 @@ export default class ModelConversionDispatcher {
 	 * * there is only one event,
 	 * * consumables object includes marker range with event name.
 	 *
+	 * If selection inside a marker is converted:
+	 * * the is only one event,
```
pjasiun: there

pjasiun: And I don't get what does this sentence mean here.

scofalik: It's in the same vein as above. It means that there is just one `attribute:` event, not multiple, like when non-collapsed range is converted.

src/conversion/model-to-view-converters.js
scofalik: Now I do.

src/view/writer.js
```diff
@@ -457,29 +456,64 @@ export function move( sourceRange, targetPosition ) {
 
 /**
  * Wraps elements within range with provided {@link module:engine/view/attributeelement~AttributeElement AttributeElement}.
+ * If a collapsed range is provided, it will be wrapped only if it is equal to view selection.
```
pjasiun: Will it?

src/view/writer.js
```diff
@@ -457,29 +456,64 @@ export function move( sourceRange, targetPosition ) {
...
  *
  * Throws {@link module:utils/ckeditorerror~CKEditorError} `view-writer-invalid-range-container`
  * when {@link module:engine/view/range~Range#start}
  * and {@link module:engine/view/range~Range#end} positions are not placed inside same parent container.
+ *
  * Throws {@link module:utils/ckeditorerror~CKEditorError} `view-writer-wrap-invalid-attribute` when passed attribute element is not
  * an instance of {module:engine/view/attributeelement~AttributeElement AttributeElement}.
  *
+ * Throws {@link module:utils/ckeditorerror~CKEditorError} `view-writer-wrap-nonselection-collapsed-range` when passed range
+ * is collapsed and different than view selection.
```
pjasiun: Nope.

src/view/writer.js
scofalik: Not anymore.

src/conversion/modelconversiondispatcher.js
Reinmar: Please, please, please, pretty please – a paragraph should be followed by a blank line. Write markdown just like you'd like it to render (and you always have a margin between a paragraph and a list).

scofalik: They feel so disconnected though :O.

Reinmar: That's how it should be. If they miss each other they can send each other letters.

pjasiun:
If you search for `selectionAttribute` in the whole project you will find some docs to fix.

PR #1252, opened 2018-01-24
Introduce Schema#addChildCheck and Schema#addAttributeCheck

### LINKED ISSUE #1234: Schema: add helpers for callback checks
After the `Schema` refactor, we are able to add custom callbacks to `checkChild` and `checkAttribute` methods, which are decorated. This lets us define very custom rules (mostly for disabling stuff).

This was a very needed feature, but I'd like to propose some improvements. Instead of using raw decorator mechanism, we could have helpers (or _callback registering functions_) that would do some additional stuff. So I propose `Schema#addChildCheck()` and `Schema#addAttributeCheck()`.

How does the usual callback look like now? Straight outta docs:

```
schema.on( 'checkChild', ( evt, args ) => {
	const context = args[ 0 ];
	const child = args[ 1 ];

	// Normalize child too (it can be a string or a node).
	const childDefinition = schema.getDefinition( child );

	if ( context.endsWith( 'bar foo' ) && childDefinition.name == 'listItem' ) {
		// Prevent next listeners from being called.
		evt.stop();
 		// Set the checkChild()'s return value.
 		evt.return = true;
 	}
 }, { priority: 'high' } );
```

What is wrong with it?

1. Parameters are not passed directly. You need to retrieve them from `args` array. There is a quick workaround for it, to use `( evt, [ context, child ] ) => {` instead of `( evt, args ) => {`, but it would be still better if we could use just `( context, child ) => {`.

2. `child` is not normalized. Right now, `context` is normalized in `highest` priority callback. `child` could be normalized to `childDefinition` there as well, so it is also doable and can be transparent.

3. `evt.stop()` and `evt.return = true` have to be used in order to stop further checking and return that the check didn't pass. It is a bit unelegant, it would be better if the callback could just return `false`.

4. `{ priority: 'high' }. I thought that every callback has to be added with `'high'` priority, but in fact it doesn't matter, as the custom check can easily be called after the default check. It doesn't matter much. Still, this is used in, I think, all examples, while the truth is that neither check should be more or less important than other.

To sum up.

Problems 1, 2 and 4 can be easily solved. Problem 3 cannot be easily solved cause that's how events work. However, if we let the callback just `return false;`, we:
1. would have to either `return true;` at the end of each callback, or treat `undefined` return value same as `true` (so check didn't pass only if callback returned exactly `false`).
2. would lose a possibility to "force check pass" by doing `evt.stop(); evt.return = true;`. However this is a very unlikely that we will ever want to do it.

After proposed changes, the callback from the beginning of this post would look like this:

```
schema.addChildCheck( ( context, childDefinition ) => {
	if ( context.endsWith( 'bar foo' ) && childDefinition.name == 'listItem' ) {
		return false;
 	}
 } );
```

Alternatively, without introducing callback registerting function, we can do it like this:

```
// `childDefinition` is normalized in previous callback.
schema.on( 'checkChild', ( evt, [ context, childDefinition ] ) => {
	if ( context.endsWith( 'bar foo' ) && childDefinition.name == 'listItem' ) {
		// Prevent next listeners from being called.
		evt.stop();
 		// Set the checkChild()'s return value.
 		evt.return = true;
 	}
 } );
```

So it is not bad either. However I thought that there are already multiple things to be fixed so we maybe should think about introducing such a helper.

Reinmar:

Internal: Introduced `Schema#addChildCheck()` and `Schema#addAttributeCheck()`. Closes #1234.

---

### Additional information

Marked as internal because this whole API is new.

I'll add accompanying PRs with changes in other repos.

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -423,8 +423,8 @@ export default class Schema {
 	 * Example:
 	 *
 	 *		// Disallow heading1 directly inside a blockQuote.
-	 *		schema.addChildCheck( ( ctx, childDefinition ) => {
-	 *			if ( ctx.endsWith( 'blockQuote' ) && childDefinition.name == 'heading1' ) {
+	 *		schema.addChildCheck( ( context, childDefinition ) => {
+	 *			if ( context.endsWith( 'blockQuote' ) && childDefinition.name == 'heading1' ) {
 	 *				return false;
 	 *			}
 	 *		} );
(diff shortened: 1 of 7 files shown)


Reinmar:
Other branches:

* https://github.com/ckeditor/ckeditor5-basic-styles/compare/t/ckeditor5-engine/1234
* https://github.com/ckeditor/ckeditor5-block-quote/compare/t/ckeditor5-engine/1234
* https://github.com/ckeditor/ckeditor5-image/compare/t/ckeditor5-engine/1234
* https://github.com/ckeditor/ckeditor5-link/compare/t/ckeditor5-engine/1234
* https://github.com/ckeditor/ckeditor5-list/compare/t/ckeditor5-engine/1234
* https://github.com/ckeditor/ckeditor5-paragraph/compare/t/ckeditor5-engine/1234
* https://github.com/ckeditor/ckeditor5-typing/compare/t/ckeditor5-engine/1234

Ready for a review.

PR #1253, opened 2018-01-24
Introduce `Schema#checkMerge()`

### LINKED ISSUE #1251: Introduce `Schema#checkMerge()`
`Schema#checkMerge()` will check whether two elements can be merged.

See more: https://github.com/ckeditor/ckeditor5-engine/issues/1240.

pomek:

Internal: Schema#checkMerge(). Closes #1251.

---

### Additional information

Marked as internal because this whole API is new.

### DIFF
--- a/tests/model/schema.js
+++ b/tests/model/schema.js
@@ -645,13 +645,14 @@ describe( 'Schema', () => {
 		beforeEach( () => {
 			schema.register( '$root' );
 			schema.register( '$block', {
-				allowIn: '$root'
+				allowIn: '$root',
+				isBlock: true
+			} );
+			schema.register( '$text', {
+				allowIn: '$block'
 			} );
 			schema.register( 'paragraph', {
 				allowWhere: '$block'
-			} );
-			schema.register( '$text', {
-				allowIn: 'paragraph'
 			} );
 			schema.register( 'blockQuote', {
 				allowWhere: '$block',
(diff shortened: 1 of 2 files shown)


tests/model/schema.js
```diff
@@ -641,6 +641,48 @@ describe( 'Schema', () => {
...
+		} );
+
+		it( 'returns true if a block can be merged with other block', () => {
+			const listItem = new Element( 'listItem' );
+			const listItemToMerge = new Element( 'listItem', null, [
+				new Text( 'xyz' )
+			] );
+
+			expect( schema.checkMerge( listItem, listItemToMerge ) ).to.equal( true );
+		} );
```
Reinmar: Please add a test where `elementToMerge` has more children and only one of them (not the first one) will not be allowed in the `baseElement`.

Also, another test where `baseElement` already contains disallowed elements, but `elementToMerge` contains only allowed ones. This test will verify stability of the implementation. Such situation is by definition invalid, but current (sufficient) implementation of `checkMerge()` wouldn't reject it, so let's be sure that future implementations won't do that too.

pomek: Missing tests have been added in ddf9e54.

Reinmar:
I realised that `Writer#merge()` accepts a position, so it'd be good if `checkMerge()` also accepted a position (in addition to two elements).

pomek:
@Reinmar, done. Can you check again?

Reinmar:
One thing only – "before/after **the** merge position". I'll fix it on master.

Issue #1250, opened 2018-01-23
Modify SchemaContext


pjasiun:
At this moment, the schema is not mutable and there is no easy way to create one schema based on another. The simplest way right now is to create a new definition and the new schema based on it. Even the current docs do not look very simple:

https://github.com/ckeditor/ckeditor5-engine/blob/d00c24c0fefad47f6a48abe8bd43c71ee8fddf58/src/model/schema.js#L1018

You also, need to remember about all these fixes done in the `SchemaContext` creator if you want to get items from the position manually:

https://github.com/ckeditor/ckeditor5-engine/blob/d00c24c0fefad47f6a48abe8bd43c71ee8fddf58/src/model/schema.js#L877-L890

This is why I think that schema `SchemaContext` should have methods like `add`/`delete` or `push`/`pop` to create one context based one another. These methods could modify the context object, but if we want to keep a context immutable (as @Reinmar suggested) these methods could return the new context object (do not change the current one).

Also, `checkChild` and `checkAttribute` should accept `SchemaContext` instance. To do so, we could make `SchemaContext` constructor accepts `SchemaContext` instance as a parameter. In such case, the same (or even the same instance) should be returned from the constructor.

Reinmar:
Off topic:

```js
	constructor( context ) {
		if ( Array.isArray( context ) ) {
			if ( context[ 0 ] && typeof context[ 0 ] != 'string' && context[ 0 ].is( 'documentFragment' ) ) {
				context.shift();
			}
		}
		else {
			// `context` is item or position.
			// Position#getAncestors() doesn't accept any parameters but it works just fine here.
			context = context.getAncestors( { includeSelf: true } );

			if ( context[ 0 ].is( 'documentFragment' ) ) {
				context.shift();
			}
		}

		this._items = context.map( mapContextItem );
	}
```

Why is the `documentFragment` check implemented twice? It can be done once, after the `if-else` part, once `context` is an array.

Reinmar:
> Also, checkChild and checkAttribute should accept SchemaContext instance. To do so, we could make SchemaContext constructor accepts SchemaContext instance as a parameter. In such case, the same (or even the same instance) should be returned from the constructor.

I'd go with the same instance. Since it's immutable anyway, that doesn't matter.

pjasiun:
We decided to call it `concat`and it will accept any form of the schema definition, which can be string now too (it is useful also in other cases). Such string will be treated as a name of a single element.

[...]

PR #1256, opened 2018-01-29
Refactored view->model conversion to use position.

### LINKED ISSUE #1213: Convert view to model using positions
Context: view to model conversion.

Right now converters build nodes which are passed to the parent converters which build parent element with children it gets as an array of elements. There are couple problems with this:
 - the structure is not built until all top-level elements are converted, we can not use live ranges during the conversion because of it,
 - we need to pass "context" object to children which know nothing more about the structure than this context, can not check siblings, also context needs to be updated in every converter.

After closing https://github.com/ckeditor/ckeditor5-engine/issues/858 we should try to fix it.

Parent converter should pass to the child converter position which is the conversion context. Child converter should return the range with the converted fragment. Parent converter can execute its action before or after child converters.

Consider an example: `<p><b>foo</b></p>`

- `<p>` converter creates `<paragraph>` and pass the position inside it to `<b>` converter (`<paragraph>^</paragraph>`),
- `<b>` converter call `foo` converter first and pass the position it gets from `<p>` converter,
- `foo` converter insert text on the position and returns range with inserted element (`<paragraph>[foo]</paragraph>`),
- `<b>` will add attribute to the range.

oskarwrobel:

Feature: Convert view to model using position. Closes #1213. Closes https://github.com/ckeditor/ckeditor5-engine/issues/1250.

BREAKING CHANGE: `DataController#parse`, `DataController#toModel`, `ViewConversionDispatcher#convert` gets `SchemaContextDefinition` as a contex instead of `String`.
BREAKING CHANGE: `ViewConversionApi#splitToAllowedParent` has been introduced.
BREAKING CHANGE: `ViewConversionApi#storage` has been introduced.
BREAKING CHANGE: `ViewConsumable` has been merged to `ViewConversionApi`.
BREAKING CHANGE: Format od data object passed across conversion callback has been changed.
BREAKING CHANGE: `Schema#findAllowedParent` has been introduced.
BREAKING CHANGE: `SchemaContext#concat` has been introduced.

---

Constelation: https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-engine/1213

Related PR's:
- https://github.com/ckeditor/ckeditor5-paragraph/pull/29
- https://github.com/ckeditor/ckeditor5-image/pull/169
- https://github.com/ckeditor/ckeditor5-list/pull/91

Related branches (small changes):
- https://github.com/ckeditor/ckeditor5-clipboard/tree/t/ckeditor5-engine/1213


src/conversion/viewconversiondispatcher.js
```diff
@@ -137,55 +136,71 @@ export default class ViewConversionDispatcher {
...
+			// Top element of context tree is marked by a `isContextTree` attribute.
+			const position = contextToPosition( context, writer );
+
 			// Store writer in current conversion as a conversion API.
 			this.conversionApi.writer = writer;
 
-			this.fire( 'viewCleanup', viewItem );
+			// Create set for split elements. We need to remember this elements, because at the end of conversion
+			// we want to remove all empty elements that was created as a result of the split.
+			this.conversionApi.splitElements = new Set();
```
oskarwrobel: I'm going to move `splitElements` from conversionApi to private `ViewConversiondispatcher` property.

src/conversion/buildviewconverter.js
```diff
@@ -291,31 +292,62 @@ class ViewConverterBuilder {
...
+					// after:  <allowed><notAllowed></notAllowed>[<converted></converted><notAllowed>]</notAllowed></allowed>
+					if ( splitResult.endElement ) {
+						output.end = Position.createAt( splitResult.endElement );
+
+					// When element is converted on target position (without splitting) we need to move range
+					// after this element but we need to take into consideration that children could split our
+					// element, so we need to move range after parent of the last converted child.
+					//
+					// after: <allowed>[]</allowed>
+					// before: <allowed>[<converted><child></child></converted><child></child><converted>]</converted></allowed>
```
pjasiun: after, before? Shouldn't it be: before, after?

src/conversion/buildviewconverter.js
```diff
@@ -291,31 +292,62 @@ class ViewConverterBuilder {
...
+					} else if ( childrenOutput.end.parent !== childrenOutput.end.root ) {
+						output.end = Position.createAfter( childrenOutput.end.parent );
+
+					// Finally when we are sure that element and its parent was not split we need to put selection
+					// after element.
+					//
+					// after: <allowed>[]</allowed>
+					// before: <allowed>[<converted></converted>]</allowed>
```
pjasiun: As above.

pjasiun: Also, as far as I understand, this is a case when there are no children. Otherwise, the previous one will do the job.

src/conversion/viewconversiondispatcher.js
```diff
@@ -137,55 +136,71 @@ export default class ViewConversionDispatcher {
...
 
-			const consumable = ViewConsumable.createFrom( viewItem );
-			let conversionResult = this._convertItem( viewItem, consumable, additionalData );
+			// Additional date available between conversions.
+			// Needed when one converter needs to leave some data for oder converters.
+			this.conversionApi.data = {};
```
pjasiun: Since events still have `data` property, we should rename this one to `storage`.

src/conversion/buildviewconverter.js
oskarwrobel: Hm... I change it yesterday.

src/conversion/viewconversiondispatcher.js
```diff
@@ -137,55 +157,68 @@ export default class ViewConversionDispatcher {
...
+			// Create context tree and store position in the top element.
+			// Items will be converted according to this position.
+			this._contextPosition = contextToPosition( context, writer );
+
 			// Store writer in current conversion as a conversion API.
 			this.conversionApi.writer = writer;
 
-			this.fire( 'viewCleanup', viewItem );
+			// Custom data stored by converter for conversion process.
+			this.conversionApi.data = {};
```
pjasiun: Since events still have data property, we should rename this one to storage.

src/model/schema.js
```diff
@@ -1083,6 +1122,15 @@ export class SchemaContext {
 		return this._items[ Symbol.iterator ]();
 	}
 
+	/**
+	 * Adds new item at the to of the context.
+	 *
+	 * @param {module:engine/model/node~Node|String} item Item to add.
+	 */
+	addItem( item ) {
```
pjasiun: See https://github.com/ckeditor/ckeditor5-engine/issues/1250 and comments about immutable context.

src/model/schema.js
```diff
@@ -667,6 +674,38 @@ export default class Schema {
...
+	 * Tries to find position ancestors that allows to insert given node.
+	 * It starts searching from the given position and goes node by node to the top of the model tree
+	 * as long as {@link module:engine/model/schema~Schema#isLimit limit element} or top-most ancestor won't be reached.
+	 *
+	 * @params {module:engine/model/node~Node} node Node for which allowed parent should be found.
+	 * @params {module:engine/model/position~Position} position Position from searching will start.
+	 * @params {module:engine/model/element~Element} [limitElement] Custom limit element.
+	 * @returns {module:engine/model/element~Element|null} element Allowed parent or null if nothing was found.
+	 */
+	findAllowedParent( node, position, limitElement ) {
```
pjasiun: The limit element here is strange since the method also looks for an element with `isLimit` property. This is a limit element which is not a limit. Can we try to use a custom element as a base of the conversion which allows everything to be a child of it?

src/model/schema.js
```diff
@@ -193,11 +193,16 @@ export default class Schema {
 		this.decorate( 'checkAttribute' );
 
 		this.on( 'checkAttribute', ( evt, args ) => {
-			args[ 0 ] = new SchemaContext( args[ 0 ] );
+			if ( !( args[ 0 ] instanceof SchemaContext ) ) {
+				args[ 0 ] = new SchemaContext( args[ 0 ] );
```
pjasiun: We could move it to the `SchemaContext` constructor. If the parameter is an instance of the schema context, the constructor should return this object.

pjasiun: (in JS you can return from the class constructor)

src/model/writer.js
```diff
@@ -1141,3 +1174,10 @@ function isSameTree( rootA, rootB ) {
 
 	return false;
 }
+
+/**
+ * @typedef {Object} SplitResult
```
pjasiun: Do we need to define a typedef here? I think we can describe the return object inline, in the `split` documentation.

src/model/schema.js
oskarwrobel: We can drop this limit element and do it in a different way but still check schema according to real context instead of element that allows everything.

pjasiun: Ok, lets do it in `_splitToAllowedParent`.

src/conversion/viewconversiondispatcher.js
```diff
@@ -137,55 +157,68 @@ export default class ViewConversionDispatcher {
...
-	convert( viewItem, additionalData ) {
+	convert( viewItem, context = [ '$root' ] ) {
 		return this._model.change( writer => {
+			this.fire( 'viewCleanup', viewItem );
+
+			const consumable = ViewConsumable.createFrom( viewItem );
+
+			// Create context tree and store position in the top element.
+			// Items will be converted according to this position.
+			this._contextPosition = contextToPosition( context, writer );
```
pjasiun: `contextToPosition` -> `createContextTree`? This method is more about creating than type chage.

pjasiun: and `_contextPosition` -> `modelCursor`?

src/conversion/viewconversiondispatcher.js
```diff
@@ -137,55 +157,68 @@ export default class ViewConversionDispatcher {
...
 	 * @returns {module:engine/model/documentfragment~DocumentFragment} Model data that is a result of the conversion process
 	 * wrapped in `DocumentFragment`. Converted marker elements will be set as that document fragment's
 	 * {@link module:engine/model/documentfragment~DocumentFragment#markers static markers map}.
 	 */
-	convert( viewItem, additionalData ) {
+	convert( viewItem, context = [ '$root' ] ) {
 		return this._model.change( writer => {
+			this.fire( 'viewCleanup', viewItem );
+
+			const consumable = ViewConsumable.createFrom( viewItem );
```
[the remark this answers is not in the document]
oskarwrobel: I'm on it right now :)

src/conversion/viewconversiondispatcher.js
```diff
@@ -217,19 +250,77 @@ export default class ViewConversionDispatcher {
...
+			return null;
+		}
+
+		// When current position parent allows to insert element then return this position.
+		if ( allowedParent === position.parent ) {
+			return { position };
+		}
+
+		// Split element to allowed parent.
+		const data = this.conversionApi.writer.split( position, allowedParent );
```
pjasiun: `data` -> `splitResult`

src/conversion/viewconversiondispatcher.js
oskarwrobel: 👍 I like it.

src/conversion/viewconversiondispatcher.js
```diff
@@ -217,19 +250,77 @@ export default class ViewConversionDispatcher {
...
+
+		// Remember all elements that are created as a result of split.
+		// This is important because at the end of conversion we want to remove all empty split elements.
+		for ( const pos of data.range.getPositions() ) {
+			if ( !pos.isEqual( data.position ) ) {
+				this._splitElements.add( pos.parent );
```
pjasiun: Maybe `_splitElements` -> `_removeIfEmpty`. It does not matter anymore that these elements are results of the split. What does matter is that they should be removed if empty. Also, in the future, we may have more elements like this.

src/conversion/viewconversiondispatcher.js
```diff
@@ -217,19 +250,77 @@ export default class ViewConversionDispatcher {
...
+		if ( allowedParent === position.parent ) {
+			return { position };
+		}
+
+		// Split element to allowed parent.
+		const data = this.conversionApi.writer.split( position, allowedParent );
+
+		// Remember all elements that are created as a result of split.
+		// This is important because at the end of conversion we want to remove all empty split elements.
+		for ( const pos of data.range.getPositions() ) {
```
pjasiun: A sample in the comment would be useful to explain what this loop does.

src/model/schema.js
```diff
@@ -193,11 +193,16 @@ export default class Schema {
...
+			}
 		}, { priority: 'highest' } );
 
 		this.on( 'checkChild', ( evt, args ) => {
-			args[ 0 ] = new SchemaContext( args[ 0 ] );
+			if ( !( args[ 0 ] instanceof SchemaContext ) ) {
+				args[ 0 ] = new SchemaContext( args[ 0 ] );
+			}
```
pjasiun: As above.

oskarwrobel:
Everything fixed (I hope).

src/conversion/viewconversiondispatcher.js
```diff
@@ -137,99 +159,179 @@ export default class ViewConversionDispatcher {
...
+		}
+
+		// When current position parent allows to insert element then return this position.
+		if ( allowedParent === cursorPosition.parent ) {
+			return { position: cursorPosition };
+		}
+
+		// When allowed parent is in context tree.
+		if ( this._modelCursor.parent.getAncestors().includes( allowedParent ) ) {
+			return null;
```
pjasiun: Are you sure that null should be return here? Shouldn't we break it as much as possible in such case, and let `insertContent` do the rest?

src/model/schema.js
pjasiun: Still open issue.

oskarwrobel: eh... probably lost my change after merge.

oskarwrobel: Nope, it's github cache.

src/conversion/viewconversiondispatcher.js
oskarwrobel: When insertContent is inserting conversion result it means that it is a pasting. Pasting is converted in `$clipboardHolder` context.

We do not allow anywhere to create invalid (from schema perspective) model tree. The idea of splitting was introduced to avoid creating invalid tree.

PR #1260, opened 2018-01-31
T/730: Model#insertContent() will not produce invalid schema (when merging)

### LINKED ISSUE #730: Filler is created at the beginning of the root if image is the first element
Modify the manual test for the image feature so it's the first element in the content and load localhost:8125/tests/image/image.html.

See that the root has an inline filler at the beginning:

image

This is really odd – I don't know why would a root has an inline filler ;|

pomek:

Fix: `Model#insertContent()` will not merge nodes if a schema after merging will be incorrect. Closes ckeditor/ckeditor5#730.

### DIFF
--- a/src/model/utils/insertcontent.js
+++ b/src/model/utils/insertcontent.js
@@ -282,8 +282,8 @@ class Insertion {
 			return;
 		}
 
-		const mergeLeft = canMergeLeft( this.canMergeWith, this.model.schema );
-		const mergeRight = canMergeRight( this.canMergeWith, this.model.schema );
+		const mergeLeft = this._canMergeLeft( node, context );
+		const mergeRight = this._canMergeRight( node, context );
 		const mergePosLeft = LivePosition.createBefore( node );
 		const mergePosRight = LivePosition.createAfter( node );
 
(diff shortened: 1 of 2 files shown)


Reinmar:
Please rename this PR – it's not about paragraphs and list items about which the engine knows nothing.

src/model/utils/insertcontent.js
```diff
@@ -282,8 +282,8 @@ class Insertion {
 			return;
 		}
 
-		const mergeLeft = context.isFirst && ( node.previousSibling instanceof Element ) && this.canMergeWith.has( node.previousSibling );
-		const mergeRight = context.isLast && ( node.nextSibling instanceof Element ) && this.canMergeWith.has( node.nextSibling );
+		const mergeLeft = canMergeLeft.call( this );
```
Reinmar: Do we have to use `call()`? Can' we pass necessary things as params?

tests/model/utils/insertcontent.js
```diff
@@ -407,6 +407,25 @@ describe( 'DataController utils', () => {
 						'<paragraph>b</paragraph>'
 					);
 				} );
+
+				it( 'should not merge a paragraph wrapped in blockQuote with lists', () => {
```
Reinmar: Just a single test? Please be more creative. E.g. there should be separate tests for left merge and right merge. This gives at least 3 tests. And I'm sure you can think of more.

Reinmar:
Please also add an integration test in ckeditor5-block-quote package.

pomek:
I've added additional tests. Could you check it once again?

src/model/utils/insertcontent.js
```diff
@@ -327,6 +327,30 @@ class Insertion {
 
 		mergePosLeft.detach();
 		mergePosRight.detach();
+
+		// @param {Set} canMergeWith
+		// @param {module:engine/model/schema~Schema} schema
+		// @returns {Boolean}
+		function canMergeLeft( canMergeWith, schema ) {
```
Reinmar: This function params are odd. It gets some here but also uses the `context` and `node`. I'd prefer to move these functions out of this method's body and have more clarity about what it uses.

Reinmar: The way it's implemented currently you see function calls and miss the idea what these functions do.

pomek: Done.

PR #1265, opened 2018-02-01
SchemaContext enhancements.

### LINKED ISSUE #1263: Schema.concat()'s behaviour is confusing
What does the below code do?

```js
const ctx = new SchemaContext( foo );

const ctx2a = ctx.concat( foo.getFirstChild() );
const ctx2b = ctx.concat( new SchemaContext( foo.getFirstChild ) );
```

You may expect to have the below twice:

```
[someRoot, fooParent, foo, fooChild]
```

But you won't. The latter version will contain this:

```
[someRoot, fooParent, foo, someRoot, fooParent, foo, fooChild]
```

The reason why the first version works is that `concat()` casts the param to an array. This is very significant, but hard to spot detail, which, got me by surprise. Especially that the code misses API docs. A code sample is a must in such cases. It says more than 1000 words.

However, I'll go even further – where will be the opposite of `concat()`? What if you'll want to get just a piece of the current context? Are we going to introduce `slice()`? ;)

Finally, I don't get why we had to allow passing `SchemaContext` to `concat()`.  Why can you pass there an entire context while you can't pass a position? I'd say that you shouldn't need to do either of them but maybe I don't see some cases.

## Alternative API

I think that we should look for an alternative API. One which:

* won't create the confusion of how it works when passing a single "thing",
* ~one which won't require supporting `new SchemaContext( ctx )` and passing a context instance to the method that we'll create~ EDIT: one which won't require supporting passing an entire context instance to the method that we'll create,
* one which has a clear opposite method.

So, why not getting back to the initially proposed `push()` and `pop()`?

```js
ctx2 = ctx.push( foo.getFirstChild() );

ctx3 = ctx.push( foo ).push( bar );
```

What was wrong with it that we had to introduce`concat()`?

oskarwrobel:

Feature: Changed `SchemaContext#concat` to `SchemaContext#push`. Closes #1263.

BREAKING CHANGE: `SchemaContext#concat` has been changed to `SchemaContext#push`.

---

Related: https://github.com/ckeditor/ckeditor5-paragraph/tree/t/ckeditor5-engine/1263

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -1174,7 +1174,7 @@ export class SchemaContext {
 	 * 		const ctx3 = ctx.concat( 'other' ); // [ 'elementParent', 'element', 'other' ]
 	 *
 	 * 		// Array with above.
-	 * 		const ctx3 = ctx.concat( [ 'other', element.getChild( 0 ) ] ); // [ 'elementParent', 'element', 'other', 'elementChild ]
+	 * 		const ctx4 = ctx.concat( [ 'other', element.getChild( 0 ) ] ); // [ 'elementParent', 'element', 'other', 'elementChild ]
 	 *
 	 * @param {String|module:engine/model/node~Node|Array<String|module:engine/model/node~Node>} definition
 	 * Definition of item(s) that will be added to current context.
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -1167,13 +1167,13 @@ export class SchemaContext {
 	 *
 	 * 		const ctx = new SchemaContext( element ); // [ 'elementParent', 'element' ]
 	 *
-	 * 		// Node.
+	 * 		// A node.
 	 * 		const ctx2 = ctx.concat( element.getChild( 0 ) ); // [ 'elementParent', 'element', 'elementChild' ]
 	 *
-	 * 		// String (element name).
+	 * 		// A string (element name).
 	 * 		const ctx3 = ctx.concat( 'other' ); // [ 'elementParent', 'element', 'other' ]
 	 *
-	 * 		// Array with above.
+	 * 		// An array with above.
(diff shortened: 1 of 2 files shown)


src/model/schema.js
```diff
@@ -1165,13 +1163,25 @@ export class SchemaContext {
 
 	/**
 	 * Returns new SchemaContext instance with additional items created from provided definition.
+	 * Definition can be:
+	 *
+	 * 		const ctx = new SchemaContext( element ); // [ 'elementParent', 'element' ]
+	 *
+	 * 		// A node.
+	 * 		const ctx2 = ctx.concat( element.getChild( 0 ) ); // [ 'elementParent', 'element', 'elementChild' ]
```
pjasiun: I would write:
```js
const fooElement = writer.createElement( 'foo' );
const context2 = context.concat( element ); // [ 'elementParent', 'element', 'foo' ]
```
First, becase using element which is already in the tree is strange case for the concat method, second, to make it clear where the name come from.

src/model/schema.js
```diff
@@ -1165,13 +1163,25 @@ export class SchemaContext {
 
 	/**
 	 * Returns new SchemaContext instance with additional items created from provided definition.
+	 * Definition can be:
+	 *
+	 * 		const ctx = new SchemaContext( element ); // [ 'elementParent', 'element' ]
```
pjasiun: We should not use `ctx` in public API, including docs samples. Use `context` instead.

src/model/schema.js
oskarwrobel: This is to show that element that already has a parent will be added without this parent.

pjasiun: I would write it as a note "If the element has a parent, only the element will be added to the new context (the partent will not be included)." The sample does not show it well, especially when you have some parents in the outputs.

pjasiun: If I would see such sample I would ask myself: why should I do such thing? Isn't it better to do `new SchemaContext( element.getChild( 0 ) );`?

oskarwrobel: Changed.

PR #1268, opened 2018-02-01
Provide one API for two types of markers, improve docs

### LINKED ISSUE #1086: Markers are all you need
First, let's put apart (static) ranges, positions, and selection. These are used for different cases, for instance, to create and set the document selection.

I think that the number mechanisms we offer developers to handle the same problem is too big: liveposition, liverange and markers. It looks over-complicated and it is more than most users need. The answer to the question: what is a replacement for bookmarks from CKE4 should be simple. And I think it should be: markers. 

This is the complete solution, gives you all what you get using live ranges (or live position if the range is collapsed) but also:
 - have a name, what is very useful to get a marker, allows users to use namespaces, create multiple markers of the same "type",
 - markers collection fire event when a marker is created or removed, what is not possible to do with live ranges,
 - let you create converters to the virtual selection easily and show the marker in the DOM.

However, there are 2 problems.

First, we create markers much later than we created live ranges. This is why they are described as live ranges with names. To understand makers you need to understand live ranges first, what is wrong in my opinion. They should be described as an easy and ready to use solution, not overlay using another mechanism.

The second problem is that it should be very clear what is the difference between marker added and not added to the document.
 - Marker added to the document will be synchronized in the collaboration and handled in the undo. Is useful for solutions like spell checking or comments.
 - Marker not added to the document can be used as a bookmark or visual markers. To show results of the find, or select link when the focus is in the input (see https://github.com/ckeditor/ckeditor5-link/issues/13).

They also need to use different API.

We should make it very clear. We could even call these two types of markers differently. What does not help is the fact that the second group is also added to the `MarkersCollection` which is a property of the document, so they are somehow added to the document. Maybe we should reconsider this, for instance, `MarkersCollection` could be a property in the editing controller, not the document.

ma2ciek:

Other: Provided one API for two types of markers, improved docs. Closes #1086.

---

### Additional information

### DIFF
--- a/src/model/markercollection.js
+++ b/src/model/markercollection.js
@@ -122,7 +122,7 @@ export default class MarkerCollection {
 		const marker = new Marker( markerName, liveRange, managedUsingOperations );
 
 		this._markers.set( markerName, marker );
-		this.fire( 'add:' + markerName, marker );
+		this.fire( 'set:' + markerName, marker );
 
 		return marker;
 	}
@@ -211,10 +211,10 @@ export default class MarkerCollection {
 	}
 
 	/**
-	 * Fired whenever marker is added to `MarkerCollection`.
+	 * Fired whenever marker is added or updated in `MarkerCollection`.
 	 *
-	 * @event add
-	 * @param {module:engine/model/markercollection~Marker} The added marker.
+	 * @event set
+	 * @param {module:engine/model/markercollection~Marker} The set marker.
 	 */
 
 	/**
@@ -246,7 +246,7 @@ mix( MarkerCollection, EmitterMixin );
  * using common prefixes, separated with `:`, for example: `user:john` or `search:3`. That's useful in term of creating namespaces
  * for custom elements (e.g. comments, highlights).
  *
- * There're two types of markers. Both type of them should be added /updated by {@link module:engine/model/writer~Writer#setMarker}
+ * There're two types of markers. Both type of them should be added / updated by {@link module:engine/model/writer~Writer#setMarker}
  * and removed by {@link module:engine/model/writer~Writer#removeMarker} methods. Both of them are stored in the {@link ~MarkerCollection}.
  *
  * 1. Markers not added to document (default ones). They can be used as bookmarks or visual markers.
--- a/src/model/markercollection.js
+++ b/src/model/markercollection.js
@@ -99,28 +99,30 @@ export default class MarkerCollection {
 	 * set is different, the marker in collection is removed and then new marker is added. If the range was same, nothing
 	 * happens and `false` is returned.
 	 *
+	 * @protected
 	 * @fires module:engine/model/markercollection~MarkerCollection#event:add
 	 * @fires module:engine/model/markercollection~MarkerCollection#event:remove
 	 * @param {String|module:engine/model/markercollection~Marker} markerOrName Name of marker to add or Marker instance to update.
 	 * @param {module:engine/model/range~Range} range Marker range.
+	 * @param {Boolean} managedUsingOperations Specifies whether the marker is managed using operations.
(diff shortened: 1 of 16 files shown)


pjasiun:
The `managedUsingOperations` property is missing.

src/model/markercollection.js
```diff
@@ -78,28 +96,30 @@ export default class MarkerCollection {
...
-			this.remove( markerName );
+			this._remove( markerName );
 		}
 
 		const liveRange = LiveRange.createFromRange( range );
-		const marker = new Marker( markerName, liveRange );
+		const marker = new Marker( markerName, liveRange, managedUsingOperations );
 
 		this._markers.set( markerName, marker );
 		this.fire( 'add:' + markerName, marker );
```
pjasiun: I think that we should rename this to `set` because it is fired also when marker value changes.

src/model/writer.js
```diff
@@ -795,17 +860,22 @@ export default class Writer {
 		if ( !newRange ) {
 			// If `newRange` is not given, treat this as synchronizing existing marker.
 			// Create `MarkerOperation` with `oldRange` set to `null`, so reverse operation will remove the marker.
-			addMarkerOperation( this, name, null, currentRange );
+			addMarkerOperation( this, markerName, null, currentRange );
 		} else {
 			// Just change marker range.
-			addMarkerOperation( this, name, currentRange, newRange );
+			addMarkerOperation( this, markerName, currentRange, newRange );
```
pjasiun: Maybe `applyMarkerOperation`? It's strange that you add a marker. Add to what? It also sounds like it is about adding markers what is not true.

src/model/markercollection.js
ma2ciek: I'm thinking about the whole event concept used here. Is it useful? If you manage the `Marker`, then you have direct access to it. If you don't have the access, then you won't have access to the `Marker`'s name. There are zero use cases, for now, only tests.

Maybe `this.fire( 'add', marker );` would be better? Then you'd have the power to access all markers and filter needed marker or marker group.

src/model/writer.js
ma2ciek: I agree. WIll fix it.

pjasiun:
Consider such scenario: marker was created using operation but is not using operations anymore (it is set with the same range but without `usingOperation` flag). Will the operation to remove this marker be created? It should be removed using operation and created directly using `_set`.

Alternatively, we can decide to not support such case because I do not have any valid case for it, but then an exception should be thrown. Otherwise, we will have a very ugly bug when we will try to do so.

src/model/markercollection.js
ma2ciek: Hmm. Nevermind. I misunderstood this problem.

ma2ciek:
> Consider such scenario: marker was created using operation but is not using operations anymore (it is set with the same range but without `usingOperation` flag). Will the operation to remove this marker be created? It should be removed using operation and created directly using _set.
>
>Alternatively, we can decide to not support such case because I do not have any valid case for it, but then an exception should be thrown. Otherwise, we will have a very ugly bug when we will try to do so.

I think, that enabling type change of the Marker can easily be a developer's unintended typo, so I'm for the error in such case.

ma2ciek:
I added an exception when marker type changes.

pjasiun:
I mean it is a pretty common case that one may want to change the type from not `usingOperation` to `usingOperation`. For instance, if there was some preview which was later confirmed. This is what happens when you add a comment. Such type change need to be supported. The other change (a marker is no longer tracked by the document) is also theoretically possible, but is not that common and can be not supported for now.

ma2ciek:
I restored that ability and handled the scenario when marker changes from managed using operations to the "independent".

Issue #1243, opened 2018-01-16
ViewElementDefinition and priority


Reinmar:
```js
 *
 *		viewToModelElement( {
 *			model: 'heading1',
 *			view: {
 *				name: 'p',
 *				attribute: {
 *					'data-heading': 'true'
 *				},
 *				// You may need to use a high-priority listener to catch elements
 *				// which are handled by other (usually – more generic) converters too.
 *				priority: 'high'
 *			}
 *		}, [ editor.data.viewToModel ] );
```

The priority property used here causes a lot of problems, because:

* it can be confused with element's priority,
* it's actually not part of `ViewElementDefinition` (because it's element-focused and `priority` would have a different meaning there) so it's missed both in the docs for `ViewElementDefinition` and `ConverterDefinition`.

I'd move it either to `ConverterDefinition` or remove it completely (claiming that if you need priorities you should use the imperative API).

The first question that comes to my mind is why was this added to `ConverterDefinition#view` property instead of being defined directly in the converter definition, like this:

```js
 *
 *		viewToModelElement( {
 *			model: 'heading1',
 *			view: {
 *				name: 'p',
 *				attribute: {
 *					'data-heading': 'true'
 *				}
 *			},
 *			// You may need to use a high-priority listener to catch elements
 *			// which are handled by other (usually – more generic) converters too.
 *			priority: 'high'
 *		}, [ editor.data.viewToModel ] );
```

cc @pjasiun @jodator

pjasiun:
I agree, however, the api may change after https://github.com/ckeditor/ckeditor5-engine/issues/1236 so you can hold implementing it.

jodator:
The issues hera are
- the `view` is duplicated to `acceptsAlso` (only in `acceptsAlso` the priority is required AFAIR)
- probably not all `acceptsAlso` requires priority

Given above we could probably move priority from `view`to the same level as `model` & `view` (the second example from @Reinmar post) but this would require to override every `acceptAlso` / `view` for `view-to-model` conversion. This  shouldn't be problem for `new` definitions (they will just have `high` priority instead of `default`) and will override features defaults (like in headings examples).

Example configurations that requires priorities:
```js
const headingOptions: [
    { model: 'paragraph', title: 'Paragraph', class: 'ck-heading_paragraph' },
    { model: 'heading1', view: 'h1', title: 'Heading 1', class: 'ck-heading_heading1' },
[...]

PR #1274, opened 2018-02-02
T/1236 Conversion helpers

### LINKED ISSUE #1236: Conversion builder refactoring
There are 2 problems with the current conversion builder.

First, since the API is chainable, not declarative, it is hard to pass parameters to the config. Recently we realized that we need to make plugins like font size or headings configurable and conversion builder is nothing similar to what you want to pass to the config.

Second, since conversion builder mix all converters together it is hard to learn, based on in, how to create your custom converter.

This is why converters should be more specific. In the F2F talk with @Reinmar and @scofalik we agreed on the solution simmilar to what @szymonkups [proposed some time ago](https://github.com/ckeditor/ckeditor5-engine/issues/736#issuecomment-341670469).

```js
editor.conversion.for( 'model' ).add( elementToElement( 'paragraph', 'p' ) )
```

There will be specific model to view converters like:
```js
elementToElement
attributeToElement
attributeToAttribute
markerToElement
markerToHighlight
```
And corresponding view to model converters.

`editor.conversion` will have predefined sets of dispatchers:

```js
editor.conversion = new Conversion();
editor.conversion.register( 'data', editor.data.moveToView );
editor.conversion.register( 'editing', editor.editing.moveToView );
editor.conversion.register( 'model', [ editor.editing.moveToView, editor.data.moveToView ] );
editor.conversion.register( 'view', editor.data.viewToModel );
// The type of the conversion might be taken based on the constructor type.
```

Conversion function should get configuration defined the way it will be simple to use it with the [conversion definition](https://github.com/ckeditor/ckeditor5-engine/blob/master/src/conversion/definition-based-converters.js).

scofalik:

Other: Conversion utilities refactor. Closes #1236.

---

### Additional information

* `ModelConverterBuilder` and `ViewConverterBuilder` are removed,
* `definition-based-converters` are removed,
* `conversion.Conversion` class is introduced,
* new converter functions, basing on declarative config, are introduced,
* other related changes.


pjasiun:
Notes:

 - We should documented view parameters definition better, there are 2 types of them: `ViewElementDefinition` and `ViewMatcherDefinition`. The second one is a superset of the first one and allows additionally use regular expressions and callbacks. This is the one which should be used as a parameter of the view matcher as well as `alternativeView`.
 - It means also that `alternativeView` should accept callbacks. This seems to be the only place where callback will be useful in the top-level configuration. 
 -  `alternativeView` is a bad name, it is not any kind of alternative for the defined view definition, it will be never used instead of it. It is view formats which will be accepted as view element and converted to it. I remember we had a discussion about this property with @jodator and @reinmar. If we want to use `view` in the name we can choose `acceptableView`.
 - Priorities will be handled in the separate ticket (https://github.com/ckeditor/ckeditor5-engine/issues/1243).
 - The only common case for consumables was converting attributes only for the specific element. This should be the default behavior: `attributeToAttriute` converter should expect element name and convert only this element attributes. `attributeToAttriute` converter which works for all elements might causes conflicts.
 - Because during the conversion we have naming conflicts and using `modelToView` and `viewToModel` is inconvenient, we decided to use `downcast` and `upcast` to call these 2 types of the conversion.

Reinmar:
> alternativeView is a bad name, it is not any kind of alternative for the defined view definition, it will be never used instead of it. It is view formats which will be accepted as view element and converted to it. I remember we had a discussion about this property with @jodator and @Reinmar. If we want to use view in the name we can choose acceptableView.

In the current API this is called `acceptAlso`. `acceptableView` is worse than this because it loses the "also" word which clarifies what is it about.

Why did you move the `acceptAlso` out of the `view` part of converter definitions?

scofalik:
Honestly, I didn't like `acceptAlso` parameter, cause it doesn't state exactly what it means. I guess it was better in one-way-converters (from view to model) cause it is more obvious there. In this case, we are talking about model-to-view and view-to-model helper that sets both conversions.

That parameter name seemed to me lacking ingenuity :P. So I wanted to propose something better, although I understand how `alternativeView` is not technically correct.

I can change back to `acceptAlso`, no biggie, but I'd rather find something that satisfies us all :).

jodator:
> In this case, we are talking about model-to-view and view-to-model helper that sets both conversions.

But the many views model-to-view conversion doesn't make sense. This `acceptAlso` parameter makes sense only for view-to-model conversion.

scofalik:
Yes. of course. As I said, no biggie for me, I may change it back.

Reinmar:
First of all, I wouldn't be worried if one object which defines model and view forms of some feature is used for conversion in both direction. It may contain some excess information in this case (e.g. information which is only used by conversion in one direction), but that's not a big deal IMO. This object is, at least by me, understood as "feature definition". 

As for the `acceptAlso` name, I think it was pretty good. It told quite well what it did. That it's accepting some input and that it happens "also" (i.e. additionally) to some base form.

Perhaps, what might play even better is something with "upcast" in it. Like `upcastAlso`. Then, it will even more clearly say what's accepted.

Reinmar:
I saw the `modelElementIsViewElement()` and `modelAttributeIsViewAttribute()` functions and I don't like their names. For two reasons, actually:

* First of all, and this is what came to my mind immediately, because they sound like incorrectly named `isModelElementViewElement()`.
* But also because a model element is not a view element. From semantical POV this is an incorrect statement. They are converted to each other. But they are not the same thing.

Finally, this is not a correct function name from because it's not based on a verb. It states some fact, but it doesn't say **what it does**.

I'd rather go in directions lilke `addElementConverters()` and `addAttributeConverters()`.

Reinmar:
BTW, are we going with the upcast/downcast proposal?

scofalik:
> BTW, are we going with the upcast/downcast proposal?

I guess 🙄 .

At least `upcastAlso` sounds good.

Reinmar:
> > BTW, are we going with the upcast/downcast proposal?
> 
> I guess 🙄 .

 If so, instead of `addElementConverters()` we could *maybe* risk something like `castElement()` or `upcastDowncastElement()` to continue the `upcast/downcast` wording. Unfortunately, in this specific example the "converter" word plays better but it'd be good to have those names following one scheme...

scofalik:
> I'd rather go in directions lilke addElementConverters() and addAttributeConverters().

The point of those functions was that they say, maybe not technically correct, but explains that "this model element is [represented as] a view element".

As much as you can do `addElementsConverter()`, it's tougher to came up with a name for model-attribute-to-view-element utility.

scofalik:
We have those:

Highest level utilities, that define conversion in two ways. Let's call them "definitions" cause they mostly base on declarative config (definition):

* `modelElementIsViewElement` for model-element-to-from-view-element-conversion,
* `modelAttributeIsViewElement` for model-attribute-to-from-view-element-conversion,
* `modelAttributeIsViewAttribute` for model-attribute-to-from-view-attribute-conversion.

These are the only ways we can convert.

Then we have middle level utilities. They specify one way conversion. They are a bit more flexible but you have to use two. Let's call them "helpers":

* `mtvElementToElement`,
* `mtvAttributeToElement`,
* `mtvAttributeToAttribute`,
* `vtmElementToElement`,
* `vtmElementToAttribute`,
* `vtmAttributeToAttribute`.

Then we have the good ol' converters we had that actually do stuff, but I think and PJ agrees that there might be no use for them, we even thought of making them protected.

pjasiun:
> If so, instead of addElementConverters() we could maybe risk something like castElement() or upcastDowncastElement() to continue the upcast/downcast wording. Unfortunately, in this specific example the "converter" word plays better but it'd be good to have those names following one scheme...

`cast` is not a good word here, in my opinion. If we want something this way, we can use `bind` (`bindElementToElement`), `convert` (`convertElementToElement`) or just `to` (`modelElementToViewElement`, or `elementToElement`). Since one-way binding will use the additional word to define conversion direction (`upcastElementToElement`) I like the simple option and do not use `model` or `view` in the name. The only problem is `attributeToElement`, but since it makes sense only in one direction I think that also there we could avoid additional words.

scofalik:
So, following PJs proposal:

Highest level utilities:

* `modelElementIsViewElement` -> `elementToElement`
* `modelAttributeIsViewElement` -> `attributeToElement`
* `modelAttributeIsViewAttribute` -> `attributeToAttribute`

Then we have middle level utilities:

* `model-to-view-helpers.elementToElement` -> `downcastElementToElement`,
* `model...attributeToElement` -> `downcastAttributeToElement`,
* `model...attributeToAttribute` -> `downcastAttributeToAttribute`,
* `view-to-model-helpers...elementToElement` -> `upcastElementToElement`,
* `view...elementToAttribute` -> `upcastElementToAttribute`,
* `view...attributeToAttribute` -> `upcastAttributeToAttribute`

?

Reinmar:
Middle level utils are fine. But as for highest level utils I really miss there a verb. The question is – how do we call the sum of upcasting and downcasting? :D "Bind" and "convert" as PJ proposed are fine and I'd rather go with one of them. But if you prefer the no-verb version more, then I'd be fine as well.

scofalik:
The thing here is that you don't convert anything by calling the function. You just set up conversion rules. I mean... I guess it is also true for middle-level functions. The most right would be `setConversionFromElementToElement` or `setDowncastFromElementToElement`. But then we are back to super long awkward names.

Reinmar:
> But then we are back to super long awkward names.
> 

Exactly. And there's no point in going back there. The names should be understandable and as short as possible at the same time. 

BTW, `setConversion...` is so "engineered" that it has no meaning anymore.

pjasiun:
I'm fine with `elementToElement`, in the future, we may have it as `editor.conversion.elementToElement`, what is nice.

But I also understand your point @Reinmar that a verb is missing. `bindElementToElement` or `convertElementToElement` are fine for me too.

So, everything is cool for me. Well, everything except `setConversionFromElementToElement` and `setDowncastFromElementToElement`. ;)

scofalik:
I've pushed commit with most improvements discussed here.

pjasiun:
We should merge `downcast-helpers` and `downcast-converters` files. Even if such separation makes sense, they represent different levels of the API, it is very misleading. Usually, what you are looking for, will be something like `downcastElementToElement` and I would look for in `downcast-converters`. If we don't want to have a giant file, we can move everything about markers to the separate file, so we should have:
```
upcast-converters.js
upcast-selection-converters.js
upcast-markers-converters.js
downcast-converters.js
downcast-selection-converters.js
downcast-markers-converters.js
two-way-converters.js
```

Reinmar:
If all those helpers will be exposed in `editor.conversion`, then does the file structure matters for other people? Isn't it only for us?

Although, I agree that `-helpers` and `-converters` suffixes give me no clue what they are all about.

pjasiun:
But it will be confusing even for us. Also, now every converter is split into 2 files and wherever you work with it you need to jump between them.

scofalik:
One note: there's a slight problem with extracting markers functions from other functions cause all of them use common helper (private) functions.

scofalik:
Pushed merged files.

PR #1276, opened 2018-02-05
`Schema#getLimitElement()` will work with multi-ranges selection

### LINKED ISSUE #1275: `Schema#getLimitElement()` does not work for multi-ranges selection
`Schema#getLimitElement()` returns `null` for multi-ranges selection if one of them has the root element as the limit element. 

That's the reason why the editor crashes in some cases on Firefox (https://github.com/ckeditor/ckeditor5/issues/678).

pomek:

Fix: `Schema#getLimitElement()` will return a proper limit element (the root element) if one of the selection's ranges have the root element as the limit element. Closes #1275.

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -581,12 +581,22 @@ export default class Schema {
 	getLimitElement( selection ) {
 		// Find the common ancestor for all selection's ranges.
 		let element = Array.from( selection.getRanges() )
-			.reduce( ( node, range ) => {
-				if ( !node ) {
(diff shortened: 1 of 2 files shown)


Reinmar:
Wouldn't this work:

```js
return element.getCommonAncestor( range.getCommonAncestor(), { includeSelf: true } );
```

(I added the options param there)

All tests pass.

pomek:
Thx. Fixed.

PR #1286, opened 2018-02-08
T/1281: DocumenSelection#change:range will be fired once for multi-ranges selection

### LINKED ISSUE #1281: DocumentSelection validates its ranges before they are updated
`DocumentSelection` contains live ranges. Every live range is updated at `model#applyOperation`. Then, when it's updated, on `change:range` document selection validates that new range positions are correct.

This works fine with a single range in the selection, but it breaks badly with multiple ranges, because this happens:

1. First selection range gets updated on `model#applyOperation`.
2. This triggers `range1#change:range` (BTW, this event name is a bit odd).
3. This is delegated to `selection#change:range`.
4. On `selection#change:range` the selection starts validating its ranges.
5. All ranges... at once. So, also the second range which wasn't updated yet.

It seems to me that the selection should validate only the range which has changed.

cc @scofalik

pomek:

Fix: `DocumenSelection#change:range` event will be fired only once after multiple selection live ranges have changed. Closes #1281.


Reinmar:
If you're adding a new param to an existing event, then you need to clarify when it's available in the `change:range` event. I can see that you extended only one place where `change:range` is fired. What about the other place (inside `_fixGraveyardSelection()`)? Besides, docs are missing for this event.

I was trying to analyse this code a bit and found a couple of odd places:

* `_fixGraveyardSelection()` is executed inside `liveRange#change:range` event. The same listener for that event fires `LiveSelection#change:range`. So, since `_fixGraveyardSelection()` fires `LiveSelection#change:range` too, then does it mean that this event is fired twice?
* Is `LiveSelection#change:range` fired only when one of its existing ranges change? Is it intentional that it isn't fired also when range is added/removed from the selection?
* Does `directChange` param of the `LiveSelection#change:range` event make any sense if it's always `false`? When would it be true?

cc @scofalik

pomek:
> `_fixGraveyardSelection()` is executed inside `liveRange#change:range` event. The same listener for that event fires `LiveSelection#change:range`. So, since `_fixGraveyardSelection()` fires `LiveSelection#change:range` too, then does it mean that this event is fired twice?

In our case, `_fixGraveyardSelection()` isn't executed. It means that `LiveSelection#change:range` is not called twice. 

> Is `LiveSelection#change:range` fired only when one of its existing ranges change?

Yes. When a range is adding or removing, this event is not fired.

> Does `directChange` param of the `LiveSelection#change:range` event make any sense if it's always `false`? When would it be true?

I have the same feelings.

As you noticed, the range validation can be moved from `LiveSelection#change:range` to `liveRange#change:range`. Does it mean that `LiveSelection` shouldn't listen to the `change:range` event anymore?

Reinmar:
> In our case, _fixGraveyardSelection() isn't executed. It means that LiveSelection#change:range is not called twice.

I asked generally. Not specifically.

> Does it mean that LiveSelection shouldn't listen to the change:range event anymore?

It should still listen to `change:range` after this PR.  Otherwise, you'd change the API because `DocumentSelection#change:range` would disappear. We can, however, ask a question whether it should delegate that event in general, but I think that it should

cc @scofalik still :P

scofalik:
> * `_fixGraveyardSelection()` is executed inside `liveRange#change:range` event. The same listener for that event fires `LiveSelection#change:range`. So, since `_fixGraveyardSelection()` fires `LiveSelection#change:range` too, then does it mean that this event is fired twice?

You are right. Not only that but if there will be multiple ranges in a selection and they will get changed by the same operation, there might be multiple `change:range` events on selection too. This could be fixed to have only one `change:range` event for selection, that is fired after all live ranges are updated.

* Is `LiveSelection#change:range` fired only when one of its existing ranges change? Is it intentional that it isn't fired also when range is added/removed from the selection?

`change:range` event is called in `model.Selection#_setRanges`.

* Does `directChange` param of the `LiveSelection#change:range` event make any sense if it's always `false`? When would it be true?

`directChange` is set to `true` in some events fired in `model.Selection` (like `_setRanges`).

Reinmar:
I missed the fact that `LiveSelection` extends normal `Selection` so also its events. Thanks.

> You are right. Not only that but if there will be multiple ranges in a selection and they will get changed by the same operation, there might be multiple change:range events on selection too. This could be fixed to have only one change:range event for selection, that is fired after all live ranges are updated.

@pomek, you may consider fixing this too in this PR.

pomek:
The easiest solution is replacing (in `LiveSelection` class) `this.fire( 'change:range', { directChange: false } );` with this:

```js
if ( this.getLastRange().isEqual( liveRange ) ) {
	this.fire( 'change:range', { directChange: false } );
}
```

But I assume it isn't a correct way to fix that. Could you guys give me some tip?

Reinmar:
I had a doubt whether only a single `change:range` should be fired on the selection when document has been changed, but it makes sense to fire this event just once. That's because, as we actually learnt in this ticket, the selection isn't even correct until all ranges are updated.

scofalik:
My suggestion on how to fix it:

1. In document selection, keep listener listening to live range change. When a live range is changed. However, instead of fixing selection, store data that is used to fix range (I assume this would be live range instance and old range parameter).
2. Add one listener that will be fired after all live ranges are updated. Check where and with what priority live range listeners are hooked and it after them. In that new listener, check if any data from point 1. was stored, and if so, use it to fix selection and clear the stored data.

I hope that will be enough.

pomek:
> However, instead of fixing selection, store data that is used to fix range

Unfortunately, I don't understand where the selection is "fixed". `LiveRange`s which are part of the `DocumentSelection` are updating on `Model#applyOperation` event: https://github.com/ckeditor/ckeditor5-engine/blob/4b7d4355dbe4c7d754c985d8c1928aff4def89d4/src/model/liverange.js#L122-L137

Then the `LiveRange#change:range` event is fired. It is handled by `LiveSelection`: https://github.com/ckeditor/ckeditor5-engine/blob/4b7d4355dbe4c7d754c985d8c1928aff4def89d4/src/model/documentselection.js#L620-L628

and fired once again (informs the whole selection about change in range) https://github.com/ckeditor/ckeditor5-engine/blob/4b7d4355dbe4c7d754c985d8c1928aff4def89d4/src/model/documentselection.js#L458-L475

Maybe I didn't understand anything but at this moment I don't what and where to do.

scofalik:
The point is not to have this fixing callback fired multiple times: https://github.com/ckeditor/ckeditor5-engine/blob/4b7d4355dbe4c7d754c985d8c1928aff4def89d4/src/model/documentselection.js#L620-L628

To do so, we need just one callback, added in `DocumentSelection` to `applyOperation` event, with `lowest` priority. The selection is fixed in the linked callback (`this._fixGraveyardSelection( liveRange, data.sourcePosition );`).

To be able to fix selection in an unrelated callback (the callback will no longer listen to live range changes so you don't have any live range information in it) you need to store the changed ranges (and any needed data) in the linked callback (instead of calling `_fixGraveyardSelection`).

Hope that's more clear.

pomek:
Ok. Thanks to your suggestion I was able to do it the right way. Let's wait for Travis. If everything will work as I expect, the `r-` label will be removed.

PR #1287, opened 2018-02-08
T/ckeditor5 typing/92: Add support for "word" unit in `modifySelection()` helper.

jodator:

Feature: Add support for "word" unit in `modifySelection()` helper.

---

### Additional information

* This is a preview PR required for fixes in https://github.com/ckeditor/ckeditor5-typing/issues/92.


jodator:
I had some troubles with modifing `getCorrectPosition()` to some sane form.

I does too much now - maybe I should refactor it and extract `word` logic to other method?

src/model/utils/modifyselection.js
```diff
@@ -13,6 +13,8 @@ import Range from '../range';
 import { isInsideSurrogatePair, isInsideCombinedSymbol } from '@ckeditor/ckeditor5-utils/src/unicode';
 import DocumentSelection from '../documentselection';
 
+const wordBoundaryCharacters = ' ,.-():\'"';
```
Reinmar: Is `'` a word boundary? What about "it's"?

src/model/utils/modifyselection.js
```diff
@@ -31,6 +33,7 @@ import DocumentSelection from '../documentselection';
  *  For example `𨭎` is represented in `String` by `\uD862\uDF4E`. Both `\uD862` and `\uDF4E` do not have any meaning
  *  outside the pair (are rendered as ? when alone). Position between them would be incorrect. In this case, selection
  *  extension will include whole "surrogate pair".
+ *  * `'word'` - moves selection by whole word.
```
Reinmar: "a whole word"

src/model/utils/modifyselection.js
```diff
@@ -79,11 +82,13 @@ export default function modifySelection( model, selection, options = {} ) {
 }
 
 // Checks whether the selection can be extended to the the walker's next value (next position).
+// @param {{ walker, unit, isForward, schema }} data
+// @param {{ item, nextPosition, type}} value
```
Reinmar: I guess that `value` is of the `TreeWalkerValue` type

src/model/utils/modifyselection.js
```diff
@@ -117,17 +122,48 @@ function tryExtendingTo( data, value ) {
...
 
 	if ( textNode ) {
-		const data = textNode.data;
+		let data = textNode.data;
 		let offset = walker.position.offset - textNode.startOffset;
+		let isAtNodeBoundary = offset === ( isForward ? textNode.endOffset : 0 );
 
-		while ( isInsideSurrogatePair( data, offset ) || ( unit == 'character' && isInsideCombinedSymbol( data, offset ) ) ) {
+		while (
+			isInsideSurrogatePair( data, offset ) ||
```
Reinmar: This could be a separate function.

tests/model/utils/modifyselection.js
```diff
@@ -411,6 +411,402 @@ describe( 'DataController utils', () => {
...
+					modifySelection( model, doc.selection, { unit: 'word', direction: 'backward' } );
+
+					expect( stringify( doc.getRoot(), doc.selection ) ).to.equal( '<p>[f]oo</p>' );
+					expect( doc.selection.isBackward ).to.true;
+				} );
+
+				test(
+					'expands selection to the word start',
+					'<p>foo bar[b]az</p>',
+					// TODO: '<p>foo [barb]az</p>',
```
Reinmar: ?

Reinmar: Why this TODO and other TODOs?

Reinmar:
I was pleased to see a great test suite. As for the code, you may try to improve it a bit – especially when it comes to overly long conditions. But it's completely internal stuff so it can be far from perfection.

Reinmar:
I've tested it manually and it works very well.

src/model/utils/modifyselection.js
jodator: It looks that it is (as far as I can test it in textarea and Gnome Text editor:

`it's a trap` - the text will stop: on spaces and on `'`.

Also check this docs: https://www.unicode.org/reports/tr29/#Default_Word_Boundaries.

We might think of expanding this list to full unicode support but I've used the most common ones for the text.

jodator: Yes it is - check the behovior on textarea or in text editor:

```
it's a trap
```
In the above the cursor will stop on spaces and on `'`.

Also check this docs: https://www.unicode.org/reports/tr29/#Default_Word_Boundaries.

We might think of expanding this list to full unicode support but I've used the most common ones for the text.

jodator: Hmm... we should expand the list of `boundaryCharacters` with at least: `?|;` and probably some more common interpuntion characters.

Reinmar: @wwalc told me that we need to differentiate between "it's" (which is made of two words so the caret should stop after "s") and "Peter's" where it's a single word which should be removed at once.

Reinmar: I'm joking of course. Since a single quote is popular in English and e.g. in macOS it's not treated as a word boundary, I'd not treat it as such.

tests/model/utils/modifyselection.js
jodator: I've forgot to remove them while working.

src/model/utils/modifyselection.js
jodator: So I'd go with:
```js
// current:                    ' ,.-():\'"';
const wordBoundaryCharacters = ' ,.?!:;"-()';
```

PR #1288, opened 2018-02-08
Keep the same marker instance when marker is updated.

### LINKED ISSUE #1284: The same marker should be used when I change range or option
Now, when I set a new range or option on a marker a new instance of the marker is created and old is destroyed.

It means that for instance such code will not work as expected:
```js
const myMarker = writer.setMarker( 'foo', range );

writer.setMarker( myMarker, newRange );

const range = myMarker.getRange(); // throw error that marker is destroyed
```

It should be fixed. As long as `MarkerCollection` contains marker with given name `writer.setMarker` should update this instance.

### LINKED ISSUE #1285: managedUsingOperations getter should throw when marker is removed
When a maker is removed from the collection an exception should be thrown whenever one tries to use it. It is already implemented for all methods.

```js
const myMarker = writer.setMarker( 'foo', range );

writer.removeMarker( myMarker );

myMarker.managedUsingOperations // marker-destroyed should be thrown
```

oskarwrobel:

Other: Keep the same marker instance when marker is updated. Closes #1284. Closes #1285.

BREAKING CHANGE: `MarkerCollection#event:set` and `MarkerCollection#event:remove` have been replaced by `MarkerCollection#event:update`.

### DIFF
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -131,10 +131,10 @@ export default class EditingController {
 			}
 		}, { priority: 'high' } );
 
-		// If a marker is removed through `model.Model#markers` directly (not through operation), just remove it (if
-		// it was not removed already).
-		this.listenTo( model.markers, 'remove', ( evt, marker ) => {
-			if ( !removedMarkers.has( marker.name ) ) {
+		// If an existing marker is updated through `model.Model#markers` directly (not through operation),
+		// just remove it (if it was not removed already).
+		this.listenTo( model.markers, 'update', ( evt, marker, oldRange ) => {
+			if ( oldRange && !removedMarkers.has( marker.name ) ) {
 				removedMarkers.add( marker.name );
 				this.view.change( writer => {
 					this.downcastDispatcher.convertMarkerRemove( marker.name, marker.getRange(), writer );
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -134,10 +134,11 @@ export default class EditingController {
 		// If an existing marker is updated through `model.Model#markers` directly (not through operation),
 		// just remove it (if it was not removed already).
 		this.listenTo( model.markers, 'update', ( evt, marker, oldRange ) => {
-			if ( oldRange && !removedMarkers.has( marker.name ) ) {
+			if ( oldRange && !marker.managedUsingOperations ) {
 				removedMarkers.add( marker.name );
+
 				this.view.change( writer => {
-					this.downcastDispatcher.convertMarkerRemove( marker.name, marker.getRange(), writer );
+					this.downcastDispatcher.convertMarkerRemove( marker.name, oldRange, writer );
 				} );
 			}
 		} );
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -133,7 +133,7 @@ export default class EditingController {
 
 		// If an existing marker is updated through `model.Model#markers` directly (not through operation), just remove it.
 		this.listenTo( model.markers, 'update', ( evt, marker, oldRange ) => {
-			if ( oldRange && !marker.managedUsingOperations ) {
+			if ( oldRange && !removedMarkers.has( marker.name ) ) {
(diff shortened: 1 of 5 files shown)


src/model/markercollection.js
```diff
@@ -79,37 +78,47 @@ export default class MarkerCollection {
 	 * Creates and adds a {@link ~Marker marker} to the `MarkerCollection` with given name on given
 	 * {@link module:engine/model/range~Range range}.
 	 *
-	 * If `MarkerCollection` already had a marker with given name (or {@link ~Marker marker} was passed) and the range to
-	 * set is different, the marker in collection is removed and then new marker is added. If the range was same, nothing
-	 * happens and `false` is returned.
+	 * If `MarkerCollection` already had a marker with given name (or {@link ~Marker marker} was passed), the marker in
+	 * collection is updated and {module:engine/model/markercollection~MarkerCollection#event:update} event is fired but only
```
scofalik: forgot `@link`.

src/model/markercollection.js
```diff
@@ -79,37 +78,47 @@ export default class MarkerCollection {
...
 	 *
 	 * @protected
-	 * @fires module:engine/model/markercollection~MarkerCollection#event:set
-	 * @fires module:engine/model/markercollection~MarkerCollection#event:remove
+	 * @fires module:engine/model/markercollection~MarkerCollection#event:update
 	 * @param {String|module:engine/model/markercollection~Marker} markerOrName Name of marker to set or marker instance to update.
 	 * @param {module:engine/model/range~Range} range Marker range.
 	 * @param {Boolean} managedUsingOperations Specifies whether the marker is managed using operations.
-	 * @returns {module:engine/model/markercollection~Marker} `Marker` instance added to the collection.
+	 * @returns {module:engine/model/markercollection~Marker} Added or updated `Marker` instance.
```
scofalik: `Marker` instance which was added or updated.

src/model/markercollection.js
```diff
@@ -79,37 +78,47 @@ export default class MarkerCollection {
...
 
-			if ( oldRange.isEqual( range ) && managedUsingOperations === oldMarker.managedUsingOperations ) {
-				return oldMarker;
+			if ( !oldRange.isEqual( range ) ) {
+				oldMarker._attachLiveRange( LiveRange.createFromRange( range ) );
+				hasChanged = true;
 			}
 
-			this._remove( markerName );
+			if ( !!managedUsingOperations !== oldMarker.managedUsingOperations ) {
```
scofalik: I wonder whether this "Boolean casting" is needed here. If `managedUsingOperations` is a required parameter it should always be `true` or `false`. If it is not, maybe it should have a default value. If the value is taken from somewhere else, maybe it should be cast before the call. Lastly, if `oldMarker.managedUsingOperations` is always a `Boolean` I think you don't need to be overly safe.

Was there an error with condition written as `managedUsingOperations != oldMarker.managedUsingOperations`?

scofalik: BTW. if you cast to `Boolean` on condition you probably should also cast when setting `managedUsingOperations` in a line below.

src/controller/editingcontroller.js
```diff
@@ -130,12 +130,11 @@ export default class EditingController {
 			}
 		}, { priority: 'high' } );
 
-		// If a marker is removed through `model.Model#markers` directly (not through operation), just remove it (if
-		// it was not removed already).
-		this.listenTo( model.markers, 'remove', ( evt, marker ) => {
-			if ( !removedMarkers.has( marker.name ) ) {
+		// If an existing marker is updated through `model.Model#markers` directly (not through operation), just remove it.
+		this.listenTo( model.markers, 'update', ( evt, marker, oldRange ) => {
+			if ( oldRange && !marker.managedUsingOperations ) {
```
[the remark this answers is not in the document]
oskarwrobel: The reason why I changed it was "not through operation" in comment. I'll remove this condition.

src/model/markercollection.js
oskarwrobel: I changed this param to optional and set false as default.

PR #1296, opened 2018-02-12
View refactoring

### LINKED ISSUE #1210: Update view API after changes in the model
The only proper way of changing model will be now the model writer in the change block. 

From time to time we have a problem with render being not synchronized (see https://github.com/ckeditor/ckeditor5-engine/issues/654#issuecomment-332188800 or https://github.com/ckeditor/ckeditor5-engine/issues/1148). We can solve it in the same way, introducing `view.change` block (see the draft here https://gist.github.com/pjasiun/74b5c4e22b2c60ec17d82918a9663fd0#file-engine-js-L124-L136). Model to view conversion should be done in this block as well as any other view modification. When all changes are done and there are changes in the renderer, view document should be rendered.

`view.change` should be the only way to get view writer. It should be changed to the class instead of a set of function. It should get all methods to modify the view tree, similar to the model writer, including attributes changes and selection. View writer API should be similar to the model writer with the exception that view writer does not need batches.

At the same time, we should consider splitting view and the document. Making view a general controller of the view and the document focused on what is rendered.

szymonkups:

Other: Refactoring of the view API. Closes #1210.

BREAKING CHANGE: `view.Writer` is no longer an object literal with functions but a class.
BREAKING CHANGE: View controller `view.View` is introduced. Changes to the view document tree structure should be done by using writer provided to callback in `view.change()` method.
BREAKING CHANGE: View document is now separated from the DOM. `view.Renderer`, `view.DomConverter` and observers are moved to `view.View`. 
BREAKING CHANGE: `view#event:render` is introduced to indicate a moment when all changes are applied and document may be rendered to the DOM.
BREAKING CHANGE: Downcast converter helpers no longer accepts view elements instances as constructors are now protected. Callbacks using view writer should be used.

---

### Additional information

Main CKEditor 5 repo contains [ckeditor5-engine/1210 branch](https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-engine/1210) which have modified `mgit.json` with other changed repositories on correct branches.


src/conversion/downcast-converters.js
```diff
@@ -45,17 +41,23 @@ import cloneDeep from '@ckeditor/ckeditor5-utils/src/lib/lodash/cloneDeep';
  *
  * 		downcastElementToElement( {
  * 			model: 'heading',
- * 			view: modelElement => new ViewContainerElement( 'h' + modelElement.getAttribute( 'level' ) )
+ * 			view: ( modelItem, consumable, conversionApi ) => {
+ * 				const viewWriter = conversionApi.writer;
+ *
+ * 				return viewWriter.createContainerElement( 'h' + modelElement.getAttribute( 'level' ) );
```
pjasiun: `modelElement` -> `modelItem`

src/conversion/downcast-converters.js
```diff
@@ -118,7 +118,11 @@ export function downcastElementToElement( config, priority = 'normal' ) {
  *		] );
  *
  * 		downcastAttributeToElement( 'bold', {
- * 			view: modelAttributeValue => new ViewAttributeElement( 'span', { style: 'font-weight:' + modelAttributeValue } )
+ * 			view: ( modelAttributeValue, data, consumable, conversionApi ) => {
+ * 				const viewWriter = conversionApi.writer;
+ *
+ * 				return	viewWriter( 'span', { style: 'font-weight:' + modelAttributeValue } );
```
pjasiun: `return[tab]` -> `return[space]`

src/conversion/downcast-converters.js
scofalik: Not that related, but there is an issue that we should try to unify parameters passed to callbacks in downcast and upcast. `( modelItem, viewWriter ) => { ... }` would look better.

src/conversion/downcast-converters.js
```diff
@@ -355,17 +358,19 @@ function _normalizeToElementConfig( config, ViewElementClass ) {
...
 	// Build `.view` property.
-	// It is expected to be either creator function or view element instance.
-	if ( typeof config.view == 'string' ) {
-		// If `.view` is a string, create a proper view element instance out of given `ViewElementClass` and name given in `.view`.
-		config.view = new ViewElementClass( config.view );
-	} else if ( typeof config.view == 'object' && !( config.view instanceof ViewElementClass ) ) {
+	// It is expected to be either string, element definition or creator function.
+	if ( typeof view == 'string' ) {
+		// If `.view` is a string, create a function that returns view element instance out of given `ViewElementClass`.
+		config.view = () => new ViewElementClass( view );
```
scofalik: Shouldn't that be build using writer too?

scofalik: BTW., we also need to change `ViewElementClass` to some kind of flag or something that would then result in using a proper method from `view.Writer`.

src/conversion/downcast-converters.js
```diff
@@ -474,37 +475,33 @@ function _getCreatorForArrayConfig( config ) {
...
+ * {@link module:engine/conversion/downcastdispatcher~DowncastDispatcher#event:insert insert event}.
  * It's expected that the function returns a {@link module:engine/view/element~Element}.
  * The result of the function will be inserted to the view.
  *
  * The converter automatically consumes corresponding value from consumables list, stops the event (see
  * {@link module:engine/conversion/downcastdispatcher~DowncastDispatcher}) and bind model and view elements.
  *
- *		modelDispatcher.on( 'insert:paragraph', insertElement( new ViewElement( 'p' ) ) );
- *
  *		modelDispatcher.on(
```
scofalik: `modelDispatcher` => `downcastDispatcher`

src/conversion/downcast-converters.js
```diff
@@ -537,8 +534,9 @@ export function insertText() {
 			return;
 		}
 
+		const viewWriter = conversionApi.writer;
 		const viewPosition = conversionApi.mapper.toViewPosition( data.range.start );
-		const viewText = new ViewText( data.item.data );
+		const viewText = viewWriter.createText( data.item.data );
```
scofalik: `viewWriter` mixed with `conversionApi.writer`. Let's use one :P

scofalik: Okay it seems that I was who mixed things :) In this function we use only `viewWriter`. But other uses `conversionApi.writer` directly.

src/conversion/downcast-selection-converters.js
```diff
@@ -113,16 +112,19 @@ export function convertCollapsedSelection() {
...
 			if ( range.isCollapsed ) {
 				// Position might be in the node removed by the view writer.
 				if ( range.end.parent.document ) {
-					viewWriter.mergeAttributes( range.start );
+					conversionApi.writer.mergeAttributes( range.start );
 				}
 			}
 		}
-		conversionApi.viewSelection.setTo( null );
+		writer.setSelection( null );
```
scofalik: Gotcha! I knew they will be mismatched somewhere :P

src/conversion/downcastdispatcher.js
```diff
@@ -126,23 +129,26 @@ export default class DowncastDispatcher {
 	 * Takes {@link module:engine/model/differ~Differ model differ} object with buffered changes and fires conversion basing on it.
 	 *
 	 * @param {module:engine/model/differ~Differ} differ Differ object with buffered changes.
+	 * @param {module:engine/view/writer~Writer} writer View writer that should be used to modify view document.
 	 */
-	convertChanges( differ ) {
+	convertChanges( differ, writer ) {
+		this.conversionApi.writer = writer;
```
scofalik: Do we need to set it up here if it is set up in every method called later?

BTW. I don't have a better solution for now, but it seems kind of awkward how this writer is obtained and then passed down to deeper methods in dispatcher. I am sure that you thought of different ways of solving that so I don't necessarily want to start a discussion, just leaving it here as a thought.

src/conversion/downcast-converters.js
szymonkups: I was thinking about it but wasn't sure if going with a flag is a good option. Other than that this is private function used in engine and constructors will be protected to that package and we know what we are doing here. Buuut on the other hand (when I am thinking about it now after you mentioned it) changing it to writer will show that this is the proper practice everywhere :D I will take care of it.

src/conversion/downcast-converters.js
pjasiun: For sure, but it should be done in a follow-up.

src/conversion/downcastdispatcher.js
szymonkups: I got used to it during the work on it, but it may look a little awkward. I couldn't find better solution either. Editing controller is listening for changes in the document and then starts to converting changes in `view.change()` block, passing writer down to the dispatcher and from there to the `conversionApi` which can be used by converters.

src/view/writer.js
```diff
@@ -12,704 +12,993 @@ import ContainerElement from './containerelement';
...
+		return this.insert( targetPosition, nodes );
 	}
 
-	return insert( targetPosition, nodes );
-}
+	/**
+     * Wraps elements within range with provided {@link module:engine/view/attributeelement~AttributeElement AttributeElement}.
+     * If a collapsed range is provided, it will be wrapped only if it is equal to view selection.
+     *
+     * If `viewSelection` was set and a collapsed range was passed, if the range is same as selection, the selection
```
scofalik: This comment seems outdated cause `viewSelection` is no longer set in a parameter.

src/conversion/downcast-converters.js
```diff
@@ -787,9 +783,12 @@ export function wrap( elementCreator ) {
 			return;
 		}
 
+		const viewWriter = conversionApi.writer;
+		const viewSelection = viewWriter.document.selection;
+
 		if ( data.item instanceof ModelSelection || data.item instanceof DocumentSelection ) {
 			// Selection attribute conversion.
-			viewWriter.wrap( conversionApi.viewSelection.getFirstRange(), newViewElement, conversionApi.viewSelection );
+			viewWriter.wrap( viewSelection.getFirstRange(), newViewElement, viewSelection );
```
scofalik: `view.Writer#wrap` does not take third parameter anymore.

This also means that maybe this method now can be simplified. I'd check for that.

src/conversion/downcast-converters.js
szymonkups: I will wait with this in follow up, after we unify callbacks for `downcastElementToElement` and `downcastAttributeToElement` - it will be easier to use `view.Writer` then.

src/controller/datacontroller.js
```diff
@@ -143,17 +145,18 @@ export default class DataController {
 		const modelRange = ModelRange.createIn( modelElementOrFragment );
 
 		const viewDocumentFragment = new ViewDocumentFragment();
+		const viewWriter = new ViewWriter( new ViewDocument() );
 		this.mapper.bindElements( modelElementOrFragment, viewDocumentFragment );
 
-		this.downcastDispatcher.convertInsert( modelRange );
+		this.downcastDispatcher.convertInsert( modelRange, viewWriter );
```
pjasiun: We should make it symmetrical. Since `downcastDispatcher` do not have `view` as a property and `change` is called outside, `upcastDispatcher` may have no `model` and also get `writer` as a parameter.

src/controller/datacontroller.js
```diff
@@ -143,17 +145,18 @@ export default class DataController {
 		const modelRange = ModelRange.createIn( modelElementOrFragment );
 
 		const viewDocumentFragment = new ViewDocumentFragment();
+		const viewWriter = new ViewWriter( new ViewDocument() );
```
pjasiun: A comment is needed why there is no change block used here.

src/view/writer.js
```diff
@@ -12,704 +12,993 @@ import ContainerElement from './containerelement';
...
+	 * represented in other way, for example by applying proper CSS class.
+	 *
+	 * Additionally fake's selection label can be provided. It will be used to describe fake selection in DOM (and be
+	 * properly handled by screen readers).
+	 *
+	 * @param {Boolean} [value=true] If set to true selection will be marked as `fake`.
+	 * @param {Object} [options] Additional options.
+	 * @param {String} [options.label=''] Fake selection label.
+	 */
+	setFakeSelection( value, options ) {
```
pjasiun: I don't like this method name. Since we use `setSelection` to set ranges I would expect that this method does something with ranges. The better name would be `fakeSelection`. However,  it is still not perfect. You chnage selection direction using `setSelection`, but need to use the separate method to make it fake. We could change `backwardSelection` parameter of the selection to the `option` parameter and use it also to set fake selections. I will create a follow-up about it.

src/view/observer/mutationobserver.js
```diff
@@ -240,16 +240,16 @@ export default class MutationObserver extends Observer {
 			// Anchor and focus has to be properly mapped to view.
 			if ( viewSelectionAnchor && viewSelectionFocus ) {
 				viewSelection = new ViewSelection();
-				viewSelection.setTo( viewSelectionAnchor );
-				viewSelection.setFocus( viewSelectionFocus );
+				viewSelection._setTo( viewSelectionAnchor );
```
pjasiun: We need a separate `Selection` class with the public API and `DocumentSelection` used by document, simillary to what we have in the model.

src/view/editableelement.js
```diff
@@ -84,11 +84,10 @@ export default class EditableElement extends ContainerElement {
 			isFocused => isFocused && document.selection.editableElement == this
 		);
 
-		// Update focus state before each rendering. Rendering should not change neither the selection nor the value of
-		// document.isFocused property.
-		this.listenTo( document, 'render', () => {
+		// Update focus state based on selection changes.
+		this.listenTo( document.selection, 'change', () => {
```
pjasiun: Can we use `view#change` event here?

src/view/observer/mutationobserver.js
pjasiun: Same for all other observers which uses protected methods.

src/view/view.js
```diff
@@ -0,0 +1,372 @@
...
+
+import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
+import log from '@ckeditor/ckeditor5-utils/src/log';
+import mix from '@ckeditor/ckeditor5-utils/src/mix';
+import { scrollViewportToShowTarget } from '@ckeditor/ckeditor5-utils/src/dom/scroll';
+import { injectUiElementHandling } from './uielement';
+import { injectQuirksHandling } from './filler';
+
+/**
+ * Editor's view controller class.
```
pjasiun: I think we can improve this class docs. I will work on it.

src/view/writer.js
```diff
@@ -12,704 +12,993 @@ import ContainerElement from './containerelement';
...
+	 * Creates new {@link module:engine/view/element~Element element}.
+	 *
+	 *		writer.createElement( 'paragraph' );
+	 *		writer.createElement( 'paragraph', { 'alignment': 'center' } );
+	 *
+	 * @param {String} name Name of the element.
+	 * @param {Object} [attributes] Elements attributes.
+	 * @returns {module:engine/view/element~Element} Created element.
+	 */
+	createElement( name, attributes ) {
```
pjasiun: This method can be removed, since `domConverter` changes view document directly.

src/view/writer.js
```diff
@@ -12,704 +12,993 @@ import ContainerElement from './containerelement';
...
+		return new Element( name, attributes );
+	}
 
-		move( sourceRange, targetPosition );
+	/**
+	 * Creates new {@link module:engine/view/attributeelement~AttributeElement}.
+	 *
+	 *		writer.createAttributeElement( 'paragraph' );
+	 *		writer.createAttributeElement( 'paragraph', { 'alignment': 'center' } );
```
pjasiun: Paragraph is not an attribute element, so the example is wrong. You can use `strong` here.

src/view/writer.js
```diff
@@ -12,704 +12,993 @@ import ContainerElement from './containerelement';
...
- * @returns {module:engine/view/position~Position} Position after merge.
- */
-export function mergeAttributes( position ) {
-	const positionOffset = position.offset;
-	const positionParent = position.parent;
+	/**
+	 * Creates new {@link module:engine/view/editableelement~EditableElement}.
+	 *
+	 *		writer.createEditableElement( document, 'paragraph' );
+	 *		writer.createEditableElement( document, 'paragraph', { 'alignment': 'center' } );
```
pjasiun: I think `div` is a better example.

src/view/writer.js
```diff
@@ -12,704 +12,993 @@ import ContainerElement from './containerelement';
...
+	}
 
-	// When inside text node - nothing to merge.
-	if ( positionParent.is( 'text' ) ) {
-		return position;
+	/**
+	 * Creates new {@link module:engine/view/emptyelement~EmptyElement}.
+	 *
+	 *		writer.createEmptyElement( 'paragraph' );
+	 *		writer.createEmptyElement( 'paragraph', { 'alignment': 'center' } );
```
pjasiun: Paragraph is not an empty element, so the example is wrong. You can use `img` here.

src/view/view.js
```diff
@@ -0,0 +1,374 @@
...
+			callback( this._writer );
+		} else {
+			this._ongoingChange = true;
+
+			callback( this._writer );
+			this._render();
+
+			this._ongoingChange = false;
+
+			this.document.fire( 'change' );
```
pjasiun: As we talk, I am not sure about the order here. `change` event is fired after the rendering, what can be a suprose for someone. I think that rendering should be the result of the `change` event (listener with the `low` priority). Also both `change` and `render` should be inside `_ongoingChange` block, so there will be no infinite loop if one add some more change. Also if `view.change` blok is open after the rendering, in the `change` event, exception should be thrown.

src/view/editableelement.js
szymonkups: I would stay with `selection#change`. EditableElement's focus state depends on document focus and selection position, so it's updated whenever `document.isFocused` observable changes or view selection fires `#change` event.

PR #1298, opened 2018-02-12
Using model.Writer instead of constructor in tests and docs of conversion utils

### LINKED ISSUE #1291: Conversion utilities: Docs and tests create model elements using contructor
Follow-up to #1274.

In docs and tests we use `new ModelElement`. We should use `writer.createElement` instead.

scofalik:

Internal: Using `model.Writer` instead of `model.Element` constructor in tests and docs of conversion utils. Closes #1291.


pjasiun:
Fixed by https://github.com/ckeditor/ckeditor5-engine/pull/1302.

PR #1300, opened 2018-02-15
Two-steps caret movement

### LINKED ISSUE #1289: Implement two-step caret movement at styled text boundaries
Background: see https://github.com/ckeditor/ckeditor5-link/issues/72.

* We’ll implement only the “additional step on arrow key press” functionality.
   * You’re at a link boundary. The link is on.
       * NOTE: currently, the link is on if you’re at the end of the link. After this change it should be true also for the beginning of the link if the selection was moved there by using the left arrow key, but not if set there using mouse. This doesn't have to be a part of MVP – can be added later.
   * You press the right arrow key (if you’re at the end) and the link becomes disabled, but the caret doesn’t move.
   * You can press the left arrow key now to enable the link, but that won’t move the caret either.
   * This should work the same way on both ends of this style.
* We need the same behaviour for highlight because removing highlight from collapsed selection removes it completely (the whole highlight), so you can’t easily stop typing with highlight. Hence, let’s implement it for all styles.
* As the second step, after MVP, we should add a visual indicator that you’re typing with link. But this applies only to the link feature as it’s the most common feature which requires this indication.
* Behaviour with multiple styles:
   * `<b><i>x[]</b></i>x` –> `<b><i>x</b></i>[]x` (gets out of all styles at once)
   * `<b><i>x[]</b>x</i>` –> `<b><i>x</b>[]x</i>`
* There's an issue with the current proposal to listen on `keydown` (see https://github.com/ckeditor/ckeditor5-link/compare/hackathon-link-end) – selection attributes will be refreshed on next `applyOperation` which may fail when:
   * Collab changes come. Link may be restored.
   * You press Ctrl+B. Link may be restored.
* IDEA: Implement `DocumentSelection#overrideGravity()`. The gravity should be overridden until the current user changes the selection. More or less. We can also allow defining how long the gravity is overridden  by specifying your own callback or returning a function to be called to revert the gravity. Etc.
* UNSURE: This becomes an engine feature. Where should the whole implementation land? How can features configure whether this works for a certain attribute? Should this be configurable? What would happen if you’d be at the boundary of two styles – one which works like this and the other which doesn’t?

oskarwrobel:

Feature: Introduced two-steps caret movement mechanism. Closes #1289.

---

### Additional information


oskarwrobel:
I found two issues (not a problem for Link feature):
1. (minor) overriding gravity on `keyup` do it after caret move, so for a short moment selection is set according to not overridden gravity and that is noticeable (blinking toolbar button).

feb-16-2018 00-14-08

2. (major) when two attribute elements touch each other and two-steps caret behavior is enabled for both then is not possible to move caret from one to another while moving left
```html
`<u>foo</u><i>{}bar</i>`
```
feb-16-2018 00-10-26

oskarwrobel:
I'm afraid that second issue (from above) makes this implementation unacceptable.

Reinmar:
> (minor) overriding gravity on keyup do it after caret move, so for a short moment selection is set according to not overridden gravity and that is noticeable (blinking toolbar button).

Why is the gravity overriden on `keyup` and not `keydown`? 

> (major) when two attribute elements touch each other and two-steps caret behavior is enabled for both then is not possible to move caret from one to another while moving left

Yeah, this is bad. But if the first implementation will about links only we could postpone this issue. Thanks to that we could close this issue first and move to marking a link which is being edited. Only when we'll want to add this feature to the highlight feature we'll need to fix this scenario. So, I'd recommend treating it as a followup.

oskarwrobel:
> Why is the gravity overriden on keyup and not keydown?

Because while moving left we need to override gravity after caret is moved, but keydown is fired before the move. There are a few more options to handle that but `keyup` seemed to be the best solution.

Reinmar:
Why do we have to override the gravity after the caret is moved? Because we'd do that before, then we'd re-render the content or something and block the native caret move?

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * 		<p>foo<strong>bar</strong>[]biz<p> `<-` <p>foo<strong>bar[]</strong>biz<p>
+ * 		<p>foo<strong>[]bar</strong>biz<p> `<-` <p>foo[]<strong>bar</strong>biz<p>
+ *
+ * @param {module:core/editor/editor~Editor} editor The Editor instance.
+ * @param {module:utils/dom/emittermixin~Emitter} emitter The emitter to which this behavior should be added.
+ * @param {String} attribute Attribute for which behaviour will be added.
+ */
+export default function bindTwoStepCaretToAttribute( editor, emitter, attribute ) {
+	const model = editor.model;
```
Reinmar: AFAICS, the editor isn't needed here – we only need to pass the view and the model to this function. Let's make this function engine-only.

oskarwrobel:
Moving right is simple because first step is consistent with the gravity:
`fo{}o<a>bar</a>biz` <kbd>-></kbd> `foo{}<a>bar</a>biz` <kbd>-></kbd> `foo<a>{}bar</a>biz`

But moving left is different:
`foo<a>bar</a>b{}iz` <kbd><-</kbd>
and because of gravity carret goes inside attribute element
`foo<a>bar{}</a>biz` 
but we don't want to type with link, so we need to override the gravity `foo<a>bar</a>{}biz`

oskarwrobel:
We talked with @pjasiun and we had a few more ideas but like I wrote `keyup` seemed to be the right option.

oskarwrobel:
But because of these blinking toolbar buttons I think we should try to do it in some other way. If we decide that this PR is good enough to be merged I'll handle it in a followup.

Reinmar:
I know that we have to override the gravity, but I still don't understand why not sooner than on `keyup`. In other words – your comment doesn't answer the "why" part ;)

So – what would happen if we overridden the gravity on `keydown`?

oskarwrobel:
Changing selection range directly resets overridden gravity so when you override it on keydown and then move the caret it will reset. This is something I need to change.

Reinmar:
In other words – overridden gravity should not be reset on the first selection change following the overriding. Then, if `overrideGravity()` returned a callback through which you could control when revert happens, you might skip that first selection change. Am I right?

oskarwrobel:
Above issues are fixed:

feb-16-2018 17-37-52

oskarwrobel:
I found another two things to improve.

1. I don't check attribute value I check only if node has attribute, so e.g. two different link elements that stick each other will be treated like the same element (easy to fix).
2. When attribute element starts in one block and ends in the other then two steps is required to move between them.
```html
<p>fo<a href="url">o</a></p>
<p><a href="url">{}b</a>ar</p>
```
Left arrow needs to be pressed more than once to move caret to `<a>` from the first `<p>`. But TBH I'm not sure if it isn't an expected behavior 🤔

tests/manual/twostepscarret.js
```diff
@@ -0,0 +1,31 @@
+/**
```
Reinmar: Incorrectly named – `two-step-caret.js` is the right name (because these are three separate words).

src/model/documentselection.js
```diff
@@ -139,6 +139,10 @@ export default class DocumentSelection {
 		return this._selection.isBackward;
 	}
 
+	get isGravityOverridden() {
```
Reinmar: Misses docs.

src/model/documentselection.js
```diff
@@ -388,6 +392,27 @@ export default class DocumentSelection {
 		return this._selection._getStoredAttributes();
 	}
 
+	/**
+	 * Temporarily and partially disables default gravity behaviour that tries to get attributes from nodes surrounding the caret.
```
Reinmar: "behavior" in AmE.

And what does "partially" mean here?

Besides, it'd be good to make some minimal introduction here to what is that gravity.

> Temporarily changes the gravity of the selection from left to right. The gravity defines from which direction the selection inherits its attributes. If it's the default left gravity, the selection (after being moved by the user) inherits attributes from its left hand side. This method allows to temporarily override this behavior by forcing the gravity to the right.

src/model/documentselection.js
```diff
@@ -388,6 +392,27 @@ export default class DocumentSelection {
...
+	 *
+	 * @see module:engine/model/writer~Writer#overrideSelectionGravity
+	 * @protected
+	 * @param {Boolean} [customRestore=false] When `true` then gravity won't be restored automatically.
```
Reinmar: What does automatically mean here? What to do if you pass `false` here?

src/model/documentselection.js
```diff
@@ -388,6 +392,27 @@ export default class DocumentSelection {
...
+	 */
+	_overrideGravity( customRestore ) {
+		this._selection.overrideGravity( customRestore );
+	}
+
+	/**
+	 * Restore overridden gravity.
```
Reinmar: > Restores {@link xyz overridden gravity}.

src/model/documentselection.js
```diff
@@ -465,6 +490,13 @@ class LiveSelection extends Selection {
 		// @member {Array} module:engine/model/liveselection~LiveSelection#_hasChangedRange
 		this._hasChangedRange = false;
 
+		// When is set as `true` then selection attributes on node before the caret won't be taken
```
Reinmar: When set to `true`...

src/model/documentselection.js
```diff
@@ -601,6 +633,26 @@ class LiveSelection extends Selection {
 		}
 	}
 
+	overrideGravity( customRestore ) {
```
Reinmar: Docs?

src/model/documentselection.js
```diff
@@ -601,6 +633,26 @@ class LiveSelection extends Selection {
...
+					this.restoreGravity();
+					evt.off();
+				}
+			} );
+		}
+
+		this._updateAttributes();
+	}
+
+	restoreGravity() {
```
Reinmar: Docs?

src/model/documentselection.js
Reinmar: I can see that you wrote more thorough docs in the writer. Still, you can update these to what I described and you should link, in such a case, to the writer. Make it easier for developers to find the docs.

src/model/writer.js
```diff
@@ -1016,6 +1016,35 @@ export default class Writer {
...
+	 * 		<$text bold="true" linkHref="url">bar[]</$text><$text bold="true">biz</$text>
+	 *
+	 * Selection attribute keys before override will be equal `[ 'bold', 'linkHref' ]`
+	 * Selection attribute keys after override will be equal `[ 'bold' ]`
+	 *
+	 * As default gravity is automatically restored just after a direct selection change event but this behaviour
+	 * can be disabled by passing `true` flag as param.
+	 *
+	 * @param {Boolean} [customRestore=false] When `true` then gravity won't be restored until
+	 * {@link ~Writer#overrideSelectionGravity} will be called directly.
```
Reinmar: restoreSelectionGravity

src/model/writer.js
```diff
@@ -1016,6 +1016,35 @@ export default class Writer {
 		}
 	}
 
+	/**
+	 * Disables default gravity behaviour that tries to get attributes from nodes surrounding the caret. When gravity is
```
Reinmar: You can update based on the description I wrote for the protected method.

src/model/documentselection.js
oskarwrobel: > and you should link, in such a case, to the writer

there is a link
```
* @see module:engine/model/writer~Writer#overrideSelectionGravity
```

src/model/writer.js
```diff
@@ -1016,6 +1016,35 @@ export default class Writer {
...
+	 * marked as overridden then attributes from the node before the caret won't be taken into consideration while
+	 * updating selection attributes.
+	 *
+	 * For the following model fragment:
+	 *
+	 * 		<$text bold="true" linkHref="url">bar[]</$text><$text bold="true">biz</$text>
+	 *
+	 * Selection attribute keys before override will be equal `[ 'bold', 'linkHref' ]`
```
Reinmar: * Default gravity: selection will have the `bold` and `linkHref` attributes.
* Overridden gravity: ...

src/model/writer.js
```diff
@@ -1016,6 +1016,35 @@ export default class Writer {
...
+	 * Selection attribute keys after override will be equal `[ 'bold' ]`
+	 *
+	 * As default gravity is automatically restored just after a direct selection change event but this behaviour
```
Reinmar: By default the selection's gravity is automatically restored just after a direct selection change ( 
 [[[WHAT IS DIRECT SELECTION CHANGE?]]]  ) but you can pass `customRestore = true` in which case you will have to call ... manually.

src/model/writer.js
```diff
@@ -1016,6 +1016,35 @@ export default class Writer {
...
+	 * {@link ~Writer#overrideSelectionGravity} will be called directly.
+	 */
+	overrideSelectionGravity( customRestore ) {
+		this.model.document.selection._overrideGravity( customRestore );
+	}
+
+	/**
+	 * Restore overridden gravity to default.
+	 */
+	restoreSelectionGravity() {
```
Reinmar: What would happen if we had two features which independently override and restore gravity? They might start to conflict with each other. Instead of:

1. A: override
1. A: restore
1. B: override
1. B: restore

The sequence might look like this:


1. A: override
1. B: override
1. A: restore
1. B: restore

In which case, the B feature's decision to override the gravity might be immediately cancelled by feature A which decided that it's time to restore the gravity.

What I'd propose is that `overrideSelectionGravity()` returned an identifier of "overriding" and this identifier will need to be passed to `restoreSelectionGravity()`.

Normally, I'd propose that `overrideSelectionGravity()` returned a callback, but we must enforce calling it in the `change()` block, so that's why the identifier.

Reinmar: Ofc, the default gravity will be restored only after all features restored gravity.

Actually, this could be done even simpler by just having a counter of how many times `override()` was called. This could be ok too.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -0,0 +1,139 @@
...
+			return;
+		}
+
+		// This implementation works only for collapsed selection.
+		if ( !modelSelection.isCollapsed ) {
+			return;
+		}
+
+		// User tries to expand selection, so two-steps movement is not necessary.
+		if ( data.shiftKey ) {
```
Reinmar: What if the user tries to jump over the whole word with alt/ctrl?

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -0,0 +1,139 @@
...
+			// If gravity is already overridden then do nothing.
+			// It means that we already enter `foo<a>{}bar</a>biz` or left `foo<a>bar</a>{}biz` text with attribute
+			// and gravity will be restored just after caret movement.
+			if ( modelSelection.isGravityOverridden ) {
+				return;
+			}
+
+			// If caret sticks to the bound of Text with attribute it means that we are going to
+			// enter `foo{}<a>bar</a>biz` or leave `foo<a>bar{}</a>biz` the text with attribute.
+			if ( isStickToAttribute( position.nodeAfter, position.nodeBefore, attribute ) ) {
```
Reinmar: stick is a verb so `sticksToAttribute()` or something like `atAttributeBoundary()`?

src/model/documentselection.js
Reinmar: Sorry, I forgot I found it there in the meantime :D

Reinmar:
The first batch of code review done. I also played with this manually. It works really nice but I found two issues:

* that alt/ctrl modifiers don't stop this behaviour to kick in,
* that in `<p>x</p><p><u>[]x</u></p>` when you press the left arrow, it's blocked, but the gravity doesn't change (the underline button is still on).

oskarwrobel:
> that in `<p>x</p><p><u>[]x</u></p>` when you press the left arrow, it's blocked, but the gravity doesn't change (the underline button is still on).

I was planning to handle that in a followup - https://github.com/ckeditor/ckeditor5-engine/issues/1301

src/model/documentselection.js
oskarwrobel: `LiveSelecton` class is private and has no docs.
```
// `LiveSelection` is used internally by {@link module:engine/model/documentselection~DocumentSelection} and shouldn't be used directly.
```

PR #1302, opened 2018-02-16
T/1295 Further improvements to conversion helpers

### LINKED ISSUE #1290: Conversion utilities: specify model element name in downcastAttributeToAttribute
Follow-up to #1274.

When refactoring Image plugin, I noticed, that new converters are missing one functionality. Right now, we support upcasting an attribute from given view element, but we don't support same thing for downcasting.

This means that we can't state that we define conversion of, for example, `src` attribute that is set on model `image` element.

### LINKED ISSUE #1291: Conversion utilities: Docs and tests create model elements using contructor

### LINKED ISSUE #1292: Conversion utilities: better handling for classes and styles in downcast
Follow-up to #1274.

Right now there is no way to properly and easily handle setting and removing `class`es and `style`s during downcasting.

You can of course use the utilities and (un)set `class` attribute but this will overwrite already set classes.

### LINKED ISSUE #1293: Conversion utilities: move two-way converters as methods of `Conversion` class
It would be very useful in tests and simple features to writer:
```
editor.conversion.elementToElement( { model: 'paragraph', view: 'p' } );
```

No imports.

I think there is no point in exporting these functions as separate utils too, so the whole content of  `two-way-conveters` should be moved to the `Conversion` class.

### LINKED ISSUE #1294: Conversion utilities: Unify parameters received by callbacks in downcast and upcast configuration
Follow-up to #1274.

Right now, in element-to-element conversion, if we pass a callback creating an element, the callback is called with different parameters for upcasting and downcasting.

For downcasting it is `( modelItem, consumable, conversionApi )`.

For upcasting it is `( viewItem, modelWriter )`.

We should unify them. Callbacks for downcasting are fired like this because of the good-ol' converters we already had. We need to think whether we really need `consumable` and whole `conversionApi` in those callbacks. Maybe we will settle on: `( item, writer, conversionApi )`?

### LINKED ISSUE #1295: Conversion utilities: abandon multiples values attribute converters
We based our conversion utils, among others, on definition-based-converters. This util was focused on the specific case, where the model attribute is an enum with several possible values. This lead to the API which is very inconvenient in other cases.

Refactoring converters we realized that attributes values come in 4 types:
 - boolean (like bold), 
 - string (link img href),
 - object (like img srcset),
 - enum (like font family).

Most of them need only a single converter (often a callback).

Because we focus on creating conversion utilities the way that you can define multiple values they look like this for boolean:
```js
editor.conversion.for( 'downcast' ).add( downcastAttributeToElement( 'bold', { view: 'strong' } ) );
editor.conversion.for( 'upcast' ).add( upcastElementToAttribute( { view: 'strong', model: 'bold' } ) )
```

And like this for callbacks:
```js
editor.conversion.for( 'downcast' )
	.add( downcastAttributeToElement( 'linkHref', { view: linkHref => createLinkElement( linkHref ) } ) );

editor.conversion.for( 'upcast' )
	.add( upcastElementToAttribute( {
		view: { name: 'a', attribute: { href: true } },
		model: {
			key: 'linkHref',
			value: viewElement => viewElement.getAttribute( 'href' )
		}
	} ) );
```

They are very unsymmetrical: in `downcastElementToElement` and all `upcast*` methods you pass object with `model` and `view` properties, but for `downcastAttributeToElement` and `downcastAttributeToAttribute` parameters are very different. I updated multiple converters to the new API and had to check docs every time to be sure which one should be defined which way because as long as you do not use enum it is not natural to write converters like this.

Also, since `upcastElementToAttribute` accepts only a single attribute value to be set, this does not solve any problem.

My proposal is to do it this way. 

For `downcastAttributeToElement`:

```js
downcastAttributeToElement( { model: 'bold', view: 'strong' } );

downcastAttributeToElement( { model: 'bold', view: 'strong' }, 'high' );

downcastAttributeToElement( { model: 'bold', view: new ViewAttributeElement( 'strong' ) } );

downcastAttributeToElement( {
	model: 'bold',
	view: {
		name: 'span',
		class: 'bold'
	}
} );

downcastAttributeToElement( {
	model: {
		key: 'styled',
		value: 'dark'
	}
	view: {
		name: 'span',
		class: [ 'styled', 'styled-dark' ]
	}
} );

downcastAttributeToElement( {
	model: {
		key: 'fontSize',
	 	value: 'big'
	}
	view: {
		name: 'span',
		style: {
			'font-size': '1.2em'
		}
	}
} );
downcastAttributeToElement( {
	model: {
		key: 'fontSize',
	 	value: 'small'
	}
	view: {
		name: 'span',
		style: {
			'font-size': '0.8em'
		}
	}
} );

downcastAttributeToElement( {
	model: 'bold',
	view: modelAttributeValue => new ViewAttributeElement( 'span', { style: 'font-weight:' + modelAttributeValue } )
} );
```
Check the current API here: https://github.com/ckeditor/ckeditor5-engine/blob/fd128a1c7ee5aef3e5aab788726af89c88059b64/src/conversion/downcast-converters.js#L78-L122

For `downcastAttributeToAttribute`:

```js
downcastAttributeToAttribute( { model: 'source', view: 'src' } );

downcastAttributeToAttribute( { model: 'source', view: 'src' }, 'high' );

downcastAttributeToAttribute( {
	model: 'stylish',
	view: {
		key: 'class',
		value: 'styled'
	}
} );

downcastAttributeToAttribute( {
	model: {
		key: 'styled',
		value: 'dark'
	} 
	view: {
		key: 'class',
		value: 'styled styled-dark'
	}
} );

downcastAttributeToAttribute( {
	model: {
		key: 'styled',
		value: 'dark'
	} 
	view: {
		key: 'class',
		value: 'styled-dark'
	}
} );
downcastAttributeToAttribute( {
	model: {
		key: 'styled',
		value: 'light'
	}
	view: {
		key: 'class',
		value: 'styled-light'
	}
} );

downcastAttributeToAttribute( {
	model: 'style',
	view: attributeValue => ( { key: 'class', value: 'style-' + attributeValue } )
} );
```
Check the current API here: https://github.com/ckeditor/ckeditor5-engine/blob/fd128a1c7ee5aef3e5aab788726af89c88059b64/src/conversion/downcast-converters.js#L155-L195.

Then we have plugins configuration and two-way config.

Besides the problems mentioned above (inconvenient configuration for booleans, string and object attributes) there are 2 more issues. First, the config might be giant https://github.com/ckeditor/ckeditor5-engine/blob/fd128a1c7ee5aef3e5aab788726af89c88059b64/src/conversion/two-way-converters.js#L169-L230 or https://github.com/ckeditor/ckeditor5-engine/blob/fd128a1c7ee5aef3e5aab788726af89c88059b64/src/conversion/two-way-converters.js#L298-L327.

Second, is that if we will keep `model` string value to be the attribute key, we will get 2 confusing configuration:
 - in one configuration `model` attributes key,
 - in the other configuration `model` attributes value.

This is why, plugins configuration should be defined different way, to make it clear that this is attribute value and make it compatible with the `downcast*` configuration:
```js
[
	{
		model: {
			value: 'right'
		},
		view: {
			key: 'class',
			value: 'align-right'
		}
	},
	{
		model: {
			value: 'center'
		},
		view: {
			key: 'class',
			value: 'align-center'
		},
		upcastAlso: {
			style: {
				'text-align': 'center'
			}
		}
	}
]
```

Then, we may have a util to add `model.key` property to the configuration and execute it. As soon as someone finds a good name for such helper, I am fine with it, but since all it will do is:
```js
for( const config in editor.config.get( 'align' ) ) {
	config.model.key = 'align';

	editor.conversion.attributeToElement( config );
}
```
I think that all plugins which have to do so can do it directly in the plugin code.

Note that some plugins may do more and this code might be very specific with all normalization. Also, in some cases we need to define additionally `model.name`: https://github.com/ckeditor/ckeditor5-engine/issues/1290.

### LINKED ISSUE #1305: Usage of writer in _normalizeToElementConfig
`_normalizeToElementConfig` method should use writer to create view elements. Is should be changed after https://github.com/ckeditor/ckeditor5-engine/issues/1294.

scofalik:

Other: Introduced several improvements to conversion helpers. Closes #1295. Closes #1293. Closes #1292. Closes #1291. Closes #1290. Closes #1305.

---

### Additional information

Basing and waiting for #1296. After #1296 is closed, I'll make PRs in other repositories related to this PR.


scofalik:
Related PRs:

https://github.com/ckeditor/ckeditor5-alignment/pull/22
https://github.com/ckeditor/ckeditor5-basic-styles/pull/66
https://github.com/ckeditor/ckeditor5-block-quote/pull/20
https://github.com/ckeditor/ckeditor5-heading/pull/97
https://github.com/ckeditor/ckeditor5-image/pull/179
https://github.com/ckeditor/ckeditor5-link/pull/168
https://github.com/ckeditor/ckeditor5-list/pull/96
https://github.com/ckeditor/ckeditor5-paragraph/pull/32
https://github.com/ckeditor/ckeditor5-widget/pull/33

src/conversion/conversion.js
```diff
@@ -102,6 +114,347 @@ export default class Conversion {
 		};
 	}
 
+	/**
+	 * Sets up converters between the model and the view which convert a model element to a view element (and vice versa).
+	 * For example, model `<paragraph>Foo</paragraph>` is `<p>Foo</p>` in the view.
+	 *
+	 * `definition.model` is a `String` with a model element name to converter from/to.
```
pjasiun: I would put it togheter with the parameter desctription. When it is here, I ask myself why you tell me what `model` is, but not what `view` is.

scofalik: It's here because in the parameter definition specifies that input parameter is `ConverterDefinition`. I am not sure if it will be readable if such a long description would be put there.

src/conversion/conversion.js
```diff
@@ -102,6 +114,347 @@ export default class Conversion {
...
+			);
+		}
+	}
+
+	/**
+	 * Sets up converters between the model and the view which convert a model attribute to a view element (and vice versa).
+	 * For example, model text node with data `"Foo"` and `bold` attribute is `<strong>Foo</strong>` in the view.
+	 *
+	 * `definition.model` parameter specifies what model attribute should be converted from/to. It can be a `{ key, value }` object
+	 * describing attribute key and value to convert or a `String` specifying just attribute key (then `value` is set to `true`).
```
pjasiun: As above.

src/conversion/conversion.js
```diff
@@ -102,6 +114,347 @@ export default class Conversion {
...
+	 *
+	 *		// Use `upcastAlso` to define other view elements that should be also converted to `paragraph` element.
+	 *		conversion.elementToElement( {
+	 *			model: 'paragraph',
+	 *			view: 'p',
+	 *			upcastAlso: [
+	 *				'div',
+	 *				{
+	 *					// Any element with `display: block` style.
+	 *					name: /./,
```
pjasiun: I think we should improve it, like we did for attributes in https://github.com/ckeditor/ckeditor5-engine/issues/1239. It should be no `name` in this case.

src/conversion/conversion.js
```diff
@@ -102,6 +114,347 @@ export default class Conversion {
...
+	 *				const match = fontSize.match( /(\d+)\s*px/ );
+	 *
+	 *				if ( !match ) {
+	 *					return null;
+	 *				}
+	 *
+	 *				const size = Number( match[ 1 ] );
+	 *
+	 *				if ( size > 26 ) {
+	 *					return { name: true, style: [ 'font-size' ] };
```
pjasiun: There should be a comment that this is information what should be consumed with the link to the full documentation. `name: true` looks complcated with no explanation.

src/conversion/conversion.js
```diff
@@ -102,6 +114,347 @@ export default class Conversion {
...
+	 *					style: {
+	 *						'font-weight': 'bold'
+	 *					}
+	 *				},
+	 *				viewElement => {
+	 *					const fontWeight = viewElement.getStyle( 'font-weight' );
+	 *
+	 *					if ( viewElement.is( 'span' ) && fontWeight && /\d+/.test() && Number( fontWeight ) > 500 ) {
+	 *						return {
+	 *							name: true,
```
pjasiun: As above.

src/conversion/conversion.js
```diff
@@ -102,6 +114,347 @@ export default class Conversion {
...
+			);
+		}
+	}
+
+	/**
+	 * Sets up converters between the model and the view which convert a model attribute to a view attribute (and vice versa).
+	 * For example, `<image src='foo.jpg'></image>` is converted to `<img src='foo.jpg'></img>` (same attribute key and value).
+	 *
+	 * `definition.model` parameter specifies what model attribute should be converted from/to.
+	 * It can be a `{ key, values, [ name ] }` object or a `String`, which will be treated like `{ key: definition.model }`.
```
pjasiun: `{ key, [ values ], [ name ] }` (value is optional too)

src/conversion/conversion.js
```diff
@@ -129,6 +482,23 @@ export default class Conversion {
...
+ * for different functions that accepts `ConverterDefinition`. See the description of a function to learn how to set it.
+ * @property {module:engine/view/elementdefinition~ElementDefinition|Object} view Definition of a view element to convert from/to.
+ * If `model` describes multiple values, `view` is an object that assigns those values (`view` object keys) to view element definitions
+ * (`view` object values).
+ * @property {module:engine/view/matcher~MatcherPattern|Array.<module:engine/view/matcher~MatcherPattern>} [upcastAlso]
+ * Any view element matching `upcastAlso` will also be converted to model. If `model` describes multiple values, `upcastAlso`
+ * is an object that assigns those values (`upcastAlso` object keys) to {@link module:engine/view/matcher~MatcherPattern}s
+ * (`upcastAlso` object values).
+ * @property {module:utils/priorities~PriorityString} [priority] Conversion priority.
+ */
```
pjasiun: I think you should not use `/**` in private methods. Use `//`.

src/conversion/conversion.js
pjasiun: I know, but users will not know it. You can write that all other parameters are defined in `ConverterDefinition`, or move it after samples close to the parameter definition.

src/conversion/conversion.js
scofalik: The comment might be helpful. I haven't added it cause, AFAIR, it is commented in `MatcherPattern` description.

src/conversion/conversion.js
```diff
@@ -102,6 +114,347 @@ export default class Conversion {
...
+	 *					const size = Number( match[ 1 ] );
+	 *
+	 *					if ( viewElement.is( 'span' ) && size < 10 ) {
+	 *						return { name: true, style: [ 'font-size' ] };
+	 *					}
+	 *
+	 *					return null;
+	 *				}
+	 *			}
+	 *		} );
```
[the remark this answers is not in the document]
scofalik: Thumbs up for...? :P

src/conversion/conversion.js
scofalik: As we are at it, I wanted to ask you, whether it is better to name that property `values` or `value` (it is always an array).

src/conversion/conversion.js
scofalik: I will move it lower.

scofalik:
IDK why this build: https://travis-ci.org/ckeditor/ckeditor5/builds/343343886 is still failing on docs, I've fixed all docs problems and pushed commits and I am clean on my local machine.

src/conversion/conversion.js
scofalik: This is a tricky example, actually.

First of all, for `Matcher` purposes, specifying something like:

```
{
	name: true,
	class: ...
}
```

is just stupid, cause the name could be omitted. It says "any name" and every element has a name.

OTOH, the `name` is in upcast definition intentionally, so that `name` is returned in a match so it will be consumed.

I think that if we want to improve that, we should add `name` to matches object (the one returned by `Matcher#match`) in `elementToElement` and `attributeToElement` (so whenever there is a view element is converted.

scofalik: Okay I did as described above ☝️

src/conversion/conversion.js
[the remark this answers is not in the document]
pjasiun: For a very nice structure :)

PR #1310, opened 2018-02-19
Corrected offests transformation in model.Differ when multiple change items interfere with each other

### LINKED ISSUE #1309: Differ: changes' offsets are incorrectly recalculated when multiple changes are done in one element
Original issue: https://github.com/ckeditor/ckeditor5/issues/849

The issue is caused by an incorrect change transformation. It happens when there are three changes in an element and:
- the second change interfered with first change,
- the third change interfered both with first change and the second change.

In that scenario, the second change has not been correctly transformed by the second change.

scofalik:

Fix: Corrected offests transformation in `model.Differ` when multiple change items interfere with each other. Closes #1309. Closes https://github.com/ckeditor/ckeditor5/issues/849.

### DIFF
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -547,7 +547,23 @@ export default class Differ {
 	 * @param {Array.<Object>} changes An array containing all the changes done on that element.
 	 */
 	_handleChange( inc, changes ) {
-		inc.newHowMany = inc.howMany;
+		// We need a helper variable that will store how many nodes are to be still handled for this change item.
+		// `nodesToHandle` (how many nodes still need to be handled) and `howMany` (how many nodes were affected)
+		// needs to be differentiated.
+		//
+		// This comes up when there are multiple changes that are affected by `inc` change item.
+		//
+		// For example: assume two insert changes: `{ offset: 2, howMany: 1 }` and `{ offset: 5, howMany: 1 }`.
+		// Assume that `inc` change is remove `{ offset: 2, howMany: 2, nodesToHandle: 2 }`.
+		//
+		// Then, we:
+		// - "forget" about first insert change (it is "eaten" by remove),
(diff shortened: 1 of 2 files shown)


src/model/differ.js
```diff
@@ -547,6 +547,8 @@ export default class Differ {
 	 * @param {Array.<Object>} changes An array containing all the changes done on that element.
 	 */
 	_handleChange( inc, changes ) {
+		inc.newHowMany = inc.howMany;
```
pjasiun: Some comment what `newHowMany` is and why it is needed is needed.

pjasiun:
Missing tests.

Mgsy:
After switching to this branch, there is the error in the console on init:
```
article.js:24 TypeError: editor.conversion.elementToElement is not a function
    at BlockQuoteEngine.init (blockquoteengine.js:42)
    at <anonymous>
```

scofalik:
@pjasiun I will add a test.
@Mgsy are you sure you have all branches on `master`?

Mgsy:
>@Mgsy are you sure you have all branches on master?

Yes, all branches are on `master`. I've removed all packages and installed it once again and the problem still exists.

scofalik:
First, I had to refactor all the tests, because they were incorrect after engine refactoring. They worked by a coincidence.

Other than that, I rebased this branch with `master` and set all other branches on current `master`s. Tests are all green. I also added the missing test for this case.

Mgsy:
The crash is gone, but data is restored incorrectly:

bug_ckeditor5

It will be a part of this PR or should I report it in the [Undo umbrella](https://github.com/ckeditor/ckeditor5-undo/issues/65)?

scofalik:
Yes, I already reported the bug in the undo ticket.

PR #1313, opened 2018-02-20
View render() and change() methods create new change block

### LINKED ISSUE #1312: Calling view.change() and view.render() on 'render' event with low priority
During the view refactoring we created a mechanism that prevents changing the view after rendering to the DOM is started. It throws special exception when we call `View#render()` or `View#change()` after rendering to DOM is done:

```js
// Listening on `low` priority when rendering to DOM is over.
view.on( 'render' , () => {
   // This will throw because View#render event on `low` priority indicates that
   // rendering is over for this change() block.
   view.render(); // or view.change( () => {} );
}, { priority: 'low' } );
```

Calling `View#render()` or `View#change()` should be possible - it should just start another rendering process. 

Motivation:
There are situations when we want to listen to the `render` event on `low` priority and re-render the view afterwards. For example: balloon panel wants to re-position or close itself after DOM is rendered and focus back the main editable (focusing means that view should be re-rendered).

CC: @pjasiun.

szymonkups:

Other: View render() and change() methods create new change block when called after rendering.
 Closes #1312.

---
Please close together with depending branch: [ckeditor5-image:t/ckeditor5-engine/1312](https://github.com/ckeditor/ckeditor5-image/tree/t/ckeditor5-engine/1312).

### DIFF
--- a/src/view/view.js
+++ b/src/view/view.js
@@ -319,7 +319,19 @@ export default class View {
 	change( callback ) {
 		if ( this._renderingInProgress || this._postFixersInProgress ) {
 			// TODO: better description
-			throw new CKEditorError( 'incorrect-view-change' );
+			/**
+			 * Thrown when there is an attempt to make changes to the view tree when it is in incorrect state. This may
+			 * cause some unexpected behaviour and inconsistency between the DOM and the view.
+			 * This may be caused by:
+			 *   * calling {@link #change} or {@link #render} during rendering process,
+			 *   * calling {@link #change} or {@link #render} inside of
+			 *   {@link module:engine/view/document~Document#registerPostFixer post fixer function}.
+			 */
+			throw new CKEditorError(
+				'cannot-change-view-tree: ' +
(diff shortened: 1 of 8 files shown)


src/view/view.js
```diff
@@ -343,8 +340,7 @@ export default class View {
 		// Render only if no ongoing changes are in progress. If there are some, view document will be rendered after all
 		// changes are done. This way view document will not be rendered in the middle of some changes.
 		if ( !this._ongoingChange ) {
```
pjasiun: Do we need this if here now?

pjasiun: `this._assertRenderingInProgress();` is not needed here too.

pjasiun:
I'm afraid that the implementation is incorrect. Right now, if I add `low` listener to the `render` event, another `render` event will be fired imminently, before the first render event is done. It might lead to bugs in some edge cases. I pushed a test for it.

If `render` or `change` is called in `low` or `lowest` listener, a flag should be set, and the event should be fired again later.

PR #1321, opened 2018-02-23
model.Writer#createText should throw when trying to create empty text node

### LINKED ISSUE #1320: model.Writer should not be letting creating empty text nodes

scofalik:

Other: `model.Writer#createText` should throw when trying to create empty text node. Closes #1320.

---

### Additional information

Related to https://github.com/ckeditor/ckeditor5-link/pull/170

### DIFF
--- a/src/model/writer.js
+++ b/src/model/writer.js
@@ -85,14 +85,25 @@ export default class Writer {
 	/**
 	 * Creates a new {@link module:engine/model/text~Text text node}.
 	 *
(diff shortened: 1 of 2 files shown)


Reinmar:
It's better to abort (or throw) once an empty text node is used somewhere. E.g. one might want to create an empty text node to append data to it later on and then insert it.

Therefore, I prefer https://github.com/ckeditor/ckeditor5-engine/pull/1322.

PR #1322, opened 2018-02-26
Prevent model.Writer from inserting empty text nodes

### LINKED ISSUE #1320: model.Writer should not be letting creating empty text nodes

scofalik:

Fix: Prevent `model.Writer` from inserting empty text nodes. Closes #1320.

---

### Additional information

Alternative solution to: https://github.com/ckeditor/ckeditor5-engine/pull/1321

### DIFF
--- a/src/model/writer.js
+++ b/src/model/writer.js
@@ -156,6 +156,7 @@ export default class Writer {
 	insert( item, itemOrPosition, offset ) {
 		this._assertWriterUsedCorrectly();
 
+		// Empty text node should not be added to the model. See #1320.
 		if ( item instanceof Text && item.data == '' ) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


src/model/writer.js
```diff
@@ -222,6 +226,10 @@ export default class Writer {
 	 * third parameter is a {@link module:engine/model/item~Item model item}.
 	 */
 	insertText( text, attributes, itemOrPosition, offset ) {
+		if ( text == '' ) {
```
Reinmar: Is this necessary?

Reinmar:
I added tests for inserting an array of items and those empty text nodes there are normalized. That's weird that inserting a single text node fails.

Reinmar:
I think that the patch should be somewhere in text nodes normalization.

scofalik:
This PR is a bit outdated - I'll create a new one.

PR #1325, opened 2018-02-26
Writer#setSelection and Selection#setTo improvements

### LINKED ISSUE #1308: writer.setSelection improvements
There are two problems with the current `writer.setSelection` implementation (both model and view).

1. It missleanding when single `element` is passed as a parameter. `writer.setSelection( image )` means `writer.setSelection( image, 0 )` and will put a collapsed selection in the offset `0` in the image. It is unexpected.
2. It is not possible to set a fake selection using this method.

Updated API should looks like this:
```js
witer.setSelection( paragraph, 'in' );
// Do: witer.setSelection( Range.createIn( paragraph ) );

witer.setSelection( paragraph, 'on' );
// Do: witer.setSelection( Range.createOn( paragraph ) );

witer.setSelection( paragraph, 0 );
witer.setSelection( paragraph, 'after' );
// Works the same way as now.

witer.setSelection( paragraph );
// Throws an exception.

witer.setSelection( range, { backward: true } );
witer.setSelection( paragraph, 'in', { backward: true } );
// Set backward selection. It should be object instead of the boolean value.

viewWiter.setSelection( range, { fake: true, label: 'Foo' } );
// Set fake selection. With 'Foo' as the label. 
```

ma2ciek:

Other: Improved `Writer#setSelection()` and `Selection#setTo()` methods. Closes #1308.

---

### Additional information

BREAKING CHANGE: `Writer#setFakeSelection()` method is removed. You should use `Writer#setSelection()` method with additional options (`fake` and `label`) to achieve the same effect.
BREAKING CHANGE: `Writer#setSelection( item )` now requires second parameter. It can be offset or `in` / `on` to create selection in / on the item.

Affected repos:
* https://github.com/ckeditor/ckeditor5-typing/tree/t/ckeditor5-engine/1308
* https://github.com/ckeditor/ckeditor5-link/tree/t/ckeditor5-engine/1308
* https://github.com/ckeditor/ckeditor5-list/tree/t/ckeditor5-engine/1308
* https://github.com/ckeditor/ckeditor5-paragraph/tree/t/ckeditor5-engine/1308
* https://github.com/ckeditor/ckeditor5-undo/tree/t/ckeditor5-engine/1308
* https://github.com/ckeditor/ckeditor5-widget/tree/t/ckeditor5-engine/1308
* https://github.com/ckeditor/ckeditor5-enter/tree/t/ckeditor5-engine/1308
* https://github.com/ckeditor/ckeditor5-clipboard/tree/t/ckeditor5-engine/1308

### DIFF
--- a/src/model/selection.js
+++ b/src/model/selection.js
@@ -56,19 +56,26 @@ export default class Selection {
 	 *		const position = new Position( root, path );
 	 *		const selection = new Selection( position );
 	 *
-	 * 		// Creates selection at the start position of given element.
+	 * 		// Creates selection inside the node.
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		selection.setTo( paragraph, 'in', { backward } );
+	 *
+	 *		// Creates selection on the node.
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		selection.setTo( paragraph, 'on', { backward } );
+	 *
+	 * 		// Creates selection at the start position of the given element.
 	 *		const paragraph = writer.createElement( 'paragraph' );
 	 *		const selection = new Selection( paragraph, offset );
 	 *
 	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection|
 	 * module:engine/model/position~Position|module:engine/model/element~Element|
 	 * Iterable.<module:engine/model/range~Range>|module:engine/model/range~Range|null} selectable
-	 * @param {Object|Number|'before'|'end'|'after'|'on'|'in'} [optionsOrPlaceOrOffset]
-	 * @param {Boolean} [optionsOrPlaceOrOffset.backward]
+	 * @param {Number|'before'|'end'|'after'|'on'|'in'} [placeOrOffset]
 	 * @param {Object} [options]
 	 * @param {Boolean} [options.backward]
 	 */
-	constructor( selectable, optionsOrPlaceOrOffset, options ) {
+	constructor( selectable, placeOrOffset, options ) {
 		/**
 		 * Specifies whether the last added range was added as a backward or forward range.
 		 *
@@ -94,7 +101,7 @@ export default class Selection {
 		this._attrs = new Map();
 
 		if ( selectable ) {
-			this.setTo( selectable, optionsOrPlaceOrOffset, options );
+			this.setTo( selectable, placeOrOffset, options );
 		}
 	}
 
@@ -304,52 +311,51 @@ export default class Selection {
 	 * {@link module:engine/model/element~Element element}, {@link module:engine/model/position~Position position},
 	 * {@link module:engine/model/range~Range range}, an iterable of {@link module:engine/model/range~Range ranges} or null.
 	 *
-	 *		// Sets ranges from the given range.
+	 *		// Sets selection to the given range.
 	 *		const range = new Range( start, end );
 	 *		selection.setTo( range, { backward } );
 	 *
-	 *		// Sets ranges from the iterable of ranges.
+	 *		// Sets selection to given ranges.
 	 * 		const ranges = [ new Range( start1, end2 ), new Range( star2, end2 ) ];
 	 *		selection.setTo( ranges, { backward } );
 	 *
-	 *		// Sets ranges from the other selection.
+	 *		// Sets selection to other selection.
 	 *		// Note: It doesn't copies selection attributes.
 	 *		const otherSelection = new Selection();
 	 *		selection.setTo( otherSelection );
 	 *
-	 * 		// Sets ranges from the given document selection's ranges.
+	 * 		// Sets selection to the document selection.
 	 *		// Note: It doesn't copies selection attributes.
 	 *		const documentSelection = new DocumentSelection( doc );
 	 *		selection.setTo( documentSelection );
 	 *
-	 * 		// Sets range at the given position.
+	 * 		// Sets collapsed selection at the given position.
 	 *		const position = new Position( root, path );
 	 *		selection.setTo( position );
 	 *
-	 * 		// Sets range at the position of given node and offset.
+	 * 		// Sets collapsed selection at the position of given node and offset.
 	 *		const paragraph = writer.createElement( 'paragraph' );
 	 *		selection.setTo( paragraph, offset );
 	 *
-	 *		// Sets range inside the node.
+	 *		// Sets selection inside the node.
 	 *		const paragraph = writer.createElement( 'paragraph' );
-	 *		selection.setTo( paragraph, 'in' );
+	 *		selection.setTo( paragraph, 'in', { backward } );
 	 *
-	 *		// Sets range on the node.
+	 *		// Sets selection on the node.
 	 *		const paragraph = writer.createElement( 'paragraph' );
-	 *		selection.setTo( paragraph, 'on' );
+	 *		selection.setTo( paragraph, 'on', { backward } );
 	 *
-	 * 		// Removes all ranges.
+	 * 		// Clears selection. Removes all ranges.
 	 *		selection.setTo( null );
 	 *
 	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection|
 	 * module:engine/model/position~Position|module:engine/model/node~Node|
 	 * Iterable.<module:engine/model/range~Range>|module:engine/model/range~Range|null} selectable
-	 * @param {Object|Number|'before'|'end'|'after'|'on'|'in'} [optionsOrPlaceOrOffset]
-	 * @param {Boolean} [optionsOrPlaceOrOffset.backward]
+	 * @param {Number|'before'|'end'|'after'|'on'|'in'} [placeOrOffset]
 	 * @param {Object} [options]
 	 * @param {Boolean} [options.backward]
 	 */
-	setTo( selectable, optionsOrPlaceOrOffset, options ) {
+	setTo( selectable, placeOrOffset, options ) {
 		if ( selectable === null ) {
 			this._setRanges( [] );
 		} else if ( selectable instanceof Selection ) {
@@ -359,33 +365,34 @@ export default class Selection {
 			// It can't be imported here, because it would lead to circular imports.
 			this._setRanges( selectable.getRanges(), selectable.isBackward );
 		} else if ( selectable instanceof Range ) {
-			this._setRanges( [ selectable ], !!optionsOrPlaceOrOffset && !!optionsOrPlaceOrOffset.backward );
+			this._setRanges( [ selectable ], !!placeOrOffset && !!placeOrOffset.backward );
 		} else if ( selectable instanceof Position ) {
 			this._setRanges( [ new Range( selectable ) ] );
 		} else if ( selectable instanceof Node ) {
 			const backward = !!options && !!options.backward;
 			let range;
 
-			if ( optionsOrPlaceOrOffset == 'in' ) {
+			if ( placeOrOffset == 'in' ) {
 				range = Range.createIn( selectable );
-			} else if ( optionsOrPlaceOrOffset == 'on' ) {
+			} else if ( placeOrOffset == 'on' ) {
 				range = Range.createOn( selectable );
-			} else if ( optionsOrPlaceOrOffset !== undefined ) {
-				range = Range.createCollapsedAt( selectable, optionsOrPlaceOrOffset );
+			} else if ( placeOrOffset !== undefined ) {
+				range = Range.createCollapsedAt( selectable, placeOrOffset );
 			} else {
 				/**
-				 * Required second parameter when setting selection to node.
+				 * selection.setTo requires the second parameter when the first parameter is a node.
 				 *
 				 * @error model-selection-setTo-required-second-parameter
 				 */
 				throw new CKEditorError(
-					'model-selection-setTo-required-second-parameter: Required second parameter when setting selection to node.' );
+					'model-selection-setTo-required-second-parameter: ' +
+					'selection.setTo requires the second parameter when the first parameter is a node.' );
 			}
 
 			this._setRanges( [ range ], backward );
 		} else if ( isIterable( selectable ) ) {
 			// We assume that the selectable is an iterable of ranges.
-			this._setRanges( selectable, optionsOrPlaceOrOffset && !!optionsOrPlaceOrOffset.backward );
+			this._setRanges( selectable, placeOrOffset && !!placeOrOffset.backward );
 		} else {
 			/**
 			 * Cannot set selection to given place.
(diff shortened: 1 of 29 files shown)


src/model/selection.js
```diff
@@ -322,36 +326,66 @@ export default class Selection {
...
 	 *
 	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection|
-	 * module:engine/model/position~Position|module:engine/model/element~Element|
+	 * module:engine/model/position~Position|module:engine/model/node~Node|
 	 * Iterable.<module:engine/model/range~Range>|module:engine/model/range~Range|null} selectable
-	 * @param {Boolean|Number|'before'|'end'|'after'} [backwardSelectionOrOffset]
+	 * @param {Object|Number|'before'|'end'|'after'|'on'|'in'} [optionsOrPlaceOrOffset]
+	 * @param {Boolean} [optionsOrPlaceOrOffset.backward]
+	 * @param {Object} [options]
+	 * @param {Boolean} [options.backward]
```
pjasiun: We should documentation simple:
```
* @param {Object|Number|'before'|'end'|'after'|'on'|'in'} [placeOrOffset]
* @param {Object} [options]
* @param {Boolean} [options.backward]
```

Since the second parameter is optional it is clear that the third parameter can the second parameter too. Even if it is not clear, there are examples in docs.

pjasiun: Also, you can add some descriptions to these parameters.

src/model/selection.js
```diff
@@ -322,36 +326,66 @@ export default class Selection {
...
+			} else if ( optionsOrPlaceOrOffset !== undefined ) {
+				range = Range.createCollapsedAt( selectable, optionsOrPlaceOrOffset );
+			} else {
+				/**
+				 * Required second parameter when setting selection to node.
+				 *
+				 * @error model-selection-setTo-required-second-parameter
+				 */
+				throw new CKEditorError(
+					'model-selection-setTo-required-second-parameter: Required second parameter when setting selection to node.' );
```
pjasiun: "selection.setTo require the second parameter when the first parameter is a node."

src/model/writer.js
```diff
@@ -925,24 +925,35 @@ export default class Writer {
 	 *		const position = new Position( root, path );
 	 *		writer.setSelection( position );
 	 *
-	 * 		// Sets collapsed range at the given offset in element.
+	 * 		// Sets range at the position of given node and offset.
 	 *		const paragraph = writer.createElement( 'paragraph' );
 	 *		writer.setSelection( paragraph, offset );
 	 *
+	 *		// Sets range inside the node.
+	 *		const paragraph = writer.createElement( 'paragraph', { backward } );
```
pjasiun: What `backward` property of an element do? ;)

ma2ciek: Hmm, that's a good question! 😄

pjasiun: Also, I think that we should skip this line. It is not correct to create a selection on a newly created element, which is not yet attached to the tree.

src/model/selection.js
ma2ciek: I've added them to the view, probably forget to add it here too. I'll fix this in model & view selections.

src/view/selection.js
```diff
@@ -55,15 +55,31 @@ export default class Selection {
...
 	 *		const selection = new Selection( position );
 	 *
-	 * 		// Creates selection at the start position of given element.
+	 * 		// Sets collapsed range at the position of given item and offset.
 	 *		const paragraph = writer.createElement( 'paragraph' );
-	 *		const selection = new Selection( paragraph, offset );
+	 *		selection.setTo( paragraph, offset );
+	 *
+	 *		// Sets range inside the item.
+	 *		const paragraph = writer.createElement( 'paragraph' );
```
pjasiun: There is no `createElement` method in the view writer.

src/view/selection.js
```diff
@@ -55,15 +55,31 @@ export default class Selection {
...
+	 *		selection.setTo( paragraph, 'in', { backward } );
+	 *
+	 *		// Sets range on the item.
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		selection.setTo( paragraph, 'on', { backward } );
 	 *
```
pjasiun: Fake selection samples are missing.

src/model/selection.js
ma2ciek: Should the second parameter allow 'Object'? As I understand, it should be `{Number|'before'|'end'|'after'|'on'|'in'} [placeOrOffset]` now.

pjasiun: Yep. The information that it is optional should be enough. If it is `Object` it means there is no `placeOrOffset` and the parameter is `options`. We can not do this in the code but we can do it in docs.

ma2ciek: Hmm, I've checked it and TS throws an error when two parameters are optional and only the second is passed as first (Only valid invocation is `fn( undefined, secondArgument )`. Maybe that's why I intuitively didn't like this idea. Hope we won't have problems with it as there was an idea to generate some TS declaration files in the future.

ma2ciek:
Ready for the review.

src/model/selection.js
```diff
@@ -55,16 +56,26 @@ export default class Selection {
...
 	 *		const selection = new Selection( paragraph, offset );
 	 *
 	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection|
 	 * module:engine/model/position~Position|module:engine/model/element~Element|
-	 * Iterable.<module:engine/model/range~Range>|module:engine/model/range~Range} [selectable]
-	 * @param {Boolean|Number|'before'|'end'|'after'} [backwardSelectionOrOffset]
+	 * Iterable.<module:engine/model/range~Range>|module:engine/model/range~Range|null} selectable
+	 * @param {Number|'before'|'end'|'after'|'on'|'in'} [placeOrOffset]
+	 * @param {Object} [options]
+	 * @param {Boolean} [options.backward]
```
pjasiun: Docs are missing.

src/model/selection.js
```diff
@@ -300,58 +311,85 @@ export default class Selection {
...
 	 *		const position = new Position( root, path );
 	 *		selection.setTo( position );
 	 *
-	 * 		// Sets range at the position of given element and optional offset.
-	 *		const paragraph = writer.createElement( 'paragraph' );
+	 * 		// Sets collapsed selection at the position of the given node and an offset.
 	 *		selection.setTo( paragraph, offset );
 	 *
-	 * 		// Removes all ranges.
+	 *		// Sets selection inside the given  node.
```
pjasiun: Double space.

pjasiun: Also, it would be nice to write what does it mean to create the selection in the element. See who it is described here:

https://github.com/ckeditor/ckeditor5-engine/blob/a0ad8fee1606f5a0f42e70c96d8f1ceca6c0f919/src/model/range.js#L730-L731

src/model/selection.js
```diff
@@ -300,58 +311,85 @@ export default class Selection {
...
+	 *		selection.setTo( paragraph, 'in', { backward } );
+	 *
+	 *		// Sets selection on the given node.
```
pjasiun: As above.

src/model/writer.js
```diff
@@ -902,47 +902,54 @@ export default class Writer {
...
 	 *		const position = new Position( root, path );
 	 *		writer.setSelection( position );
 	 *
-	 * 		// Sets collapsed range at the given offset in element.
-	 *		const paragraph = writer.createElement( 'paragraph' );
+	 * 		// Sets collapsed selection at the position of the given node and an offset.
 	 *		writer.setSelection( paragraph, offset );
 	 *
-	 * 		// Removes all ranges.
+	 *		// Sets selection inside the given node.
```
pjasiun: As above.

src/model/writer.js
```diff
@@ -902,47 +902,54 @@ export default class Writer {
...
+	 *		writer.setSelection( paragraph, 'in', { backward } );
+	 *
+	 *		// Sets selection on the given node.
```
pjasiun: As above.

src/view/selection.js
```diff
@@ -41,29 +41,41 @@ export default class Selection {
...
 	 *		// Creates selection at the given range.
 	 *		const range = new Range( start, end );
-	 *		const selection = new Selection( range, isBackwardSelection );
+	 *		const selection = new Selection( range, { backward, fake, label } );
```
pjasiun: I don't think we should create a fake selection in the all samples. Keep it simple and use `fake` and `label` only in the sample of the fake selection.

src/view/selection.js
```diff
@@ -41,29 +41,41 @@ export default class Selection {
...
 	 *
 	 *		// Creates selection at the given ranges
 	 * 		const ranges = [ new Range( start1, end2 ), new Range( star2, end2 ) ];
-	 *		const selection = new Selection( ranges, isBackwardSelection );
+	 *		const selection = new Selection( ranges, { backward, fake, label } );
```
pjasiun: As above.

src/view/selection.js
```diff
@@ -41,29 +41,41 @@ export default class Selection {
...
 	 *
 	 *		// Creates selection from the other selection.
 	 *		const otherSelection = new Selection();
 	 *		const selection = new Selection( otherSelection );
 	 *
 	 * 		// Creates selection at the given position.
 	 *		const position = new Position( root, path );
-	 *		const selection = new Selection( position );
+	 *		const selection = new Selection( position, { fake, label } );
```
pjasiun: As above.

src/view/selection.js
```diff
@@ -41,29 +41,41 @@ export default class Selection {
...
+	 *		selection.setTo( paragraph, offset, { fake, label } );
+	 *
+	 *		// Sets selection inside the item.
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		selection.setTo( paragraph, 'in', { backward, fake, label } );
+	 *
+	 *		// Sets selection on the item.
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		selection.setTo( paragraph, 'on', { backward, fake, label } );
 	 *
```
pjasiun: There should be a dedicated sample with the fake selection.

src/view/selection.js
```diff
@@ -41,29 +41,41 @@ export default class Selection {
...
 	 * 		// Creates selection at the given position.
 	 *		const position = new Position( root, path );
-	 *		const selection = new Selection( position );
+	 *		const selection = new Selection( position, { fake, label } );
 	 *
-	 * 		// Creates selection at the start position of given element.
+	 * 		// Sets collapsed selection at the position of given item and offset.
 	 *		const paragraph = writer.createElement( 'paragraph' );
-	 *		const selection = new Selection( paragraph, offset );
+	 *		selection.setTo( paragraph, offset, { fake, label } );
```
pjasiun: As above.

src/view/selection.js
```diff
@@ -41,29 +41,41 @@ export default class Selection {
...
+	 *
+	 *		// Sets selection inside the item.
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		selection.setTo( paragraph, 'in', { backward, fake, label } );
```
pjasiun: As above.

src/view/selection.js
```diff
@@ -41,29 +41,41 @@ export default class Selection {
...
+	 *
+	 *		// Sets selection on the item.
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		selection.setTo( paragraph, 'on', { backward, fake, label } );
```
pjasiun: As above.

src/view/selection.js
```diff
@@ -388,70 +398,97 @@ export default class Selection {
...
-	 *		// Sets ranges from the given range.
+	 * This method provides option to create a fake selection.
+	 * Fake selection does not render as browser native selection over selected elements and is hidden to the user.
+	 * This way, no native selection UI artifacts are displayed to the user and selection over elements can be
+	 * represented in other way, for example by applying proper CSS class.
+	 *
+	 *		// Sets selection to the given range.
 	 *		const range = new Range( start, end );
-	 *		selection.setTo( range, isBackwardSelection );
+	 *		selection.setTo( range, { backward, fake, label } );
```
pjasiun: As in the constructor.

src/view/writer.js
```diff
@@ -36,35 +36,54 @@ export default class Writer {
...
+	 *
+	 * Additionally fake's selection label can be provided. It will be used to describe fake selection in DOM (and be
+	 * properly handled by screen readers).
+	 *
+	 * Usage:
+	 *
+	 *		// Sets selection to the given range.
 	 *		const range = new Range( start, end );
-	 *		writer.setSelection( range, isBackwardSelection );
+	 *		writer.setSelection( range, { backward, fake, label } );
```
pjasiun: As in `view.selection` class.

ma2ciek:
I fixed mentioned issues.

src/model/selection.js
```diff
@@ -55,16 +56,34 @@ export default class Selection {
...
+	 * Creates a range inside an {@link module:engine/model/element~Element element} which starts before the first child of
+ 	 * that element and ends after the last child of that element.
+	 *
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		const selection = new Selection( paragraph, 'in' );
+	 *
+	 * Creates a range on an {@link module:engine/model/item~Item item} which starts before the item and ends just after the item.
+	 *
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		const selection = new Selection( paragraph, 'on' );
```
pjasiun: Let's remove `const paragraph = writer.createElement( 'paragraph' );` as we do it in other cases.

src/view/selection.js
```diff
@@ -55,15 +55,45 @@ export default class Selection {
...
 	 *
+	 * Creates a range inside an {@link module:engine/view/element~Element element} which starts before the first child of
+ 	 * that element and ends after the last child of that element.
+	 *
+	 *		const paragraph = writer.createElement( 'paragraph' );
+	 *		const selection = new Selection( paragraph, 'in' );
+	 *
+	 * Creates a range on an {@link module:engine/view/item~Item item} which starts before the item and ends just after the item.
+	 *
+	 *		const paragraph = writer.createElement( 'paragraph' );
```
pjasiun: Remove `const paragraph = writer.createElement( 'paragraph' );`.

src/view/selection.js
```diff
@@ -55,15 +55,45 @@ export default class Selection {
...
 	 *
-	 * 		// Creates selection at the start position of given element.
+	 * 		// Creates collapsed selection at the position of given item and offset.
 	 *		const paragraph = writer.createElement( 'paragraph' );
 	 *		const selection = new Selection( paragraph, offset );
 	 *
+	 * Creates a range inside an {@link module:engine/view/element~Element element} which starts before the first child of
+ 	 * that element and ends after the last child of that element.
+	 *
+	 *		const paragraph = writer.createElement( 'paragraph' );
```
pjasiun: Remove `const paragraph = writer.createElement( 'paragraph' );`.

PR #1328, opened 2018-02-27
Changed public API to protected

### LINKED ISSUE #738: Consider making the insert/append/remove view/model API protected
Currently I can't think of a case in which you should use methods like `Node.remove()` or `Element.appendChildren()`. In all cases which I know you either use deltas (when operating on the model) or model/view writer (when converting).

So, we can make this API protected. In other words, view and model nodes should be immutable, unless you're using the writer or deltas.

Also, we should consider simplifying the writer (at least the model one), because it requires working with ranges which is a pain. I'm constantly changing:

```js
el.appendChildren( foo );
```

into:

```js
modelWriter.insert( ModelRange.createOn( el ), foo );
```

Related issue: https://github.com/ckeditor/ckeditor5-engine/issues/737.

pomek:

Other: Methods which modify the model's and view's tree are now protected and shouldn't be used directly in the code. Instead of an instance of `Writer` should be used. Closes #738.

---

Other repositories which should be merged along with this PR:

* https://github.com/ckeditor/ckeditor5-undo/pull/81
* https://github.com/ckeditor/ckeditor5-list/pull/98
* https://github.com/ckeditor/ckeditor5-clipboard/pull/44
* https://github.com/ckeditor/ckeditor5-image/pull/182
* https://github.com/ckeditor/ckeditor5-typing/pull/141

### DIFF
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -66,6 +66,7 @@ export default class DataController {
 		 * cleared directly after the data are converted. However, the mapper is defined as a class property, because
 		 * it needs to be passed to the `DowncastDispatcher` as a conversion API.
 		 *
+		 * @readonly
 		 * @member {module:engine/conversion/mapper~Mapper}
 		 */
 		this.mapper = new Mapper();
--- a/src/conversion/conversion.js
+++ b/src/conversion/conversion.js
@@ -29,6 +29,10 @@ export default class Conversion {
 	 * Creates new Conversion instance.
 	 */
 	constructor() {
+		/**
+		 * @private
+		 * @member {Map}
+		 */
 		this._dispatchersGroups = new Map();
 	}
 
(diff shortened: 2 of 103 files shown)


pomek:
Changed `mgit.json` in the main repo (for CI purposes) – https://github.com/ckeditor/ckeditor5/compare/ci/ckeditor5-engine/738?expand=1.

Travis: https://travis-ci.org/ckeditor/ckeditor5/builds/346812638

pomek:
9 tests fail because of the issue with Differ – https://github.com/ckeditor/ckeditor5-engine/issues/1326.

pjasiun:
It would be great to have `@see` and a link to the proper writer method next to each protected methods.

pjasiun:
`text.data` should be readonly.

pjasiun:
All `element.clone`, `node.clone`, `text.clone` should be protected now too.

pjasiun:
Nodelist constructor should be protected.

pjasiun:
Please check all properties and add `@readonly` wherever it is needed.

I found these properties with missing`@readonly` flag:
 - `model.document`
 - `model.schema`
 - `element.name`

But there might be more.

pomek:
@pjasiun and @szymonkups, fixes that you proposed and missing `@readonly` tag has been introduced. Could you guys review it once again?

I will fix the rest packages.

PR #1330, opened 2018-02-28
Introduced composition observer

### LINKED ISSUE #1329: Introduce composition observer
While working on improving IME handling, there is a need to react on composition state changes. The composition observer should allow listening to composition events (`compositionstart`, `compositionupdate`, `compositionstop`) and also set proper document state marking that composition is taking place at the moment.

f1ames:

Feature: Introduced composition observer. Closes #1329.

---

### Additional information

It mostly reuses the code from #861, but seems to be enough to cover this feature.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -104,6 +104,13 @@ export default class Renderer {
 		this.isFocused = false;
 
 		/**
+		 * Indicates if composition takes places inside view document.
+		 *
+		 * @member {Boolean}
+		 */
+		this.isComposing = false;
+
+		/**
 		 * DOM element containing fake selection.
 		 *
 		 * @private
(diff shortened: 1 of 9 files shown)


src/view/renderer.js
```diff
@@ -103,6 +103,13 @@ export default class Renderer {
 		 */
 		this.isFocused = false;
 
+		/**
+		 * Indicates if composition takes places inside view document.
+		 *
+		 * @member {Boolean}
+		 */
+		this.isComposing = false;
```
szymonkups: Why this is needed here? It is not used anywhere in the renderer.

f1ames: Not needed, it was a leftover (well, it probably will be used in the related task, but definitely should not be added in this PR) - removed.

PR #1343, opened 2018-03-07
Update model selection attributes after each change

### LINKED ISSUE #1267: Selection should not retain attributes after parent block removal by an external change
1. Go to http://localhost:8125/ckeditor5-link/tests/manual/tickets/113/1.html.
2. Take the 2nd scenario.
3. Write some text.

**Expected:**

The selection should not retain previous (`linkHref`) attribute. The whole editing context has changed and most likely users will not expect that.

**Actual:**

Still writing in the same link despite that it's parent `paragraph` has been removed from the model.

kapture 2018-02-01 at 14 55 05

Spotted in https://github.com/ckeditor/ckeditor5-link/pull/165.

**Note**: It's not about the link. Other attributes like bold, etc. are also retained.

szymonkups:

Fix: `model.DocumentSelection` should update it's attributes after each change, including external changes. Closes #1267.

---
Close together with: https://github.com/ckeditor/ckeditor5-undo/pull/83 and https://github.com/ckeditor/ckeditor5-autoformat/pull/52.

### DIFF
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -19,10 +19,6 @@ import toMap from '@ckeditor/ckeditor5-utils/src/tomap';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 import log from '@ckeditor/ckeditor5-utils/src/log';
 
-const attrOpTypes = new Set(
-	[ 'addAttribute', 'removeAttribute', 'changeAttribute', 'addRootAttribute', 'removeRootAttribute', 'changeRootAttribute' ]
-);
-
 const storePrefix = 'selection:';
 
 /**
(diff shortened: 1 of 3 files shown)


src/model/documentselection.js
```diff
@@ -1019,24 +999,30 @@ function getAttrsIfCharacter( node ) {
...
-	let changeParent = null;
+//
+// @private
+// @param {module:engine/model/model~Model} model
+// @param {module:engine/model/batch~Batch} batch
+function clearAttributesStoredInElement( model, batch ) {
+	const differ = model.document.differ;
+
+	for ( const entry of differ.getChanges() ) {
+		if ( entry.type != 'insert' || !entry.position.parent ) {
```
scofalik: `!entry.position.parent` - is there any situation when this happens? 🤔

PR #1347, opened 2018-03-08
T/1226 Refactored how markers removal is converted from the model to the view

### LINKED ISSUE #1226: Markers removal
Markers removal is now implemented similarly to the attribute removal: based on marker name/descriptor converters decide what kind of elements should be removed and unwraps similar elements.

The problem is that markers do not work like attributes. For attributes it is not important which instance of an element should be removed: if `<b>` suppose to be removed from `foo` it does not matter which `<b>` is removed as long as `foo` is unbolded. If a comment or users selection should be removed it is very important to remove the proper element.

Also, markers values change based on data stored outside the document model much more often than attributes do. In fact, this is what markers were created for: to show some external data on the document (like comments or users selection). The problem is that it means that markers descriptors can change asynchronously. For instance, if a comment is selected marker gets a different color. Because of it, the marker to remove is not the same as the marker which was added. It leads to many bugs and hacks.

The third problem is the current way of conversation, introduced together with differ. Because now conversion is delayed it is very hard to find the range of the makers in the view, when model changed.

This is why I think that markers should be removed in a different way. When elements are added to the view they should be stored and exactly these elements should be removed. The only information sent to the converters should be the name of the marker to be removed because the descriptor and range can be invalid at the moment of removal.

After this short introduction, let's move to details.

There are 2 mechanisms to be changed to implement it.

First is view writer. View writer is designed to work with attributes. It splits attribute elements with no information about their connection. The only way to unwrap is by giving range and element template - all similar elements are removed, what is fine for attributes but not for markers. Whenever view writer split view element it should store information about this elements parts. Then it should be possible to unwrap certain element (and all its parts) using a reference to the element instance, not by using range and template. The proper set of view writer methods should be provided.

Second, is the way how converters work. Markers converters should store information what element has been added to the view and what element should be removed. Then it should be enough to say which marker should be removed, based on its name. Range and descriptor should not be needed when a marker is removed.

scofalik:

Other: Refactored how markers removal is converted from the model to the view. Closes #1226.


scofalik:
Related PR https://github.com/ckeditor/ckeditor5-link/pull/178

src/view/writer.js
```diff
@@ -891,6 +924,25 @@ export default class Writer {
...
+	/**
+	 * For given {@link module:engine/view/attributeelement~AttributeElement attribute element}, returns all other
+	 * attribute elements with the same {@link module:engine/view/attributeelement~AttributeElement#id id} that
+	 * were inserted by `Writer` to the view tree.
+	 *
+	 * @param {module:engine/view/attributeelement~AttributeElement} element
+	 * @returns {Set.<module:engine/view/attributeelement~AttributeElement>|null} Set containing all the attribute elements
+	 * with the same `id` or `null` if given element had no `id`.
+	 */
+	getAllClonedElements( element ) {
```
pjasiun: The getter should not be a writer method. It should be a method of an element. You should not require creating `view.change` block to call getter, because this block causes rendering.

src/conversion/mapper.js
```diff
@@ -239,6 +276,35 @@ export default class Mapper {
...
+
+		if ( !boundElements ) {
+			return null;
+		}
+
+		const elements = new Set();
+
+		for ( const element of boundElements ) {
+			if ( element.is( 'attributeElement' ) ) {
+				for ( const clone of element.getCustomProperty( 'clonedElements' ) ) {
```
pjasiun: Shouldn't `getAllClonedElements` be used here?

src/view/writer.js
```diff
@@ -1265,6 +1328,201 @@ export default class Writer {
...
+
+		let group = this._cloneGroups.get( id );
+
+		if ( !group ) {
+			group = new Set();
+			this._cloneGroups.set( id, group );
+		}
+
+		group.add( element );
+		this.setCustomProperty( 'clonedElements', group, element );
```
pjasiun: I would use a protected property instread of a custom property.

pjasiun: Only make sure it is cloned well ;)

src/view/attributeelement.js
```diff
@@ -53,15 +60,26 @@ export default class AttributeElement extends Element {
...
-	 * @return {Number}
+	 * @returns {Number}
 	 */
 	get priority() {
 		return this._priority;
 	}
 
+	/**
+	 * Element identifier. If set, it is used by {@link module:engine/view/element~Element#isSimilar},
+	 * and then two elements are considered similar if, and only if they have the same `_id`.
```
pjasiun: `id`

src/view/attributeelement.js
```diff
@@ -34,34 +35,91 @@ export default class AttributeElement extends Element {
...
+	/**
+	 * Returns all {@link module:engine/view/attributeelement~AttributeElement attribute elements} that has the
+	 * same {@link module:engine/view/attributeelement~AttributeElement#id id} and are in the view tree (were not removed).
+	 *
+	 * Note: If this element has been removed from the tree, returned set will not include it.
+	 *
+	 * Throws {@link module:utils/ckeditorerror~CKEditorError attribute-element-get-elements-with-same-id-no-id}
+	 * if this element has no `id`.
+	 *
+	 * @returns {Set.<module:engine/view/attributeelement~AttributeElement>|null} Set containing all the attribute elements
```
pjasiun: It will never return `null`.

Issue #1342, opened 2018-03-06
Rendering improvements


Reinmar:
## Avoid unnecessary structure updates

Background: #1334.

How does selection conversion work?

1. MODEL: `<p><$text bold=true>xx</text></p>`  + `selection[bold=true]`
2. VIEW:
	1. `<p><b>xx</b></p>`
	2. `<p>[]<b>xx</b></p>` (after position conversion)
	3. `<p><B>[]</B><b>xx</b></p>` (during `writer.wrap( selection )` )
	4. `<p><B>[]xx</B></p>` (after attribute conversion; we're merging to the left, so a new element `<B>` is left)
3. RENDERING:
	* actual DOM: `<p><b>xx</b></p>`
	* expected DOM `<p><B>XX</B></p>` (we need to replace the entire `<b>` element; IDEA: don't replace `<b>` with `<B>` because they are similar!)

BUT... WHAT IF SOMETHING ELSE HAS CHANGED?

* actual DOM: `<p><b>xx</b></p>`
* expected DOM `<p><B>XXY</B></p>` (don't replace `<b>` with `<B>` but recursively go deeper to apply other possible changes, see #1125)

Possible solution: 

1. Make `diff( actual, expected, sameNode )` skip elements which have the same names and attributes (see `view.Element#isSimilar()`).
2. As a result, we won't replace `<b>` with `<B>`, which means that the view to DOM mapping will be outdated – the new view `<B>` element will be mapped to the new DOM `<B>` element which is not in the tree (because the old DOM `<b>` element was left there). Which means that we should update bindings after skipping such replacements. How to do that? Perhaps we can simply and blindly set new mappings after rendering: purge existing bindngs (in a limited scope), `updatedDomElements.forEach( mapToUpdatedViewElement )`.

--

## Deep structure updates

Background: https://github.com/ckeditor/ckeditor5-engine/issues/1125

1. `<p><a>Text<b>X[]</b></a></p>`
2. `<p><a>Text<b>X</b></a><b>Y[]</b></p>` (after typing "Y"; a lot mutations)
3. `<p><a>Text<b>XY</b></a></p>` (the model and the view after inserting "Y")

* ACTUAL DOM:		`<p><a>Text<b>X</b></a><b>Y[]</b></p>`
* EXPECTED DOM: 	`<p><a>Text<b>XY</b></a></p>`

---

## Fillers

[...]

PR #1355, opened 2018-03-13
Disable inline filling character removal during composition

### LINKED ISSUE #898: Disable inline filling character removal during composition
Extracted from https://github.com/ckeditor/ckeditor5-engine/pull/896#issuecomment-290169270.

> I feel that the next thing we should do is work on disabling filling char and element removal when composition takes place. Since things are transparent to the view (they are not visible there) I think that they could be kept longer if needed without much of a trouble. Am I right @pjasiun?
>
> Also, I remember that the filling char is removed on keydown so it doesn't affect caret movement. This should also be prevented during composition.

### LINKED ISSUE #1409: Inline filler (ZWS) is removed while typing inside empty inline element
I have completely changed this issue description, see why in https://github.com/ckeditor/ckeditor5-engine/issues/1409#issuecomment-391973289 below.

1. Go to https://ckeditor5.github.io/.
1. Replace whole content with `Foo Bar` and place selection on the end.
1. Turn on `Bold`.
    * Now inline filler is inserted.
1. Type something.
    * Inline filler is removed.

## Expected

Inline filler is not removed until selection is moved from text node.

## Actual

Inline filler is removed while typing inside empty inline element.

There are cases when it behaves correctly. Follow the below steps (also shown in below gif) to see how it should behave:

1. Turn `Bold` off.
    * Now inline filler is inserted.
1. Type something.
    * Inline filler is not removed.

filler

f1ames:

Fix: Prevent inline filler removal during composition. Closes #898. Closes #1409.

---

### Additional information

There is still one issue (#1349) which is not resolved by this fix due to native Chrome bug.

This PR adjusted the way how `inlineFiller` is handled in non-empty text nodes. Previously there was an assumption that in nodes like `FILLERtext{}`, the filler should not be touched. But there were no real cases producing such situations.
Blocking inline filler removal during composition is now such case (and the only we know ATM) so it seems reasonable to just adjust this behaviour, so the filler gets removed also in such situations.


Reinmar:
@Mgsy could you test whether this PR doesn't introduce some regressions? Feel free to talk with @f1ames to learn what might got changed.

f1ames:
Just to describe in more details changes in this PR so it may be helpful during testing / looking for regressions.

There are two main changes in the `renderer` here:

### Preventing inline filler removal during composition

Every time composition is started inside element which has inline filler (so basically empty inline element), inline filler will not be removed until composition is finished. Before this change the inline filler was simply removed every time first character was inserted inside such elements. Now if character is inserted by composition, the filler will not be removed. That means the modified behaviour is only observable during composition.

The one problem I could think of is that renderer is still not perfect when it comes to handling composition and it tent to break it in some cases (especially when composing inside empty elements). Sometimes in such cases the `compositionend` event will not be fired and inline filler may not be removed properly (but it will be removed later when composition is properly ended).
This is more general problem and it needs other improvements (so renderer will not break composition).

### Adjusting the general behaviour when filler is removed

It is already mentioned in the PR description:

> This PR adjusted the way how inlineFiller is handled in non-empty text nodes. Previously there was an assumption that in nodes like FILLERtext{}, the filler should not be touched. But there were no real cases producing such situations.

When renderer encounters already rendered text node with content like `FILLERtext{}`, it automatically removes (in general, not only during composition) inline filler. This was basically needed for properly handling filler removal after composition.

The previous behaviour [assumed that filler in such cases should not be touched](https://github.com/ckeditor/ckeditor5-engine/blob/686625658d0267850a5d7ac175dbb70854786813/src/view/renderer.js#L303) because it was left there on purpose. However, such situations will not have place as it is removed right after something is inserted in the empty node with filler. So if there is somewhere any case like above in which inline filler should not be touched in a content like `FILLERtext{}` (apart from composition ofc), the regression could occur there.

Mgsy:
I've noticed one regression. Putting the selection in a link inserts the inline filler. The caret becomes invisible, the link is split and after typing some text, more `<a>` elements appear.

bug_cke5

Mgsy:
## Steps to reproduce
1. Open [the article sample](http://localhost:8125/ckeditor5-core/tests/manual/article.html).
1. Put the caret in `Paragraph`.
1. Start the composition.
1. Press <kbd>Shift</kbd> + <kbd>Arrow up</kbd>.

## Current result
The inline filler has been inserted to the `Heading 1`. It doesn't disappear after finishing the composition.

## GIF
[Click](https://user-images.githubusercontent.com/20988892/37895166-cd2782e2-30e0-11e8-8e59-e3c22c592705.gif).

src/view/renderer.js
```diff
@@ -104,6 +104,13 @@ export default class Renderer {
 		 */
 		this.isFocused = false;
 
+		/**
+		 * Indicates if composition takes places inside view document.
```
Reinmar: "Indicates whether text composition takes place in the document."

(you can also update `isFocused` description)

src/view/renderer.js
```diff
@@ -182,7 +189,8 @@ export default class Renderer {
 		// There was inline filler rendered in the DOM but it's not
 		// at the selection position any more, so we can remove it
 		// (cause even if it's needed, it must be placed in another location).
-		if ( this._inlineFiller && !this._isSelectionInInlineFiller() ) {
+		// Filler should not be touched during composition to not break it.
+		if ( this._inlineFiller && !this.isComposing && !this._isSelectionInInlineFillerOnlyNode() ) {
```
Reinmar: Could you explain the change from `isSelectionInInlineFiller()` to `isSelectionInInlineFillerOnlyNode()`?

tests/view/renderer.js
```diff
@@ -646,6 +646,80 @@ describe( 'Renderer', () => {
 			expect( domSelection.getRangeAt( 0 ).collapsed ).to.be.true;
 		} );
 
+		it( 'should add and remove inline filler in case <p>foo<b>[]bar</b></p>', () => {
```
Reinmar: I feel that these two tests are too long. Can't they be split into more concrete parts?

Reinmar: E.g. there's just one new test for the whole change with not removing the filler during composition while there are at least two things to check – whether the filler is not touched during composition and that it's removed when it can be removed.

I mean – I know there are some pretty long tests in this file already, but your tests are twice as long :D

Reinmar:
After some bold/italic buttons clicking, typing and moving the caret around (by clicking) I got two fillers:

image

And once there are two+ fillers, only one of them is skipped on left/right arrow keys. The other require 7 key presses.

Reinmar:
Another interesting observation – sometimes, selecting text from right to left causes adding the filler to every text node in that selection :D But I can't repeat it in a stable manner.

Also, to be clear, multiple fillers is not something we certainly need to avoid. However, I wonder whether we shouldn't remove the previous filler if the selection needs a new one. So, to never allow to have two fillers at the same time. WDYT, @f1ames? Could we do that?

Reinmar:
BTW, should I notice any composition case getting fixed? Or are all of them still broken because we're re-rendering too big part of the tree?

f1ames:
> BTW, should I notice any composition case getting fixed? Or are all of them still broken because we're re-rendering too big part of the tree?

The second one, so there are not visible improvements due to how rerendering works now.

f1ames:
> I've noticed one regression. Putting the selection in a link inserts the inline filler. The caret becomes invisible, the link is divided and after typing some text, more <a> elements appear.

@Mgsy for some reason I am not able to reproduce this issue. Are there any specific steps needed?

---

What I noticed looking at the provided gif is that in you case "link selection span" is inside `a` element. 

image

When I run similar test locally I see that `span` is outside `a` 🤔 

image

---

> Putting the selection in a link inserts the inline filler. 

Also I don't see inline filler on the provided gif, only span marking link selection. Inline filler is shown in a browser dev tools like

image


Checked on `http://localhost:8125/ckeditor5-core/tests/manual/article.html` with Chrome (Version 65.0.3325.162).

f1ames:
I was able to reproduce the 2nd issue (only on windows as on macOS `shift + up` behaves a little different). The main issue here is that `shift + up` doesn't cause `compositionend`. I will look for a workaround for this.

f1ames:
> So, to never allow to have two fillers at the same time. WDYT, @f1ames? Could we do that?

I think that would be a valid approach.

src/view/renderer.js
f1ames: Described it in more detail in comment above - https://github.com/ckeditor/ckeditor5-engine/pull/1355#issuecomment-375609711 in **Adjusting the general behaviour when filler is removed** section.

tests/view/renderer.js
f1ames: The test is super long, I think I tried to omit some initialization code repetition and merged two checks into one test... I will split it and introduce more test to cover other cases if needed.

Mgsy:
>@Mgsy for some reason I am not able to reproduce this issue. Are there any specific steps needed?

Fortunately, after pulling latest changes I can't reproduce it too 🎉

f1ames:
I redesigned this solution a bit, so now it handles more cases. In the ideal world, handling inline filler should work like:
1. Inline filler is inserted in the empty inline element (currently selected).
    * When something is typed, inline filler is removed.
    * Or when composition starts, inline filler is not removed. After composition ends (by changing selection or inserting composed characters), filler is removed.

However, there are other cases:
* Extending selection during composition (mentioned in the https://github.com/ckeditor/ckeditor5-engine/pull/1355#issuecomment-376087328 above).
* Composition is not properly ended and then moving selection may cause strange inline filler behaviour.

---

The proposed solution handles all above situations:
* For regular typing, inline filler is removed as it was before.
* During composition in empty inline element, filler is not removed. It is removed after composition ends.
* If selection is extended during composition, inline filler would not be touched (because you can go back to composing). If selection would be moved (thus ending composition) filler is removed too.
* If there is a case that for some reason (e.g. broken composition), selection is moved to a different node during composition (different from the one in which composition started) without firing `compositionend`, the filler will also be removed.

The only issue here is that if composition is not ended properly (last point above), regular typing in empty inline element will not remove filler immediately (because `renderer` still thinks composition takes place). The filler will be removed after selection changes or composition ends. So it is still guaranteed that there will be only one inline filler and that it is removed right after selection changes or composition ends. This is an edge case (now it is caused by rerendering element nodes in which composition takes place) and such handling should be sufficient so nothing gets broken and inline fillers are still properly handled.

Mgsy:
I've found another case.

## Steps to reproduce
1. Open the article sample.
1. Put the caret at the end of the `Paragraph`.
1. Press <kbd>Ctrl</kbd> + <kbd>B</kbd>.
1. Start the composition and type something.
1. Press <kbd>Ctrl</kbd> + <kbd>B</kbd>.
1. Press <kbd>Arrow up</kbd>.

## Current result
- After step 5, the inliner filler is moved outside of the `<strong>` element
- After step 6, the inline filler appears at the beginning of the block

## GIF
bug_letters

f1ames:
@Mgsy I see it's Windows only (e.g. on macOS you cannot change styling during composition). As for step **6.**, it behaves as described but with <kbd>Shift</kbd> + <kbd>Arrow up</kbd> for me.

Step **5.** is correct and behaves the same without composition. While you are on end of paragraph and after `strong` element, filler is needed there:


The thing is, we assumed that we shouldn't remove inline filler during composition and try to not rerender any part of the text node in which composition takes place. Here when switching off bold we need to render it which breaks composition. I think it will be good to report as separate issue (that you can change styling during composition which breaks it). Still, I will take a look on reported issue with fillers management if it can be improved.

Mgsy:
>@Mgsy I see it's Windows only (e.g. on macOS you cannot change styling during composition). As for step 6., it behaves as described but with Shift + Arrow up for me.

You're right, it can't be reproduced on macOS. And of course in the last step I meant <kbd>Shift</kbd> + <kbd>Arrow up</kbd> :) I'll edit the scenario.

f1ames:
I improved `inline filler` handling during composition for non-collapsed selection (fixing https://github.com/ckeditor/ckeditor5-engine/pull/1355#issuecomment-381516509).

The complexity of this PR (the whole solution basically) is mainly caused by some composition edge cases (as mentioned in https://github.com/ckeditor/ckeditor5-engine/pull/1355#issuecomment-380430752). So apart from simple cases (composition starts and ends in the same element), it handles extending selection during composition (cases mentioned by @Mgsy) and also deals with not properly ended composition (when `compositionend` is not properly fired).

Broken composition may occur when element in which composition takes place is rerendered, but also during collaborative editing. So even if we are planning to fix as many cases as possible it still may happen that composition will break at some point and renderer should be able to handle such situations without spoiling editing experience.

f1ames:
Ready for re-review, cc @Mgsy @Reinmar.

@Mgsy when testing <kbd>Shift</kbd> + <kbd>Arrow up</kbd> please keep #1416 in mind.

src/view/renderer.js
```diff
@@ -180,23 +187,43 @@ export default class Renderer {
...
 		if ( this._inlineFiller ) {
-			inlineFillerPosition = this._getInlineFillerPosition();
+			if ( !this.isComposing && !this._isSelectionInInlineFiller( false ) ) {
+				// There was inline filler rendered in the DOM but it's not at the selection position any more, so we can remove
+				// it if not during composition (cause even if it's needed, it must be placed in another location).
+				this._removeInlineFiller();
+			} else if ( this.isComposing ) {
+				// When selection has 0 ranges, `isCollapsed` returns false, but here
+				// we are only interested in non-collapsed selection (so with at least 1 range).
+				if ( !this.selection.isCollapsed && this.selection.rangeCount > 0 ) {
```
Reinmar: In such cases, to make the code more readable is good to split the code into two steps:

```js
// When selection has 0 ranges, `isCollapsed` returns false, but here
// we are only interested in non-collapsed selection (so with at least 1 range).
const isNotCollapsed = !this.selection.isCollapsed && this.selection.rangeCount > 0;

// Explain what we do if it's not collapsed and someone's composing.
if ( isNotCollapsed ) {
```

src/view/renderer.js
```diff
@@ -180,23 +187,43 @@ export default class Renderer {
...
+					// During composition selection may be extended and its start/end moved to a different text
+					// node (using 'shift + up' in Chrome on Windows during composition). In such situations
+					// filler should not be moved or deleted (because it is possible to continue composing).
+					inlineFillerPosition = this._getExistingInlineFillerPosition();
+				} else if ( !this._isValidCompositionSelection() ) {
```
Reinmar: This function is used only in this place so it could be made more contextual. Like `_checkCompositionWasNotEndedCorrectly()`.

Reinmar:
We've just talked with @f1ames that this PR doesn't actually change the behaviour much when typing (the filler is maintained if the mapping worked) because of this:

https://github.com/ckeditor/ckeditor5-engine/blob/868d79b260f627987ab48b3ae6a212dcea585e0d/src/view/renderer.js#L321-L325

It's even reported in https://github.com/ckeditor/ckeditor5-engine/issues/1409.

This PR will make filler stay longer if the composition takes place. But if the selection left the text node in which the filler was, this PR will also make sure that the filler is removed (because we might've encountered an incorrectly terminated composition). So, to sum up, the only thing which this PR changes is the fact that the filler will be kept if the composition takes place when the selection is non-empty (on master it's enough that the selection is not empty for the filler to be removed).

Therefore, we need to reconsider whether merging this PR makes sense. Perhaps it's enough to keep the current behaviour (keeping the filler as long as the selection stays in the same text node) + extending this behaviour to non empty selection + extending `jumpOverInlineFiller()` for these scenarios + fixing mapping early so `isInlineFillerInSelection()` can rely on that.

f1ames:
See also https://github.com/ckeditor/ckeditor5-engine/issues/1342#issuecomment-391422791 which extends the above comment.

Reinmar:
It turned out that this change isn't needed because the renderer was designed to not remove the inline filler while the selection is in the same text node anyway. That was buggy but since we fixed it in https://github.com/ckeditor/ckeditor5-engine/pull/1424 it works fine. There's not need for any additional blocking. At least we think so now...

Reinmar:
Please **do not remove the branch**.

PR #1357, opened 2018-03-14
View selection split into Selection and DocumentSelection

### LINKED ISSUE #1304: Split view selection into Selection & DocumentSelection
Follow-up to #1210.

Some observers, like selection or mutation observers, create an instance of `Selection` and put it in the event. Now `Selection` is `read-only` so it can not be created nor modified without using a private API.

We should solve it the same way we do it the model: rename current `Selection` to `DocumentSelection` and introduce another `Selection` class, with public write API.

szymonkups:

Other: View selection is now split onto Selection and DocumentSelection. Closes #1304 .

BREAKING CHANGE: Introduced view.DocumentSelection. It has protected API and can be modified only by view writer. Observers creating instance of selection (like SelectionObserver, MutationObserver) use view.Selection class now. 

---

### Additional information

Please close together with:
 * https://github.com/ckeditor/ckeditor5-typing/pull/144
 * https://github.com/ckeditor/ckeditor5-image/pull/185


src/view/documentselection.js
```diff
@@ -0,0 +1,382 @@
...
+	 *		// Creates selection from the document selection.
+	 *		const documentSelection = new DocumentSelection();
+	 *		const selection = new DocumentSelection( documentSelection );
+	 *
+	 * 		// Creates selection at the given position.
+	 *		const position = new Position( root, path );
+	 *		const selection = new DocumentSelection( position );
+	 *
+	 *		// Creates collapsed selection at the position of given item and offset.
+	 *		const paragraph = writer.createElement( 'paragraph' );
```
ma2ciek: There's no `ViewWriter#createElement()` method - it was my mistake in the Selection class.

src/view/selection.js
```diff
@@ -16,21 +16,23 @@ import Node from './node';
 import Element from './element';
 import count from '@ckeditor/ckeditor5-utils/src/count';
 import isIterable from '@ckeditor/ckeditor5-utils/src/isiterable';
+import DocumentSelection from './documentselection';
```
ma2ciek: I remember, that in the model I had some problem with it. The two-way importing with extending other class AFAIR. (In the model there is also a `LiveSelection` which inherits from the `Selection`).

ma2ciek: I've added there something like duck typing to prevent this. But If there's no problem with the current implementation, it's not worth to change it.

src/view/selection.js
```diff
@@ -51,12 +53,16 @@ export default class Selection {
 	 *		const otherSelection = new Selection();
 	 *		const selection = new Selection( otherSelection );
 	 *
+	 *		// Creates selection from the document selection.
+	 *		const documentSelection = new DocumentSelection();
```
pjasiun: I would remove this line since one never will create a new DocumentSelection.

PR #1366, opened 2018-03-16
Introduced deep children text nodes rerendering

### LINKED ISSUE #1093: Renderer does not re-render spaces in text nodes correctly
There are (at least) two cases when spaces are not correctly from view text nodes to dom text nodes after a change.

Those errors are caused by the fact, that during re-render only the changed dom text node is updated. https://github.com/ckeditor/ckeditor5-engine/pull/1083 changes the way spaces are converter to `&nbsp;`. Now, changing to `&nbsp;` bases on the data of the previous text node. This means that whenever a text node is changed, next touching text node has to be re-rendered, etc.

Mentioned cases:

Case 1:

1. DOM is `<p>x<strong> y</strong></p>`.
2. Type space after `x`.
3. DOM becomes `<p>x <strong> y</strong></p>`.
4. Only one space is rendered.

Case 2:

1. DOM is `<p><strong>&nbsp;y</strong></p>`.
2. Put caret before `&nbsp;` and turn off bold.
3. Type `x`.
4. DOM becomes `<p>x<strong>&nbsp;y</strong></p>`.
5. Incorrect `&nbsp;` stays in DOM.

We know that there is a "bug" in `DomConverter.viewChildrenToDom`. View elements mapped to DOM elements are not re-rendered. Those elements might have text nodes that need refreshing and this is the moment when the bug occurs. Unfortunately if we want to change `viewChildrenToDom` behavior, we will do a bit more processing. Hopefuly, this won't have negative impact on editor.

PS.
@Reinmar told me to hide those tests in case we need them: https://gist.github.com/scofalik/7d33adb45b82722bf5f6ca59ae8439d8

### LINKED ISSUE #1125: Rendering deeper changes fails to update text nodes
Example:

1. Starting view and DOM: `<p><a>Text<b>X</b></a></p>`.
2. Let's say that you type something at the end of this content which makes browser to create a lot of mutations and inserting the text in the wrong place (at the end of `<p>`) instead of at the end of `<b>` where we want to render it. It means that at this point we have the `<p>`'s children marked to be re-rendered.
3. Now, in the model everything is fine so we end up with such a view to be rendered: `<p><a>Text<b>XY</b></a></p>`.
4. What happens now is that the renderer tries to update first `<a>`'s children and then `<p>`'s children. It also wants to update "X" text node but since `<a>` will be rendered anyway, this is blocked.
5. `_updateChildren( <a> )` is executed but it takes `<b>` through `viewToDom()` which then uses `mapViewToDom()` (because in the view that's still the same element) and we get the old... `<b>X</b>` DOM element.
6. This leaves us with an "X" instead of "XY".

This needs to be fixed in `viewToDom()` – it seems that it needs to somehow update the entire tree (not only take the element from the mapper) or mark its contents to be re-rendered by the renderer after it does its job.

Causes the following issues:

* https://github.com/ckeditor/ckeditor5-engine/issues/1093
* https://github.com/ckeditor/ckeditor5-typing/issues/120

### LINKED ISSUE #1342: Rendering improvements

f1ames:

Other: Introduced deep children text nodes rerendering. Closes #1125. Closes #1093. Closes ckeditor/ckeditor5-typing#120. 

---

### Additional information

The proposed solution is very similar to what was initially proposed by @scofalik - https://github.com/ckeditor/ckeditor5-engine/issues/1125#issuecomment-332791847. The logic marking text nodes is placed in `renderer._updateChildren` as this is the place where element nodes are modified and based on that modifications we can decide which elements' text nodes should be considered for rerendered.

The mutation observer change is also described here - https://github.com/ckeditor/ckeditor5-engine/issues/1125#issuecomment-373376368.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -198,18 +198,18 @@ export default class Renderer {
 			this.markedChildren.add( inlineFillerPosition.parent );
 		}
 
-		for ( const node of this.markedTexts ) {
-			if ( !this.markedChildren.has( node.parent ) && this.domConverter.mapViewToDom( node.parent ) ) {
-				this._updateText( node, { inlineFillerPosition } );
-			}
-		}
-
 		for ( const element of this.markedAttributes ) {
 			this._updateAttrs( element );
 		}
 
 		for ( const element of this.markedChildren ) {
 			this._updateChildren( element, { inlineFillerPosition } );
+		}
+
+		for ( const node of this.markedTexts ) {
+			if ( !this.markedChildren.has( node.parent ) && this.domConverter.mapViewToDom( node.parent ) ) {
+				this._updateText( node, { inlineFillerPosition } );
+			}
 		}
 
 		// Check whether the inline filler is required and where it really is in the DOM.
@@ -498,6 +498,8 @@ export default class Renderer {
 				nodesToUnbind.add( actualDomChildren[ i ] );
 				remove( actualDomChildren[ i ] );
 			} else { // 'equal'
+				// Inserted nodes are already up to date, so we mark to sync those which was not rerendered (#1125).
+				this._markDescendantTextToSync( domConverter.domToView( expectedDomChildren[ i ] ) );
 				i++;
 			}
 		}
@@ -528,6 +530,25 @@ export default class Renderer {
 
 			// Not matching types.
 			return false;
+		}
+	}
+
+	/**
+	 * Marks all text nodes in a tree starting from passed element to be synced.
+	 *
+	 * @private
+	 * @param {module:engine/view/element~Element} viewElement View element whose all 'text' children should
+	 * be marked to sync. If element is 'text' itself it will be marked too.
+	 */
+	_markDescendantTextToSync( viewElement ) {
+		if ( viewElement ) {
+			if ( viewElement.is( 'text' ) ) {
+				this.markedTexts.add( viewElement );
+			} else if ( viewElement.is( 'element' ) ) {
+				for ( const child of viewElement.getChildren() ) {
+					this._markDescendantTextToSync( child );
+				}
+			}
 		}
 	}
 
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -541,14 +541,16 @@ export default class Renderer {
 	 * @private
 	 * @param {module:engine/view/node~Node} viewNode View node to sync.
 	 */
-	_markDescendantTextToSync( viewElement ) {
-		if ( viewElement ) {
-			if ( viewElement.is( 'text' ) ) {
-				this.markedTexts.add( viewElement );
-			} else if ( viewElement.is( 'element' ) ) {
-				for ( const child of viewElement.getChildren() ) {
-					this._markDescendantTextToSync( child );
-				}
+	_markDescendantTextToSync( viewNode ) {
+		if ( !viewNode ) {
+			return;
+		}
+
+		if ( viewNode.is( 'text' ) ) {
+			this.markedTexts.add( viewNode );
+		} else if ( viewNode.is( 'element' ) ) {
+			for ( const child of viewNode.getChildren() ) {
+				this._markDescendantTextToSync( child );
 			}
 		}
 	}
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -498,7 +498,7 @@ export default class Renderer {
 				nodesToUnbind.add( actualDomChildren[ i ] );
 				remove( actualDomChildren[ i ] );
 			} else { // 'equal'
-				// Inserted nodes are already up to date, so we mark to sync those which was not rerendered (#1125).
+				// Force updating text nodes inside elements which did not change and do not need to be re-rendered (#1125).
 				this._markDescendantTextToSync( domConverter.domToView( expectedDomChildren[ i ] ) );
 				i++;
 			}
@@ -534,11 +534,12 @@ export default class Renderer {
 	}
 
 	/**
-	 * Marks all text nodes in a tree starting from passed element to be synced.
+	 * Marks text nodes to be synced.
+	 *
+	 * If a text node is passed, it will be marked. If an element is passed, all descendant text nodes inside it will be marked.
 	 *
 	 * @private
-	 * @param {module:engine/view/element~Element} viewElement View element whose all 'text' children should
-	 * be marked to sync. If element is 'text' itself it will be marked too.
+	 * @param {module:engine/view/node~Node} viewNode View node to sync.
 	 */
 	_markDescendantTextToSync( viewElement ) {
 		if ( viewElement ) {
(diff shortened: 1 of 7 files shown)


Reinmar:
@Mgsy Could you check this PR? There's a wide range of test cases so we should ba careful.

f1ames:
This PR introduces mechanism which forces rerendering of sibling `text nodes` of the `text node` which was modified.

One visible issue of previous behaviour is described in #1093 - in general, renderer failed when changes in one `text node` required rerendering of some of its sibling (e.g. spaces rerendering in siblings). And the introduced change takes care of that. Here `renderer._updateText` function was not changed so rendering happens in the same manner (text is rerendered only if it different than existing ones). The only difference is that it is executed for more text nodes than before.

The other change which may have some impact is [change in rendering order](https://github.com/ckeditor/ckeditor5-engine/pull/1366/files#diff-4c87133ed830fc4ea0646426deba32cfL201). Previously, `text nodes` were rerendered first and `element nodes` after that. Now due to the logic which was applied (_rerender children `text nodes` of synced `element nodes`_), the order was changed so `element nodes` are rerendered before `text nodes`. So **it may affect any operation which does more complex changes (so updating text/elements or its attributes at the same time) and probably we should look for any regressions in such cases**.

The last thing was a change in [processing mutations](https://github.com/ckeditor/ckeditor5-engine/pull/1366/files#diff-57e194ad013563b5752afa147f447d67L209) already described in https://github.com/ckeditor/ckeditor5-engine/issues/1125#issuecomment-373376368. Here it affects how mutations are generated (previously `view element` was passed with its children to mutation event and now it's passed without it). **The scenarios which may get affected by this change are some bigger content modifications (when some bigger part of content, especially with deep structure, gets changed or partially updated) and also cases (I'm not sure if there are any of such situations) when editing one element inside editor changes the other one not in the subtree (so not a child but sibling or any not directly structurally connected element)**.

Mgsy:
I've spent some time testing this one. It fixes mentioned issues and I haven't noticed any regressions caused by this fix.

Reinmar:
There's one thing that would make us feel safer – a test which checks rendering one big batch of various type of changes – in text nodes, removing, adding elements in various places on various levels of a tree. The troublesome part is how text updates mix with element updates due to shifts in indexes.

f1ames:
> a test which checks rendering one big batch of various type of changes – in text nodes, removing, adding elements in various places on various levels of a tree...

Makes perfect sense, I will try to create such test so we could check if the change in rendering order had some unexpected effects.

src/view/renderer.js
```diff
@@ -531,6 +533,25 @@ export default class Renderer {
...
+	 */
+	_markDescendantTextToSync( viewElement ) {
+		if ( viewElement ) {
+			if ( viewElement.is( 'text' ) ) {
+				this.markedTexts.add( viewElement );
+			} else if ( viewElement.is( 'element' ) ) {
+				for ( const child of viewElement.getChildren() ) {
+					this._markDescendantTextToSync( child );
+				}
+			}
```
scofalik: I wonder when this else path is taken (what else could `viewElement` be), but CC is 100%.

f1ames: Theoretically it shouldn't be anything else so simple `if - else` should be sufficient. But checking it in both case is more future proof.

f1ames:
@scofalik after merging `master` one unit test from model (looks irrelevant to the changes - `model/document/should call all already processed callbacks again if a callback returned true`) started to fail on Travis 🤔 But it seems `master` itself passes on Travis.

scofalik:
For me it looks good. I've pushed some slight changes in docs.

f1ames:
And I pushed tests @Reinmar requested, so now we just need to figure out why this one test is failing on Travis.

f1ames:
This test also fails on `ckeditor5-engine#master` - https://travis-ci.org/ckeditor/ckeditor5-engine/builds/359328562#L2272 so it is not related to this PR.

PR #1368, opened 2018-03-19
Introduced .isBefore() and .isAfter() in model.Node and view.Node

### LINKED ISSUE #1365: Introduce `isBefore()` and `isAfter()` to view and model nodes
Similarly to what we have in `model.Position` and `view.Position`, we could use `model.Node#isBefore` and `view.Node#isBefore`. Same for `#isAfter`. Of course, the methods would compare two nodes and tell if one is before the other one in the tree. If nodes are in different trees, the methods would return `false` (just like in positions).

This might be useful for example in sorting. Now we have to create positions before items and compare positions. It's not as much of hassle but is also prone to errors (if you'll try to create a position before detached element (root) it will throw).

scofalik:

Feature: Introduced `#isBefore` and `#isAfter` in `model.Node` and `view.Node`. Additionally, `model.Node#is` and `view.Node#is` and `view.Node#getPath` were added. Closes #1365.

### DIFF
--- a/src/model/node.js
+++ b/src/model/node.js
@@ -9,6 +9,7 @@
 
 import toMap from '@ckeditor/ckeditor5-utils/src/tomap';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+import compareArrays from '@ckeditor/ckeditor5-utils/src/comparearrays';
 
 /**
  * Model node. Most basic structure of model tree.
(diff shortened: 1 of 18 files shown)


src/view/node.js
```diff
@@ -231,6 +217,23 @@ export default class Node {
...
 	 * @returns {Boolean}
 	 */
+	is( type ) {
+		return type == 'node';
+	}
+
+	/**
+	 * Clones this node.
+	 *
+	 * @method #clone
```
oskarwrobel: Should be protected.

src/model/node.js
```diff
@@ -273,6 +274,63 @@ export default class Node {
...
+			return false;
+		}
+
+		// Return `false` if it is impossible to compare nodes.
+		if ( this.root !== node.root ) {
+			return false;
+		}
+
+		// In other cases, just check if the `node` is before, and return the opposite.
+		return !this.isBefore( node );
```
oskarwrobel: clever :)

src/view/position.js
```diff
@@ -243,61 +243,39 @@ export default class Position {
...
 			return 'same';
 		}
 
-		// If positions have same parent.
-		if ( this.parent === otherPosition.parent ) {
-			return this.offset - otherPosition.offset < 0 ? 'before' : 'after';
-		}
+		// Get path from root to position's parent element. "Indexify" the paths (change elements to indices).
+		const thisPath = this.getAncestors().map( element => element.index );
+		const otherPath = otherPosition.getAncestors().map( element => element.index );
```
oskarwrobel: WDYT about using just introduced `Node#getPath()`?

```javascript
const thisPath = this.parent.is( 'node' ) ? this.parent.getPath() : [];
const otherPath = otherPosition.parent.is( 'node' ) ? otherPosition.parent.getPath() : []
```
and shifting `null` from array won't be necessary.

oskarwrobel:
I left 2 small comments but works and looks fine in general 👍

scofalik:
Fixed according to comments.

PR #1372, opened 2018-03-20
conversion.downcast-converters.changeAttribute should not consume if element creators returned null

### LINKED ISSUE #1369: changeAttribute converter consumes even if element creators returns null
As in title, `changeAttribute` converter first consumes consumables even if it is not going to do anything because `elementCreator`s returned `null`. This means that `changeAttribute` cannot be "aborted" using a proper (if statement) in a callback.

In contrary, `wrap` does it correctly:

```
return ( evt, data, conversionApi ) => {
	const oldViewElement = elementCreator( data.attributeOldValue, conversionApi.writer );
	const newViewElement = elementCreator( data.attributeNewValue, conversionApi.writer );

	if ( !oldViewElement && !newViewElement ) {
		return;
	}

	if ( !conversionApi.consumable.consume( data.item, evt.name ) ) {
		return;
	}
	
	// ... convert ...
};
```

scofalik:

Fixed: `conversion.downcast-converters.changeAttribute` should not consume if element creators returned `null.` Closes #1369.

### DIFF
--- a/tests/conversion/downcast-converters.js
+++ b/tests/conversion/downcast-converters.js
@@ -615,7 +615,7 @@ describe( 'downcast-converters', () => {
 		} );
 	} );
 
-	describe( 'setAttribute', () => {
+	describe( 'changeAttribute', () => {
 		it( 'should convert attribute insert/change/remove on a model node', () => {
 			const modelElement = new ModelElement( 'paragraph', { class: 'foo' }, new ModelText( 'foobar' ) );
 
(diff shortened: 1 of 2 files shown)


tests/conversion/downcast-converters.js
```diff
@@ -686,6 +686,26 @@ describe( 'downcast-converters', () => {
 			// No attribute set.
 			expect( viewToString( viewRoot ) ).to.equal( '<div><p>foobar</p></div>' );
 		} );
+
+		it( 'should not convert or consume if element creator returned null', () => {
+			const spy = sinon.spy();
```
jodator: I'd use `const callback = sinon.stub().returns( null )` here - tests will be shorter.

tests/conversion/downcast-converters.js
```diff
@@ -686,6 +686,26 @@ describe( 'downcast-converters', () => {
...
+
+			const modelElement = new ModelElement( 'paragraph', { class: 'foo' }, new ModelText( 'foobar' ) );
+
+			model.change( writer => {
+				writer.insert( modelElement, modelRootStart );
+			} );
+
+			expect( viewToString( viewRoot ) ).to.equal( '<div><p class="foo">foobar</p></div>' );
+
+			expect( spy.called ).to.be.true;
```
jodator: AFAIR it's better to use `sinon.assert.called( spy )` - there were some issues with `expect()` & sinon.

scofalik:
Fixed according to suggestions.

PR #1375, opened 2018-03-22
Fix: Upcasting element with multiple attributes should not consume th…

### LINKED ISSUE #1348: Upcasting an element that has multiple attributes may not work
When trying to write additional converters for link-related attributes I realized that I am unable to do it using the new, refactored helpers. Here's why:

1. The original converter, declared in Link feature, converts `linkHref` model attribute to `<a href="">` view element. And the other way around.
2. Upcasting `<a href="">` causes consuming not only `href` attribute but also tagname `a` is consumed.
3. Another converter (let's say for `linkRel`) does not convert because `a` is already consumed.

We checked with @jodator how it works in Font Size and Font Family plugins. Unfortunately there's a bug. `<span class="text-huge" style="font-family: ... ">` is not read correctly - only `class` is read and converted to font size.

We need some kind of workaround for this.

jodator:

Fix: Upcasting element with multiple attributes should not consume the element name by default. Closes #1348.

---

### DIFF
--- a/src/conversion/upcast-converters.js
+++ b/src/conversion/upcast-converters.js
@@ -138,7 +138,7 @@ export function upcastElementToAttribute( config ) {
 
 	_normalizeModelAttributeConfig( config );
 
-	const converter = _prepareToAttributeConverter( config, true );
+	const converter = _prepareToAttributeConverter( config );
 
 	const elementName = _getViewElementNameFromConfig( config );
 	const eventName = elementName ? 'element:' + elementName : 'element';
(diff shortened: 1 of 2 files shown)


Reinmar:
So, you decided to change the default behaviour without adding an option to override it?

jodator:
@Reinmar more or less yes. I've get the feeling that it was the outcome of the discussion in #1348 as the overriding this is not on a near horizon - so we'd implement that if needed.

Reinmar:
Yep, I think so too. Just confirming it.

PR #1381, opened 2018-03-26
Introduced model.Model#convertToDeltas

### LINKED ISSUE #1380: Introduce model.Model#convertToDeltas
Introduce a function that will take a `model.DocumentFragment` instance and convert it to an array of deltas which, when applied, will re-create the document fragment. Should return one `InsertDelta` with all the nodes and zero or more `MarkerDelta`s to re-create all the markers that were set on the `model.DocumentFragment`.

This is required by Letters and also should be useful for any compression algorithms.

scofalik:

Feature: Introduced `model.Model#convertToDeltas`. Closes #1380.

### DIFF
--- a/src/model/model.js
+++ b/src/model/model.js
@@ -21,6 +21,12 @@ import mix from '@ckeditor/ckeditor5-utils/src/mix';
 import deltaTransform from './delta/transform';
 import ModelElement from './element';
 import ModelRange from './range';
+import ModelPosition from './position';
+
+import InsertDelta from './delta/insertdelta';
+import MarkerDelta from './delta/markerdelta';
+import InsertOperation from './operation/insertoperation';
+import MarkerOperation from './operation/markeroperation';
 
 import insertContent from './utils/insertcontent';
 import deleteContent from './utils/deletecontent';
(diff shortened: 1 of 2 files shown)


src/model/model.js
```diff
@@ -382,6 +388,48 @@ export default class Model {
...
+	 * inserts all the nodes and zero or more {@link module:engine/model/operation/markeroperation~MarkerOperation}, depending
+	 * on {@link module:engine/model/documentfragment~DocumentFragment#markers markers set on the document fragment}.
+	 *
+	 * @param {module:engine/model/documentfragment~DocumentFragment} documentFragment Document fragment to be converted to deltas.
+	 * @param {String} [rootName] Name of a root element which will be base for the operations. The returned insert
+	 * operation will insert the nodes inside the given root. If not set, the main root will be used.
+	 * @returns {Array.<module:engine/model/operation/operation~Operation>} Operations which, when applied, will recreate given
+	 * `documentFragment` inside the specified root.
+	 */
+	convertToDeltas( documentFragment, rootName ) {
```
Reinmar: If you can think of some other name, which avoids the word "convert", I'm sure @pjasiun would be happy :P

pjasiun: Yep. And `transformToDeltas` won't be good either.

src/model/model.js
```diff
@@ -382,6 +388,48 @@ export default class Model {
...
+	/**
+	 * Creates deltas from given {@link module:engine/model/documentfragment~DocumentFragment}.
+	 *
+	 * Returned array contains exactly one {@link module:engine/model/operation/insertoperation~InsertOperation} which
+	 * inserts all the nodes and zero or more {@link module:engine/model/operation/markeroperation~MarkerOperation}, depending
+	 * on {@link module:engine/model/documentfragment~DocumentFragment#markers markers set on the document fragment}.
+	 *
+	 * @param {module:engine/model/documentfragment~DocumentFragment} documentFragment Document fragment to be converted to deltas.
+	 * @param {String} [rootName] Name of a root element which will be base for the operations. The returned insert
+	 * operation will insert the nodes inside the given root. If not set, the main root will be used.
```
pjasiun: Can't you just create the tree in the `RootElement`?

src/model/model.js
pjasiun: In the F2F talk, we agreed it will have no parameters and will serialize model to deltas. It can be called `getAsDeltas`

scofalik:
Eventually, we agreed with PJ that we can create a `model.Model` instance and rename the method to `getAsDeltas()`. But then we concluded that we could just take the deltas straight from the history. So this method is not needed for now.

Reinmar:
My pleasure :D

PR #1388, opened 2018-03-28
T/1299: Create writer#updateMarker() method.

### LINKED ISSUE #1299: model.writer#setMarker should not change marker's options when they are not set
When creating a marker you can set whether the marker is managed through operations:
```
const marker = modelWriter.setMarker( 'myMarker', myRange, { usingOperations: true } );
```
However, at the moment, the method works in a way, that you need to pass that object every time because when not set, the option will be overwritten:
```
modelWriter.setMarker( marker, myOtherRange ); // Will set `usingOperations` to `false`.
```
Instead, when options object is not set, `setMarker` call should not change any options. Only this should change an option:
```
modelWriter.setMarker( marker, myOtherRange, { usingOperations: false } );
```

jodator:

Feature: Introduce `writer#updateMarker()` method. Closes #1299.

BREAKING CHANGE: The `writer#setMarker()` method is used only to create a new marker and it does not accept a `marker` instance as a parameter. To update existing marker use `writer#updateMarker()` method.
BREAKING CHANGE: The `options.usingOperation` option in `writer#setMarker()` is now a required one.
BREAKING CHANGE: The `range` parameter was removed. Use `options.range` instead.

---

### Additional information

Other repositories that required changes:
- ckeditor5-link: https://github.com/ckeditor/ckeditor5-link/tree/t/ckeditor5-engine/1299
- ckeditor5-undo: https://github.com/ckeditor/ckeditor5-undo/tree/t/ckeditor5-engine/1299

Suggested merge message for other repositores:
```
Aligned code to changes in writer#setMarker method.
```

### DIFF
--- a/src/model/writer.js
+++ b/src/model/writer.js
@@ -884,7 +884,6 @@ export default class Writer {
 	 * @param {module:engine/model/range~Range} [options.range] Marker range to update.
 	 * @param {Boolean} [options.usingOperation] Flag indicated whether the marker should be added by MarkerOperation.
 	 * See {@link module:engine/model/markercollection~Marker#managedUsingOperations}.
-	 * @returns {module:engine/model/markercollection~Marker} Marker that was set.
 	 */
 	updateMarker( markerOrName, options ) {
 		this._assertWriterUsedCorrectly();
@@ -902,30 +901,43 @@ export default class Writer {
 			throw new CKEditorError( 'writer-updateMarker-marker-not-exists: Marker with provided name does not exists.' );
 		}
 
-		const range = options && options.range;
-
+		const newRange = options && options.range;
 		const hasUsingOperationDefined = !!options && typeof options.usingOperation == 'boolean';
-		const usingOperation = !!options && !!options.usingOperation; // : currentMarker.managedUsingOperations;
 
-		if ( !usingOperation ) {
-			// If marker changes to a marker that do not use operations then we need to create additional operation
-			// that removes that marker first.
-			if ( hasUsingOperationDefined && currentMarker.managedUsingOperations ) {
-				applyMarkerOperation( this, markerName, currentMarker.getRange(), null );
+		if ( !hasUsingOperationDefined && !newRange ) {
+			/**
+			 * One of options is required - provide range or usingOperations.
+			 *
+			 * @error writer-updateMarker-wrong-options
+			 */
+			throw new CKEditorError( 'writer-updateMarker-wrong-options: One of options is required - provide range or usingOperations.' );
+		}
+
+		if ( hasUsingOperationDefined && options.usingOperation !== currentMarker.managedUsingOperations ) {
+			// The marker type is changed so it's necessary to create proper operations.
+			if ( options.usingOperation ) {
+				// If marker changes to a managed one treat this as synchronizing existing marker.
+				// If marker changes to a managed one treat this as synchronizing existing marker.
+				// Create `MarkerOperation` with `oldRange` set to `null`, so reverse operation will remove the marker.
+				applyMarkerOperation( this, markerName, null, newRange ? newRange : currentMarker.getRange() );
+			} else {
+				// If marker changes to a marker that do not use operations then we need to create additional operation
+				// that removes that marker first.
+				const currentRange = currentMarker.getRange();
+				applyMarkerOperation( this, markerName, currentRange, null );
+
+				// Although not managed the marker itself should stay in model and its range should be preserver or changed to passed range.
+				this.model.markers._set( markerName, newRange ? newRange : currentRange );
 			}
-
-			this.model.markers._set( markerName, range, hasUsingOperationDefined ? usingOperation : currentMarker.managedUsingOperations );
 
 			return;
 		}
 
-		if ( !range ) {
-			// If `range` is not given, treat this as synchronizing existing marker.
-			// Create `MarkerOperation` with `oldRange` set to `null`, so reverse operation will remove the marker.
-			applyMarkerOperation( this, markerName, null, currentMarker.getRange() );
+		// Marker's type doesn't change so update it accordingly.
+		if ( currentMarker.managedUsingOperations ) {
+			applyMarkerOperation( this, markerName, currentMarker.getRange(), newRange );
 		} else {
-			// Just change marker range.
-			applyMarkerOperation( this, markerName, currentMarker.getRange(), range );
+			this.model.markers._set( markerName, newRange );
 		}
 	}
 
--- a/src/model/writer.js
+++ b/src/model/writer.js
@@ -903,16 +903,18 @@ export default class Writer {
 		}
 
 		const range = options && options.range;
-		const usingOperation = !!options && !!options.usingOperation;
+
+		const hasUsingOperationDefined = !!options && typeof options.usingOperation == 'boolean';
+		const usingOperation = !!options && !!options.usingOperation; // : currentMarker.managedUsingOperations;
 
 		if ( !usingOperation ) {
 			// If marker changes to a marker that do not use operations then we need to create additional operation
 			// that removes that marker first.
-			if ( currentMarker.managedUsingOperations ) {
+			if ( hasUsingOperationDefined && currentMarker.managedUsingOperations ) {
 				applyMarkerOperation( this, markerName, currentMarker.getRange(), null );
 			}
 
-			this.model.markers._set( markerName, range, usingOperation );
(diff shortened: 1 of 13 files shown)


oskarwrobel:
Since we have a separate method for updating marker we can rename setMarker to addMarker.

src/model/writer.js
```diff
@@ -789,101 +860,71 @@ export default class Writer {
...
+	 * The `options.usingOperation` parameter lets you change if the marker should be managed by operations or not. See
 	 * {@link module:engine/model/markercollection~Marker marker class description} to learn about the difference between
 	 * markers managed by operations and not-managed by operations. It is possible to change this option for an existing marker.
 	 * This is useful when a marker have been created earlier and then later, it needs to be added to the document history.
 	 *
-	 * Create/update marker directly base on marker's name:
+	 * Update marker directly base on marker's name:
 	 *
-	 * 		setMarker( markerName, range );
+	 * 		setMarker( markerName, { range } );
```
pjasiun: It should be `updateMarker`.

src/model/writer.js
```diff
@@ -789,101 +860,71 @@ export default class Writer {
...
-	 *
-	 * Create marker with a unique id using operation:
-	 *
-	 * 		setMarker( range, { usingOperation: true } );
-	 *
-	 * Create marker directly without using operations:
-	 *
-	 * 		setMarker( range )
+	 * 		setMarker( marker, { range, usingOperation: true } );
+	 * 		setMarker( markerName, { range, usingOperation: true } );
```
pjasiun: It should be updateMarker.

tests/model/writer.js
```diff
@@ -2544,16 +2596,22 @@ describe( 'Writer', () => {
...
-			marker = writer.setMarker( markerOrNameOrRange, rangeOrManagedUsingOperations, options );
+			marker = writer.setMarker( name, options );
 		} );
 
 		return marker;
 	}
 
+	function updateMarker( markerOrNameOrRange, rangeOrManagedUsingOperations, options ) {
+		model.enqueueChange( batch, writer => {
+			writer.updateMarker( markerOrNameOrRange, rangeOrManagedUsingOperations, options );
```
pjasiun: `updateMarker` has different parameters.

src/model/writer.js
```diff
@@ -781,6 +780,78 @@ export default class Writer {
...
+	 *
+	 * Note: For efficiency reasons, it's best to create and keep as little markers as possible.
+	 *
+	 * @see module:engine/model/markercollection~Marker
+	 * @param {String} name Name of a marker to create - must be unique.
+	 * @param {Object} options
+	 * @param {Boolean} options.usingOperation Flag indicated whether the marker should be added by MarkerOperation.
+	 * @param {module:engine/model/range~Range} options.range Marker range.
+	 * See {@link module:engine/model/markercollection~Marker#managedUsingOperations}.
```
pjasiun: This `see` is to `usingOperation`. Not to `range`.

src/model/writer.js
```diff
@@ -789,101 +860,71 @@ export default class Writer {
...
+			 * Marker with provided name does not exists.
+			 *
+			 * @error writer-updateMarker-marker-not-exists
+			 */
+			throw new CKEditorError( 'writer-updateMarker-marker-not-exists: Marker with provided name does not exists.' );
 		}
 
-		const currentMarker = this.model.markers.get( markerName );
+		const range = options && options.range;
+		const usingOperation = !!options && !!options.usingOperation;
```
pjasiun: Maybe I do not understand something or was not clear enough, but if I call `writer.updateMarker( name, { range } );` (with `usingOperation` not specified), then `usingOperation` should **not change**. So only range should change. This code looks like it changes `usingOperation` to `false` in such case. Also, I can not find a test which checks this.

src/model/writer.js
```diff
@@ -789,101 +860,73 @@ export default class Writer {
...
-			if ( !newRange ) {
-				/**
-			 	 * Range parameter is required when adding a new marker.
-				 *
-				 * @error writer-setMarker-no-range
-				 */
-				throw new CKEditorError( 'writer-setMarker-no-range: Range parameter is required when adding a new marker.' );
-			}
+		const hasUsingOperationDefined = !!options && typeof options.usingOperation == 'boolean';
+		const usingOperation = !!options && !!options.usingOperation; // : currentMarker.managedUsingOperations;
```
pjasiun: `// : currentMarker.managedUsingOperations;`?

tests/model/writer.js
```diff
@@ -1952,128 +1952,159 @@ describe( 'Writer', () => {
...
 
-		it( 'should use operations when having set usingOperation to true', () => {
+		it( 'should accept empty range parameter if marker instance is passed and usingOperation is set to true', () => {
+			const marker = addMarker( 'name', { range, usingOperation: true } );
 			const spy = sinon.spy();
 
 			model.on( 'applyOperation', spy );
 
-			setMarker( 'name', range, { usingOperation: true } );
+			updateMarker( marker, { usingOperation: true } );
```
pjasiun: You don't change range nor `usingOperation`. I do not get what this test is about.

tests/model/writer.js
```diff
@@ -2086,25 +2117,62 @@ describe( 'Writer', () => {
...
 			const spy = sinon.spy();
 			model.on( 'applyOperation', spy );
 
-			setMarker( 'name', range );
-			const marker = setMarker( 'name', range, { usingOperation: true } );
+			addMarker( 'name', { range, usingOperation: true } );
+
+			sinon.assert.calledOnce( spy );
+
+			updateMarker( 'name', { range } );
```
pjasiun: As above. You don't change range nor usingOperation. I do not get what this test is about.

pjasiun:
Tests for `updateMarker` are bad. First, there is no point to set the same range or `usingOperation` value in `addMarker` and then in `updateMarker` in that many tests. It might be a good idea to have one or two tests like this to be sure that nothing bad happens if the range or `usingOperation` parameter is the same, but it should be an edge case.

At the same time, there are missing tests to check if `updateMarker` really update markers. It should be possible to change the value of `usingOperation` or `range` or both using this method and this is what should be deeply tested.

src/model/writer.js
```diff
@@ -789,101 +860,73 @@ export default class Writer {
...
-	 *
-	 * Create marker with a unique id using operation:
-	 *
-	 * 		setMarker( range, { usingOperation: true } );
-	 *
-	 * Create marker directly without using operations:
-	 *
-	 * 		setMarker( range )
+	 * 		updateMarker( marker, { range, usingOperation: true } );
+	 * 		updateMarker( markerName, { range, usingOperation: true } );
```
pjasiun: There should be no `range` in a sample to show that you can update only `usingOperation: true` parameter.

src/model/writer.js
```diff
@@ -789,101 +860,73 @@ export default class Writer {
...
+			// If marker changes to a marker that do not use operations then we need to create additional operation
 			// that removes that marker first.
-			if ( currentMarker && currentMarker.managedUsingOperations && !usingOperation ) {
+			if ( hasUsingOperationDefined && currentMarker.managedUsingOperations ) {
 				applyMarkerOperation( this, markerName, currentMarker.getRange(), null );
 			}
 
-			return this.model.markers._set( markerName, newRange, usingOperation );
-		}
+			this.model.markers._set( markerName, range, hasUsingOperationDefined ? usingOperation : currentMarker.managedUsingOperations );
```
pjasiun: What if `range` is not defined?

jodator:
OK to have this clear, the `updateMarker()` should
1. `updateMarker( marker, { range } )` - change marker's range (usingOperation as defined in `addMarker()`
2. `updateMarker( marker, { usingOperation: true } )` - change marker's `mangedUsingOperation` to `true` (if `usingOperation` was not defined in `addMarker()`
3. `updateMarker( marker, { usingOperation: false } )` - change marker's `mangedUsingOperation` to `false` (if `usingOperation` was defined in `addMarker()`
4. `updateMarker( marker, { range, usingOperation )` - change marker's range and change it's `managedUsingOperation` if it is different then current marker's `mangedUsingOperation`.

jodator:
@pjasiun OK I've revisited the test and I hope that all cases are handled. In particular I've added (changed) that if a marker was created not using operations and we update it with a new range the operations that is registered is for the updated range directly).

And vice versa if a manged marker is updated to be not managed by operations with a new range the operation that removes the marker is for current range but the marker is updated to a new range.

PR #1390, opened 2018-03-30
Added missing `.ck` class to placeholder.

dkonopka:

Fix: Added missing `.ck` class to placeholder. Closes https://github.com/ckeditor/ckeditor5/issues/931.

### DIFF
--- a/src/view/placeholder.js
+++ b/src/view/placeholder.js
@@ -54,7 +54,7 @@ export function detachPlaceholder( view, element ) {
 	}
 
 	view.change( writer => {
-		writer.removeClass( 'ck-placeholder', element );
+		writer.removeClass( 'ck ck-placeholder', element );
 		writer.removeAttribute( 'data-placeholder', element );
 	} );
 }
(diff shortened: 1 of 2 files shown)


src/view/placeholder.js
```diff
@@ -54,7 +54,7 @@ export function detachPlaceholder( view, element ) {
 	}
 
 	view.change( writer => {
-		writer.removeClass( 'ck-placeholder', element );
+		writer.removeClass( 'ck ck-placeholder', element );
 		writer.removeAttribute( 'data-placeholder', element );
```
oleq: I'm not super sure but I'd rather use `writer.removeClass( [ 'foo', 'bar' ], ... );` https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_engine_view_writer-Writer.html#function-removeClass. The same for other methods. It feels safer, though I don't know the consequences of the way you used it.

dkonopka: After that change, test is getting timeout locally and in Travis too 😞

PR #1391, opened 2018-04-03
Fix for placeholder `ck` class removal

szymonkups:

Fix: Placeholder is removing `ck` CSS class only if it was added by that placeholder. Closes https://github.com/ckeditor/ckeditor5-image/issues/198.

### DIFF
--- a/src/view/placeholder.js
+++ b/src/view/placeholder.js
@@ -34,7 +34,14 @@ export function attachPlaceholder( view, element, placeholderText, checkFunction
 	}
 
 	// Store information about element with placeholder.
-	documentPlaceholders.get( document ).set( element, { placeholderText, checkFunction } );
+	documentPlaceholders.get( document ).set( element, {
+		placeholderText,
+		checkFunction,
+
+		// Store information if 'ck' class is already present and don't remove it when detaching placeholder.
+		// https://github.com/ckeditor/ckeditor5-image/issues/198#issuecomment-377542222
+		removeCkClass: !element.hasClass( 'ck' )
+	} );
 
 	// Update view right away.
 	view.render();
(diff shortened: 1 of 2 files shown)


Reinmar:
Unless someone's against, https://github.com/ckeditor/ckeditor5-image/issues/198#issuecomment-378211139 contains a solution which differs from the one proposes in this PR so I'm R-ing it.

src/view/placeholder.js
```diff
@@ -48,13 +55,22 @@ export function attachPlaceholder( view, element, placeholderText, checkFunction
  */
 export function detachPlaceholder( view, element ) {
 	const document = element.document;
+	let removeCkClass = false;
```
Reinmar: Why is this variable defined here? Why can't `view.change()` be in the `if()`?

src/view/placeholder.js
```diff
@@ -34,7 +34,14 @@ export function attachPlaceholder( view, element, placeholderText, checkFunction
...
 
 	// Store information about element with placeholder.
-	documentPlaceholders.get( document ).set( element, { placeholderText, checkFunction } );
+	documentPlaceholders.get( document ).set( element, {
+		placeholderText,
+		checkFunction,
+
+		// Store information if 'ck' class is already present and don't remove it when detaching placeholder.
+		// https://github.com/ckeditor/ckeditor5-image/issues/198#issuecomment-377542222
+		removeCkClass: !element.hasClass( 'ck' )
```
Reinmar: Whether to remove the `ck` class when detaching the placeholder.

Reinmar: "if 'ck' class is already present" is exactly what the code below is so it's unnecessary to additionally write this in a comment.

src/view/placeholder.js
```diff
@@ -105,8 +121,8 @@ function updateSinglePlaceholder( writer, element, info ) {
 
 	// If checkFunction is provided and returns false - remove placeholder.
 	if ( checkFunction && !checkFunction() ) {
-		if ( element.hasClass( 'ck', 'ck-placeholder' ) ) {
-			writer.removeClass( [ 'ck', 'ck-placeholder' ], element );
+		if ( element.hasClass( 'ck-placeholder' ) ) {
```
Reinmar: I like that you decided to not remove `.ck` when the placeholder is not needed. It's simpler and totally reasonable – the `.ck-placeholder` class is responsible for showing the thing. The `.ck` class makes for the base.

src/view/placeholder.js
```diff
@@ -50,11 +57,16 @@ export function detachPlaceholder( view, element ) {
...
 	if ( documentPlaceholders.has( document ) ) {
+		const info = documentPlaceholders.get( document ).get( element );
+		if ( info.removeCkClass ) {
+			view.change( writer => writer.removeClass( 'ck', element ) );
+		}
+
 		documentPlaceholders.get( document ).delete( element );
 	}
 
 	view.change( writer => {
```
Reinmar: We have two `view.change()` calls which mean rendering the thing twice. Can't we have one big `view.change()`? It doesn't actually matter, but I'm asking out of curiosity.

Reinmar:
We have some doubts whether using the `.ck` class in the content makes sense in general. @szymonkups will report a ticket about that because we may need to rethink this. But let's fix the tests for now  by merging this PR.

Reinmar:
Followup in https://github.com/ckeditor/ckeditor5/issues/936.

PR #1394, opened 2018-04-04
Removed the "model.Writer#setTextData()" method

### LINKED ISSUE #1363: Should model.Writer#setTextData() be removed?
I can't see it used... and it shouldn't be used at all because it changes the text node's data directly:

```js
	/**
	 * Sets the text content for the specified `textNode`.
	 *
	 * @param {String} value New value.
	 * @param {module:engine/model/text~Text} textNode Text node that will be updated.
	 */
	setTextData( value, textNode ) {
		textNode._data = value;
	}
```

pomek:

Other: Removed the unnecesary  `model.Writer#setTextData()` and `view.Writer#setTextData()` methods. Closes #1363.

Should be merged with https://github.com/ckeditor/ckeditor5-list/pull/102.

### DIFF
--- a/src/conversion/conversion.js
+++ b/src/conversion/conversion.js
@@ -133,7 +133,7 @@ export default class Conversion {
 	 *			model: 'fancyParagraph',
 	 *			view: {
 	 *				name: 'p',
-	 *				class: 'fancy'
+	 *				classes: 'fancy'
 	 *			}
 	 *		} );
 	 *
@@ -145,7 +145,7 @@ export default class Conversion {
 	 *				'div',
 	 *				{
 	 *					// Any element with `display: block` style.
-	 *					style: {
+	 *					styles: {
 	 *						display: 'block'
 	 *					}
 	 *				}
@@ -177,7 +177,7 @@ export default class Conversion {
 	 *					// Those properties will be "consumed" during conversion.
 	 *					// See `engine.view.Matcher~MatcherPattern` and `engine.view.Matcher#match` for more.
 	 *
-	 *					return { name: true, style: [ 'font-size' ] };
+	 *					return { name: true, styles: [ 'font-size' ] };
 	 *				}
 	 *
 	 *				return null;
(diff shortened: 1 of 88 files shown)


Reinmar:
You didn't check whether this method is used somewhere... and it is.

@scofalik, I checked that we can't simply remove this bit: https://github.com/ckeditor/ckeditor5-list/blob/d2023267caa88eb383b1fab773e137c233cf5b8d/src/converters.js#L455-L462

Do you think that we could use `writer.remove()` there to remove these trailing spaces?

Reinmar:
BTW, it seems that we can also fix https://github.com/ckeditor/ckeditor5-list/issues/101 easily because it should also have access to the writer.

pomek:
> You didn't check whether this method is used somewhere... and it is.

So the `view.Writer#setTextData()` isn't used anywhere.

Reinmar:
> So the view.Writer#setTextData() isn't used anywhere.

May be true. We both got a bit confused when discussing this in #1363 with @pjasiun.

Anyway, the `model.Writer#setTextData()` should be removed not because it's barely used, but because it's dangerous. You cannot change the text like this in the model.

Reinmar:
Model writer offers a simple `writer.remove()` method which accepts a range. Based on this you can remove a piece of text. I think that it will be the most optimal solution.

A simpler solution would indeed be replacing an entire text node, but then we need to take care of its attributes. That's a bit worse to me.

pomek:
Everything sounds good but the write instance is the instance of the model but the `child` is an instance of the view. So at this moment, `model.setTextData()` changes value of the view.

scofalik:
I am for `view.Writer#remove` if it wasn't resolved yet.

Reinmar:
OK, this is a deeper problem. And it's the same problem as in https://github.com/ckeditor/ckeditor5-list/issues/101 – we need a view writer somewhere. And we need a place to clean up the view being converted.

For now, let's use protected methods of view text nodes to change their data. This will be in line with what we have reported in https://github.com/ckeditor/ckeditor5-list/issues/101.

pomek:
https://github.com/ckeditor/ckeditor5-list/pull/102

This PR is ready to review once again.

PS: I fixed failed, not related to my changes test.

PR #1396, opened 2018-04-05
.ck class should not be added to elements inside editor editable

scofalik:

Internal: `.ck` class should not be added to elements inside editor editable.

---

### Additional information

Part of https://github.com/ckeditor/ckeditor5/issues/936

### DIFF
--- a/theme/placeholder.css
+++ b/theme/placeholder.css
@@ -3,6 +3,7 @@
  * For licensing, see LICENSE.md.
  */
 
+/* See ckeditor/ckeditor5#936. */
 .ck.ck-placeholder, .ck .ck-placeholder {
 	&::before {
 		 content: attr(data-placeholder);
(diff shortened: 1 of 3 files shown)


theme/placeholder.css
```diff
@@ -3,9 +3,11 @@
  * For licensing, see LICENSE.md.
  */
 
-.ck.ck-placeholder::before {
-	content: attr( data-placeholder );
+.ck.ck-placeholder, .ck .ck-placeholder {
```
Reinmar: Missing comment.

theme/placeholder.css
```diff
@@ -3,9 +3,11 @@
...
+	&::before {
+		 content: attr(data-placeholder);
```
Reinmar: You broke code style.

Reinmar: Or actually you fixed it :P

Reinmar:
I'm so proud of us.

image

PR #1403, opened 2018-04-10
view.Writer should deeply add and remove attribute elements to/from their clone groups

### LINKED ISSUE #1401: view.Writer should deeply unlink attribute elements from their groups when they are removed
Currently, `view.AttributeElement`s which have `id` property set are linked together in `view.Writer` in so called "clone groups". All elements with the same id that are still in document tree are inside the group.

When elements are removed from the document tree, they are removed from the groups. However, the function that does so is not looking recursively for the elements.

That means that in following scenarios it will fail to remove `<span class="b">`:

1. In following scenario when selection `[]` is removed.
```html
<p>f[oo</p>
<p>b<span class="b"a</span>r</p>
<p>b]az</p>
```

2. In following scenario when `<span class="a">` is unwrapped:
```html
<p>
    <span class="a">a<span class="b">bc</span>xyz</span>
</p>
```

scofalik:

Fix: `view.Writer` should deeply add and remove `view.AttributeElement`s to/from their clone groups. Closes #1401.

### DIFF
--- a/tests/view/writer/writer.js
+++ b/tests/view/writer/writer.js
@@ -13,7 +13,7 @@ import createViewRoot from '../_utils/createroot';
 describe( 'Writer', () => {
 	let writer, attributes, root, doc;
 
-	before( () => {
+	beforeEach( () => {
 		attributes = { foo: 'bar', baz: 'quz' };
 		doc = new Document();
 		root = createViewRoot( doc );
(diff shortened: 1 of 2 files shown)


tests/view/writer/writer.js
```diff
@@ -43,6 +43,9 @@ describe( 'Writer', () => {
 
 	describe( 'setSelectionFocus()', () => {
 		it( 'should use selection._setFocus method internally', () => {
+			const position = ViewPosition.createAt( root );
```
scofalik: This seems unrelated but this test started to fail after changing from `before` to `beforeEach`.

PR #1405, opened 2018-04-10
model.Differ did not handle attribute change transformation correctly

### LINKED ISSUE #1404: model.Differ does misses some cases for attribute changes transformations
Well, the thing is simple: right now we are handling just one out of four possible cases for attribute changes. We need to handle all four "intersecting" scenarios:

```
{   [   }   ]    -->    {   }[      ]
{   [   ]   }    -->    {           }
[   {   }   ]    -->    [           ]
[   {   ]   }    -->    [      ]{   }
```

Where `{}` is incoming change and `[]` is already buffered chage.

scofalik:

Fix: `model.Differ` did not handle attribute change transformation correctly. Closes #1404.

### DIFF
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -690,6 +690,7 @@ export default class Differ {
 			}
 
 			if ( inc.type == 'attribute' ) {
+				// In case of attribute change, `howMany` should be kept same as `nodesToHandle`. It's not an error.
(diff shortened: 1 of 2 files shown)


scofalik:
Self R-, I need to think about one more case that just came to my mind.

scofalik:
Okay, all should be fine.

PR #1406, opened 2018-04-16
t/1301: The bindTwoStepCaretToAttribute should not fail in more complex cases

### LINKED ISSUE #1301: Two-steps caret movement when attribute starts at the beginning of block
Follow-up #1289.

Caret does not "leave" attribute element while moving left when text starts at the beginning of block element.

```html
<p><a href="">{}foo</a></p>
```

Pressing left arrow does not move selection before element `<a/>`.

### LINKED ISSUE #1346: Re–entering the link selection mode takes an additional key press
1. Go to http://localhost:8125/ckeditor5-core/tests/manual/article.html
2. Place the caret before the link.
3. Press right arrow key until the link gets highlighted.
4. Press the left arrow key to leave the link.
5. Repeat 3.

**Expected**:

The same number of left/right key presses is used to enter and leave.

**Actual**:

When re-entering the link (4.), there's one additional ghost step before the link is highlighted.

kapture 2018-03-08 at 16 06 03

Related to https://github.com/ckeditor/ckeditor5-engine/issues/1301 (?)

cc @Reinmar @szymonkups

oleq:

Fix: The `bindTwoStepCaretToAttribute` behavioral helper should not fail in more complex cases. Closes #1301. Closes #1346. Closes ckeditor/ckeditor5#937.  Closes ckeditor/ckeditor5#922.  Closes ckeditor/ckeditor5#946.

---

### Additional information

**Note**: A thorough testing required before merging. @Mgsy?

### DIFF
--- a/src/utils/bindtwostepcarettoattribute.js
+++ b/src/utils/bindtwostepcarettoattribute.js
@@ -225,7 +225,7 @@ class TwoStepCaretHandler {
 
 		// DON'T ENGAGE 2-SCM if gravity is already overridden. It means that we just entered
 		//
-		// 		<paragraph>foo{}<$text attribute>bar</$text>baz</paragraph>
+		// 		<paragraph>foo<$text attribute>{}bar</$text>baz</paragraph>
 		//
 		// or left the attribute
 		//
@@ -236,7 +236,8 @@ class TwoStepCaretHandler {
 			return;
 		}
 
-		// DON'T ENGAGE 2-SCM when the selection is at the beginning of an attribute AND already has it:
+		// DON'T ENGAGE 2-SCM when the selection is at the beginning of the block AND already has the
+		// attribute:
 		// * when the selection was initially set there using the mouse,
 		// * when the editor has just started
 		//
@@ -257,26 +258,28 @@ class TwoStepCaretHandler {
 		if ( isBetweenDifferentValues( position, attribute ) && this._hasAttribute ) {
 			this._preventCaretMovement( data );
 			this._removeSelectionAttribute();
-		} else {
-			// ENGAGE 2-SCM when entering an attribute:
-			//
-			// 		<paragraph>foo{}<$text attribute>bar</$text>baz</paragraph>
-			//
-			if ( isAtStartBoundary( position, attribute ) ) {
-				this._preventCaretMovement( data );
-				this._overrideGravity();
 
-				return;
-			}
+			return;
+		}
 
-			// ENGAGE 2-SCM when leaving an attribute:
-			//
-			//		<paragraph>foo<$text attribute>bar{}</$text>baz</paragraph>
-			//
-			if ( isAtEndBoundary( position, attribute ) && this._hasAttribute ) {
-				this._preventCaretMovement( data );
-				this._overrideGravity();
-			}
+		// ENGAGE 2-SCM when entering an attribute:
+		//
+		// 		<paragraph>foo{}<$text attribute>bar</$text>baz</paragraph>
+		//
+		if ( isAtStartBoundary( position, attribute ) ) {
+			this._preventCaretMovement( data );
+			this._overrideGravity();
+
+			return;
+		}
+
+		// ENGAGE 2-SCM when leaving an attribute:
+		//
+		//		<paragraph>foo<$text attribute>bar{}</$text>baz</paragraph>
+		//
+		if ( isAtEndBoundary( position, attribute ) && this._hasAttribute ) {
+			this._preventCaretMovement( data );
+			this._overrideGravity();
 		}
 	}
 
--- a/src/utils/bindtwostepcarettoattribute.js
+++ b/src/utils/bindtwostepcarettoattribute.js
@@ -380,8 +380,6 @@ class TwoStepCaretHandler {
 			if ( position.isAtStart && isAtBoundary( position, attribute ) ) {
 				if ( this._hasSelectionAttribute ) {
 					this._removeSelectionAttribute();
-
-					return;
 				}
 
 				return;
--- a/src/utils/bindtwostepcarettoattribute.js
+++ b/src/utils/bindtwostepcarettoattribute.js
@@ -37,20 +37,11 @@ import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
  * @param {String} attribute Attribute for which this behavior will be added.
  */
 export default function bindTwoStepCaretToAttribute( view, model, emitter, attribute ) {
+	const twoStepCaretHandler = new TwoStepCaretHandler( model, emitter, attribute );
 	const modelSelection = model.document.selection;
-	let overrideUid;
-	let skipNextChangeRange = false;
 
 	// Listen to keyboard events and handle cursor before the move.
 	emitter.listenTo( view.document, 'keydown', ( evt, data ) => {
-		const arrowRightPressed = data.keyCode == keyCodes.arrowright;
-		const arrowLeftPressed = data.keyCode == keyCodes.arrowleft;
-
-		// When neither left or right arrow has been pressed then do noting.
-		if ( !arrowRightPressed && !arrowLeftPressed ) {
-			return;
-		}
-
 		// This implementation works only for collapsed selection.
 		if ( !modelSelection.isCollapsed ) {
 			return;
@@ -62,39 +53,60 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
 			return;
 		}
 
+		const arrowRightPressed = data.keyCode == keyCodes.arrowright;
+		const arrowLeftPressed = data.keyCode == keyCodes.arrowleft;
+
+		// When neither left or right arrow has been pressed then do noting.
+		if ( !arrowRightPressed && !arrowLeftPressed ) {
+			return;
+		}
+
 		const position = modelSelection.getFirstPosition();
 
 		if ( arrowRightPressed ) {
-			handleArrowRightPress( position, data );
+			twoStepCaretHandler.handleForwardMovement( position, data );
 		} else {
-			handleArrowLeftPress( position, data );
+			twoStepCaretHandler.handleBackwardMovement( position, data );
 		}
 	} );
+}
 
-	emitter.listenTo( modelSelection, 'change:range', ( evt, data ) => {
-		if ( skipNextChangeRange ) {
-			skipNextChangeRange = false;
+class TwoStepCaretHandler {
+	constructor( model, emitter, attribute ) {
+		this.model = model;
+		this.attribute = attribute;
 
-			return;
-		}
+		this._modelSelection = model.document.selection;
+		this._overrideUid = null;
+		this._skipNextChangeRange = false;
 
-		// Skip automatic restore when the gravity is not overridden — simply, there's nothing to restore
-		// at this moment.
-		if ( !overrideUid ) {
-			return;
-		}
+		emitter.listenTo( this._modelSelection, 'change:range', ( evt, data ) => {
+			if ( this._skipNextChangeRange ) {
+				this._skipNextChangeRange = false;
 
-		// Skip automatic restore when the change is indirect AND the selection is at the attribute boundary.
-		// It means that e.g. if the change was external (collaboration) and the user had their
-		// selection around the link, its gravity should remain intact in this change:range event.
-		if ( !data.directChange && isAtBoundary( modelSelection.getFirstPosition(), attribute ) ) {
-			return;
-		}
+				return;
+			}
 
-		restoreGravity( model );
-	} );
+			// Skip automatic restore when the gravity is not overridden — simply, there's nothing to restore
+			// at this moment.
+			if ( !this._isGravityOverridden ) {
+				return;
+			}
 
-	function handleArrowRightPress( position, data ) {
+			// Skip automatic restore when the change is indirect AND the selection is at the attribute boundary.
+			// It means that e.g. if the change was external (collaboration) and the user had their
+			// selection around the link, its gravity should remain intact in this change:range event.
+			if ( !data.directChange && isAtBoundary( this._modelSelection.getFirstPosition(), attribute ) ) {
+				return;
+			}
+
+			this._restoreGravity();
+		} );
+	}
+
+	handleForwardMovement( position, data ) {
+		const attribute = this.attribute;
+
 		// DON'T ENGAGE 2-SCM if gravity is already overridden. It means that we just entered
 		//
 		// 		<paragraph>foo{}<$text attribute>bar</$text>baz</paragraph>
@@ -104,7 +116,7 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
 		// 		<paragraph>foo<$text attribute>bar</$text>{}baz</paragraph>
 		//
 		// and the gravity will be restored automatically.
-		if ( overrideUid ) {
+		if ( this._isGravityOverridden ) {
 			return;
 		}
 
@@ -114,7 +126,7 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
 		//
 		//		<paragraph><$text attribute>{}bar</$text>baz</paragraph>
 		//
-		if ( position.isAtStart && modelSelection.hasAttribute( attribute ) ) {
+		if ( position.isAtStart && this._hasAttribute ) {
 			return;
 		}
 
@@ -126,17 +138,17 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
 		//
 		// 		<paragraph><$text attribute="1">foo</$text>{}<$text attribute="2">bar</$text></paragraph>
 		//
-		if ( isBetweenDifferentValues( position, attribute ) && modelSelection.hasAttribute( attribute ) ) {
-			preventCaretMovement( data );
-			removeSelectionAttribute( model );
+		if ( isBetweenDifferentValues( position, attribute ) && this._hasAttribute ) {
+			this._preventCaretMovement( data );
+			this._removeSelectionAttribute();
 		} else {
 			// ENGAGE 2-SCM when entering an attribute:
 			//
 			// 		<paragraph>foo{}<$text attribute>bar</$text>baz</paragraph>
 			//
 			if ( isAtStartBoundary( position, attribute ) ) {
-				preventCaretMovement( data );
-				overrideGravity( model );
+				this._preventCaretMovement( data );
+				this._overrideGravity();
 
 				return;
 			}
@@ -145,16 +157,18 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
 			//
 			//		<paragraph>foo<$text attribute>bar{}</$text>baz</paragraph>
 			//
-			if ( isAtEndBoundary( position, attribute ) && modelSelection.hasAttribute( attribute ) ) {
-				preventCaretMovement( data );
-				overrideGravity( model );
+			if ( isAtEndBoundary( position, attribute ) && this._hasAttribute ) {
+				this._preventCaretMovement( data );
+				this._overrideGravity();
 			}
 		}
 	}
 
-	function handleArrowLeftPress( position, data ) {
+	handleBackwardMovement( position, data ) {
+		const attribute = this.attribute;
+
 		// When the gravity is already overridden...
-		if ( overrideUid ) {
+		if ( this._isGravityOverridden ) {
 			// ENGAGE 2-SCM & REMOVE SELECTION ATTRIBUTE
 			// when about to leave one attribute value and enter another:
 			//
@@ -164,10 +178,10 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
 			//
 			// 		<paragraph><$text attribute="1">foo</$text>{}<$text attribute="2">bar</$text></paragraph>
 			//
-			if ( isBetweenDifferentValues( position, attribute ) && modelSelection.hasAttribute( attribute ) ) {
-				preventCaretMovement( data );
-				restoreGravity( model );
-				removeSelectionAttribute( model );
+			if ( isBetweenDifferentValues( position, attribute ) && this._hasAttribute ) {
+				this._preventCaretMovement( data );
+				this._restoreGravity();
+				this._removeSelectionAttribute();
 			}
 
 			// ENGAGE 2-SCM when at any boundary of the attribute:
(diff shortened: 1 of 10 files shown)


Mgsy:
## Steps to reproduce
1. Open the article sample.
1. Select the last character in the `Paragraph` word and apply the bold.
1. Put the caret before the bolded character.
1. Insert a link.
1. Put the caret inside the link and press <kbd>Enter</kbd>.
1. Undo.
1. Press <kbd>Arrow left</kbd> few times.

## Current result
The caret is stuck.

## GIF
[Click](https://user-images.githubusercontent.com/20988892/38809482-ea3c8aae-4183-11e8-8eb9-2759128cae61.gif)

It's a regression.

Reinmar:
@scofalik, could you review this?

scofalik:
AFAIK @oleq is fighting with a nasty bug ATM.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
+		// 		<paragraph><$text attribute="1">foo{}</$text><$text attribute="2">bar</$text></paragraph>
+		//
+		// but DON'T when already in between of them (no attribute selection):
+		//
+		// 		<paragraph><$text attribute="1">foo</$text>{}<$text attribute="2">bar</$text></paragraph>
+		//
+		if ( isBetweenDifferentValues( position, attribute ) && this._hasAttribute ) {
+			this._preventCaretMovement( data );
+			this._removeSelectionAttribute();
+		} else {
```
scofalik: Is this `else` needed here?

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
-		const position = modelSelection.getFirstPosition();
+		// ENGAGE 2-SCM when about to leave one attribute value and enter another:
+		//
+		// 		<paragraph><$text attribute="1">foo{}</$text><$text attribute="2">bar</$text></paragraph>
+		//
+		// but DON'T when already in between of them (no attribute selection):
+		//
+		// 		<paragraph><$text attribute="1">foo</$text>{}<$text attribute="2">bar</$text></paragraph>
+		//
+		if ( isBetweenDifferentValues( position, attribute ) && this._hasAttribute ) {
```
scofalik: Wouldn't this condition be met in this case:

```html
<paragraph><$text attribute="1">foo</$text><$text attribute="2">{}bar</$text>
```

oleq:
> Open the article sample.
> Put the caret at the end of the Paragraph word.
> Open the link balloon, type one character, i.e. x and add it.
> Put the caret at the end of the link.
> Press Arrow left.

It's not a bug. Setting selection using the mouse in a boundary situation (link is first/last in the block) will always put it inside the link.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
+// @param {module:engine/model/position~Position} position
+// @param {String} attribute
+function isAtStartBoundary( position, attribute ) {
+	const prevNode = position.nodeBefore;
+	const nextNode = position.nodeAfter;
 	const isAttrInNext = nextNode ? nextNode.hasAttribute( attribute ) : false;
 	const isAttrInPrev = prevNode ? prevNode.hasAttribute( attribute ) : false;
 
-	if ( isAttrInNext && isAttrInPrev && nextNode.getAttributeKeys( attribute ) !== prevNode.getAttribute( attribute ) ) {
+	if ( ( !isAttrInPrev && isAttrInNext ) || isBetweenDifferentValues( position, attribute ) ) {
```
scofalik: I am not sure if I like `isBetweenDifferentValues` here, because it repeats some code from before. Maybe

`isAttrInNext && ( !isAttrInPrev || prevNode.getAttribute( attribute ) != nextNode.getAttribute( attribute ) )` ?

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
+
+// @param {module:engine/model/position~Position} position
+// @param {String} attribute
+function isAtEndBoundary( position, attribute ) {
+	const prevNode = position.nodeBefore;
+	const nextNode = position.nodeAfter;
+	const isAttrInNext = nextNode ? nextNode.hasAttribute( attribute ) : false;
+	const isAttrInPrev = prevNode ? prevNode.hasAttribute( attribute ) : false;
+
+	if ( ( isAttrInPrev && !isAttrInNext ) || isBetweenDifferentValues( position, attribute ) ) {
```
scofalik: Same as above.

src/utils/bindtwostepcarettoattribute.js
scofalik: Okay I guess that in this scenario gravity is already overridden, so it would return earlier. If you think it is worthy to left a note about this, please do so. If not, it's fine.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
+			//
+			// 		<paragraph><$text attribute="1">foo</$text><$text attribute="2">{}bar</$text></paragraph>
+			//
+			// but DON'T when already in between of them (no attribute selection):
+			//
+			// 		<paragraph><$text attribute="1">foo</$text>{}<$text attribute="2">bar</$text></paragraph>
+			//
+			if ( isBetweenDifferentValues( position, attribute ) && this._hasAttribute ) {
+				this._preventCaretMovement( data );
+				this._restoreGravity();
```
scofalik: I wonder whether the selection should be restored in case. I guess there's a difference when this would be your model:

```html
<paragraph><$text attribute="1" bold="true">foo</$text><$text attribute="2">{}bar</$text></paragraph>
```

Do we want the bold or not.

scofalik: I don't say that current solution is incorrect, I just wonder.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
+				this._preventCaretMovement( data );
+				this._setSelectionAttributeFromTheNodeBefore( position );
 
 				return;
 			}
 
-			// If we are here we need to check if caret is a one character before the text with attribute bound
-			// `foo<a>bar</a>b{}iz` or `foo<a>b{}ar</a>biz`.
-			const nextPosition = position.getShiftedBy( -1 );
+			// DON'T ENGAGE 2-SCM if gravity is already overridden. It means that we have already entered
```
scofalik: Outdated comment?

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
+					return;
+				}
+			}
+
+			// REMOVE SELECTION ATRIBUTE when restoring gravity towards a non-existent content at the
+			// beginning of the block.
+			//
+			// 		<paragraph>{}<$text attribute>bar</$text></paragraph>
+			//
+			if ( position.isAtStart && isAtBoundary( position, attribute ) ) {
```
scofalik: I am not sure if `isAtBoundary` is needed here if we check `this._hasAttribute` anyway.

scofalik: BTW. wouldn't using `isAtStartBoundary` more precise and a tiny little bit faster? :trollface:

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
+		} );
+	}
+
+	/**
+	 * Prevents the caret movement in the view by calling `preventDefault` on the event data.
+	 *
+	 * @private
+	 */
+	_preventCaretMovement( data ) {
+		data.preventDefault();
```
scofalik: Did you give it a thought to also cancel the event? I wonder if canceling the event wouldn't be correct. Of course, it works now as it is, so we don't have to change it, but IDK. We kind of cancel the event by handling it in quite a custom way so IDK. Both canceling and not canceling may create issues in future.

Reinmar: You can verify how important this is by checking how this works next to widgets which also try to handle left/right arrow keys ;)

Reinmar: PS. I think we should cancel the event and make sure that 2scm handles it.

scofalik:
## Review

I finished reviewing the code state as it is now. AFAIK @oleq is still fighting with some bugs but there is progress in this matter. After the bugs are fixed we could merge this PR.

This saying, in my opinion, this solution is too complicated at this point. I think there are two causes for that:

1. @oleq inherited this code from @oskarwrobel and there was not much time to rethink the architecture of this solution.
2. We had too few cases on our hands.

As much as I am for merging this PR (since it will make the whole feature more stable) I also think that we should discuss this feature once again, having the full view and discovering some cases that we didn't think about at the beginning.

## Proposal

Personally, I think that we might be better dropping "override/restore" gravity as it brings a lot of complications. Dropping this solution and managing attributes directly through writer API may simplify this feature. Additionally, we will get rid of weird situations, where gravity overriding might do more than we'd like to, for example:

`<$text linkHref="foo.html">foo{}</$text><$text bold="true">bar</$text>`

If we override gravity to the ride, we start typing bold. Is it correct? Maybe - I am not saying it is not. But it is more mess and potential problems.

This feature seems like it shouldn't really be that difficult, though I guess that we made it too complicated by using wrong tool (gravity) to approach it.

I have a feeling that there are just three things that should be taken into consideration:

1. Does the selection has the attribute?
2. Will the selection "leave the attribute"?
3. Will the selection "enter the attribute"?

Which leads to those three scenarios:

1. Has attribute + is leaving -> Remove the attribute, prevent caret move.
2. Doesn't have attribute + is entering -> Add the attribute, prevent caret move.
3. If approaching from the right side, remove the attribute when approaching the attribute (because the gravity is to the left). _This however is tricky*_.

### How would it work for different cases?

**Between attributes:**

`<$text attribute="1">fo{}o</$text><$text attribute="2">bar</$text>`

-> not leaving or entering
`<$text attribute="1">foo{}</$text><$text attribute="2">bar</$text>`

-> has attribute + is leaving, remove attribute, prevent caret move
`<$text attribute="1">foo</$text>{}<$text attribute="2">bar</$text>`

-> does not have attribute + is entering, add attribute, prevent caret move
`<$text attribute="1">foo</$text><$text attribute="2">{}bar</$text>`

-> not leaving or entering
`<$text attribute="1">foo</$text><$text attribute="2">b{}ar</$text>`

<- not leaving or entering
`<$text attribute="1">foo</$text><$text attribute="2">{}bar</$text>`

<- has attribute + is leaving, remove attribute, prevent caret move
`<$text attribute="1">foo</$text>{}<$text attribute="2">bar</$text>`

<- does not have attribute + is entering, add attribute, prevent caret move
`<$text attribute="1">foo{}</$text><$text attribute="2">bar</$text>`

<- not leaving or entering
`<$text attribute="1">fo{}o</$text><$text attribute="2">bar</$text>`

**Inside multiple attributes using 2SCM:**
_The trick is to handle each attribute separately and cancel handling if one of them was handled. So the `keydown` event should be stopped after 2SCM is used._

`<$text a="1" b="2">fo{}o</$text>bar`

-> not leaving or entering
`<$text a="1" b="2">foo{}</$text>bar`

-> has attribute `a` + is leaving, remove attribute, prevent caret move
`<$text a="1" b="2">foo</$text><$text b="2">{}</$text>bar`

-> has attribute `b` + is leaving, remove attribute, prevent caret move
`<$text a="1" b="2">foo</$text>{}bar`

-> not leaving or entering
`<$text a="1" b="2">foo</$text>b{}ar`

**The beginning of an element:**

`<paragraph><$text a="1">f{}oo</$text></paragraph>`

<- not leaving or entering
`<paragraph><$text a="1">{}foo</$text></paragraph>`

<- has attribute + is leaving, remove attribute, prevent caret move
`<paragraph>{}<$text a="1">foo</$text></paragraph>`

**The end of an element:**

`<paragraph><$text a="1">fo{}o</$text></paragraph>`

-> not leaving or entering
`<paragraph><$text a="1">foo{}</$text></paragraph>`

-> has attribute + is leaving, remove attribute, prevent caret move
`<paragraph><$text a="1">foo</$text>{}</paragraph>`

**Approaching from the right side:**

`<paragraph><$text a="1">foo</$text>b{}ar</paragraph>`

<- approaching from the right side, remove attribute `a`, remove attribute `b`
`<paragraph><$text a="1" b="2">foo</$text>{}bar</paragraph>`

<- does not have attribute `a` + is entering, add attribute, prevent caret move
`<paragraph><$text a="1" b="2">foo</$text><$text a="1">{}</$text>bar</paragraph>`

<- does not have attribute `b` + is entering, add attribute, prevent caret move
`<paragraph><$text a="1" b="2">foo{}</$text>bar</paragraph>`

<- not leaving or entering
`<paragraph><$text a="1" b="2">fo{}o</$text>bar</paragraph>`

### Other questions regarding the solution

**How to check "is leaving" and "is entering"?**

This seems easy. We already check if the selection is at attribute boundary so we would just add checking if proper key was pressed.

**Clearing attributes set by 2SCM?**

There's one question about clearing attributes set by 2SCM. However maybe we don't need to do anything?

When the selection is moved through API (key press, mouse click, `directChange == true`) then the attributes are cleared anyway. This means that we don't need to care about additional clearing.
* - However this messes up our solution when the caret approaches element from the right, because the attribute would be removed first and then selection would clear that. So for that case we would need more additional handling. I believe it is doable, though.

When the selection is moved because of changes to model tree (changes in background, collaboration, `directChange == false`) we don't want those changes to affect selection anyway. The only exception would be if the selection is "removed" or the attribute is removed/changed.

scofalik:
Now, as I am thinking about the solution I describe, I got an idea, that maybe we could implement gravity managing as an additional separate feature. It would be much simpler. Basically, if a user clicks left arrow key, gravity is set to right, and when a user clicks right arrow key, gravity is set to right.

This would cover two things:

1. We would not have the described problem in "approach from the right side" scenario.
2. We (maybe) would give users a better UX for attributes for which we don't want a 2SCM.

How would it work?

`fo{}o<$text bold="true">bold</$text>bar`

-> keep gravity to the left
`foo{}<$text bold="true">bold</$text>bar`

-> keep gravity to the left
`foo<$text bold="true">b{}old</$text>bar`

<- keep gravity to the right
`foo<$text bold="true">{}bold</$text>bar`

<- keep gravity to the right
`fo{}o<$text bold="true">bold</$text>bar`

The gravity would be also reset on `change:range` with `directChange` set to `true`.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -49,93 +108,425 @@ export default function bindTwoStepCaretToAttribute( view, model, emitter, attri
...
-		// When user tries to expand selection or jump over the whole word or to the beginning/end then
-		// two-steps movement is not necessary.
-		if ( data.shiftKey || data.altKey || data.ctrlKey ) {
+		// DON'T ENGAGE 2-SCM when the selection is at the beginning of an attribute AND already has it:
+		// * when the selection was initially set there using the mouse,
+		// * when the editor has just started
+		//
+		//		<paragraph><$text attribute>{}bar</$text>baz</paragraph>
+		//
+		if ( position.isAtStart && this._hasAttribute ) {
```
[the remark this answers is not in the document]
oleq: Bug in docs, OFC.

src/utils/bindtwostepcarettoattribute.js
oleq: In this case

```
<$text attribute="1" bold="true">foo</$text><$text attribute="2">{}bar</$text>
```

<kbd>←</kbd>: the selection has `bold="true"` only (2-SCM active for `attribute`). It's a typing between two attribute boundaries of a different value case.

<kbd>←</kbd>: the selection has `attribute="1" bold="true"`

If the selection lost `bold="true"` in the first <kbd>←</kbd> step I'd personally find it confusing. Is that what you wanted to discuss?

src/utils/bindtwostepcarettoattribute.js
Reinmar: I love the second point. Can we use `priorities.high + 1`? Does `on()` allow specifying number still? I think it does because it should be using `priorities.get()`.

scofalik: Well, that's what priorities are for!

src/utils/bindtwostepcarettoattribute.js
scofalik: I wonder if the selection should have bold after the first left-arrow press.

scofalik: But this is nothing critical.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -496,44 +530,33 @@ function isAtBoundary( position, attribute ) {
...
-	const prevNode = position.nodeBefore;
-	const nextNode = position.nodeAfter;
-	const isAttrInNext = nextNode ? nextNode.hasAttribute( attribute ) : false;
-	const isAttrInPrev = prevNode ? prevNode.hasAttribute( attribute ) : false;
-
-	if ( ( !isAttrInPrev && isAttrInNext ) || isBetweenDifferentValues( position, attribute ) ) {
-		return true;
-	}
+	const { nodeBefore, nodeAfter } = position;
+	const isAttrBefore = nodeAfter ? nodeAfter.hasAttribute( attribute ) : false;
```
scofalik: I find those names confusing now. `isAttrBefore` checks `nodeAfter`. What does `isAttrBefore` mean? If anything, I'd suggest `isBeforeAttr`.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -496,44 +530,33 @@ function isAtBoundary( position, attribute ) {
...
-	const prevNode = position.nodeBefore;
-	const nextNode = position.nodeAfter;
-	const isAttrInNext = nextNode ? nextNode.hasAttribute( attribute ) : false;
-	const isAttrInPrev = prevNode ? prevNode.hasAttribute( attribute ) : false;
-
-	if ( ( isAttrInPrev && !isAttrInNext ) || isBetweenDifferentValues( position, attribute ) ) {
-		return true;
-	}
+	const { nodeBefore, nodeAfter } = position;
+	const isAttrBefore = nodeAfter ? nodeAfter.hasAttribute( attribute ) : false;
```
scofalik: As above.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -496,44 +530,33 @@ function isAtBoundary( position, attribute ) {
...
 
 // @param {module:engine/model/position~Position} position
 // @param {String} attribute
 function isBetweenDifferentValues( position, attribute ) {
-	const prevNode = position.nodeBefore;
-	const nextNode = position.nodeAfter;
-	const isAttrInNext = nextNode ? nextNode.hasAttribute( attribute ) : false;
-	const isAttrInPrev = prevNode ? prevNode.hasAttribute( attribute ) : false;
+	const { nodeBefore, nodeAfter } = position;
+	const isAttrBefore = nodeAfter ? nodeAfter.hasAttribute( attribute ) : false;
```
scofalik: As above.

src/utils/bindtwostepcarettoattribute.js
```diff
@@ -380,6 +397,8 @@ class TwoStepCaretHandler {
 			if ( position.isAtStart ) {
 				if ( this._hasSelectionAttribute ) {
 					this._removeSelectionAttribute();
```
scofalik: Why aren't we preventing caret move in this case? 🤔

src/utils/bindtwostepcarettoattribute.js
oleq: My bad. I changed it automatically and somehow make a mistake.

src/utils/bindtwostepcarettoattribute.js
oleq: Would that matter in this case? It's a start of the block anyway and we're moving left.

scofalik: Doesn't it matter in case of `<paragraph>foo</paragraph><paragraph><$text a="1">{}foo</$text></paragraph>`? Won't we get to the previous paragraph if we won't stop caret movement?

scofalik: I tested this in manual test and for some reason it works 🤷‍♂️

oleq: Hm, you're right. It should be prevented. But the funny thing is that it works. As long as the `keydown` event is stopped.

scofalik: lol :)

apr-18-2018 16-01-40

scofalik:
Things I found in manual test (http://localhost:8125/ckeditor5-engine/tests/manual/two-step-caret.html):

1. `[]<b><i>biz</i></b>` after first right-arrow keypress becomes `<b><i>[]biz</i></b>` and then additional keypress does nothing.

2. Minor thing, in one scenario a button on toolbar flickers:
apr-18-2018 16-06-56

3. This scenario from manual test fails, however I don't know how important it is, really:
```
Attributes set explicit
Press right arrow once again
selection should be at the same position
underline button should be not selected
(*) bold button should stay selected
```

4. I don't like that there's no intermediate step between `<u>bar</u><i>biz</i>`.

5. Filler bug is back:
apr-18-2018 16-20-28

6. You were supposed to merge http://localhost:8125/ckeditor5-engine/tests/manual/tickets/1301/1.html with http://localhost:8125/ckeditor5-engine/tests/manual/two-step-caret.html

7. I miss an image in http://localhost:8125/ckeditor5-engine/tests/manual/two-step-caret.html test to see how 2SCM works with widgets.

oleq:
> []<b><i>biz</i></b> after first right-arrow keypress becomes <b><i>[]biz</i></b> and then additional keypress does nothing.

I'm unable to reproduce it.

> Minor thing, in one scenario a button on toolbar flickers:

It's because the engine sets attributes and we remove them :/

> This scenario from manual test fails, however I don't know how important it is, really:

We sacrificed this when choosing `_refreshAttributes()` over `_updateAttributes` in `LiveSelection`.

> I don't like that there's no intermediate step between <u>bar</u><i>biz</i>.

Me too. But I'm afraid there's no easy way to do that. Different `bindTwoStepCaretToAttribute` "instances" would need to talk to each other to do that. ATM when we leave (arrow right) `<u>bar[]</u><i>biz</i>` we prevent the movement and override gravity (`u` attribute handler does that). To create that additional "empty" step we'd need to remove the selection attribute (`i`) too at the same time. And it's up to the `i` attribute handler`. It would need to know what happened in the `u` handler to approach the situation properly.

Reinmar:
apr-18-2018 17-27-43

Reinmar:
OK, the above is not caused by 2scm.

PR #1413, opened 2018-04-19
Introduced "ElementDefinition#priority"

### LINKED ISSUE #1408: Conversion: Allows specyfing the priority for view element
When I was working on https://github.com/ckeditor/ckeditor5-highlight/issues/17 I noticed that there is no option for specifying the priority for the view element that will be created.

It means we cannot use two-way conversion methods like `elementToElement()`, `attributeToElement()` or `attributeToAttribute()`. Instead of we have to care about `upcast` and `downcast` conversion manually. It means we cannot use the [helpers](https://github.com/ckeditor/ckeditor5-engine/blob/23afe78d736d9e90046835e06f3112b283777c8f/src/conversion/conversion.js#L600-L629) that are using in two-way conversion. 

Would be nice to have an option for that. It simplify changes the priority element. We will have a single property instead of [the whole block of code (which does not care about `upcastAlso` option](https://github.com/ckeditor/ckeditor5-highlight/blob/1f4bd3bb023b62d5ed613812e7d826ea4101499a/src/highlightediting.js#L103-L150).

pomek:

Feature: Introduce `ElementDefinition#priority` property which allows specifying the priority of created element during the downcast conversion. Closes #1408.

BREAKING CHANGE: In order to avoid collision with the `engine/view/elementdefinition~ElementDefinition#priority` property, the conversion priority option has been renamed from `priority` to `converterPriority`.

---

Should be merged along with https://github.com/ckeditor/ckeditor5-heading/pull/107.

### DIFF
--- a/src/conversion/conversion.js
+++ b/src/conversion/conversion.js
@@ -164,7 +164,7 @@ export default class Conversion {
 	 *		conversion.elementToElement( { model: 'paragraph', view: 'p' } );
 	 *
 	 *		// Override other converters by specifying converter definition with higher priority.
-	 *		conversion.elementToElement( { model: 'paragraph', view: 'div', priority: 'high' } );
+	 *		conversion.elementToElement( { model: 'paragraph', view: 'div', converterPriority: 'high' } );
 	 *
 	 *		// View specified as an object instead of a string.
 	 *		conversion.elementToElement( {
@@ -237,7 +237,7 @@ export default class Conversion {
 				upcastElementToElement( {
 					model,
 					view,
-					priority: definition.priority
+					converterPriority: definition.converterPriority
 				} )
 			);
 		}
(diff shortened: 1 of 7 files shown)


scofalik:
Other than that I am fine with those changes.

pomek:
My editor shows me some outdated files and I didn't change everything. I'm giving R-.

src/conversion/conversion.js
```diff
@@ -579,7 +579,7 @@ export default class Conversion {
  * Any view element matching `upcastAlso` will also be converted to model. If `model` describes multiple values, `upcastAlso`
  * is an object that assigns those values (`upcastAlso` object keys) to {@link module:engine/view/matcher~MatcherPattern}s
  * (`upcastAlso` object values).
- * @property {module:utils/priorities~PriorityString} [priority] Conversion priority.
+ * @property {module:utils/priorities~PriorityString} [converterPriority] Conversion priority.
```
scofalik: I would change the param description to "Converter priority". Looks weird now.

PR #1421, opened 2018-05-29
Introduce `Document#change:data` event

### LINKED ISSUE #1418: Mark model changes which do not affect data
There are model changes which do not affect data: selection change, some markers. At the moment, there is no easy way to learn if the change touches data what makes integrations much harder.

First, we should introduce marker API to tell if that marker change touches data. I propose to call it `affectsData` (`true` by default). It should be a property in `options` parameter in `addMarker` and `updateMarker`. If the value is set it should be the same for all further operations, as long as it is not changed.

Then, we need to have an easy way to get only these `change` events which affect data. I propose to introduce `change:data` event. It will be called for all changes which affect data. If the change will contain only selection change, or markers which do not affect data, then `change` event should be fired. In all other cases, when data can change `change:data` should be fired.

ma2ciek:

Feature: Introduced `ModelDocument#change:data` event. Closes #1418.

---

### Additional information

### DIFF
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -189,6 +189,7 @@ export default class Differ {
 	 * @param {module:engine/model/range~Range|null} oldRange Marker range before the change or `null` if the marker has just
 	 * been created.
 	 * @param {module:engine/model/range~Range|null} newRange Marker range after the change or `null` if the marker was removed.
+	 * @param {Boolean} affectsData Flag indicating whether marker affects the data model.
 	 */
 	bufferMarkerChange( markerName, oldRange, newRange, affectsData ) {
 		const buffered = this._changedMarkers.get( markerName );
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -247,11 +247,11 @@ export default class Differ {
 	}
 
 	/**
-	 * Checks whether some of buffered marker affects the data model or whether some element will change.
+	 * Checks whether some of buffered marker affects the data model or whether some element changed.
 	 *
-	 * @returns {Boolean} `true` if buffered markers or changes in elements will affect the data model.
+	 * @returns {Boolean} `true` if buffered markers or changes in elements affects the data model.
 	 */
-	willDataChange() {
+	hasDataChanges() {
 		for ( const [ , change ] of this._changedMarkers ) {
 			if ( change.affectsData ) {
 				return true;
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -190,13 +190,14 @@ export default class Differ {
 	 * been created.
 	 * @param {module:engine/model/range~Range|null} newRange Marker range after the change or `null` if the marker was removed.
 	 */
-	bufferMarkerChange( markerName, oldRange, newRange ) {
+	bufferMarkerChange( markerName, oldRange, newRange, affectsData ) {
 		const buffered = this._changedMarkers.get( markerName );
 
 		if ( !buffered ) {
 			this._changedMarkers.set( markerName, {
 				oldRange,
-				newRange
+				newRange,
+				affectsData
 			} );
 		} else {
 			buffered.newRange = newRange;
@@ -241,6 +242,21 @@ export default class Differ {
 		}
 
 		return result;
+	}
+
+	/**
+	 * Checks whether some of buffered marker can affect the data model.
+	 *
+	 * @returns {Boolean} `true` if buffered markers can change the data model.
+	 */
+	containsMarkersAffectingData() {
+		for ( const [ , change ] of this._changedMarkers ) {
+			if ( change.affectsData ) {
+				return true;
+			}
+		}
+
+		return false;
 	}
 
 	/**
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -247,18 +247,19 @@ export default class Differ {
 	}
 
 	/**
-	 * Checks whether some of buffered marker can affect the data model.
+	 * Checks whether some of buffered marker affects the data model or whether some element will change.
 	 *
-	 * @returns {Boolean} `true` if buffered markers can change the data model.
+	 * @returns {Boolean} `true` if buffered markers or changes in elements will affect the data model.
 	 */
-	containsMarkerAffectingData() {
+	willDataChange() {
 		for ( const [ , change ] of this._changedMarkers ) {
 			if ( change.affectsData ) {
 				return true;
 			}
 		}
 
-		return false;
+		// If markers do not affect the data model, check whether there are some changes in elements.
+		return this._changesInElement.size > 0;
 	}
 
 	/**
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -189,7 +189,7 @@ export default class Differ {
 	 * @param {module:engine/model/range~Range|null} oldRange Marker range before the change or `null` if the marker has just
 	 * been created.
 	 * @param {module:engine/model/range~Range|null} newRange Marker range after the change or `null` if the marker was removed.
-	 * @param {Boolean} affectsData Flag indicating whether marker affects the data model.
+	 * @param {Boolean} affectsData Flag indicating whether marker affects the editor data.
 	 */
 	bufferMarkerChange( markerName, oldRange, newRange, affectsData ) {
 		const buffered = this._changedMarkers.get( markerName );
@@ -247,9 +247,9 @@ export default class Differ {
 	}
 
 	/**
-	 * Checks whether some of buffered marker affects the data model or whether some element changed.
+	 * Checks whether some of buffered marker affects the editor data or whether some element changed.
 	 *
-	 * @returns {Boolean} `true` if buffered markers or changes in elements affects the data model.
+	 * @returns {Boolean} `true` if buffered markers or changes in elements affects the editor data.
 	 */
 	hasDataChanges() {
 		for ( const [ , change ] of this._changedMarkers ) {
@@ -258,7 +258,7 @@ export default class Differ {
 			}
 		}
 
-		// If markers do not affect the data model, check whether there are some changes in elements.
+		// If markers do not affect the data, check whether there are some changes in elements.
 		return this._changesInElement.size > 0;
 	}
 
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -171,7 +171,7 @@ export default class Differ {
 				for ( const marker of this._markerCollection.getMarkersIntersectingRange( range ) ) {
 					const markerRange = marker.getRange();
 
-					this.bufferMarkerChange( marker.name, markerRange, markerRange );
+					this.bufferMarkerChange( marker.name, markerRange, markerRange, marker.affectsData );
 				}
 
 				break;
@@ -201,6 +201,7 @@ export default class Differ {
 			} );
 		} else {
 			buffered.newRange = newRange;
+			buffered.affectsData = affectsData;
 
 			if ( buffered.oldRange == null && buffered.newRange == null ) {
 				// The marker is going to be removed (`newRange == null`) but it did not exist before the first buffered change
@@ -249,7 +250,7 @@ export default class Differ {
 	 *
 	 * @returns {Boolean} `true` if buffered markers can change the data model.
 	 */
-	containsMarkersAffectingData() {
+	containsMarkerAffectingData() {
 		for ( const [ , change ] of this._changedMarkers ) {
 			if ( change.affectsData ) {
 				return true;
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -152,7 +152,7 @@ export default class Document {
 			if ( !this.differ.isEmpty || hasSelectionChanged ) {
 				this._callPostFixers( writer );
 
-				if ( this.differ.willDataChange() ) {
+				if ( this.differ.hasDataChanges() ) {
 					this.fire( 'change:data', writer.batch );
 				} else {
 					this.fire( 'change', writer.batch );
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -152,7 +152,7 @@ export default class Document {
 			if ( !this.differ.isEmpty || hasSelectionChanged ) {
 				this._callPostFixers( writer );
 
-				if ( isBatchAffectingData( writer.batch ) || this.differ.containsMarkersAffectingData() ) {
+				if ( isBatchAffectingData( writer.batch ) || this.differ.containsMarkerAffectingData() ) {
 					this.fire( 'change:data', writer.batch );
 				} else {
 					this.fire( 'change', writer.batch );
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -147,7 +147,7 @@ export default class Document {
 		// Wait for `_change` event from model, which signalizes that outermost change block has finished.
 		// When this happens, check if there were any changes done on document, and if so, call post fixers,
 		// fire `change` event for features and conversion and then reset the differ.
-		// Fire `change:data` event when at least one operation changes the data model.
+		// Fire `change:data` event when at least one operation or buffered marker changes the data model.
 		this.listenTo( model, '_change', ( evt, writer ) => {
 			if ( !this.differ.isEmpty || hasSelectionChanged ) {
 				this._callPostFixers( writer );
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -385,7 +385,7 @@ export default class Document {
 	 *		} );
 	 *
 	 * If, however, you only want to be notified about the data changes, then use the
-	 * {@link module:engine/model/document~Document#event:change:data change:data event},
+	 * {@link module:engine/model/document~Document#event:change:data change:data} event,
 	 * which fires for document structure changes and marker changes (which affects the data).
 	 *
 	 *		model.document.on( 'change:data', () => {
@@ -397,18 +397,18 @@ export default class Document {
 	 */
 
 	/**
-	 * Fired when the data changes:
+	 * Fired when the data changes, which includes:
 	 * * document structure changes,
 	 * * marker changes (which affects the data).
 	 *
-	 * If you want to be notified about the data changes, then simply listen to this event:
+	 * If you want to be notified about the data changes, then listen to this event:
 	 *
 	 *		model.document.on( 'change:data', () => {
 	 *			console.log( 'The data has changed!' );
 	 *		} );
 	 *
 	 * If you would like to listen to all changes, then check the
-	 * {@link module:engine/model/document~Document#event:change change event}.
+	 * {@link module:engine/model/document~Document#event:change change} event.
 	 *
 	 * @event change:data
 	 * @param {module:engine/model/batch~Batch} batch The batch that was used in the executed changes block.
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -173,7 +173,7 @@ export default class Document {
 			if ( oldRange === null ) {
 				// If this is a new marker, add a listener that will buffer change whenever marker changes.
 				marker.on( 'change', ( evt, oldRange ) => {
-					this.differ.bufferMarkerChange( marker.name, oldRange, marker.getRange(), affectsData );
+					this.differ.bufferMarkerChange( marker.name, oldRange, marker.getRange(), marker.affectsData );
 				} );
 			}
 		} );
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -147,7 +147,7 @@ export default class Document {
 		// Wait for `_change` event from model, which signalizes that outermost change block has finished.
 		// When this happens, check if there were any changes done on document, and if so, call post fixers,
 		// fire `change` event for features and conversion and then reset the differ.
-		// Fire `change:data` event when at least one operation or buffered marker changes the data model.
+		// Fire `change:data` event when at least one operation or buffered marker changes the data.
 		this.listenTo( model, '_change', ( evt, writer ) => {
 			if ( !this.differ.isEmpty || hasSelectionChanged ) {
 				this._callPostFixers( writer );
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -384,13 +384,12 @@ export default class Document {
 	 *			console.log( 'The Document has changed!' );
 	 *		} );
 	 *
-	 * If, however, you only want to be notified about structure changes, then check whether the
-	 * {@link module:engine/model/differ~Differ differ} contains any changes:
+	 * If, however, you only want to be notified about the data changes, then use the
+	 * {@link module:engine/model/document~Document#event:change:data change:data event},
+	 * which fires for document structure changes and marker changes (which affects the data).
 	 *
-	 *		model.document.on( 'change', () => {
-	 *			if ( model.document.differ.getChanges().length > 0 ) {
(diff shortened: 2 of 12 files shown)


src/model/differ.js
```diff
@@ -243,6 +246,21 @@ export default class Differ {
 		return result;
 	}
 
+	/**
+	 * Checks whether some of buffered marker can affect the data model.
+	 *
+	 * @returns {Boolean} `true` if buffered markers can change the data model.
+	 */
+	containsMarkerAffectingData() {
```
Reinmar: Do we need it? I feel that we'll use it in one place. Instead, I feel that it'd be good to have one differ's method which will do all this: `isBatchAffectingData( writer.batch ) || this.differ.containsMarkerAffectingData()`.

src/model/markercollection.js
```diff
@@ -88,9 +88,10 @@ export default class MarkerCollection {
 	 * @param {String|module:engine/model/markercollection~Marker} markerOrName Name of marker to set or marker instance to update.
 	 * @param {module:engine/model/range~Range} range Marker range.
 	 * @param {Boolean} [managedUsingOperations=false] Specifies whether the marker is managed using operations.
+	 * @param {Boolean} [affectsData=true] Specifies whether the marker changes the data model.
```
Reinmar: Marker doesn't change anything itself. Its the presence of the marker that does it. Also, it's not about the data model. All kinds of markers are in the data model. It's about the persistence of markers. Maybe something like this:

> Specifies whether the marker affects the data produced by the data pipeline (is persisted in the editor's data).

This repeats a at least in two more place in this file.

src/model/markercollection.js
```diff
@@ -331,17 +338,25 @@ class Marker {
 		 */
 		this._managedUsingOperations = managedUsingOperations;
 
+		/**
+		 * Flag indicates if the marker changes the data model.
+		 *
+		 * @protected
```
Reinmar: Why did you make it protected? It does not need to be used in the tests (although, it is) AFAICS.

src/model/markercollection.js
```diff
@@ -382,7 +410,7 @@ class Marker {
 	}
 
 	/**
-	 * Returns a range that represents current state of marker.
+	 * Returns range that represents current state of marker.
```
Reinmar: "a range"
"the current state of the marker"

src/model/writer.js
```diff
@@ -794,6 +794,10 @@ export default class Writer {
 	 * {@link module:engine/model/markercollection~Marker marker class description} to learn about the difference between
 	 * markers managed by operations and not-managed by operations.
 	 *
+	 * The `options.affectData` parameter, which defaults to true, allows you to create marker, that doesn't affects the data model.
```
Reinmar:
```
defaults to `true`
```

No comma before "that"

"does not affect" (no "s")

src/model/writer.js
```diff
@@ -794,6 +794,10 @@ export default class Writer {
...
+	 * It impacts on performance, because changes in markers with set `options.affectData:false` don't cause firing the
```
Reinmar: It impacts the performance because...

do not cause

src/model/writer.js
```diff
@@ -802,14 +806,19 @@ export default class Writer {
 	 *
 	 *		addMarker( markerName, { range, usingOperation: true } );
 	 *
+	 * Create marker that doesn't affects the data model:
```
Reinmar: I started commenting on this so I'll continue – we don't use contractions in the docs ("doesn't" should be changed to "does not").

src/model/writer.js
```diff
@@ -802,14 +806,19 @@ export default class Writer {
...
 	 * Note: For efficiency reasons, it's best to create and keep as little markers as possible.
 	 *
 	 * @see module:engine/model/markercollection~Marker
 	 * @param {String} name Name of a marker to create - must be unique.
 	 * @param {Object} options
-	 * @param {Boolean} options.usingOperation Flag indicated whether the marker should be added by MarkerOperation.
+	 * @param {Boolean} options.usingOperation Flag indicating that the marker should be added by MarkerOperation.
 	 * See {@link module:engine/model/markercollection~Marker#managedUsingOperations}.
 	 * @param {module:engine/model/range~Range} options.range Marker range.
+	 * @param {Boolean} [options.affectsData=true] Flag indicating that the marker changes the data model.
```
Reinmar: Why does it default to `true`?

If we want it to default to `true`, we should think about changing its name so it can default to `false`.

Reinmar:
So far I haven't found anything big. The concept and implementation look generally good. But there a couple of things we should improve before merging this PR.

src/model/writer.js
ma2ciek: I remember that we couldn't think of a better name with PJ. The opposite to `affectsData` could be `transparent` (but that name is reserved) or smth like `notAffectingData`. If you have some proposals, I'm open to them :)

I agree that parameters that default to false are better, but on the other hand, descriptive parameter names are also important.

Reinmar: > Why does it default to true?

This question is still open ;)

src/model/differ.js
ma2ciek: What do you mean exactly? Because the `Differ` knows nothing about batches. I can check every operation that passes to the `bufferOperation()` and from that point, I can mark that some operation will change the data model. But I'm not sure if you mean this.

Reinmar: I don't know relations between Differ and batches, so I may not make much sense. But what I noticed is that you created a method which will only be used once in its lifetime. It's not very useful, it checks just one thing and is used to do a far more useful thing (check if the data will change), but itself... it's just useless.

So, we have two options:

1. Make it protected. 
2. Expose some useful functionality in the `Differ` – e.g. a method which could check if the data will change. I'm not sure if this is doable, but I thought that perhaps `isBatchAffectData()` could be moved here. Or maybe there's some other simple solution.

src/model/markercollection.js
ma2ciek: It's used to check whether the getter reflects its changes. I'll fix that test as for now it's artificial a bit. But I see, that live ranges are used in tests too few times and these tests would be hard to change, so I'll mark that property as protected.

src/model/differ.js
ma2ciek: I think that I can check differs' changed elements and if there is a changed element or marker that affects data, I can say that buffered changes affect data (I'll ask @scofalik tomorrow to be sure it's correct). But how to call that method? `doChangesAffectData`, `areChangesAffectingData`. I'm not sure about the batch in the method name. It looks like an expected parameter and the `Differ` doesn't use the `batch` or the `delta` terminology.

Reinmar: `willChangeData()`?

Reinmar: Or `willDataChange()`... not sure which sounds better.

ma2ciek: `willDataChange` sound good to me. Thanks!

ma2ciek:
I've asked @scofalik for the review of the Differ's usage in my implementation.

I'd stick with the `affectsData` as IMO it's a good and meaningful name.

Reinmar:
> I'd stick with the affectsData property as IMO it's a good and meaningful name.

Agree. It's a good name. But:

> > Why does it default to true?
>
> This question is still open ;)

ma2ciek:
> Why does it default to true? This question is still open ;)

Most of the markers potentially can have some impact on the produced data. Actually, the only two ones for now which do not affect the data model are `user:range` and `user:position`. So it should be up to the developer marking a marker as not affecting data.

Reinmar:
That's because we implemented such features so far. Consider e.g. spellchecking, all kinds of content quality checks, etc. I wouldn't say that the majority will affect data. I'd say that we don't know.

At the same time, this is a simple option to understand. And markers will not be used by many developers as this is a quite advanced feature. So I don't see a big problem with defaulting to `false` here.

cc @scofalik

src/model/differ.js
scofalik: AFAIC understand you want to have a `Differ` method that will tell you if anything in data changed? @ma2ciek you can do it as you said - check `_changesInElement.size`. For markers data you will need to iterate through `_changedMarkers` and see if there are any markers that change model. Pretty straightforward.

scofalik:
To be backward compatible I'd err on the side of having markers change data as the default. I'd rather change the name and has it default to `false` than keep it as is (the name and the default) or keep the name but default to `false` instead.

However, I agree that we don't know what will be the more common use and I agree that defaulting to `true` stinks :). 🤷‍♂️

ma2ciek:
Maybe we could name the opposite `dataTransparent`?

Reinmar:
It seems that I and @scofalik both are ok with `[affectsData=false]`? Why don't you like it, @ma2ciek?

> Maybe we could name the opposite `dataTransparent`?

`dataTransparent` is faaaar less understandable than `affectsData`.

scofalik:
@Reinmar - I don't like `[affectsData=true]` but I also dislike a bit `[affectsData=false]` as it is not backward compatible.

I don't think there's a good solution here though, cause it is hard to come up with something like "notAffectingData" but without using negation (negations in variable names are evil).

Few more proposals:

`transparentForData`
`notAffectingData`
`visualOnly`
`(for)editingOnly`

Reinmar:
> I also dislike a bit [affectsData=false] as it is not backward compatible.

How is it not backwards compatible? I thought we're adding a feature here (`change:data`). Old markers will still fire a simple `change` event, so nothing will change here.

In other words, we're adding a completely BC, opt-in feature here. Unless I miss something.

Reinmar:
> It is not backward compatible in a way that from now the markers that were already defined are "not affecting data".

I see your point. But you that know this is purely theoretical because, at this stage,  we know all the markers ever used in CKEditor 5 :D

scofalik:
Still, this is not backward compatible because markers that affect data will be marked as not affecting data. We need to change it. It will be incorrect.

AFAIR there are no markers used in CKEditor 5 features outside of collaboration, but still. It's not a lot of work, so if you go with `[affectsData=false]` please fix it where needed 🙏

ma2ciek:
It should be ready now. I'll change the markers in the `ckeditor5-collaboration` as a follow-up when we update the engine's hash in CF.

src/model/document.js
```diff
@@ -391,6 +396,28 @@ export default class Document {
...
+	 * was executed and the document has been changed during that block's execution.
+	 *
+	 * This event fires before the {@link module:engine/model/model~Model#change} event.
+	 *
+	 * The changes which this event will cover include:
+	 *
+	 * * document structure changes,
+	 * * marker changes, which affects the data model.
+	 *
+	 * If you want to be notified about the document's model changes, then simply listen to this event like this:
```
Reinmar: "document's model changes" makes no sense. If anything, this is "model's document".  Still, it doesn't make sense because you don't use this event to listen to document changes, but to listen to changes which affect the data. So to "data changes".

src/model/document.js
```diff
@@ -391,6 +396,28 @@ export default class Document {
...
+	 * Fired after each {@link module:engine/model/model~Model#enqueueChange `enqueueChange()` block},
+	 * which changes the data model or the outermost {@link module:engine/model/model~Model#change `change()` block}
+	 * was executed and the document has been changed during that block's execution.
+	 *
+	 * This event fires before the {@link module:engine/model/model~Model#change} event.
+	 *
+	 * The changes which this event will cover include:
+	 *
+	 * * document structure changes,
+	 * * marker changes, which affects the data model.
```
Reinmar: "marker changes (which affect the data)"

src/model/differ.js
```diff
@@ -243,6 +246,22 @@ export default class Differ {
 		return result;
 	}
 
+	/**
+	 * Checks whether some of buffered marker affects the data model or whether some element will change.
+	 *
+	 * @returns {Boolean} `true` if buffered markers or changes in elements will affect the data model.
+	 */
+	willDataChange() {
```
scofalik: I don't like the name of this function. I'd rather go with something like `hasDataChanges` or `dataChangeRegistered` or `registeredDataChanges`. IDK.

src/model/document.js
```diff
@@ -161,14 +166,14 @@ export default class Document {
 		// Buffer marker changes.
 		// This is not covered in buffering operations because markers may change outside of them (when they
 		// are modified using `model.markers` collection, not through `MarkerOperation`).
-		this.listenTo( model.markers, 'update', ( evt, marker, oldRange, newRange ) => {
+		this.listenTo( model.markers, 'update', ( evt, marker, oldRange, newRange, affectsData ) => {
```
scofalik: Do we need `affectsData` being passed in the event? Can't we simply use `marker.affectsData`?

src/model/document.js
```diff
@@ -391,6 +396,28 @@ export default class Document {
 	 * @event change
 	 * @param {module:engine/model/batch~Batch} batch The batch that was used in the executed changes block.
 	 */
+
+	/**
+	 * Fired after each {@link module:engine/model/model~Model#enqueueChange `enqueueChange()` block},
+	 * which changes the data model or the outermost {@link module:engine/model/model~Model#change `change()` block}
+	 * was executed and the document has been changed during that block's execution.
+	 *
+	 * This event fires before the {@link module:engine/model/model~Model#change} event.
```
Reinmar: Is this true? Aren't listeners for both these events combined together?

src/model/document.js
```diff
@@ -391,6 +396,28 @@ export default class Document {
...
+	 *
+	 * The changes which this event will cover include:
+	 *
+	 * * document structure changes,
+	 * * marker changes, which affects the data model.
+	 *
+	 * If you want to be notified about the document's model changes, then simply listen to this event like this:
+	 *
+	 *		model.document.on( 'change:data', () => {
+	 *			console.log( 'The document's data has changed!' );
```
Reinmar: "The data has changed" should be enough.

src/model/markercollection.js
```diff
@@ -88,9 +88,11 @@ export default class MarkerCollection {
 	 * @param {String|module:engine/model/markercollection~Marker} markerOrName Name of marker to set or marker instance to update.
 	 * @param {module:engine/model/range~Range} range Marker range.
 	 * @param {Boolean} [managedUsingOperations=false] Specifies whether the marker is managed using operations.
+	 * @param {Boolean} [affectsData=false] Specifies whether the marker affects the data produced by the data pipeline
+	 * (is persisted in the editor's data).
 	 * @returns {module:engine/model/markercollection~Marker} `Marker` instance which was added or updated.
 	 */
-	_set( markerOrName, range, managedUsingOperations = false ) {
+	_set( markerOrName, range, managedUsingOperations = false, affectsData = false ) {
```
scofalik: We could consider moving the last two parameters to an `options` object to keep params count low.

src/model/document.js
```diff
@@ -391,6 +396,28 @@ export default class Document {
 	 * @event change
 	 * @param {module:engine/model/batch~Batch} batch The batch that was used in the executed changes block.
 	 */
+
+	/**
+	 * Fired after each {@link module:engine/model/model~Model#enqueueChange `enqueueChange()` block},
```
Reinmar: You can now update the last example in the `Document#change`'s docs. It's outdated now. These events should be properly cross-linked between each other.

src/model/writer.js
```diff
@@ -882,18 +895,22 @@ export default class Writer {
 	 *
 	 *		updateMarker( marker, { usingOperation: true } );
 	 *
+	 * Change marker's option (inform engine, that the marker does not affect the data model):
```
pjasiun: *anymore

src/model/document.js
Reinmar: I recommend moving most of the documentation to `Document#change` and clarifying there, in one place, which event to use and when. The other event's docs should only link to `Document#change`.

src/model/writer.js
```diff
@@ -905,43 +922,49 @@ export default class Writer {
...
+		if ( !hasUsingOperationDefined && !options.range && !affectsDataDefined ) {
 			/**
-			 * One of options is required - provide range or usingOperations.
+			 * One of options is required - provide range, usingOperations or affectsData.
 			 *
 			 * @error writer-updateMarker-wrong-options
 			 */
-			throw new CKEditorError( 'writer-updateMarker-wrong-options: One of options is required - provide range or usingOperations.' );
+			throw new CKEditorError(
+				'writer-updateMarker-wrong-options: One of options is required - provide range, usingOperations or affectsData.'
```
Reinmar: "One of the options"

src/model/writer.js
```diff
@@ -867,7 +877,10 @@ export default class Writer {
 	 * The `options.usingOperation` parameter lets you change if the marker should be managed by operations or not. See
 	 * {@link module:engine/model/markercollection~Marker marker class description} to learn about the difference between
 	 * markers managed by operations and not-managed by operations. It is possible to change this option for an existing marker.
-	 * This is useful when a marker have been created earlier and then later, it needs to be added to the document history.
+	 *
+	 * The `options.affectData` parameter allows you to inform the engine, that this marker does not affects the data model.
+	 * It impacts the performance, because changes in markers with set `options.affectData:false` do not cause firing the
```
Reinmar: That's quite irrelevant here.

Instead, you could rather explain what does it mean to affect the data. And that it affects how the `change` or `change:data` is fired.

src/model/differ.js
Reinmar: `hasDataChanges()` sounds fine

src/model/document.js
ma2ciek: If someone listens to both `change` and `change:data` events, the `change:data` will fire before the `change` event. Because each fired `change:...` event will fire the `change` event too.

tests/model/document.js
```diff
@@ -321,42 +321,158 @@ describe( 'Document', () => {
...
+
+			doc.on( 'change:data', spy );
+
+			model.change( writer => {
+				const range = Range.createFromParentsAndOffsets( root, 2, root, 4 );
+				writer.addMarker( 'name', { range, usingOperation: false, affectsData: true } );
+			} );
+
+			sinon.assert.calledOnce( spy );
+		} );
```
pjasiun: There should be an integrational test with `updateMarker`:

`addMarker` (`affectsData`: `false`) -> `change` event is fired (ensure it is not `change:data`)
`updateMarker` (`affectsData`: `true`) -> `change:data` event is fired
`updateMarker` (`affectsData`: `false`) -> `change` event is fired (ensure it is not `change:data`)

src/model/markercollection.js
ma2ciek: I was considering it. But it'd be backward incompatible (for the `Writer` class).

tests/model/document.js
```diff
@@ -321,42 +321,158 @@ describe( 'Document', () => {
...
+
+			doc.on( 'change', () => {
+				callOrder.push( 2 );
+			} );
+
+			model.change( writer => {
+				writer.insertText( 'foo', doc.getRoot(), 0 );
+			} );
+
+			expect( callOrder ).to.deep.equal( [ 1, 2 ] );
```
pjasiun: I do not get this test. There are 2 events fires? I believe there is only `'change:data'` fired and you attached two listeners to it (`change` listener will also listen on `change:data`).

pjasiun: However, there should be a test that it should fire change:data when text is inserted.

pjasiun:
Does it closes also https://github.com/ckeditor/ckeditor5-engine/issues/1418?

ma2ciek:
> Does it closes also #1418?

It closes mainly that ticket.

src/model/document.js
Reinmar: image

image

Only the order of registering these listeners matter.

So, as I wrote, the listeners are combined when the event is fired. Or, in other words, when you listen to `x:y` you also listen to `x`. But only one event is fired.

ma2ciek: Right, right, right. I was completely wrong. 👍

tests/model/document.js
ma2ciek: I removed the above test. I didn't know the exact implementation and was wrong there 😕

ma2ciek:
Ready for the final review.

src/model/writer.js
```diff
@@ -794,6 +794,10 @@ export default class Writer {
 	 * {@link module:engine/model/markercollection~Marker marker class description} to learn about the difference between
 	 * markers managed by operations and not-managed by operations.
 	 *
+	 * The `options.affectData` parameter, which defaults to `true`, allows you to create marker, that does not affects the data model.
+	 * It impacts the performance, because changes in markers with set `options.affectData:false` do not cause firing the
+	 * {@link module:engine/model/document~Document#event:change:data `change:data`} event.
```
pjasiun: I would change this sentence. I am not sure if firing `change:data` event has an impact on the performance. I would focus on:
 - `affectData = true` fires `change:data`, `affectData = false` fires `change` event,
 - `affectData` should be true when the maker change changes data return by `editor.getData`.

src/model/writer.js
```diff
@@ -794,6 +794,10 @@ export default class Writer {
 	 * {@link module:engine/model/markercollection~Marker marker class description} to learn about the difference between
 	 * markers managed by operations and not-managed by operations.
 	 *
+	 * The `options.affectData` parameter, which defaults to `true`, allows you to create marker, that does not affects the data model.
```
pjasiun: defaults to `true`?

src/model/writer.js
```diff
@@ -867,7 +877,10 @@ export default class Writer {
 	 * The `options.usingOperation` parameter lets you change if the marker should be managed by operations or not. See
 	 * {@link module:engine/model/markercollection~Marker marker class description} to learn about the difference between
 	 * markers managed by operations and not-managed by operations. It is possible to change this option for an existing marker.
-	 * This is useful when a marker have been created earlier and then later, it needs to be added to the document history.
+	 *
+	 * The `options.affectData` parameter allows you to inform the engine, that this marker affects the data model.
```
pjasiun: `affects the data model` It does not affects the data model. It affects data, which you get from editor. `affects the editor data` would be better.

src/model/writer.js
```diff
@@ -867,7 +877,10 @@ export default class Writer {
...
+	 * It impacts the performance, because changes in markers with set `options.affectData:false` do not cause firing the
+	 * {@link module:engine/model/document~Document#event:change:data `change:data`} event.
```
pjasiun: As above.

src/model/writer.js
```diff
@@ -882,18 +895,22 @@ export default class Writer {
 	 *
 	 *		updateMarker( marker, { usingOperation: true } );
 	 *
+	 * Change marker's option (inform the engine, that the marker does not affect the data model anymore):
```
pjasiun: not data model

src/model/writer.js
```diff
@@ -794,6 +794,11 @@ export default class Writer {
 	 * {@link module:engine/model/markercollection~Marker marker class description} to learn about the difference between
 	 * markers managed by operations and not-managed by operations.
 	 *
+	 * The `options.affectData` parameter, which defaults to `false`, allows you to create a marker, that affects the data.
```
pjasiun: allows you to define if a marker affects the data

PR #1424, opened 2018-06-01
Avoid unnecessary DOM structure updates in renderer

### LINKED ISSUE #898: Disable inline filling character removal during composition

### LINKED ISSUE #1334: Improve 'sameNodes' comparator inside 'renderer._updateChildren'
Stumbled upon when debugging composition:

1. Set editor content to `<strong>^Bold</strong>`:
```
Model: <paragraph><$text bold="true">[]Bold</$text></paragraph>
View: <container:p><attribute:strong>{}Bold</attribute:strong></container:p>
```
2. Type anything, e.g. <kbd>a</kbd>

The [`renderer._updateChildren`](https://github.com/ckeditor/ckeditor5-engine/blob/a0ad8fee1606f5a0f42e70c96d8f1ceca6c0f919/src/view/renderer.js#L465) method checks [diff between `actualDomChildren` and `expectedDomChildren`](https://github.com/ckeditor/ckeditor5-engine/blob/a0ad8fee1606f5a0f42e70c96d8f1ceca6c0f919/src/view/renderer.js#L487) returning `[ 'insert', 'delete' ]` set of operations, while both `actualDomChildren` and `expectedDomChildren` represents exactly same HTML - `<strong>aBold</strong>`.

If the `sameNodes` comparator would be able to handle such situations we would be able to skip reinserting same nodes (should also help with composition). Preventing situations like (the intermediate state after `insert` and before `delete`):

image

Checked on `ckeditor5-core/tests/manual/article.html` manual test with Chrome 64.0.3282.167.

If this issue was already discussed in the past somewhere just point me to the right direction, but I haven't found any similar ones. Could help with https://github.com/ckeditor/ckeditor5/issues/802

### LINKED ISSUE #1349: [Chrome] The macOS accent panel places inserted characters far after initial selection
Extracted from https://github.com/ckeditor/ckeditor5-typing/issues/95 as I found an exact cause and it deserves a separate issue as it is quite different than others.

The issue:
mar-29-2017 19-13-56

As mentioned in https://github.com/ckeditor/ckeditor5-typing/issues/95#issuecomment-371516735, this issue is caused by the native Chrome bug which moves the selection if some characters are removed from focused text node (while triggered by listening on event like `keyup`, `compostion*`, etc).

The issue was reported here - https://bugs.chromium.org/p/chromium/issues/detail?id=820067 and has been partially solved in Chrome 65 (just in time as it was released few day ago) 🎉 However, for this case it still happens (see above report for more detials).

---

The real problem with it, is that usually we may block removing `inlineFiller` during composition, however in this case `keyup` is fired before `compositionstart` (so renderer is unaware of composition which may/will take place) and removes the `inlineFiller` as character was already inserted, which triggers the native issue moving caret / inserting character in a wrong place.

### LINKED ISSUE #1409: Inline filler (ZWS) is removed while typing inside empty inline element

### LINKED ISSUE #1417: Avoid unnecessary structure updates in renderer
Extracted from #1342:

> Background: #1334.
>
> How does selection conversion work?
>
> 1. MODEL: `<p><$text bold=true>xx</text></p>`  + `selection[bold=true]`
>  2. VIEW:
	1. `<p><b>xx</b></p>`
	2. `<p>[]<b>xx</b></p>` (after position conversion)
	3. `<p><B>[]</B><b>xx</b></p>` (during `writer.wrap( selection )` )
	4. `<p><B>[]xx</B></p>` (after attribute conversion; we're merging to the left, so a new element `<B>` is left)
> 3. RENDERING:
	* actual DOM: `<p><b>xx</b></p>`
	* expected DOM `<p><B>XX</B></p>` (we need to replace the entire `<b>` element; IDEA: don't replace `<b>` with `<B>` because they are similar!)
> 
> BUT... WHAT IF SOMETHING ELSE HAS CHANGED?
> 
> * actual DOM: `<p><b>xx</b></p>`
> * expected DOM `<p><B>XXY</B></p>` (don't replace `<b>` with `<B>` but recursively go deeper to apply other possible changes, see #1125)
> 
> Possible solution: 
> 
> 1. Make `diff( actual, expected, sameNode )` skip elements which have the same names and attributes (see `view.Element#isSimilar()`) - it is described in more details in #1334.
> 2. As a result, we won't replace `<b>` with `<B>`, which means that the view to DOM mapping will be outdated – the new view `<B>` element will be mapped to the new DOM `<B>` element which is not in the tree (because the old DOM `<b>` element was left there). Which means that we should update bindings after skipping such replacements. How to do that? Perhaps we can simply and blindly set new mappings after rendering: purge existing bindings (in a limited scope), `updatedDomElements.forEach( mapToUpdatedViewElement )`.

Also intersting analysis of a specific case is available in https://github.com/ckeditor/ckeditor5-engine/issues/1342#issuecomment-373778804.

### LINKED ISSUE #1427: Placing caret on the beginning of the link on the line start reinserts whole link DOM element
The link highlight is a `ck-link_selected` class added to a link element. It's an attribute so it should be added by the [`renderer._updateAttrs()`](https://github.com/ckeditor/ckeditor5-engine/blob/ba3d64110fdecfd90663847c6ea03ebd8b9e24d6/src/view/renderer.js#L448) method. This happens in most cases, for example, when moving caret like:

```
1. <p><a href="123">Foo</a> B{}ar</p>
2. <p><a href="123">F{}oo</a> Bar</p>
3. <p><a href="123">Fo{}o</a> Bar</p>
```

`renderer` adds highlight class via `renderer._updateAttrs()` method (as `a` element was marked to have its attributes updated) and then the `renderer._updateChildren()` method is run. Since nothing has changed, DOM structure is not modified.

When the caret is placed on the beginning of a link (within text) like:

```
1. <p>F{}oo <a href="123">Bar</a></p>
2. <p>Foo {}<a href="123">Bar</a></p>
```
attributes are not modified (since selection is still outside of the link), or like: 

```
1. <p>Foo <a href="123">Ba{}r</a></p>
2. <p>Foo {}<a href="123">Ba{}r</a></p>
```

highlight class is simply removed in `renderer._updateAttrs()` method and no DOM rendering takes place.

---

The situation is quite different when placing selection on the beginning of the link which is a first element in the paragraph. There are two cases.

**Moving selection from regular text to the link**:

```
1. <p><a href="123">Foo</a> B{}ar</p>
2. <p><a href="123">{}Foo</a> Bar</p>
```

1. No elements are marked to have its attributes updated ([`renderer.markedAttributes`](https://github.com/ckeditor/ckeditor5-engine/blob/ba3d64110fdecfd90663847c6ea03ebd8b9e24d6/src/view/renderer.js#L66) is empty).
1. The `p` and `a` elements are marked to be updated ([`renderer.markedChildren`](https://github.com/ckeditor/ckeditor5-engine/blob/ba3d64110fdecfd90663847c6ea03ebd8b9e24d6/src/view/renderer.js#L74))<sup>**1**</sup>.
1. Children of the `p` element are updated in `renderer._updateChildren()`. Since `a` view element is different, its DOM representation gets rerendered (old `a` is replaced with new `a.ck-link_selected`).
1. Now `renderer._updateChildren()` is executed for the old `a` view element. Since it has no parent and cannot be mapped to the DOM, nothing changes.

So here the fact that link is highlighted is simply caused by the fact that `a` element was replaced, not because its attributes were updated (as it should happen IMHO).

**Moving selection within the link**:

```
1. <p><a href="123">Fo{}o</a> Bar</p>
2. <p><a href="123">{}Foo</a> Bar</p>
```

Here situation is similar but with additional `renderer._updateAttrs()` call.

1. The `a` elements is marked to have its attributes updated (`renderer.markedAttributes`)<sup>**1**</sup>.
1. The `p` and `a` elements are marked to be updated (`renderer.markedChildren`)<sup>**1**</sup>.
1. The attributes of `a` are updated in the `renderer._updateAttrs()` method. Highlight **ck-link_selected** class is removed.
1. Children of the `p` element are updated in `renderer._updateChildren()`. Since `a` view element is different, its DOM representation gets rerendered (old `a` is replaced with new `a.ck-link_selected`).
1. Now `renderer._updateChildren()` is executed for the old `a` view element. Since it has no parent and cannot be mapped to the DOM, nothing changes.

In this case, `renderer._updateAttrs()` removes highlight class and then it's added because whole `a` element is rerenderd.

**1.** When `a` (view element) is being marked to sync it has a `parent` element but when `rednerer.render()` is called, its parent is already empty (which means it was removed from view structure and replace with another element).

---

I think in all the cases above, DOM `a` element should not be reinserted, only its attributes updated. Still I'm not sure if we should mark it as a bug or enhancement.

f1ames:

Fix: Avoid unnecessary DOM structure updates in `Renderer`. Closes #1417. Closes #1409. Closes #1349. Closes #1334. Closes #898. Closes ckeditor/ckeditor5-typing#129. Closes ckeditor/ckeditor5-typing#89. Closes #1427.

---

### Additional information

See the extensive comment about the implemented solution in https://github.com/ckeditor/ckeditor5-engine/issues/1417#issuecomment-393845235.

There is also one issue which I'm unable to reproduce (from the beginning) so we should definitely ping the person who reported it to recheck after this PR gets merged - https://github.com/ckeditor/ckeditor5/issues/748.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -247,12 +247,14 @@ export default class Renderer {
 	 * @param {module:engine/view/node~Node} viewElement The view element which children mappings will be updated.
 	 */
 	_updateChildrenMappings( viewElement ) {
-		const diff = this._diffChildren( viewElement );
+		const domElement = this.domConverter.mapViewToDom( viewElement );
 
-		if ( !diff ) {
+		if ( !domElement ) {
+			// If there is no `domElement` it means that it was already removed from DOM and there is no need to process it.
 			return;
 		}
 
+		const diff = this._diffChildren( viewElement );
 		const actions = this._findReplaceActions( diff.actions, diff.actualDomChildren, diff.expectedDomChildren );
 
 		if ( actions.indexOf( 'replace' ) !== -1 ) {
@@ -551,22 +553,24 @@ export default class Renderer {
 	 * filler should be rendered.
 	 */
 	_updateChildren( viewElement, options ) {
-		const inlineFillerPosition = options.inlineFillerPosition;
-		const diff = this._diffChildren( viewElement, inlineFillerPosition );
+		const domElement = this.domConverter.mapViewToDom( viewElement );
 
-		if ( !diff ) {
+		if ( !domElement ) {
+			// If there is no `domElement` it means that it was already removed from DOM.
+			// There is no need to process it. It will be processed when re-inserted.
 			return;
 		}
 
-		const actions = diff.actions;
-		const domElement = diff.domElement;
-		const actualDomChildren = diff.actualDomChildren;
+		const inlineFillerPosition = options.inlineFillerPosition;
+		// As binding may change actual DOM children we need to do this before diffing.
 		const expectedDomChildren = this._getElementExpectedChildren( viewElement, domElement, { bind: true, inlineFillerPosition } );
+		const diff = this._diffChildren( viewElement, inlineFillerPosition );
+		const actualDomChildren = diff.actualDomChildren;
 
 		let i = 0;
 		const nodesToUnbind = new Set();
 
-		for ( const action of actions ) {
+		for ( const action of diff.actions ) {
 			if ( action === 'insert' ) {
 				insertAt( domElement, i, expectedDomChildren[ i ] );
 				i++;
@@ -600,27 +604,17 @@ export default class Renderer {
 	 * filler should be rendered.
 	 * @returns {Object|null} result
 	 * @returns {Array.<String>} result.actions List of actions based on {@link module:utils/diff~diff} function.
-	 * @returns {Node} result.domElement The DOM element representing given view element.
 	 * @returns {Array.<Node>} result.actualDomChildren Current `viewElement`'s DOM children.
 	 * @returns {Array.<Node>} result.expectedDomChildren Expected `viewElement`'s DOM children.
 	 */
 	_diffChildren( viewElement, inlineFillerPosition = null ) {
-		const domConverter = this.domConverter;
-		const domElement = domConverter.mapViewToDom( viewElement );
-
-		if ( !domElement ) {
-			// If there is no `domElement` it means that it was already removed from DOM.
-			// There is no need to process it. It will be processed when re-inserted.
-			return null;
-		}
-
+		const domElement = this.domConverter.mapViewToDom( viewElement );
 		const actualDomChildren = domElement.childNodes;
 		const expectedDomChildren = this._getElementExpectedChildren( viewElement, domElement,
 			{ withChildren: false, inlineFillerPosition } );
 
 		return {
-			actions: diff( actualDomChildren, expectedDomChildren, sameNodes.bind( null, domConverter.blockFiller ) ),
-			domElement,
+			actions: diff( actualDomChildren, expectedDomChildren, sameNodes.bind( null, this.domConverter.blockFiller ) ),
 			actualDomChildren,
 			expectedDomChildren
 		};
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -237,15 +237,17 @@ export default class Renderer {
 	}
 
 	/**
-	 * Updates viewElement children mappings. Children which were replaced in the view structure by the similar
-	 * element (same tag name) are treated as 'replaced'. Their mappings are rebind to the corresponding,
-	 * existing DOM element so they will not be replaced by a new DOM element during rendering.
+	 * Updates mapping of `viewElement`'s children.
+	 *
+	 * Children which were replaced in the view structure by similar elements (same tag name) are treated as 'replaced'.
+	 * This means that we can update their mappings so the new view elements are mapped to the existing DOM elements.
+	 * Thanks to that we won't need to re-render these elements completely.
 	 *
 	 * @private
 	 * @param {module:engine/view/node~Node} viewElement The view element which children mappings will be updated.
 	 */
 	_updateChildrenMappings( viewElement ) {
-		// We do not perform any operations on DOM here so there is no need to bind view element or convert its children.
+		// We're diffing only so we don't have to bind anything.
 		const diff = this._diffElementChildren( viewElement, { bind: false, withChildren: false } );
 
 		if ( diff ) {
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -247,8 +247,7 @@ export default class Renderer {
 	 * @param {module:engine/view/node~Node} viewElement The view element which children mappings will be updated.
 	 */
 	_updateChildrenMappings( viewElement ) {
-		// We're diffing only so we don't have to bind anything.
-		const diff = this._diffChildren( viewElement, { bind: false, withChildren: false } );
+		const diff = this._diffChildren( viewElement );
 
 		if ( !diff ) {
 			return;
@@ -282,11 +281,11 @@ export default class Renderer {
 	}
 
 	/**
-	 * Updates mappings of a given view element.
+	 * Updates mappings of a given `viewElement`.
 	 *
 	 * @private
 	 * @param {module:engine/view/node~Node} viewElement The view element which mappings will be updated.
-	 * @param {Node} domElement DOM representation of a given view element.
+	 * @param {Node} domElement The DOM element representing given view element.
 	 */
 	_updateElementMappings( viewElement, domElement ) {
 		// Because we replace new view element mapping with the existing one, the corresponding DOM element
@@ -513,7 +512,7 @@ export default class Renderer {
 	 * Checks if attributes list needs to be updated and possibly updates it.
 	 *
 	 * @private
-	 * @param {module:engine/view/element~Element} viewElement View element to update.
+	 * @param {module:engine/view/element~Element} viewElement The view element to update.
 	 */
 	_updateAttrs( viewElement ) {
 		const domElement = this.domConverter.mapViewToDom( viewElement );
@@ -552,8 +551,8 @@ export default class Renderer {
 	 * filler should be rendered.
 	 */
 	_updateChildren( viewElement, options ) {
-		const diff = this._diffChildren( viewElement,
-			{ inlineFillerPosition: options.inlineFillerPosition, bind: true, withChildren: true } );
+		const inlineFillerPosition = options.inlineFillerPosition;
+		const diff = this._diffChildren( viewElement, inlineFillerPosition );
 
 		if ( !diff ) {
 			return;
@@ -562,7 +561,7 @@ export default class Renderer {
 		const actions = diff.actions;
 		const domElement = diff.domElement;
 		const actualDomChildren = diff.actualDomChildren;
-		const expectedDomChildren = diff.expectedDomChildren;
+		const expectedDomChildren = this._getElementExpectedChildren( viewElement, domElement, { bind: true, inlineFillerPosition } );
 
 		let i = 0;
 		const nodesToUnbind = new Set();
@@ -592,23 +591,20 @@ export default class Renderer {
 	}
 
 	/**
-	 * Compares viewElement actual and expected children and actions sequence which can be used to transform
+	 * Compares `viewElement`'s actual and expected children and returns actions sequence which can be used to transform
 	 * actual children into expected ones.
 	 *
 	 * @private
-	 * @param viewElement
-	 * @param {Object} options
-	 * @param {module:engine/view/position~Position} options.inlineFillerPosition The position on which the inline
+	 * @param {module:engine/view/node~Node} viewElement The view element which children will be compared.
+	 * @param {module:engine/view/position~Position} [inlineFillerPosition=null] The position on which the inline
 	 * filler should be rendered.
-	 * @param {Boolean} options.bind If new view elements should be bind to their corresponding DOM elements.
-	 * @param {Boolean} options.withChildren If children of newly bound view elements should also be converted.
 	 * @returns {Object|null} result
-	 * @returns {Array} result.actions List of actions based on {@link module:utils/diff~diff} function.
-	 * @returns {Node} result.domElement ViewElement corresponding DOM element.
-	 * @returns {Array} result.actualDomChildren Current viewElement DOM children.
-	 * @returns {Array} result.expectedDomChildren Expected viewElement DOM children.
+	 * @returns {Array.<String>} result.actions List of actions based on {@link module:utils/diff~diff} function.
+	 * @returns {Node} result.domElement The DOM element representing given view element.
+	 * @returns {Array.<Node>} result.actualDomChildren Current `viewElement`'s DOM children.
+	 * @returns {Array.<Node>} result.expectedDomChildren Expected `viewElement`'s DOM children.
 	 */
-	_diffChildren( viewElement, options ) {
+	_diffChildren( viewElement, inlineFillerPosition = null ) {
 		const domConverter = this.domConverter;
 		const domElement = domConverter.mapViewToDom( viewElement );
 
@@ -618,17 +614,9 @@ export default class Renderer {
 			return null;
 		}
 
-		const domDocument = domElement.ownerDocument;
-		const filler = options.inlineFillerPosition;
 		const actualDomChildren = domElement.childNodes;
-		const expectedDomChildren = Array.from( domConverter.viewChildrenToDom( viewElement, domDocument, options ) );
-
-		// Inline filler element has to be created during children update because we need it to diff actual dom
-		// elements with expected dom elements. We need inline filler in expected dom elements so we won't re-render
-		// text node if it is not necessary.
-		if ( filler && filler.parent == viewElement ) {
-			this._addInlineFiller( domDocument, expectedDomChildren, filler.offset );
-		}
+		const expectedDomChildren = this._getElementExpectedChildren( viewElement, domElement,
+			{ withChildren: false, inlineFillerPosition } );
 
 		return {
 			actions: diff( actualDomChildren, expectedDomChildren, sameNodes.bind( null, domConverter.blockFiller ) ),
@@ -636,6 +624,31 @@ export default class Renderer {
 			actualDomChildren,
 			expectedDomChildren
 		};
+	}
+
+	/**
+	 * Returns expected DOM children for a given `viewElement`.
+	 *
+	 * @private
+	 * @param {module:engine/view/node~Node} viewElement View element which children will be returned.
+	 * @param {Node} domElement DOM representation of a given view element.
+	 * @param {Object} options See {@link module:engine/view/domconverter~DomConverter#viewToDom} options parameter.
+	 * @param {module:engine/view/position~Position} [options.inlineFillerPosition=null] The position on which
+	 * the inline filler should be rendered.
+	 * @returns {Array.<Node>} The `viewElement`'s expected children.
+	 */
+	_getElementExpectedChildren( viewElement, domElement, options ) {
+		const expectedDomChildren = Array.from( this.domConverter.viewChildrenToDom( viewElement, domElement.ownerDocument, options ) );
+		const filler = options.inlineFillerPosition;
+
+		// Inline filler element has to be created as it is present in a DOM, but not in a view. It is required
+		// during diffing so text nodes could be compared correctly and also during rendering to maintain
+		// proper order and indexes while updating the DOM.
+		if ( filler && filler.parent === viewElement ) {
+			this._addInlineFiller( domElement.ownerDocument, expectedDomChildren, filler.offset );
+		}
+
+		return expectedDomChildren;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Reinmar:
@Mgsy, could you check these changes? I'm mainly interested if there are some regressions. Scenarios to cover are hard to describe... all that requires bigger DOM updates. From typing in weird places, to backspace/enter on bigger non-collapsed selections, applying block quote, lists. Collaboration too. Pretty much everything.

Mgsy:
## Steps to reproduce
1. Put the caret at the beginning of a block.
1. Press <kbd>Ctrl</kbd> + <kbd>K</kbd>.
1. Insert new link.
1. Put the caret at the beginning of the link.

## Current result
The editor crashes.

## Error
```
renderer.js:540 Uncaught TypeError: Cannot read property 'attributes' of undefined
    at Renderer._updateAttrs (renderer.js:540)
    at Renderer.render (renderer.js:214)
    at View._render (view.js:391)
    at View.on (view.js:151)
    at View.change (view.js:357)
    at Document.EditingController.listenTo (editingcontroller.js:81)
    at Model.Document.listenTo (document.js:154)
```

## GIF
[Click](https://user-images.githubusercontent.com/20988892/40916306-88e34354-67ff-11e8-8c8d-79c240f4dc83.gif).

Mgsy:
## Steps to reproduce
1. Set the following data: 
```js
editor.setData( '<ol><li>Item 1<ol><li>Item 2</li></ol></li></ol><p>Paragraph</p><ol><li>Item 3<ol><li>Item 4</li></ol></li></ol>' );
```
2. Start the selection in the first item of the second list and move it through the paragraph to the second item of the first list.
3. Click on the ordered list button in the toolbar.

## Current result
`Item 3` is inside `<ol>` element: 

image

## GIF
[Click](https://user-images.githubusercontent.com/20988892/40918928-2d9fc374-6808-11e8-897a-64a70b616aeb.gif).

Mgsy:
## Firefox
### Steps to reproduce
1. Open the article sample.
1. Change your keyboard to Hiragana.
1. Put the caret inside the link.
1. Start the composition.
1. Press <kbd>Enter</kbd> to finish the composition.

### Current result
The editor crashes.

### Error
```
TypeError: domElement is undefined[Learn More] renderer.js:540
	_updateAttrs renderer.js:540
	render renderer.js:214
	_render view.js:391
	View/< view.js:151
	fire emittermixin.js:196
	change view.js:357
	EditingController/< editingcontroller.js:81
	fire emittermixin.js:196
	Document/< document.js:154
	fire emittermixin.js:196
	_runPendingChanges model.js:405
	enqueueChange model.js:207
	execute inputcommand.js:79
	decorate/< observablemixin.js:250
	fire emittermixin.js:196
	decorate/this[methodName] observablemixin.js:254
	execute commandcollection.js:67
	execute editor.js:254
	_handleTextMutation input.js:293
	handle input.js:152
	_handleMutations input.js:106
	init/< input.js:50
	fire emittermixin.js:196
	_onMutations mutationobserver.js:247
	_onMutations self-hosted:977:17
```

### GIF
[Click](https://user-images.githubusercontent.com/20988892/40923631-9381754a-6815-11e8-87ff-a4a2dbe3f375.gif).

Mgsy:
This solution causes problems with the user's selection and markers. Most of times the marker isn't rendered and a colour of the selection is incorrect. 

**GIF**
[Click](https://user-images.githubusercontent.com/20988892/40964435-636b676c-68ab-11e8-8764-3e81a9436f86.gif).

f1ames:
The problem with markers was more general issue with handling `UIElements`. The `UIElement` is specific due to the fact that its children are not stored in a view (#799). Before rendering optimisation was implemented the whole marker (marker is just a `<span>` element) was rerendered as a `p` element child.
Now as the marker element is treated as `replaced` element, only its contents were rerendered. Since `uiElement` child are not present in a view, its expected DOM children were always empty so entire contents of such element were removed during `rendering`. I think it makes sense just to skip such elements so they are processed as before - covered in https://github.com/ckeditor/ckeditor5-engine/pull/1424/commits/c974fbbb8f57bf347c7e7add060c299a7b47ce8e.

f1ames:
Ready for 2nd review 🎉  cc @Mgsy @Reinmar

src/view/renderer.js
```diff
@@ -494,52 +643,78 @@ export default class Renderer {
...
+	 * @param {Array} expectedDom Expected DOM children.
+	 * @returns {Array} Actions array modified with `replace` actions.
+	 */
+	_findReplaceActions( actions, actualDom, expectedDom ) {
+		// If there is no both 'insert' and 'delete' actions, no need to check for replaced elements.
+		if ( actions.indexOf( 'insert' ) === -1 || actions.indexOf( 'delete' ) === -1 ) {
+			return actions;
+		}
+
+		function areSimilar( domNode1, domNode2 ) {
```
Reinmar: These helper fns should not be defined in this scope. Every execution of this method creates them.

src/view/renderer.js
```diff
@@ -19,6 +19,7 @@ import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 import isText from '@ckeditor/ckeditor5-utils/src/dom/istext';
 import fastDiff from '@ckeditor/ckeditor5-utils/src/fastdiff';
+import isNode from '@ckeditor/ckeditor5-utils/src/dom/isnode';
```
Reinmar: Move up, to be right after/before `isText()`.

src/view/renderer.js
```diff
@@ -494,52 +643,78 @@ export default class Renderer {
...
-				// Force updating text nodes inside elements which did not change and do not need to be re-rendered (#1125).
-				this._markDescendantTextToSync( domConverter.domToView( expectedDomChildren[ i ] ) );
-				i++;
+				skipActions.push( 'delete' );
+				actualSlice.push( actualDom[ counter.equal + counter.delete ] );
+			} else { // equal
+				if ( expectedSlice.length && actualSlice.length ) {
+					newActions = newActions.concat( calculateReplaceActions( actualSlice, expectedSlice ) );
+				} else if ( expectedSlice.length || actualSlice.length ) {
+					newActions = newActions.concat( skipActions );
```
scofalik: Is this needed? Wouldn't `calculateReplaceActions`, or rather `diff`, return `[ 'insert', 'insert', ... ]` if `actualSlice` is empty and `[ 'delete', 'delete', ... ]` if `expectedSlice` is empty?

scofalik: Especially since you don't have this below (in line 713) which is the same situation as meeting `equal`.

src/view/renderer.js
```diff
@@ -473,19 +536,92 @@ export default class Renderer {
...
+	/**
+	 * Compares viewElement actual and expected children and actions sequence which can be used to transform
+	 * actual children into expected ones.
+	 *
+	 * @private
+	 * @param viewElement
+	 * @param {Object} options
+	 * @param {module:engine/view/position~Position} options.inlineFillerPosition The position on which the inline
+	 * filler should be rendered.
+	 * @param {Boolean} options.bind If new view elements should be bind to their corresponding DOM elements.
```
Reinmar: This is smelly. It violates the single responsibility principle – why does `diffSth()` bind anything?

I noticed this because I couldn't understand the comment above:

> // We do not perform any operations on DOM here so there is no need to bind view element or convert its children.

It didn't have any sense for me in that context because I saw that the line is `const diff = this._diffElementChildren()`.

src/view/renderer.js
```diff
@@ -473,19 +536,92 @@ export default class Renderer {
...
+	 * @param {Boolean} options.withChildren If children of newly bound view elements should also be converted.
+	 * @returns {Object|null} result
+	 * @returns {Array} result.actions List of actions based on {@link module:utils/diff~diff} function.
+	 * @returns {Node} result.domElement ViewElement corresponding DOM element.
+	 * @returns {Array} result.actualDomChildren Current viewElement DOM children.
+	 * @returns {Array} result.expectedDomChildren Expected viewElement DOM children.
+	 */
+	_diffElementChildren( viewElement, options ) {
```
Reinmar: `_diffChildren()` would be enough. Or maybe `_diffDomChildren()`.

Reinmar:
One generic thing which I don't like in some of the `Render`'s code is that we have methods which don't use `this` in any way. In such a case I think it'd be better to make them a simple functions.

The same, in fact, applies to the `DomConverter` where I carelessly added two new methods while they could be simple fns too.

src/view/renderer.js
```diff
@@ -243,6 +236,67 @@ export default class Renderer {
...
+	 * @param {module:engine/view/node~Node} viewElement The view element which children mappings will be updated.
+	 */
+	_updateChildrenMappings( viewElement ) {
+		// We do not perform any operations on DOM here so there is no need to bind view element or convert its children.
+		const diff = this._diffElementChildren( viewElement, { bind: false, withChildren: false } );
+
+		if ( diff ) {
+			const actions = this._findReplaceActions( diff.actions, diff.actualDomChildren, diff.expectedDomChildren );
+
+			if ( actions.indexOf( 'replace' ) !== -1 ) {
```
Reinmar: Make it an early return. Easier to read when there's less nesting.

src/view/renderer.js
```diff
@@ -243,6 +236,67 @@ export default class Renderer {
...
+				const counter = { equal: 0, insert: 0, delete: 0 };
+				for ( const action of actions ) {
```
Reinmar: There should be a blank line before `for()`/`if()`/`while`/`function`. Odd that we don't have this covered in ESLint.

src/view/renderer.js
```diff
@@ -243,6 +236,67 @@ export default class Renderer {
...
+					if ( action === 'replace' ) {
+						const insertIndex = counter.equal + counter.insert;
+						const deleteIndex = counter.equal + counter.delete;
+						const viewChild = viewElement.getChild( insertIndex );
+
+						// The 'uiElement' is a special one and its children are not stored in a view (#799),
+						// so we cannot use it with replacing flow (since it uses view children during rendering
+						// which will always result in rendering empty element).
+						if ( viewChild && !viewChild.is( 'uiElement' ) ) {
```
Reinmar: This function grew a bit long. So, perhaps contents of this `if()` would make a reasonable function itself. It only accepts 3 params – the 3 vars that you define above. That's why I found it a good place to split the code a bit.

src/view/renderer.js
```diff
@@ -447,6 +501,15 @@ export default class Renderer {
 	 */
 	_updateAttrs( viewElement ) {
 		const domElement = this.domConverter.mapViewToDom( viewElement );
+
+		if ( !domElement ) {
```
Reinmar: I'm not sure about this – this function should not be called if the mapping is broken. Can't we clean `Renderer#markedAttribiutes` earlier, when updating the mappings.

In other words – if you call a function, you expect it to do something. You should be more or less sure it makes sense.

src/view/renderer.js
```diff
@@ -473,19 +536,92 @@ export default class Renderer {
 	 * filler should be rendered.
 	 */
 	_updateChildren( viewElement, options ) {
+		const diff = this._diffElementChildren( viewElement,
+			{ inlineFillerPosition: options.inlineFillerPosition, bind: true, withChildren: true } );
+
+		if ( diff ) {
```
Reinmar: Again – early return.

And – how is it possible that we call this method when its children are up to date?

src/view/renderer.js
```diff
@@ -473,19 +536,92 @@ export default class Renderer {
...
+			const actions = diff.actions;
+			const domElement = diff.domElement;
+			const actualDomChildren = diff.actualDomChildren;
+			const expectedDomChildren = diff.expectedDomChildren;
+
+			let i = 0;
+			const nodesToUnbind = new Set();
```
Reinmar: Blank line missing.

src/view/renderer.js
```diff
@@ -473,19 +536,92 @@ export default class Renderer {
...
 
 		if ( !domElement ) {
 			// If there is no `domElement` it means that it was already removed from DOM.
-			// There is no need to update it. It will be updated when re-inserted.
-			return;
+			// There is no need to process it. It will be processed when re-inserted.
+			return null;
+		}
+
+		function sameNodes( actualDomChild, expectedDomChild ) {
```
Reinmar: Don't define helper functions inside other methods unless really necessary.

Reinmar:
OK, I think we've got enough for now. I haven't dug into the logic because I think it may change a bit still.

src/view/renderer.js
f1ames: > And – how is it possible that we call this method when its children are up to date?

I'm not sure what you meant here? Do you have any particular case in mind?

src/view/renderer.js
f1ames: AFAIR the main reason was that when we update bindings of a specific element (it is removed from mappings and also all its children mappings are removed). So if you would like to update `markedAttributes` during updating bindings you will have to iterate:

1. through all updated element children to check and remove them from `markedAttributes`,
2. or through `markedAttributes` and check if any elements mappings were removed.

It just seemed simpler to check it here if we anyway iterate through `markedAttributes` (and this is just an equivalent to **2nd** point just in different place). WDYT @Reinmar ?

src/view/renderer.js
f1ames: I suppose I wasn't fully aware of that, but I will keep this in mind now 👍

src/view/renderer.js
Reinmar: I meant the logic of this method – you call `updateChildren()` based on `markedElements`. My assumption is – if an element is marked its children have changed. And hence, the `diff` should not be empty in this case.

The existence of this `if()` here means that my assumption is wrong. But why? We need to have a clarity on:

1. Why is this method called for up-to-date element in the first place.
2. Can we prevent calling it for up-to-date element (so to remove this `if()`).
3. Eventually, if it's most convenient to have this `if()` here and it's really needed (there are no other straightforward solutions), we need a comment.

f1ames: One thing, even in situations when children are equal we [mark descendant text nodes to sync](https://github.com/ckeditor/ckeditor5-engine/pull/1424/files#diff-4c87133ed830fc4ea0646426deba32cfR561) so I'm not sure if we may completely skip such situations (if that's what you mean).

f1ames: Didn't noticed your comment earlier @Reinmar, so my last comment refers to previous ones.

As for the latest comment - `diff` is empty not when elements are up-to-date (then you will have `diff` result like `[ equal, equal, ... ]`), but when the `viewElement` which is passed to `updateChildren()` and then `diffChildren()` does not have corresponding DOM element (it cannot be mapped which usually means that corresponding DOM element was removed from DOM, but probably also similar cases as with attributes - https://github.com/ckeditor/ckeditor5-engine/pull/1424#discussion_r195065595). This is not a new `if()` here, as [it was there before](https://github.com/ckeditor/ckeditor5-engine/blob/3e74554fa97741189050f7a4014ee6c558645cb8/src/view/renderer.js#L479) and it was just moved to `diffChildren()` method together with comment explaining what happened - https://github.com/ckeditor/ckeditor5-engine/pull/1424/files#diff-4c87133ed830fc4ea0646426deba32cfR611.

f1ames:
I have covered all proposed changes and explained two more complex cases:

* https://github.com/ckeditor/ckeditor5-engine/pull/1424#discussion_r195127511
* https://github.com/ckeditor/ckeditor5-engine/pull/1424#discussion_r195065595

I think it is ready for further review. cc @Reinmar @scofalik

Reinmar:
Crash:

1. `<p>[Foo<strong>Bar]</strong></p>`
2. Apply a link

Reinmar:
I miss a test for the last change. We had a bug. A fix for this bug should be tested.

Reinmar:
I made quite a lot of tests and all seems fine. Checking the tests now and if all's fine I'll be merging this. Congrats! 🌮

PR #1426, opened 2018-06-07
Render block filler in empty lines in container

### LINKED ISSUE #1422: Block filler must be rendered in empty lines created by <br>
may-30-2018 11-51-21

Unfortunately, we must achieve this when rendering a block which ends with a line break:

```html
<p>...<br>BLOCK_FILLER</p>
```

`BLOCK_FILLER` is a `<br data-cke-filler=true>` inside the editing view and `&nbsp;` in the data.

So, we need to adjust the logic which decides where to place block fillers. In general, block filler is needed after a `<br>` if that `<br>` is the last node of a container element:

```html
<p>XXX<br>YYY<br>ZZZ<br>BLOCK_FILLER</p>
<p>XXX<br>YYY<br>ZZZ<br>ĄŚĆ</p>
<p><br>YYY<br>ZZZ<br>ĄŚĆ</p>
```

pomek:

Fix: Block filler will be inserted into the container if its last child is the <br> element. Closes #1422.

### DIFF
--- a/src/view/containerelement.js
+++ b/src/view/containerelement.js
@@ -83,7 +83,7 @@ function getFillerOffset() {
 	const children = [ ...this.getChildren() ];
 	const lastChild = children[ this.childCount - 1 ];
 
-	// Block filler is required after the `<br>` element if the element is the last child in the container. See #1422.
+	// Block filler is required after a `<br>` if it's the last element in its container. See #1422.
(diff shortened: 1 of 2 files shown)


Reinmar:
👍 Great tests. All works fine for me.

One missing thing will be adding integration tests for `setData()` and `getData()`. But this should be done in `@ckeditor/ckeditor5-enter`.

PR #1431, opened 2018-06-11
T/ckeditor5 table/12: Prevent partial selection of isLimit=true elements.

### LINKED ISSUE #1156: [FF] Selection split by widgets should be merged in model to one range
FF split the selection into two ranges because of the `contenteditable=false` attribute in the widget. This causes two problems. First, the selection that contains multiple ranges is not supported well yet, and second, that we can't select an image element.

### LINKED ISSUE #1176: Incorrect selection after reaching the side image with the caption
We have few cases with the incorrect selection, when it starts in the block element which is below the side image with the caption.

### Steps to reproduce
1. Go to http://localhost:8125/ckeditor5-core/tests/manual/article.html.
2. Start the selection in the block quote.
3. Move it to the image, but do it really slowly. You have to move the selection to the beginning of the image (you can reach just the left border - see at the gif below).

### Current result
The selection moves to the beginning of the document.

### GIF
bug_cke5

### LINKED ISSUE #1182: [FF] Cannot select by mouse over a widget with nested editable
The selection falls into the nested editable. ~~And it's worse than the similar bug in Chrome because here it doesn't work in both directions.~~ (EDIT: nope, it doesn't work only for forward selection. So it's the opposite of Chrome's bug ;|)

nov-13-2017 15-51-36

jodator:

Feature: Introduced a selection post-fixer. Its role is to ensure that after all changes are applied the selection is placed in a correct position. Closes #1156. Closes #1176. Closes #1182. Closes ckeditor/ckeditor5-table#11. Closes ckeditor/ckeditor5-table#12. Closes ckeditor/ckeditor5-table#15. Closes ckeditor/ckeditor5#562. Closes ckeditor/ckeditor5#611.

---

### Additional information:

* This PR is required by tables pr: https://github.com/ckeditor/ckeditor5-table/pull/39.
* Probably fixes also: ckeditor/ckeditor5#562, ckeditor/ckeditor5#611 (to check).
* I'm not sure where to place this post fixer. Right now it is `engine/controllers/` as the `EditingController` adds this post fixer.
* Selection might sometimes jump - but I did not experienced that in a while.

I've asked @Mgsy to check if mentioned bugs are fixed. For me it looks like this post fixer will also fix them.


Mgsy:
There is still a problem with the selection. If you try to select `<paragraph><image><paragraph>`, the first `<paragraph>` won't be selected. When you try to delete selected content and restore it, only the caption will be restored. Tested on Windows and macOS.

bug_cke5

Mgsy:
**Firefox**

The backwards selection makes a part of the selection blinks.

bug_cke5

Mgsy:
>Probably fixes also: ckeditor/ckeditor5#562, ckeditor/ckeditor5#611 (to check).

It seems that this one fixes ckeditor/ckeditor5#562 but unfortunately, ckeditor/ckeditor5#611 is still reproducible.

Besides that, there are still problems with selecting widgets (see previous comments). ckeditor/ckeditor5#611 is fixed.

src/controller/selectionpostfixer.js
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * **Note** If selection contains multiple ranges the method returns minimal set of ranges that are not intersecting after expanding them
+ * to select `isLimit=true` elements.
+ *
+ * See {@link module:engine/model/schema~Schema#isLimit}.
+ *
+ * @param {module:engine/model/writer~Writer} writer
+ * @param {module:engine/model/model~Model} model
+ */
+export default function selectionPostFixer( writer, model ) {
```
Reinmar: It's a function so its name should start with a verb. You can find similar logic in the view – e.g. there's `view/filler~injectQuirksHandling`. 

I'd propose to move this code to `model/utils/selection-postfixer` and expose a function called `injectSelectionPostfixer`. I'd also say that it can be used in `Model#constructor()` because:

* it will be symmetrical to `View#constructor()` injecting some behaviours,
* it's related to schema and basic schema is defined in `Model`.

jodator: K, thanks for pointing out proper placement for the logic. I'll update the PR with this change.

Reinmar:
This PR blows up 59 tests ;) Some of these tests check some commands execution on selection spanning limit elements. But some are weird – mainly, in the lists where I think we shouldn't have any problems.

Now, should we remove these tests or fix them? Theoretically, those tests check whether commands work fine for incorrect selections, but incorrect selections are automatically fixed, so these tests (and code which they test) are unnecessary. But, we post-fix the selection so it's possible to (programmatically) set incorrect selection and execute the command immediately after that. Hence, commands need to work for incorrect selections too.

src/controller/selectionpostfixer.js
Reinmar: Actually, I see that this function doesn't add a postfixer itself. Check out how `view/filler.js` is structured – this way of exposing filler's functionality encapsulates what it needs to do – how many listeners it has to register, with what params, etc.

In this case, we're talking about just one post fixer and registering it in a straightforward way via `doc.registerPostFixer()`. Therefore, on one hand, exporting the callback like you did seems sufficient. OTOH, it's hard to tell now whether this is all that we'll ever need. Hence... IDK :P I leave it up to you.

Reinmar:
Wow, this change works better than I expected. It doesn't, at least on Chrome, get in a way of doing various selections. I found only minor non-blocking issues. And, my biggest surprise is that Ctrl+A kinda started working when widget is at the beginning/end of the content :D

jodator:
I've pushed the branch `t/ckeditor5-table/12` with proper branches set in `mgit.json` in `ckeditor5` repo.

jodator:
Side note: While test are ready to be reviewed I still need to refactor the post fixer as in comments.

jodator:
@Reinmar OK so the tests are fixed. The PR in `List` feature has master merged as some list tests were broken.

Reinmar:
The documentation of all the new functions only mention limit elements. While, since you also use `getNearestSelectionRange()` we also make sure that selection doesn't end up in a place where text isn't allowed. You wrote the documentation (esp. the public one) too oriented on the limit element aspect. I'll change this a bit because the role of this tool is to fix ALL selections. Right now it means two rules, but it may be more in the future. The public docs should be focused on that generic use case for this feature.

Reinmar:
I found a series of bugs with shift+left/right around images (when starting from a collapsed selection outside of an image). For some reason these cases we working fine with tables.

I realised that we focused on limit elements, but the same rules need to be applied to object elements. You also can't have such a selection:

```html
<paragraph>xx[x</paragraph>
<image>]<caption>xxx</caption></image>
```

In the image feature only the caption is marked as a limit. Image isn't because theoretically this was a different semantics. It's an object element. But perhaps all objects are limits too? 

We may, perhaps change this in the schema, but I'm not sure yet. I checked though, that after doing `(isLimit(node) || isObject(node))` checks in the selection post-fixer most bugs disappeared. In fact,  only one set of them remained:

```html
<p>xxx</p>
<image><caption>yyy</caption></image>
<p>x[]xx</p>
```

1. Press shift+left a couple of times (e.g. 3)
2. Press shift+right the same number of times

You should be in the same place. But, you get stuck on the image. Why? Because shift+right makes such a selection:

```html
<p>xxx</p>
<image><caption>[yyy</caption></image>
<p>x]xx</p>
```

And we extend it to:

```html
<p>xxx</p>
[<image><caption>yyy</caption></image>
<p>x]xx</p>
```

Reverting what the user did by pressing shift+right. Unfortunately, to solve this we need to distinguish between left and right keys. This can only be done in the ckeditor5-widget package and needs to be a followup of this ticket.

Reinmar:
> We may, perhaps change this in the schema, but I'm not sure yet. I checked though, that after doing (isLimit(node) || isObject(node)) checks in the selection post-fixer most bugs disappeared. In fact, only one set of them remained:

OK, I decided to make `schema.isLimit()` return true for `isObject` elements. This makes quite a lot of sense – a self-contained non-dividable element limits the selection. You can never select a part of an image e.g. This allowed simplifying a couple of existing conditions which checked for `isObject || isLimit`.

Reinmar:
I spent a couple more hours on this because I realised that this PR doesn't introduce a complete set of rules. It also oversimplified some scenarios (crossing limit boundaries), which would hit us as soon as we'd use any block content inside table cells or image captions.

I extracted these scenarios to https://github.com/ckeditor/ckeditor5-engine/issues/1436 to not block this ticket and I'm running now all the tests again. Hopefully, this PR can be merged as-is at this stage.

@Mgsy, could you do last checks how this PR behaves? You'll need to do that through a setup of branches in `ckeditor5@t/ckeditor5-table/12`.

Reinmar:
This is from `master` (I'm testing a different PR):

jun-18-2018 11-06-52

How does it behave on this PR?

Mgsy:
>How does it behave on this PR?

It selects the whole widget:

bug_cke5

Reinmar:
Pity. But I'm afraid that this all we can do for now.

Reinmar:
> It seems that this one fixes ckeditor/ckeditor5#562 but unfortunately, ckeditor/ckeditor5#611 is still reproducible.

After my latest changes this issue is fixed too :)

jun-18-2018 13-31-55

Reinmar:
OK, the biggest issue I see is the one I mentioned in https://github.com/ckeditor/ckeditor5-engine/pull/1431#issuecomment-397990090. I think we'll need to work on it soon. But it's not a regression, cause on master this is bad too, so we're fine to move forward.

PR #1449, opened 2018-06-27
Element to attribute upcast should set attribute on all the elements inside the converted element

scofalik:

Fix: Element to attribute upcast should set attribute on all the elements inside the converted element.

---

### Additional information

This PR provides a fix after tests started to fail in `ckedior5-basic-styles` after merging #1443. As of now, element to attribute upcast will set attribute on all children of the element, while attribute to attribute upcast will set the attribute only on the element.

So:

`<strong><p>Foo</p></strong>` will convert to `<paragraph><$text bold="true">Foo</$text></paragraph>`

while (assuming a plugin handling `class` attribute):

`<div class="dark"><div>Foo</div></div>` will convert to `<div class="dark"><div>Foo</div></div>`

even though inside `<div>` could accept `class` attribute dark.

### DIFF
--- a/src/conversion/upcast-converters.js
+++ b/src/conversion/upcast-converters.js
@@ -74,6 +74,14 @@ export function upcastElementToElement( config ) {
  * This conversion results in setting an attribute on a model node. For example, view `<strong>Foo</strong>` becomes
  * `Foo` {@link module:engine/model/text~Text model text node} with `bold` attribute set to `true`.
  *
+ * This helper is meant to set a model attribute on all the elements that are inside the converted element:
+ *
+ *		<strong>Foo</strong>   -->   <strong><p>Foo</p></strong>   -->   <paragraph><$text bold="true">Foo</$text></paragraph>
+ *
+ * Above is a sample of HTML code, that goes through autoparagraphing (first step) and then is converted (second step).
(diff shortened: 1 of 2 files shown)


scofalik:
To clarify:

In upcasting we use `data.modeRange` which describes on which part of the model we operate. When upcasting to attribute we use that range to set the model attribute "on that range".

Before any fix, the model attribute was applied to all elements in that range. So this incorrect situation happened:

`<div class="xyz"><div>Foo</div></div>` -> `<div class="xyz"><div class="xyz">Foo</div></div>`

So, in the fix, I changed it the way that only top-level elements receive the attribute:

`<div class="xyz"><div>Foo</div></div>` -> `<div class="xyz"><div>Foo</div></div>`

This caused a regression in following scenario in autoparagraphing:

`<strong>Foo</strong>` -> `<strong><p>Foo</p></strong>` -> `<paragraph>Foo</paragraph>`

because `bold` attribute has been tried to be set only on `<paragraph>` not on the text.

So I made another fix. Now, if element-to-attribute upcast is used, it works as earlier, so everything in the range gets the attribute. It makes sense. After all, all the elements were inside the upcasted element in the original view.

However, if there is attribute-to-attribute conversion, it is assumed, that the view attribute is connected only with its element, so the attribute is set only on top-level node.

Issue #1440, opened 2018-06-19
[FF] Complex selection when flashes spanning across a widget


oleq:
A followup of https://github.com/ckeditor/ckeditor5-engine/pull/1431#issuecomment-396207242.

![kapture 2018-06-19 at 11 39 26](https://user-images.githubusercontent.com/1099479/41598872-d1413f60-73d1-11e8-8dae-19883844e05d.gif)

Reinmar:
I checked how the selection post-fixer fixes the selection:

![image](https://user-images.githubusercontent.com/156149/42282876-2554a718-7fa8-11e8-81a8-4fb44e41320b.png)

The fixed ranges ("before combining") are ok. The thing is, though, that they don't get combined. We select two ranges and I think that it may be breking Firefox. Those not combined ranges look reasonably, but still – Firefox starts doing weird things.

Another thing is that the whole selection is blinking. I think that we may need to stop re-rendering while it's being "done". E.g. I changed `toWidget()` so it stops setting `cE=false`. We stopped having the two-range selection problem but blinking was severe. So we'll have to fix two things here – setting a reasonable selection and avoiding blinking.

Reinmar:
I made a quick fix which better merges ranges:

![image](https://user-images.githubusercontent.com/156149/42283296-55c8c31a-7fa9-11e8-9219-4c6695a51242.png)

It's not valid cause it looks like this now:

```js
		if ( combinedRanges.length > 1 ) {
			const r = new Range( combinedRanges[ 0 ].start, combinedRanges[ 1 ].end );
			combinedRanges = [ r ];
		}
```

But it prevented breaking the selection.

![jul-04-2018 16-44-44](https://user-images.githubusercontent.com/156149/42283400-9bef3248-7fa9-11e8-9745-d9d903645e5b.gif)

Reinmar:
In fact... I think that we can do a quick fix here to always combine multi-range selections into a single-range selection. This only applies to Firefox and if it won't break anything in tables (which I think it won't), then we'll at least prevent breaking the selection. It will still blink, but we can handle this separately.

jodator:
@Reinmar The proposed fix looks OK in this scenario. What about actual mutli-range selection?

I think that we cannot blindly merge mutli selection into one big selection:

1. Select some parts of text before image:
![selection_002](https://user-images.githubusercontent.com/247363/42309600-56dea0e0-8039-11e8-830f-5ca13722312c.png)

[...]

PR #1450, opened 2018-06-28
T/1436: Fixing selection that cross limit node boundaries.

### LINKED ISSUE #1436: Automatic selection post-fixer – part 2
When reviewing https://github.com/ckeditor/ckeditor5-engine/pull/1431 I spotted a couple of issues which we weren't able to fix back then. Mainly – element selection of non-objects and correct way of fixing selections crossing limit boundaries.

jodator:

Other: Fixing selection that cross limit node boundaries. Closes #1436.

Other: The `schema.getLimitElement()` method now accepts also `Range` and `Position` as a parameter.

---

### Additional information

This PR comes with some fixes in other repos all branches are available at https://github.com/ckeditor/ckeditor5/tree/%23t/ckeditor5-engine/1436:

* https://github.com/ckeditor/ckeditor5-enter/pull/55
* https://github.com/ckeditor/ckeditor5-list/pull/106
* https://github.com/ckeditor/ckeditor5-table/pull/81
* https://github.com/ckeditor/ckeditor5-undo/pull/87
* https://github.com/ckeditor/ckeditor5-widget/pull/45

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -582,14 +582,16 @@ export default class Schema {
 		}, { priority: 'high' } );
 	}
 
+	/* eslint-disable max-len */
 	/**
 	 * Returns the lowest {@link module:engine/model/schema~Schema#isLimit limit element} containing the entire
 	 * selection or the root otherwise.
 	 *
-	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection} selectionOrRangeOrPosition
+	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection|module:engine/model/range~Range|module:engine/model/position~Position} selectionOrRangeOrPosition
 	 * Selection which returns the common ancestor.
 	 * @returns {module:engine/model/element~Element}
 	 */
+	/* eslint-enable max-len */
 	getLimitElement( selectionOrRangeOrPosition ) {
 		let element;
 
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -586,22 +586,32 @@ export default class Schema {
 	 * Returns the lowest {@link module:engine/model/schema~Schema#isLimit limit element} containing the entire
 	 * selection or the root otherwise.
 	 *
-	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection} selection
+	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection} selectionOrRangeOrPosition
 	 * Selection which returns the common ancestor.
 	 * @returns {module:engine/model/element~Element}
 	 */
-	getLimitElement( selection ) {
-		// Find the common ancestor for all selection's ranges.
-		let element = Array.from( selection.getRanges() )
-			.reduce( ( element, range ) => {
-				const rangeCommonAncestor = range.getCommonAncestor();
+	getLimitElement( selectionOrRangeOrPosition ) {
+		let element;
 
-				if ( !element ) {
-					return rangeCommonAncestor;
-				}
+		if ( selectionOrRangeOrPosition instanceof Position ) {
+			element = selectionOrRangeOrPosition.parent;
+		} else {
+			const ranges = selectionOrRangeOrPosition instanceof Range ?
+				[ selectionOrRangeOrPosition ] :
+				Array.from( selectionOrRangeOrPosition.getRanges() );
 
-				return element.getCommonAncestor( rangeCommonAncestor, { includeSelf: true } );
-			}, null );
+			// Find the common ancestor for all selection's ranges.
+			element = ranges
+				.reduce( ( element, range ) => {
+					const rangeCommonAncestor = range.getCommonAncestor();
+
+					if ( !element ) {
+						return rangeCommonAncestor;
+					}
+
+					return element.getCommonAncestor( rangeCommonAncestor, { includeSelf: true } );
+				}, null );
+		}
 
 		while ( !this.isLimit( element ) ) {
 			if ( element.parent ) {
--- a/src/model/utils/selection-post-fixer.js
+++ b/src/model/utils/selection-post-fixer.js
@@ -94,9 +94,18 @@ function selectionPostFixer( writer, model ) {
 	if ( wasFixed ) {
 		// The above algorithm might create ranges that intersects each other when selection contains more then one range.
 		// This is case happens mostly on Firefox which creates multiple ranges for selected table.
-		const combinedRanges = combineOverlappingRanges( ranges );
+		let fixedRanges = ranges;
 
-		writer.setSelection( combinedRanges, { backward: selection.isBackward } );
+		// Fixing selection with many ranges usually breaks the selection in Firefox. As only Firefox supports multiple selection ranges
+		// we simply create one continuous range from fixed selection ranges (even if they are not adjacent).
+		if ( ranges.length > 1 ) {
+			const selectionStart = ranges[ 0 ].start;
+			const selectionEnd = ranges[ ranges.length - 1 ].end;
+
+			fixedRanges = [ new Range( selectionStart, selectionEnd ) ];
+		}
+
+		writer.setSelection( fixedRanges, { backward: selection.isBackward } );
 	}
 }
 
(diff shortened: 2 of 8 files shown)


Reinmar:
Two conditions are not covered when running just the post-fixer's tests: 

image

src/model/utils/selection-post-fixer.js
```diff
@@ -146,27 +146,87 @@ function tryFixingCollapsedRange( range, schema ) {
...
 // @param {module:engine/model/schema~Schema} schema
 // @returns {module:engine/model/range~Range|null} Returns fixed range or null if range is valid.
-function tryFixingNonCollpasedRage( range, schema ) {
-	// No need to check flat ranges as they will not cross node boundary.
-	if ( range.isFlat ) {
+function tryFixingNonCollapsedRage( range, schema ) {
+	const start = range.start;
+	const end = range.end;
+
+	// Flat range on the same text node is always valid - no need to fix:
```
Reinmar: I think that this comment isn't entirely correct – I mean the "same" part. What about:

```html
<p>f[oo<inlineWidget/>ba]r</p>
```

Also, what about:

```html
<p>[foo]</p>
```

In this case, neither the start's nor end's `textNode` getter will return something.

And there's one more case which is missed here:

```html
<p>[<inlineWidget/>foo<inlineWidget/>]</p>
```

You might've had a different goal by adding this condition – filtering some cases which might be broken by the following logic of this function. But, if the logic of such a check isn't "complete" in a bigger picture, a wider context of the entire algorithm, then it may be misleading. So, instead of patching this algorithm for a hidden purpose, let's thing wider.

E.g. what kind of non-collapsed selections don't need to be fixed? Those which meet the following criteria (all):

* `schema.getLimitElement( range.start ) == schema.getLimitElement( range.end )` – it means the selection starts and ends in the same limit element, so it does not cross any limit element boundary. If a range violates this rule, it needs to be extended to cover the lowest common limit element of both its ends (`schem.getLimitElement( range )`).

   Note: Right now, `schema.getLimitElement()` works only with entire selections, but it's not hard to "downgrade" it to support single ranges (returns LCA like for the entire selection) and single positions (returns a lowest limit element in which this position is contained) too.
* `schema.checkChild( range.start, '$text' ) && schema.checkChild( range.end, '$text' )` – both its ends are in a place where text is allowed. They don't need to be placed in/next to a text (`<p>[<inlineWidget/>...`) but it's enough that text is allowed there.

   Note: A range containing a block widget will not be rooted in a place where text is allowed, so it will pass this check. However, it doesn't mean it's incorrect – I was describing a range which is certainly correct (i.e. a possible early return check), not an algorithm for deciding whether a range is correct or not.

Reinmar: By structuring the algorithm as I described above, the rules by which it works will be easier to read. Right now, it consists of "let's try fixing this, let's try that" which is harder for me to digest.

tests/model/schema.js
```diff
@@ -1123,12 +1123,12 @@ describe( 'Schema', () => {
 			schema.extend( 'img', { allowAttributes: 'bold' } );
 			schema.extend( '$text', { allowIn: 'img' } );
 
-			setData( model, '[<p>foo<img>xxx</img>bar</p>]' );
```
Reinmar: This changes the semantics of this test. Let's better use `parse()`, especially that later on we do `selection.getRanges()` anyway.

tests/model/utils/selection-post-fixer.js
```diff
@@ -59,37 +63,41 @@ describe( 'Selection post-fixer', () => {
 		} );
 
 		it( 'should react to structure changes', () => {
+			setModelData( model, '<paragraph>[]foo</paragraph><image></image>' );
```
Reinmar: What was wrong with this test?

Reinmar: Ah, ok, you're simplifying the test.

Reinmar:
OK, so far, the most important bit is my comment about structuring this algorithm. I think you could try to restructure it so the rules by which it works are more visible. I *think* I described the outline in my comment, but of course that's just the outline. You also need the "fix it" parts. In general, it should look more or less like this:

1. check something
  * if it's wrong fix it; if possible return here; if not, continue;
2. check something else
  * if it's wrong fix it; if possible return here; if not, continue;
3. ... and so on

jodator:
@Reinmar ready to re-review.

src/model/schema.js
```diff
@@ -586,22 +586,32 @@ export default class Schema {
 	 * Returns the lowest {@link module:engine/model/schema~Schema#isLimit limit element} containing the entire
 	 * selection or the root otherwise.
 	 *
-	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection} selection
+	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection} selectionOrRangeOrPosition
```
Reinmar: You forgot to update the type.

tests/conversion/downcast-selection-converters.js
```diff
@@ -542,7 +542,7 @@ describe( 'downcast-selection-converters', () => {
 		it( 'should add a class to the selected table cell', () => {
 			test(
 				// table tr#0 |td#0, table tr#0 td#0|
```
Reinmar: Are the comments up to date now?

tests/conversion/downcast-selection-converters.js
```diff
@@ -551,9 +551,9 @@ describe( 'downcast-selection-converters', () => {
 		it( 'should not be used if selection contains more than just a table cell', () => {
 			test(
 				// table tr td#1, table tr#2
```
Reinmar: And here too?

tests/model/schema.js
```diff
@@ -983,6 +983,28 @@ describe( 'Schema', () => {
 
 			expect( schema.getLimitElement( doc.selection ) ).to.equal( root );
 		} );
+
+		it( 'accepts range as an argument', () => {
+			schema.extend( 'article', { isLimit: true } );
+			schema.extend( 'section', { isLimit: true } );
+
+			setData( model, '<div><section><article><paragraph>foo[]bar</paragraph></article></section></div>' );
```
Reinmar: Why do you set a selection anywhere if you're testing if this method works with a range?

You can use `parse()` here, btw.

tests/model/schema.js
```diff
@@ -1101,7 +1123,13 @@ describe( 'Schema', () => {
 				}
 			} );
 
-			setData( model, '<p>foo<img />bar</p>' );
+			// Parse data string to model.
+			const parsedResult = setData._parse( '[<p>foo<img />bar</p>]', model.schema, { context: [ root.name ] } );
```
Reinmar: Use the public export.

tests/model/schema.js
Reinmar: This is still not resolved.

tests/model/utils/deletecontent.js
```diff
@@ -621,7 +621,7 @@ describe( 'DataController utils', () => {
 				deleteContent( model, selection );
 
 				expect( getData( model, { rootName: 'bodyRoot' } ) )
-					.to.equal( '[<paragraph>x</paragraph>]<paragraph></paragraph><paragraph>z</paragraph>' );
+					.to.equal( '<paragraph>[x]</paragraph><paragraph></paragraph><paragraph>z</paragraph>' );
```
Reinmar: You can change the initial selection too, so this test is less confusing (we're not checking the sel post-fixer here).

tests/model/utils/deletecontent.js
```diff
@@ -644,7 +644,16 @@ describe( 'DataController utils', () => {
...
 				);
 
-				deleteContent( model, doc.selection );
+				model.change( writer => {
+					// Set selection to[<heading1>yyy</heading1>] in change() block due to selection post-fixer.
+					const range = new Range(
+						new Position( doc.getRoot( 'bodyRoot' ), [ 1 ] ),
+						new Position( doc.getRoot( 'bodyRoot' ), [ 2 ] )
+					);
+					writer.setSelection( range );
```
Reinmar: You don't have to set that selection. `deleteContent()` can work with any selection.

tests/model/utils/deletecontent.js
```diff
@@ -657,7 +666,16 @@ describe( 'DataController utils', () => {
...
 				);
 
-				deleteContent( model, doc.selection );
+				model.change( writer => {
+					// Set selection to[<heading1>yyy</heading1><paragraph>yyy</paragraph>] in change() block due to selection post-fixer.
+					const range = new Range(
+						new Position( doc.getRoot( 'bodyRoot' ), [ 1 ] ),
+						new Position( doc.getRoot( 'bodyRoot' ), [ 3 ] )
+					);
+					writer.setSelection( range );
```
Reinmar: The same as above– no need to set it. Also, it'd be good if you changed it in the `setData()` call above to something irrelevant to this test to not create a confusion.

tests/model/utils/deletecontent.js
Reinmar: Also, it'd be good if you changed it in the `setData()` call above to something irrelevant to this test to not create a confusion.

tests/model/schema.js
```diff
@@ -1101,7 +1123,13 @@ describe( 'Schema', () => {
...
+
+			// Set parsed model data to prevent selection post-fixer from running.
+			model.change( writer => {
+				writer.insert( parsedResult.model, root );
```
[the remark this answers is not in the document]
jodator: I don't really know how to do that  - AFAICS it's easiest to just use `writer.insert( node, root)`...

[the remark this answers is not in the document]
Reinmar: The method we're testing  there `schema.getValidRanges()` doesn't need to work on on a piece of a tree which is in the model. At least – I think that doesn't need that. It should be able to work on a detached tree. 

OTOH, we can leave  it with the `insert()` and test it just like it will be used most of the time (in an attached tree).

jodator:
@Reinmar the fix is on the branch already.

Reinmar:
🎉

jodator:
@Mgsy As this looks like finished could you check how the selection on widgets behaves now?

Also this one fixes selection on images in firefox: https://github.com/ckeditor/ckeditor5-engine/issues/1440 (not the flashing part though - only the selection problems).

Also on FF the multi-range selection that is fixed (ie on widgets) should be merge into one big (https://github.com/ckeditor/ckeditor5-engine/issues/1440#issuecomment-402653775).

Mgsy:
Basically, it works fine (except Edge, where a widget steals the selection ) 👌 

There is a problem with nested editables, see caption in **Chrome**:

bug_cke5

or table in **Firefox**

bug_cke5

jodator:
> There is a problem with nested editables, see the caption in Chrome:

@Mgsy you mean that it blinks? If so I don't think we could do anything in particular in this PR as this is a bit different issue (cc @Reinmar) - we need to stop selection updates somehow do prevent this. I think that a follow-up is needed here.

> or the table in Firefox:

It looks OK (selecting whole table) - could you describe what you mean here since I've might be missing something from the GIF.

Reinmar:
Agree with @jodator regarding the first part – we need to extract this to a separate ticket.

Mgsy:
>It looks OK (selecting whole table) - could you describe what you mean here since I've might be missing something from the GIF.

Yep, that is what I meant - selecting the whole table. As this behaviour is expected, the whole fix looks good to me 👌

PR #1452, opened 2018-06-29
Finished refactoring by removing unnecessary code

### LINKED ISSUE #1451: Unnecessary logic in Renderer#render()
After yesterday's [small refactoring in the renderer](https://github.com/ckeditor/ckeditor5-engine/commit/601f5f450c69000d4e6c1e2b2662528c8aa1154d) a CC dropped in the renderer:

image

The change I did yesterday was supposed to not touch any logic but apparently, it did. Namely, I think that `_updateChildren()` started rendering the filler itself. I tried to understand why it didn't render the filler previously but the code got complicated before the refactoring and I have problems tracking it down.

Anyway, it seems completely accidental that it didn't do it previously. According to the logic, `_diffChildren()` got the filler position as a param and so it was injecting filler into the `expectedDomChildren`. However, the `_updateChildren()` method didn't use the `expectedDomChildren` returned by `_diffChildren()` but another array which it created earlier... That array should also contain the filler, but maybe the duplication caused some problems. IDK. But the logic didn't make sense, for sure.

Now it seems that `_updateChildren()` and `_updateText()` always take care of rendering the filler by themselves. It seems logical that there's no need to render the filler in `render()` if these two functions took care of everything.

Reinmar:

Internal: Finished refactoring (done in the previous commit) by removing unnecessary code. Closes #1451.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -206,25 +206,12 @@ export default class Renderer {
 			}
 		}
 
-		// Check whether the inline filler is required and where it really is in the DOM.
-		// At this point in most cases it will be in the DOM, but there are exceptions.
-		// For example, if the inline filler was deep in the created DOM structure, it will not be created.
-		// Similarly, if it was removed at the beginning of this function and then neither text nor children were updated,
-		// it will not be present.
-		// Fix those and similar scenarios.
+		// Check whether the inline filler is required and where it really got rendered in the DOM.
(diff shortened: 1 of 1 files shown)


Reinmar:
R- due to https://github.com/ckeditor/ckeditor5-engine/issues/1451#issuecomment-401330426

PR #1454, opened 2018-07-02
Expand selection will work correctly with inline elements

pomek:

Fix: Expand selection mechanism will work correctly with the inline elements. Closes ckeditor/ckeditor5#1064.

---

An integration test: https://github.com/ckeditor/ckeditor5-typing/pull/164

### DIFF
--- a/src/model/utils/modifyselection.js
+++ b/src/model/utils/modifyselection.js
@@ -172,7 +172,7 @@ function getCorrectWordBreakPosition( walker, isForward ) {
 			// should expand to : 'foofoo [bar<$text bold="true">bar</$text>] bazbaz'.
 			const nextNode = isForward ? walker.position.nodeAfter : walker.position.nodeBefore;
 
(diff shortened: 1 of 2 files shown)


src/model/utils/modifyselection.js
```diff
@@ -172,7 +172,8 @@ function getCorrectWordBreakPosition( walker, isForward ) {
 			// should expand to : 'foofoo [bar<$text bold="true">bar</$text>] bazbaz'.
 			const nextNode = isForward ? walker.position.nodeAfter : walker.position.nodeBefore;
 
-			if ( nextNode ) {
+			// `nextNode` can be an inline element which is an invalid node in this case.
```
Reinmar: More like:

> Scan only text nodes. Ignore inline elements (like `<softBreak>`).

Reinmar: The "invalid" part of your comment is kinda confusing – invalid means e.g. that a node cannot exist in that place. Not a big deal, but it can be written better.

PR #1455, opened 2018-07-02
Unit test covering edge case related to 'withChildren' parameter inside Renderer added

### LINKED ISSUE #1425: Always render text nodes via 'renderer._updateText'
Extracted from https://github.com/ckeditor/ckeditor5-engine/issues/1417#issuecomment-389191257.

The simplest case to check this behaviour:

1. Open editor instance, could be e.g. http://localhost:8125/ckeditor5-core/tests/manual/article.html.
1. Select one character on the end of the paragraph like `<p>Paragrap[h]</p>`.
1. Type something.

Paragraph is marked to update in the `renderer` and `renderer._updateChildren` replaces the whole text node with the new one (containing modified text) - no matter how long text is, so it happens also for very long paragraphs. Since we have introduced smart text nodes updates (#403) it will be nice to utilize it in such cases too.

The main benefit should be visible during composition with collaboration. That's the reason I mentioned paragraph length. Having a long one may result in a few people composing inside it at once which will break quite quickly.

There are also similar cases (like bolding part of the text) which also may utilize smart text nodes updates. Case with bold is already described in the linked issue.

---

I'm not sure if this issue should be solved by the `renderer` itself or if it would be possible to adjust the type of elements which are marked to update in a `View` somewhere.

### LINKED ISSUE #1451: Unnecessary logic in Renderer#render()

f1ames:

Internal: Added `withChildren: false` parameter missing after refactoring and unit test covering that case. Closes #1451. Closes #1452.

---

### Additional information

What happens depending on `withChildren` value is described in more details in https://github.com/ckeditor/ckeditor5-engine/issues/1451#issuecomment-401356437. Added unit test shows the situation in which `withChildren: true` behaviour results in incorrect DOM structure after rendering.

I have added `Closes #1425` as #1452 is a previous PR related to this topic, so if we decided to merge this one, #1452 should be closed.

### DIFF
--- a/tests/view/renderer.js
+++ b/tests/view/renderer.js
@@ -1683,6 +1683,36 @@ describe( 'Renderer', () => {
 			);
 		} );
 
+		// #1451
+		it( 'should correctly render changed children even if direct parent is not marked to sync', () => {
+			const inputView =
+				'<container:ul>' +
+					'<container:li>Foo</container:li>' +
+					'<container:li><attribute:b>Bar</attribute:b></container:li>' +
+				'</container:ul>';
+
+			const view = parse( inputView );
+
+			viewRoot._appendChild( view );
+
+			renderer.markToSync( 'children', viewRoot );
+			renderer.render();
+
+			expect( domRoot.innerHTML ).to.equal( '<ul><li>Foo</li><li><b>Bar</b></li></ul>' );
+
+			const viewLi = view.getChild( 0 );
+			const viewLiIndented = view._removeChildren( 1, 1 ); // Array with one element.
+			viewLiIndented[ 0 ]._appendChild( parse( '<attribute:i>Baz</attribute:i>' ) );
+			const viewUl = new ViewContainerElement( 'ul', null, viewLiIndented );
+			viewLi._appendChild( viewUl );
+
+			renderer.markToSync( 'children', view );
+			renderer.markToSync( 'children', viewLi );
+			renderer.render();
+
+			expect( domRoot.innerHTML ).to.equal( '<ul><li>Foo<ul><li><b>Bar</b><i>Baz</i></li></ul></li></ul>' );
+		} );
+
 		describe( 'fake selection', () => {
 			beforeEach( () => {
 				const { view: viewP, selection: newSelection } = parse(
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -256,7 +256,7 @@ export default class Renderer {
 
 		const actualDomChildren = this.domConverter.mapViewToDom( viewElement ).childNodes;
 		const expectedDomChildren = Array.from(
-			this.domConverter.viewChildrenToDom( viewElement, domElement.ownerDocument )
+			this.domConverter.viewChildrenToDom( viewElement, domElement.ownerDocument, { withChildren: false } )
 		);
 		const diff = this._diffNodeLists( actualDomChildren, expectedDomChildren );
 		const actions = this._findReplaceActions( diff, actualDomChildren, expectedDomChildren );


tests/view/renderer.js
```diff
@@ -1683,6 +1683,36 @@ describe( 'Renderer', () => {
...
+			const view = parse( inputView );
+
+			viewRoot._appendChild( view );
+
+			renderer.markToSync( 'children', viewRoot );
+			renderer.render();
+
+			expect( domRoot.innerHTML ).to.equal( '<ul><li>Foo</li><li><b>Bar</b></li></ul>' );
+
+			const viewLi = view.getChild( 0 );
```
Reinmar: `viewLi` of `viewUl`?

Reinmar: Uh, `view` == `<ul>`. Sorry.

tests/view/renderer.js
```diff
@@ -1683,6 +1683,36 @@ describe( 'Renderer', () => {
...
+			const viewLiIndented = view._removeChildren( 1, 1 ); // Array with one element.
```
Reinmar: I don't understand what this line does. Which `<li>` does it remove? There's no indented `<li>` at this stage.

tests/view/renderer.js
```diff
@@ -1683,6 +1683,36 @@ describe( 'Renderer', () => {
...
+			viewLiIndented[ 0 ]._appendChild( parse( '<attribute:i>Baz</attribute:i>' ) );
```
Reinmar: Isn't the problem here that we use `appendChild()` on a detached node? That breaks automatic `markToSync()` calls, am I right?

I have a bad feeling about this – should renderer handle such cases? Shouldn't it be able to assume that `markToSync` is called for all nodes which are finally in the tree? That seems much easier than having `markToSync` used for half of the nodes in the tree.

cc @pjasiun

tests/view/renderer.js
```diff
@@ -1683,6 +1683,36 @@ describe( 'Renderer', () => {
 			);
 		} );
 
+		// #1451
+		it( 'should correctly render changed children even if direct parent is not marked to sync', () => {
```
Reinmar: This test is a bit too hard to read. Better variable naming + a comment what's happening inside it would help.

tests/view/renderer.js
pjasiun: Yes, but it is `View` class who binds the view structure and the renderer. I think that render should be tested together with the view and even with the view writer. Integration tests FTW! ;)

Reinmar:
I extracted the issue I found to https://github.com/ckeditor/ckeditor5-engine/issues/1456. Let's fix CC here and move forward because what we found is that we need to redesign this part of the engine :D

PR #1459, opened 2018-07-09
OT refactor

scofalik:

Other: Removed concept of deltas. Added new operations. Rewritten OT algorithms.

BREAKING CHANGE: Everything connected with deltas was removed from the editor.
BREAKING CHANGE: `model.Writer#setAttribute` (and `AttributeOperation`) now applies attribute only to the top-level nodes in the `range` (instead of all the nodes in the range).


src/model/operation/operation.js
```diff
@@ -48,11 +50,11 @@ export default class Operation {
...
-		 * {@link module:engine/model/delta/delta~Delta Delta} which the operation is a part of. This property is set by the
-		 * {@link module:engine/model/delta/delta~Delta delta} when the operations is added to it by the
-		 * {@link module:engine/model/delta/delta~Delta#addOperation} method.
+		 * {@link module:engine/model/operation/operation~Operation Operation} which the operation is a part of. This property is set by the
+		 * {@link module:engine/model/operation/operation~Operation operation} when the operations is added to it by the
+		 * {@link module:engine/model/operation/operation~Operation#addOperation} method.
 		 *
-		 * @member {module:engine/model/delta/delta~Delta} #delta
+		 * @member {module:engine/model/operation/operation~Operation} #operation
 		 */
```
oskarwrobel: Should be removed.

src/model/operation/operation.js
```diff
@@ -40,6 +40,8 @@ export default class Operation {
...
 		this.isDocumentOperation = this.baseVersion !== null;
 
+		this.batch = null;
```
oskarwrobel: Docs should be enough.

src/dev-utils/enableenginedebug.js
```diff
@@ -364,136 +353,25 @@ function enableLoggingTools() {
...
-
-		this.history = history.concat( itemToSave );
-	} );
-
-	const _deltaTransformTransform = deltaTransform.transform;
+	const _transformTransform = transform.transform;
 
-	sandbox.mock( deltaTransform, 'transform', function( a, b, context ) {
+	sandbox.mock( transform, 'transform', function( a, b, context ) {
+		debugger;
```
pjasiun: debugger

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+		);
+
+		result.push(
+			new AttributeOperation( range, attributeOperation.key, val, attributeOperation.newValue, 0 )
+		);
+	}
+}
+
+setTransformation( AttributeOperation, MergeOperation, ( a, b ) => {
+	debugger;
```
pjasiun: debugger

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+		padWithNoOps( operationsB, data.extraOpsA - data.extraOpsB );
+	}
+
+	updateBaseVersions( operationsA, data.nextBaseVersionB );
+	updateBaseVersions( operationsB, data.nextBaseVersionA );
+
+	return { operationsA, operationsB };
+}
+
+export default {
```
pjasiun: Why not export these 2 functions?

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+
+function updateBaseVersions( operations, baseVersion ) {
+	for ( const operation of operations ) {
+		operation.baseVersion = baseVersion++;
+	}
+
+	return operations;
+}
+
+function transform( a, b, context = { isStrong: false } ) {
```
pjasiun: It is not strong tranform, but A is isStrong, the parameter name is missleading.

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+	for ( const op of opsA.concat( opsB ) ) {
+		context.originalOperations.set( op, op );
+	}
+
+	context.relations = new Map();
+
+	context.wasUndone = function( op ) {
+		const originalOp = this.originalOperations.get( op );
+
+		return this.document.history.isUndoneOperation( originalOp );
```
pjasiun: Can we move `isUndoneOperation` map to scope of this file?

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+
+		if ( relationsA ) {
+			return relationsA.get( undoneB ) || null;
+		}
+
+		return null;
+	};
+}
+
+function updateRelations( context, opA, opB ) {
```
pjasiun: That these relations are used for?

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+		if ( group.has( OperationB ) ) {
+			return group.get( OperationB );
+		}
+
+		for ( const Operation of group.keys() ) {
+			if ( b instanceof Operation ) {
+				return group.get( Operation );
+			}
+		}
+	}
```
pjasiun: It would be good to simplify it because of the performance.

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+
+	for ( let i = 0; i < operationsA.length; ) {
+		const opsA = [ operationsA[ i ] ];
+
+		for ( let j = 0; j < operationsB.length; ) {
+			const opsB = [ operationsB[ j ] ];
+
+			for ( let k = 0; k < opsA.length; ) {
+				for ( let l = 0; l < opsB.length; ) {
```
pjasiun: A comment why we need 2 loops.

src/model/operation/transform.js
pjasiun: It should be nice to move try-catch to this functions and export it in the normal way.

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+		aWasUndone: context.wasUndone( opA ),
+		bWasUndone: context.wasUndone( opB ),
+		relation: context.getRelation( opA, opB )
+	};
+
+	const contextBA = {
+		isStrong: false,
+		aWasUndone: contextAB.bWasUndone,
+		bWasUndone: contextAB.aWasUndone,
+		relation: context.getRelation( opB, opA )
```
pjasiun: All of these could be calculated on demand.

src/model/operation/transform.js
pjasiun: The parameter name is not precise. It says which operation - a or b - is strong. Current name might suggest that the transformation itself is strong somehow.

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+	transformSets
+};
+
+function padWithNoOps( operations, howMany ) {
+	for ( let i = 0; i < howMany; i++ ) {
+		operations.push( new NoOperation( 0 ) );
+	}
+}
+
+function getCurrentContext( context, opA, opB ) {
```
pjasiun: We could have only the global context.

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+
+	if ( !relationsA ) {
+		relationsA = new Map();
+		context.relations.set( origA, relationsA );
+	}
+
+	relationsA.set( origB, relation );
+}
+
+function updateOriginalOperations( context, oldOp, newOps ) {
```
pjasiun: `updateOriginalOperation`

src/model/operation/transform.js
```diff
@@ -1,644 +1,1895 @@
...
+	for ( const op of newOps ) {
+		context.originalOperations.set( op, originalOp );
+	}
+}
+
+function getNoOp() {
+	return [ new NoOperation( 0 ) ];
+}
+
+function _onSameLevel( aPos, bPos ) {
```
pjasiun: Move to position.

src/model/liverange.js
```diff
@@ -146,71 +141,42 @@ function bindWithDocument() {
...
+		this.start = result.start;
+		this.end = result.end;
 
-		this.fire( 'change:range', oldRange, {
-			type: changeType,
-			batch,
-			range: targetRange,
-			sourcePosition
-		} );
+		this.fire( 'change:range', oldRange, operation );
```
pjasiun: It would be nice to limit the number of data we fire in this event. Fire `data` with only necessary data instead of the operation.

src/model/operation/mergeoperation.js
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * @extends module:engine/model/operation/operation~Operation
+ */
+export default class MergeOperation extends Operation {
+	/**
+	 * Creates a merge operation.
+	 *
+	 * @param {module:engine/model/position~Position} sourcePosition Position inside the merged element. All nodes from that
+	 * element after that position will be moved to {@link ~#targetPosition}.
+	 * @param {module:engine/model/position~Position} targetPosition Position which the nodes from the merged elements will be moved to.
```
pjasiun: Alternatively, this could be positions before merged element and after to-merge element.

src/model/liverange.js
pjasiun: If we will keep it this way changes in opeartion will have to be marked as breaking changes.

src/model/operation/utils.js
```diff
@@ -133,7 +133,7 @@ export function _setAttribute( range, key, value ) {
 	_splitNodeAtPosition( range.end );
 
 	// Iterate over all items in the range.
-	for ( const item of range.getItems() ) {
+	for ( const item of range.getItems( { shallow: true } ) ) {
```
pjasiun: See https://github.com/ckeditor/ckeditor5-engine/issues/1460

src/model/operation/wrapoperation.js
```diff
@@ -0,0 +1,179 @@
...
+
+		this.graveyardPosition = elementOrPosition instanceof Element ? null : Position.createFromPosition( elementOrPosition );
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	get type() {
+		return 'wrap';
+	}
```
pjasiun: Consider if we still need types of opperations.

src/model/range.js
```diff
@@ -479,102 +469,58 @@ export default class Range {
...
-				( sourceRange.containsPosition( this.start ) || sourceRange.start.isEqual( this.start ) ) &&
-				this.containsPosition( sourceRange.end ) &&
-				this.end.isAfter( targetPosition )
-			) {
-				const start = this.start._getCombined(
-					sourcePosition,
-					targetPosition._getTransformedByDeletion( sourcePosition, howMany )
-				);
-				const end = this.end._getTransformedByMove( sourcePosition, targetPosition, howMany, false, false );
+			return new Range( end, start );
```
pjasiun: There was `x` in the range before merge operation and `yz` in the range after the merge. Does not look like the correct result.

src/model/treewalker.js
```diff
@@ -90,6 +90,8 @@ export default class TreeWalker {
 			this.position = Position.createFromPosition( this.boundaries[ this.direction == 'backward' ? 'end' : 'start' ] );
 		}
 
+		this.position.stickiness = 'toNone';
```
pjasiun: Comment needed that it is to reset the stickiness.

src/model/operation/transform.js
scofalik: 👍

src/model/operation/transform.js
scofalik: 👍

src/model/operation/transform.js
scofalik: 👍

tests/model/operation/transform/attribute.js
```diff
@@ -0,0 +1,877 @@
...
+				john.setAttribute( 'bold', true );
+				kate.setAttribute( 'italic', true );
+
+				syncClients();
+
+				expectClients( '<paragraph><$text bold="true">Foo</$text> <$text italic="true">Bar</$text></paragraph>' );
+			} );
+
+			it( 'in text at same path, then undo', () => {
+                john.setData( '<paragraph>[Foo] Bar</paragraph>' );
```
pjasiun: Wrong indentation.

tests/model/operation/transform/attribute.js
```diff
@@ -0,0 +1,877 @@
...
+				kate.setAttribute( 'italic', true );
+
+				syncClients();
+
+				expectClients( '<paragraph><$text bold="true" italic="true">Foo Bar</$text></paragraph>' );
+			} );
+
+			it( 'in element at same position', () => {
+				john.setData( '[<paragraph>Foo Bar</paragraph>]' );
+				kate.setData( '[<paragraph>Foo Bar</paragraph>]' );
```
pjasiun: I am not sure about tests where the selection is around the paragraph. Is it possible to have such selection in the real case?

src/model/operation/transform.js
scofalik: If `isStrong` will have to be a function, it will be bad in case of `transform` function:

```javascript
function transform( a, b, context = { isStrong: false } ) {
	const transformationFunction = getTransformation( a, b );

	return transformationFunction( a.clone(), b, context );
}
```

And transformation units tests, which we will still need some.

tests/model/operation/transform/attribute.js
```diff
@@ -0,0 +1,877 @@
...
+
+				john.setAttribute( 'bold', true );
+				kate.setAttribute( 'italic', true );
+
+				syncClients();
+
+				expectClients( '<paragraph bold="true" italic="true">Foo Bar</paragraph>' );
```
pjasiun: Also the result is not really correct. We use attributes on empty paragraphs to leave the information that as soon as you start typing there the text should be bolded. However, if there was already text in such paragraph, the text should get the attribute, not the paragraph itself.

tests/model/operation/transform/attribute.js
scofalik: Selection is not a case here. It is used only to not create ranges by hand.

tests/model/operation/transform/attribute.js
scofalik: The name and the value of attributes do not matter. It could easily be `foo="bar"`.

src/model/operation/transform.js
scofalik: 👍

src/model/operation/transform.js
scofalik: 👍

src/model/operation/transform.js
scofalik: 👍

src/model/treewalker.js
scofalik: 👍

src/model/operation/wrapoperation.js
scofalik: I think we do, especially if we will join `RemoveOperation` and `ReinsertOperation` to `MoveOperation`.

tests/model/operation/transform/attribute.js
pjasiun: My point is, that it is very rare that we have a selection arount the block, it is more ofter to have a selection around texts, and tests should focus on such case, but I see more tests with `[<paragraph>Foo Bar</paragraph>]` than with `<paragraph>[Foo Bar]</paragraph>`.

scofalik: As I said it is not about the selection. What is needed is the range around `<paragraph>` not setting selection there. Selection is not even checked in the output data.

tests/model/operation/transform/utils.js
```diff
@@ -0,0 +1,307 @@
...
+		this.editor = null;
+		this.document = null;
+		this.syncedVersion = 0;
+		this.orderNumber = null;
+		this.name = name;
+	}
+
+	init() {
+		return ModelTestEditor.create( {
+			plugins: [ Typing, Paragraph, BoldEditing, ListEditing, UndoEditing, BlockQuoteEditing ]
```
pjasiun: Since you operate on the model, ignore schema, do you need to load these plugins?

scofalik: Not sure 🤔

scofalik:
I've pushed branches to other repositories that need to be updated:

```
basic-styles             │ ! t/new-ot         │ fcdc6fd |
block-quote              │ ! t/new-ot         │ fd065ba │
engine                   │ ! t/new-ot-w-tests │ add12f6 │
enter                    │ ! t/new-ot         │ b5d25d3 │
font                     │ ! t/new-ot         │ f674c8d │
heading                  │ ! t/new-ot         │ 58cef04 │
image                    │ ! t/new-ot         │ 7345fa3 │
list                     │ ! t/new-ot         │ 98be076 │
paragraph                │ ! t/new-ot         │ dad7d7e │
typing                   │ ! t/new-ot         │ af8f95c │
undo                     │ ! t/new-ot         │ 665322e │
```

pjasiun:
https://github.com/ckeditor/ckeditor5-undo/blob/665322ebfb27011ae2d18f9e8a775f74017fb8bf/src/undo.js#L33-L70

> [batch A]
> [batch B]
> [batch C]
...

PR #1478, opened 2018-07-27
Introduced relations in OT

scofalik:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Fix: Introduced relations in OT which helps solving conflicts during undo. Closes #1468.

PR #1480, opened 2018-07-27
Moved transformation error logging from debug tools to OT code

### LINKED ISSUE #1464: Move try-catch on transformation method from engine debug tools to transformations code

OT is a fragile part of our code which may crash as it is very difficult to come up with all the possible scenarios. Since now, in engine debug tools we had a mechanism that put a `try-catch` wrap around transformation function.

It may be better to move it straight to transformations code and have it active all the time, not only with engine debug enabled.

Also, there's a possibility that it will simplify some tests and we could export functions in `model/operation/transform.js` as functions, not as an object.

scofalik:

Other: Moved transformation error logging from debug tools to OT code. Closes #1464.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-engine/pull/1478


scofalik:
This PR removes 2 skipped tests 🎉

pjasiun:

PR #1482, opened 2018-07-27
Improved WrapOperation implementation

### LINKED ISSUE #1467: Make element cloning in WrapOperation as it is in InsertOperation
Both `WrapOperation` and `InsertOperation` clone the nodes which they insert to keep unchanged node saved in operation.

At the moment, `WrapOperation` just clones the element saved in the operation and inserts to the model tree the cloned element. This is a bit unfortunate, as you can't do this:

```javascript
const el = writer.createElement( 'myElement' );
writer.wrap( myRange, el );

// Do anything with `el`, assuming it is in the model tree.
const pos = Position.createAfter( el );
```

`InsertOperation` does it better. When executed, it clones the saved element, but it inserts the original element and saves the clone. `WrapOperation` should be done similarly.

scofalik:

Other: `WrapOperation` will save cloned version of wrapping element and insert the passed element into the model tree. Closes #1467.

### DIFF
--- a/src/model/operation/wrapoperation.js
+++ b/src/model/operation/wrapoperation.js
@@ -148,7 +148,10 @@ export default class WrapOperation extends Operation {
 		const targetPosition = new Position( this.position.root, targetPath );
 
 		if ( this.element ) {
-			_insert( insertPosition, this.element._clone() );
+			const originalElement = this.element;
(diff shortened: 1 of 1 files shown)


pjasiun:

PR #1483, opened 2018-07-30
Schema#getValidRanges returns flat ranges

### LINKED ISSUE #1460: Representation of the selection which contains object, but not children
This is tricky, I guess.

Consider image with the caption, with such model representation:
```
 <paragraph><image>Caption</image></paragraph>
```

What selection should be used to add `source` attribute to the image object but not to the caption text? So fat we use [0,0] - [0,0,0] (`<paragraph>[<image>]Caption</image></paragraph>`). This solution, however, has some flaws.
 - It is not intuitive. When recently @oskarwrobel saw such range, he thought that it is a bug.
 - It does not work well with markers. If you want to show the marker over the object you need to use [0,0] - [0,1] (`<paragraph>[<image>Caption</image>]</paragraph>`).

This is why we might consider different approach: [setting attribute](https://docs.ckeditor.com/ckeditor5/latest/api/module_engine_model_writer-Writer.html#function-setAttributes) on a range should do not touch children. We may have a parameter to set attribute in the `deep` mode. Then [0,0] - [0,1] (`<paragraph>[<image>Caption</image>]</paragraph>`) will be the proper and intuitive, range to set the source for the image.

On the other hand, the behavior of the `schema.getValidRanges` will be a little bit tricky then. For instance, if the schema allows to add the attribute only to `<image>` it should return `<paragraph>[<image>Caption</image>]</paragraph>`. If the attribute can be added also on children it should return 2 ranges:
 - `<paragraph>[<image>Caption</image>]</paragraph>` and 
 -`<paragraph><image>[Caption]</image></paragraph>`.

It is far more than simple using tree walker and it might be tricky for all cases which uses the tree walker.

This is why, at the end of this ticket, I am not sure if we should change this behavior.

//cc @Reinmar @scofalik @oskarwrobel

scofalik:

Fix: `Schema#getValidRanges` now returns flat ranges, compatible with the new `AttributeOperation`. Closes #1460.

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -668,35 +668,60 @@ export default class Schema {
 	 * @returns {Array.<module:engine/model/range~Range>} Ranges in which the attribute is allowed.
 	 */
 	getValidRanges( ranges, attribute ) {
-		const validRanges = [];
+		ranges = convertToMinimalFlatRanges( ranges );
+		const result = [];
 
 		for ( const range of ranges ) {
-			let last = range.start;
-			let from = range.start;
-			const to = range.end;
+			const validRanges = this._getValidRangesForRange( range, attribute );
 
-			for ( const value of range.getWalker() ) {
-				if ( !this.checkAttribute( value.item, attribute ) ) {
-					if ( !from.isEqual( last ) ) {
-						validRanges.push( new Range( from, last ) );
-					}
-
-					from = value.nextPosition;
-				}
-
-				last = value.nextPosition;
-			}
-
-			if ( from && !from.isEqual( to ) ) {
-				validRanges.push( new Range( from, to ) );
-			}
+			result.push( ...validRanges );
 		}
 
-		return validRanges;
+		return result;
 	}
 
 	/**
-	 * Basing on given the `position`, finds and returns a {@link module:engine/model/range~Range range} which is
+	 * Takes a flat range and an attribute name. Traverses the range recursively and deeply to find and return all ranges
+	 * inside the given range on which the attribute can be applied.
+	 *
+	 * This is a helper function for {@link ~Schema#getValidRanges}.
+	 *
+	 * @private
+	 * @param {module:engine/model/range~Range} range Range to process.
+	 * @param {String} attribute The name of the attribute to check.
+	 * @returns {Array.<module:engine/model/range~Range>} Ranges in which the attribute is allowed.
+	 */
+	_getValidRangesForRange( range, attribute ) {
+		const result = [];
+
+		let start = range.start;
+		let end = range.start;
+
+		for ( const item of range.getItems( { shallow: true } ) ) {
+			if ( item.is( 'element' ) ) {
+				result.push( ...this._getValidRangesForRange( Range.createIn( item ), attribute ) );
+			}
+
+			if ( !this.checkAttribute( item, attribute ) ) {
+				if ( !start.isEqual( end ) ) {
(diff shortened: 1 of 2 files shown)


src/model/schema.js
```diff
@@ -668,35 +668,60 @@ export default class Schema {
...
-			let last = range.start;
-			let from = range.start;
-			const to = range.end;
-
-			for ( const value of range.getWalker() ) {
-				if ( !this.checkAttribute( value.item, attribute ) ) {
-					if ( !from.isEqual( last ) ) {
-						validRanges.push( new Range( from, last ) );
-					}
+			const validRanges = this._getValidRangesForRange( range, attribute );
```
pjasiun: Looks like a great case for a generator :)

tests/model/schema.js
```diff
@@ -1116,112 +1115,145 @@ describe( 'Schema', () => {
...
+			// This is a "hack" to allow setting any ranges in `setData` util.
+			schema.extend( '$text', { allowIn: '$root' } );
+		} );
 
-				// Allow 'bold' on $root>p.
-				if ( ctx.endsWith( '$root p' ) && attributeName == 'bold' ) {
-					return true;
-				}
-			} );
+		function test( input, attribute, output ) {
```
pjasiun: Please don't. Your meaningless `test`function wrap 4 lines of code and then you use 5 lines of code to execute it. I had to check what `foo` (the second parameter) is to ensure what is going on in tests. Instead you can wrap:

```
const sel = new Selection( validRanges );
expect( stringify( root, sel ) ).to.equal( output );
```

into a simple util or make `stringify` accept an array of ranges.

scofalik: I honestly don't agree and I really insist on keeping it the way it is. The function makes code cleaner and test cases easy to look and inspect. For example, instead of:

```javascript
test(
	'[<p>foo<img></img>bar</p>]',
	'src',
	'<p>foo[<img></img>]bar</p>'
);
```

You'd have:

```javascript
setData( model, '[<p>foo<img></img>bar</p>]' );
const validRanges = schema.getValidRanges( doc.selection.getRanges(), 'src' );

expectUtil( validRanges, '<p>foo[<img></img>]bar</p>' );
```

Which looks much worse.

It doesn't matter if I "use 5 lines to execute 4 lines" cause it's all about how clear the code looks like. I could easily write it as:

```javascript
test( '[<p>foo<img></img>bar</p>]', 'foo', '[<p>foo<img></img>bar</p>]' );
```

If I wanted but it doesn't make code easier to analyze.

So, to sum up, I strongly disagree with your comment. You wrote that you don't like something but your arguments aren't convincing.

> I had to check what foo (the second parameter) is to ensure what is going on in tests.

Alright, but after you checked (and it's not like a rocket science and can even be deduced from the context) you know it and all other cases are easier to read. Mind you, you would have to understand the first scenario no matter if there would be a test helper or not.

src/model/schema.js
scofalik: Is there any gain by here by doing that (other than being cool)?

pjasiun: Less code. Also, I think it would be nice to be able to get only the first valid range for instance for comments.

tests/model/schema.js
pjasiun: Ok, at least change the name of the function to `assertValidRangesForAttribute`.

pjasiun:

PR #1485, opened 2018-07-30
Fixed results of Range#getTransformedByMerge for edge cases

### LINKED ISSUE #1462: Range#_getTransformedByMergeOperation is incorrect

Special case transformation in `Range#_getTransformedByMergeOperation` is incorrect.

It is like this:
```javascript
if ( start.isAfter( end ) ) {
	// This happens in the following case:
	//
	// <p>abc</p>{<p>x}yz</p> -> <p>abcx}yz</p>{
	//                           <p>abcx{yz</p>}
	//
	return new Range( end, start );
}
```

Which is quite incorrect, as the new range does not include `x` as it should.

scofalik:

Fix: Result of `Range#getTransformedByMerge` for a scenario where only range's end is inside the merged element is now correct. Closes #1462.

### DIFF
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -1156,6 +1156,12 @@ setTransformation( RenameOperation, InsertOperation, ( a, b ) => {
 } );
 
 setTransformation( RenameOperation, MergeOperation, ( a, b ) => {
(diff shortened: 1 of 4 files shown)


pjasiun:

PR #1487, opened 2018-07-31
Second parameter of LiveRange#event:change is now deletionPosition not operation that changed the range

### LINKED ISSUE #1463: Rethink what data is needed in LiveRange#event:change

Before OT refactor, `LiveRange#change:event` had a `data` object with some data about the change. It was simple, because only `InsertOperation` and `MoveOperation` could change the tree. After refactor we have four new operations and the old `data` object with its properties did not fit them.

So right now we pass the whole operation with the event. This however might not be the best idea, as @pjasiun wrote:

> If we will keep it this way changes in opeartion will have to be marked as breaking changes.

scofalik:

Other: Second parameter of `LiveRange#event:change` is now `deletionPosition` not the operation that changed the range. Closes #1463.

### DIFF
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -750,14 +750,14 @@ class LiveSelection extends Selection {
 
 		const liveRange = LiveRange.createFromRange( range );
 
-		liveRange.on( 'change:range', ( evt, oldRange, operation ) => {
+		liveRange.on( 'change:range', ( evt, oldRange, deletionPosition ) => {
 			this._hasChangedRange = true;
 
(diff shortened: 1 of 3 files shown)


src/model/documentselection.js
```diff
@@ -750,14 +750,14 @@ class LiveSelection extends Selection {
 
 		const liveRange = LiveRange.createFromRange( range );
 
-		liveRange.on( 'change:range', ( evt, oldRange, operation ) => {
+		liveRange.on( 'change:range', ( evt, oldRange, deletionPosition ) => {
```
pjasiun: It should be an object in case we need to add more properties in the future.

scofalik: 👍

pjasiun:

PR #1489, opened 2018-08-01
Fix: Selection on limit-block elements inside limit elements should b…

jodator:

Fix: Selection on limit-block elements inside limit elements should be possible.

---

### Additional information

* This PR is part of [block content support in tables](https://github.com/ckeditor/ckeditor5-table/pull/97).
* Without this change you cannot `$blocks` inside table cells - mostly images - as the selection post-fixer would change selection to whole table (image is also a limit element). Now `$block` element is considered selectable inside other `isLimit` elements even though it might be `isLimit` or `isObject` alongside.


jodator:
I think that this can be discarded as we do not plan `image` right now. I've forgot about this PR. It is not required by tables#56 at the moment. I'll use it if necessary in table#63.

jodator:
This one can be dropped until we work on images properly.

PR #1490, opened 2018-08-01
Introduced multiple fixes in OT algorithms and position transformation functions

### LINKED ISSUE #1475: New OT: "Empty" wraps after OT
```javascript
john.setData( '<paragraph>Foo</paragraph>[<paragraph>Bar</paragraph>]' );
kate.setData( '<paragraph>Foo</paragraph>[]<paragraph>Bar</paragraph>' );

john.wrap( 'blockQuote' );
kate.merge();
```

In above scenario, currently, the expected result is:

```html
<paragraph>FooBar</paragraph><blockQuote></blockQuote>
```

Which obviously is not correct. The question is: what to do about it? Should we introduce a postfixer for block-quote or should we introduce some general OT algorithm that will not wrap in such a scenario?

scofalik:

Fix: Introduced multiple fixes in OT algorithms and position transformation functions. Closes #1475.

---

### Additional information

Bases on #1478 .

### DIFF
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -113,13 +113,17 @@ function transformSets( operationsA, operationsB, options ) {
 					const opA = opsA[ k ];
 					const opB = opsB[ l ];
 
-					context.aIsStrong = true;
-					const newOpA = transform( opA, opB, context );
+					const newOpA = transform( opA, opB, {
+						aIsStrong: true,
+						aWasUndone: context.wasUndone( opA ),
+						bWasUndone: context.wasUndone( opB )
+					} );
 
-					context.aIsStrong = false;
-					const newOpB = transform( opB, opA, context );
-
-					delete context.aIsStrong;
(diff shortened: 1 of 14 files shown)


src/model/operation/transform.js
```diff
@@ -361,53 +532,59 @@ setTransformation( AttributeOperation, MergeOperation, ( a, b ) => {
...
 	} else {
 		ranges = [ range ];
+
+		_flatMoveTransform( ranges, moveOp );
 	}
 
+	return ranges;
+}
+
+function _flatMoveTransform( ranges, moveOp ) {
```
pjasiun: Comments needed.

scofalik: Spoiler alert: This function is rewritten on branch `t/1488` which contains following improvements in OT. Also, I will complete all the docs and also inline comments in another followup.

pjasiun:

PR #1493, opened 2018-08-06
Internal: Changed the way of data serialization.

### LINKED ISSUE #1477: toJSON() should serialize nested objects
We are changing the way of operation serialization. JSON string won't be used anymore but we still need to reduce class instance to a pure object. We need to ensure that all `toJSON()` calls create a proper structure including nested objects.

oskarwrobel:

Internal: Changed the way of data serialization. Closes #1477.

---

### Additional information

### DIFF
--- a/src/model/node.js
+++ b/src/model/node.js
@@ -381,7 +381,11 @@ export default class Node {
 		const json = {};
 
 		if ( this._attrs.size ) {
-			json.attributes = [ ...this._attrs ];
+			json.attributes = Array.from( this._attrs ).reduce( ( result, attr ) => {
+				result[ attr[ 0 ] ] = attr[ 1 ];
+
+				return result;
+			}, {} );
 		}
 
 		return json;
--- a/src/model/node.js
+++ b/src/model/node.js
@@ -380,6 +380,8 @@ export default class Node {
 	toJSON() {
 		const json = {};
 
+		// Serializes attributes to the object.
+		// attributes = { a: 'foo', b: 1, c: true }.
 		if ( this._attrs.size ) {
 			json.attributes = Array.from( this._attrs ).reduce( ( result, attr ) => {
 				result[ attr[ 0 ] ] = attr[ 1 ];
--- a/src/model/operation/attributeoperation.js
+++ b/src/model/operation/attributeoperation.js
@@ -110,6 +110,17 @@ export default class AttributeOperation extends Operation {
 	/**
 	 * @inheritDoc
 	 */
+	toJSON() {
+		const json = super.toJSON();
+
+		json.range = this.range.toJSON();
+
+		return json;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
 	_validate() {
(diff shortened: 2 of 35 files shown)


src/model/position.js
```diff
@@ -915,7 +926,7 @@ export default class Position {
 	 */
 	static fromJSON( json, doc ) {
 		if ( json.root === '$graveyard' ) {
-			const pos = new Position( doc.graveyard, json.path );
+			const pos = new Position( doc.graveyard, Array.from( json.path ) );
```
pjasiun: Why do you need this?

src/model/node.js
```diff
@@ -381,7 +381,11 @@ export default class Node {
 		const json = {};
 
 		if ( this._attrs.size ) {
-			json.attributes = [ ...this._attrs ];
+			json.attributes = Array.from( this._attrs ).reduce( ( result, attr ) => {
+				result[ attr[ 0 ] ] = attr[ 1 ];
+
+				return result;
+			}, {} );
```
pjasiun: A comment is needed.

src/model/position.js
oskarwrobel: This will be removed.

src/model/node.js
oskarwrobel: Done.

src/model/operation/attributeoperation.js
```diff
@@ -107,6 +107,17 @@ export default class AttributeOperation extends Operation {
 		return new AttributeOperation( this.range, this.key, this.newValue, this.oldValue, this.baseVersion + 1 );
 	}
 
+	/**
+	 * @inheritDoc
+	 */
+	toJSON() {
```
scofalik: I wonder here... Theoretically, values could be arrays or objects. Will those be cloned?

scofalik: BTW... maybe we should just give up that and allow only for string, numeral and boolean values for attributes?

src/model/operation/wrapoperation.js
```diff
@@ -163,6 +163,21 @@ export default class WrapOperation extends Operation {
...
 
+	/**
+	 * @inheritDoc
+	 */
+	toJSON() {
+		const json = super.toJSON();
+
+		json.position = this.position.toJSON();
+
+		if ( json.graveyardPosition ) {
```
scofalik: Wrap operation can have graveyard position or element. Shouldn't we do `json.element = this.element.toJSON()` ?

src/model/operation/attributeoperation.js
oskarwrobel: 🤔I'm afraid in a case of objects we pass a reference to the value instead of cloning.
https://github.com/ckeditor/ckeditor5-engine/blob/6c1f87e7e4d8733e9774809374b37db0df04b215/src/model/operation/attributeoperation.js#L98

We can think about dropping other formats than primitives but we don't know if someone has already used it with objects or arrays.

PR #1494, opened 2018-08-08
Tests: New cases which include ranges in blocks in OT tests

### LINKED ISSUE #1470: Add tests for ranges in blocks
Reviewing collaboration tests I met a lot of tests were the selection is next to the block (`[]<paragraph>Foo</paragraph>`) or around the block (`[<paragraph>Foo</paragraph>]`). These are valid tests for collaboration, but the selection is not realistic. The real selection will be in the block (`<paragraph>[]Foo</paragraph>` or `<paragraph>[Foo]</paragraph>`). Tests should be review and for most of the cases where the selection is next to the block, we should add an additional test for the similar case with the selection in the block (these are such tests for some cases, but there are also missing cases).

Mgsy:

Tests: New cases which include ranges in blocks in OT tests. Closes #1470 .

---

### Additional information

* Unfortunately, some scenarios fail.
* Additionally, I moved all `remove attribute` scenarios to the `attribute.js` file.


tests/model/operation/transform/attribute.js
```diff
@@ -263,6 +278,66 @@ describe( 'transform', () => {
...
+
+				john.type( 'Bar' );
+				kate.removeAttribute( 'bold' );
+
+				syncClients();
+
+				expectClients( '<paragraph>BarFoo</paragraph>' );
+			} );
+
+			it( 'remove attribute from text with 2 attributes in same path', () => {
```
scofalik: I'd change those examples so that typing is in the remove attribute range.

tests/model/operation/transform/attribute.js
```diff
@@ -343,6 +418,87 @@ describe( 'transform', () => {
...
+
+				expectClients(
+					'<paragraph>F</paragraph>' +
+					'<paragraph>ooBar</paragraph>'
+				);
+			} );
+
+			it( 'remove attribute from text in different path', () => {
+				john.setData( '<paragraph>F[oo]</paragraph><paragraph>Bar</paragraph>' );
+				kate.setData( '<paragraph>Foo</paragraph><paragraph>[Bar]</paragraph>' );
```
scofalik: There's no attribute to remove here.

tests/model/operation/transform/attribute.js
```diff
@@ -376,6 +532,21 @@ describe( 'transform', () => {
 				);
 			} );
 
+			it( 'text into div in different path', () => {
```
scofalik: I don't like this example because it is rather unrealistic. We do not wrap text. And it is bad to have `<p><div></div></p>` structure, casue `<div>` should never be in `<p>`.

Unless it won't have a negative impact on CC, I'd drop this one.

tests/model/operation/transform/attribute.js
```diff
@@ -407,6 +578,18 @@ describe( 'transform', () => {
 					'</blockQuote>'
 				);
 			} );
+
+			it( 'text into div in same path', () => {
```
scofalik: as above

tests/model/operation/transform/attribute.js
```diff
@@ -440,6 +623,21 @@ describe( 'transform', () => {
 				);
 			} );
 
+			it( 'text in different path', () => {
```
scofalik: We will need to work on those names later cause they aren't really helpful (aren't describing exactly what happens). I understand that we don't want to make sentences of description but it would be good to have more meaningful names. BTW. I am not perfect at that either.

tests/model/operation/transform/attribute.js
```diff
@@ -600,10 +806,86 @@ describe( 'transform', () => {
...
+
+				john.remove();
+				kate.removeAttribute( 'bold' );
+
+				syncClients();
+
+				expectClients( '<paragraph>o</paragraph>' );
+			} );
+
+			it( 'remove attribute from text with 2 attributes in same path', () => {
```
scofalik: How about having a test that removes part of non-changed and part of the changed text (changed = removed attributes)?

tests/model/operation/transform/attribute.js
Mgsy: I agree with you. I added this kind of tests to treat them as theoretical cases where the range is inside an element. I believe it won't have any impact on CC, so if you're up to remove those (there are more in other files), I'm ok with that.

tests/model/operation/transform/attribute.js
```diff
@@ -871,9 +1226,85 @@ describe( 'transform', () => {
...
+
+			syncClients();
+
+			expectClients(
+				'<heading1>Foo</heading1>' +
+				'<paragraph>Bar</paragraph>'
+			);
+		} );
+
+		it( 'remove attribute from text in different path', () => {
```
scofalik: Theres no attribute here.

tests/model/operation/transform/insert.js
```diff
@@ -906,74 +933,7 @@ describe( 'transform', () => {
...
 				syncClients();
 
-				expectClients(
-					'<paragraph>BarFo<m1:start></m1:start>o<m1:end></m1:end></paragraph>'
-				);
-			} );
-		} );
-
-		// This should be moved to attribute.js.
-		describe( 'by remove attribute', () => {
```
scofalik: Why you removed those tests?

tests/model/operation/transform/marker.js
```diff
@@ -644,99 +669,6 @@ describe( 'transform', () => {
 			} );
 		} );
 
-		// Should be in attribute.js.
-		describe( 'by remove attribute', () => {
```
scofalik: Why you removed those tests?

tests/model/operation/transform/move.js
```diff
@@ -224,21 +224,6 @@ describe( 'transform', () => {
 					'</blockQuote>'
 				);
 			} );
-
-			it( 'intersecting wrap', () => {
```
scofalik: Why you removed that test?

tests/model/operation/transform/move.js
```diff
@@ -594,94 +590,33 @@ describe( 'transform', () => {
 
 				expectClients( '<paragraph>FooBar</paragraph>' );
 			} );
-		} );
-
-		describe( 'by remove attribute', () => {
```
scofalik: Why you removed those tests?

tests/model/operation/transform/remove.js
```diff
@@ -333,97 +354,7 @@ describe( 'transform', () => {
 
 				syncClients();
 
-				expectClients(
-					'<paragraph></paragraph>'
-				);
-			} );
-		} );
-
-		describe( 'by remove attribute', () => {
```
scofalik: Why you removed those tests?

tests/model/operation/transform/rename.js
```diff
@@ -15,94 +15,6 @@ describe( 'transform', () => {
 	} );
 
 	describe( 'rename', () => {
-		describe( 'by remove attribute', () => {
```
scofalik: Why you removed those tests?

tests/model/operation/transform/insert.js
Mgsy: I moved all `remove attribute` tests to the `attribute.js` file.

scofalik: Okay, didn't see that cause attribute.js was not expanded. What about the wrap test?

tests/model/operation/transform/attribute.js
Mgsy: Fixed.

tests/model/operation/transform/insert.js
Mgsy: Whoops, I see that wrap, unwrap and split don't have this kind of tests. I'll add them.

scofalik: I've meant that there is a removed test in `move.js`: `'intersecting wrap'`. I've commented on that below.

tests/model/operation/transform/move.js
Mgsy: I removed it, because wrap was on text, not element. We decided to remove those cases.

tests/model/operation/transform/insert.js
Mgsy: >Whoops, I see that wrap, unwrap and split don't have this kind of tests. I'll add them.

Done.

tests/model/operation/transform/move.js
scofalik: 👍

PR #1497, opened 2018-08-09
OT refactor followups

### LINKED ISSUE #1460: Representation of the selection which contains object, but not children

### LINKED ISSUE #1461: Have just one MoveOperation

Having just one `MoveOperation` with different `type` property value might simplify some places in the code. Right now, `RemoveOperation` and `ReinsertOperation` extends `MoveOperation` but aren't much different from it.

### LINKED ISSUE #1462: Range#_getTransformedByMergeOperation is incorrect

### LINKED ISSUE #1463: Rethink what data is needed in LiveRange#event:change

### LINKED ISSUE #1464: Move try-catch on transformation method from engine debug tools to transformations code

### LINKED ISSUE #1467: Make element cloning in WrapOperation as it is in InsertOperation

### LINKED ISSUE #1468: Finish relations mechanism in OT
Relations mechanism that brings better contextual information for OT and is used by undo has not been finished in the initial refactor PR. This issue is set as a reminder to introduce this mechanism afterward.

### LINKED ISSUE #1475: New OT: "Empty" wraps after OT

### LINKED ISSUE #1479: Error misses a description
https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/selection.js#L434

Please try to imagine why people get this error (it's 5th most popular). I think they may be using broken installations with duplicated ckeditor5-engine or native `Range()` for instance. Would be good to explain both things.

scofalik:

Fix: Additional improvements for the new OT algorithms. Added relations to better solve cases in undo. Moved error logging from debug tools to OT code. Improved cloning mechanism inside `WrapOperation`. Aligned `Schema#getValidRanges` results to changes in `AttributeOperation`. Unified `RemoveOperation` and `ReinsertOperation` to have just one `MoveOperation`. Improved `model.Range#getTransformedByMergeOperation`. Simplified `LiveRange#event:change` second parameter which is now an object containing `Position` not an `Operation`. Other improvements in OT algorithms. Improvements in docs. Closes #1468. Closes #1464. Closes #1467. Closes #1460. Closes #1461. Closes #1462. Closes #1463. Closes #1475. Closes #1479.

BREAKING CHANGE: `LiveRange#event:change` second parameter is now an object containing property `deletionPosition`. It can be `model.Position` instance, if the range was moved to the graveyard root. The position is equal to the position from which nodes were removed. Otherwise, it is set to `null`.

BREAKING CHANGE: `Schema#getValidRanges` will return only flat ranges now. If an attribute is allowed on some nodes and in those nodes children, multiple "nested" ranges will be returned.

BREAKING CHANGE: `Schema#getValidRanges` is a generator now.


scofalik:
Although changes from branch `t/1469` are included here, this PR does not close this issue as we are still discussing what we want to do with it.

scofalik:
I've pushed some fixes but they broke the CI. For sure one of the problems is that now `ckeditor5-undo` changed and I've pushed the change to `t/ckeditor5-engine/ot-followups`. But other than that I see that something else fails, even though it is okay locally.

scofalik:
https://github.com/ckeditor/ckeditor5-undo/pull/90 Okay I am stupid, tests are failing in engine too cause engine uses undo -_-

PR #1498, opened 2018-08-09
Additional improvements in OT

### LINKED ISSUE #1488: New OT: MergeOperation needs howMany property
While fixing OT algorithms I noticed that I missed a case in `UnwrapOperation` x `MergeOperation` transformation.

It is a case where `MergeOperation` merges into the unwrapped node. Since `UnwrapOperation` has `howMany` property, the property needs to be updated, which is impossible right now, as `MergeOperation` has only `targetPosition` and `sourcePosition`.

`UnwrapOperation` needs `howMany` property to be properly reversed to `WrapOperation` and to be able to properly transform a range by `UnwrapOperation`. So we can't and won't get rid off it.

The "how many" property can be "executed" either as standalone property (`howMany`) or by changing `MergeOperation#sourcePosition` to `MergeOperation#sourceRange`.

### LINKED ISSUE #1492: New OT: Position#_getTransformedByMove is incorrect for "non-moves"
In some cases current implementation of `Position#_getTransformedByMove` returns wrong values. It happens when the transformed position is the same as passed `sourcePosition` and the `targetPosition` is at the end of the moved range or at the beginning of the moved range (`sourcePosition`) so it is not really any move.

```html
<p>F^[oo]bar</p>
```

`[]` is moved to the position `^` or `]`.

What happens is: `targetPosition` after transforming by `oo` remove is at the same place as `^`. Since we don't have `insertBefore` flag anymore, `^` position is transformed like the nodes were moved before it.

So basically after this transformation the position goes to:

```html
<p>F[oo]^bar</p>
```

Which is incorrect because `oo` didn't really move.

### LINKED ISSUE #1499: New OT: Improvements in OT basing on errors found during manual testing
This is an umbrella ticket for issues that we found during manual testing. Those fixes were already done and pushed together to a common branch. This issue is for documenting purposes.

Found bugs were:

* incorrect `Position#_getTransformedByMove` when the parameters were such that there is no move (i.e. `sourcePosition` same as `targetPosition`) (#1492),
* `Range#_getTransformedByMergeOperation` was returning ranges which boundaries were in different roots (one in the original root and the other in the graveyard root), this is a part of #1476,
* our current transforming algorithm (in `transformSets`) was incorrect for some peculiar scenarios where both `operationsA` and `operationsB` were broken into multiple operations, so the whole algorithm was rewritten and now it transforms in a correct order,
* several bugs or unhandled situations in operations transformations.

scofalik:

Other: Added `MergeOperation#howMany` and `SplitOperation#howMany`. Improvements in `Position` and `Range` transforming functions. Improvements in OT algorithms. Closes #1488. Closes #1492. Closes #1499.

### DIFF
--- a/src/model/operation/mergeoperation.js
+++ b/src/model/operation/mergeoperation.js
@@ -30,7 +30,7 @@ export default class MergeOperation extends Operation {
 	 *
 	 * @param {module:engine/model/position~Position} sourcePosition Position inside the merged element. All nodes from that
 	 * element after that position will be moved to {@link ~#targetPosition}.
-	 * @param {Number} howMany
+	 * @param {Number} howMany Summary offset size of nodes which will be moved from the merged element to the new parent.
 	 * @param {module:engine/model/position~Position} targetPosition Position which the nodes from the merged elements will be moved to.
 	 * @param {module:engine/model/position~Position} graveyardPosition Position in graveyard to which the merged element will be moved.
 	 * @param {Number|null} baseVersion Document {@link module:engine/model/document~Document#version} on which operation
--- a/src/model/operation/mergeoperation.js
+++ b/src/model/operation/mergeoperation.js
@@ -30,12 +30,13 @@ export default class MergeOperation extends Operation {
 	 *
 	 * @param {module:engine/model/position~Position} sourcePosition Position inside the merged element. All nodes from that
 	 * element after that position will be moved to {@link ~#targetPosition}.
+	 * @param {Number} howMany
 	 * @param {module:engine/model/position~Position} targetPosition Position which the nodes from the merged elements will be moved to.
 	 * @param {module:engine/model/position~Position} graveyardPosition Position in graveyard to which the merged element will be moved.
 	 * @param {Number|null} baseVersion Document {@link module:engine/model/document~Document#version} on which operation
 	 * can be applied or `null` if the operation operates on detached (non-document) tree.
 	 */
-	constructor( sourcePosition, targetPosition, graveyardPosition, baseVersion ) {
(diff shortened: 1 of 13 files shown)


src/model/operation/mergeoperation.js
```diff
@@ -30,12 +30,13 @@ export default class MergeOperation extends Operation {
 	 *
 	 * @param {module:engine/model/position~Position} sourcePosition Position inside the merged element. All nodes from that
 	 * element after that position will be moved to {@link ~#targetPosition}.
+	 * @param {Number} howMany
```
pjasiun: Docs?

src/model/operation/transform.js
```diff
@@ -86,72 +90,56 @@ export function transformSets( operationsA, operationsB, options ) {
...
-					const newOpA = transform( opA, opB, contextAB );
-					const newOpB = transform( opB, opA, contextBA );
-
-					if ( options.useContext ) {
-						updateOriginalOperation( context, opA, newOpA );
-						updateOriginalOperation( context, opB, newOpB );
-
-						updateRelations( context, opA, opB );
-					}
+		for ( let j = transformFrom; j < operationsB.length; ) {
```
pjasiun: As we talked it could be simplified.

src/model/operation/transform.js
```diff
@@ -77,6 +79,10 @@ export function transformSets( operationsA, operationsB, options ) {
 		return { operationsA, operationsB };
 	}
 
+	for ( const op of operationsA ) {
+		op.transformBy = 0;
```
pjasiun: WeakMap?

PR #1500, opened 2018-08-14
Remove unnecessary setTimeouts from tests

ma2ciek:

Tests: Removed unnecessary setTimeouts from tests.

---

### Additional information

Part of the https://github.com/ckeditor/ckeditor5/issues/1202.

### DIFF
--- a/tests/view/observer/fakeselectionobserver.js
+++ b/tests/view/observer/fakeselectionobserver.js
@@ -122,7 +122,7 @@ describe( 'FakeSelectionObserver', () => {
 
 		// Wait 210ms (debounced function should be called).
 		clock.tick( 210 );
-		sinon.assert.calledOnce( spy.calledOnce );
+		sinon.assert.calledOnce( spy );
 	} );
 
 	it( 'should not fire `selectionChangeDone` event when observer will be destroyed', () => {
@@ -144,7 +144,7 @@ describe( 'FakeSelectionObserver', () => {
 		clock.tick( 110 );
 
 		// Check that event won't be called.
-		sinon.assert.notCalled( spy.notCalled );
+		sinon.assert.notCalled( spy );
 	} );
 
 	// Checks if preventDefault method was called by FakeSelectionObserver for specified key code.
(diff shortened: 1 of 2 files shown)


ma2ciek:
In the `SelectionObserver` the problem is a bit deeper, all tests actually use timeouts internally. 

```
> npm t -- -f engine/view/observer/selectionobserver.js --reporter dots

> ckeditor5@11.0.1 test /Users/maciejbukowski/workspace/ckeditor5
> node --max_old_space_size=4096 node_modules/@ckeditor/ckeditor5-dev-tests/bin/test.js "-f" "engine/view/observer/selectionobserver.js" "--reporter" "dots"

(node:17831) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
ℹ ｢wdm｣:
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣:    711 modules
ℹ ｢wdm｣: Compiled successfully.
16 08 2018 16:15:00.368:INFO [karma]: Karma v2.0.5 server started at http://0.0.0.0:9876/
16 08 2018 16:15:00.371:INFO [launcher]: Launching browser CHROME_LOCAL with unlimited concurrency
16 08 2018 16:15:00.411:INFO [launcher]: Starting browser Chrome
16 08 2018 16:15:01.538:INFO [Chrome 68.0.3440 (Mac OS X 10.13.6)]: Connected on socket d1n9JHYvPt_xz7uUAAAA with id 45653572
............
Chrome 68.0.3440 (Mac OS X 10.13.6): Executed 12 of 12 SUCCESS (2.82 secs / 1.475 secs)
```

ma2ciek:
These tests are actually pretty hard to refactor. The native `selectionchange` event is asynchronous, so using sinon's fake timers makes the code looks like spaghetti as inside test I had to use fake timers combined with native timeouts to wait for this event. The gain is ~1s, but the cost of the readability is bigger IMHO. I'd drop these comments about lodash and leave these lines in the `SelectionObserver` as they are.

Reinmar:
I didn't check what you changed at all, but one general rule – native selection and focus tests should be as real as possible. We test the actual browser behaviour and whatever we'll fake will make them less realistic. 

Of course, such tests may need to take more time then. That's ok. But that also means that we should have as little of them as possible and test our abstractions instead of native stuff whenever we can.

PR #1502, opened 2018-08-16
Fix: made `startsWithFiller` to use safe `isText` check instead of `instanceof`

dimaip:
Using `domNode instanceof Text` is not safe and may cause issues e.g. across iframes.

### DIFF
--- a/src/view/filler.js
+++ b/src/view/filler.js
@@ -6,6 +6,7 @@
 /* globals window, Text */
 
 import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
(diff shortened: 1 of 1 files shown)


ma2ciek:
Hi, @dimaip :) 

Thanks for the PR!

I have only one doubt - the iframe will persist in the DOM after your test. Could you detach the iframe after the test, so the testing environment will stay clean?

dimaip:
@ma2ciek sure, wasn't aware it's not somehow destroyed between the tests.

dimaip:
@ma2ciek ok, looks like the tests are green

PR #1503, opened 2018-08-16
Create writer for non-semantic view manipulation

### LINKED ISSUE #1501: Create writer for non-semantic view manipulation
While working on Paste from Word (see https://github.com/ckeditor/ckeditor5/issues/1184 or https://github.com/ckeditor/ckeditor5-paste-from-word) there is a need to manipulate the `view` instance which was created from pasted HTML.

The `ViewElement` class has all the necessary methods which enables to manipulate view elements, however [those methods are `protected`](https://github.com/ckeditor/ckeditor5-engine/blob/85c426a1ad39a9bb694aeb6cf82d9798bb66533b/src/view/element.js#L490) and should not be used outside. There is also a [`Writer`](https://github.com/ckeditor/ckeditor5-engine/blob/85c426a1ad39a9bb694aeb6cf82d9798bb66533b/src/view/writer.js) class, however it is strictly designed to manipulate the view which was created based on the model (so the semantic one) and should not be used for non-semantic/dirty view manipulation.

There is a need for simple `Writer` class which can be used to manipulate non-semantic view (as we initially agreed it can be a simple wrapper for all `Element` manipulation methods).

The name proposals:

* `Upcast/DowncastWriter` (so the current `Writer` will be renamed)
* `Semantic/NonsemanticWriter` (same as above)
* `RawWriter`, `LowWriter`, `LowLevelWriter`, `GenericWriter`
* `BasicWriter`, `BaseWriter`

f1ames:

Internal: Introduced writer for manipulating non-semantic view. Closes #1501.

---

### Additional information

I'm still not sure about naming, the name I used is `RawWriter`, but we have discussed also `Upcast/Downcast` writer. Maybe looking at the code it will be easier to decide. I'm not in favor of any, `upcast/downcast` doesn't fit that much as this writer has basically nothing to do with upcasting/downcasting (apart from the fact that it touches view instance which may get upcasted eventually - but lots of classes do). And `raw` is somehow descriptive, but still you need to know the context well.

### DIFF
--- a/src/view/upcastwriter.js
+++ b/src/view/upcastwriter.js
@@ -13,7 +13,7 @@ import Element from './element';
  * View upcast writer class. Provides set of methods used to properly manipulate nodes attached to
  * {@link module:engine/view/view~View view instance}. It should be only used to manipulate non-semantic view
  * (view created from HTML string). For view which was downcasted from the {@link module:engine/model/model~Model model}
- * see {@link module:engine/view/writer~Writer writer}.
+ * see {@link module:engine/view/downcastwriter~DowncastWriter downcast writer}.
  */
 export default class UpcastWriter {
 	/**
--- a/src/view/upcastwriter.js
+++ b/src/view/upcastwriter.js
@@ -34,13 +34,13 @@ export default class UpcastWriter {
 	 * and sets the parent of these nodes to this element.
 	 *
 	 * @see module:engine/view/element~Element#_appendChild
+	 * @param {module:engine/view/item~Item|Iterable.<module:engine/view/item~Item>} items Items to be inserted.
 	 * @param {module:engine/view/element~Element|module:engine/view/documentfragment~DocumentFragment} element Element
 	 * to which items will be appended.
-	 * @param {module:engine/view/item~Item|Iterable.<module:engine/view/item~Item>} items Items to be inserted.
-	 * @fires module:engine/view/node~Node#change
+	 * @fires module:engine/view/node~Node#event:change
 	 * @returns {Number} Number of appended nodes.
 	 */
-	appendChild( element, items ) {
+	appendChild( items, element ) {
 		return element._appendChild( items );
 	}
 
@@ -49,14 +49,14 @@ export default class UpcastWriter {
 	 * this element.
 	 *
 	 * @see module:engine/view/element~Element#_insertChild
+	 * @param {Number} index Offset at which nodes should be inserted.
+	 * @param {module:engine/view/item~Item|Iterable.<module:engine/view/item~Item>} items Items to be inserted.
 	 * @param {module:engine/view/element~Element|module:engine/view/documentfragment~DocumentFragment} element Element
 	 * to which items will be inserted.
-	 * @param {Number} index Position where nodes should be inserted.
-	 * @param {module:engine/view/item~Item|Iterable.<module:engine/view/item~Item>} items Items to be inserted.
-	 * @fires module:engine/view/node~Node#change
+	 * @fires module:engine/view/node~Node#event:change
 	 * @returns {Number} Number of inserted nodes.
 	 */
-	insertChild( element, index, items ) {
+	insertChild( index, items, element ) {
 		return element._insertChild( index, items );
 	}
 
@@ -64,14 +64,14 @@ export default class UpcastWriter {
 	 * Removes number of child nodes starting at the given index and set the parent of these nodes to `null`.
 	 *
 	 * @see module:engine/view/element~Element#_removeChildren
+	 * @param {Number} index Offset from which nodes will be removed.
+	 * @param {Number} howMany Number of nodes to remove.
 	 * @param {module:engine/view/element~Element|module:engine/view/documentfragment~DocumentFragment} element Element
-	 * from which children will be removed.
-	 * @param {Number} index Number of the first node to remove.
-	 * @param {Number} [howMany=1] Number of nodes to remove.
-	 * @fires module:engine/view/node~Node#change
-	 * @returns {Array.<module:engine/view/node~Node>} The array of removed nodes.
+	 * which children will be removed.
+	 * @fires module:engine/view/node~Node#event:change
+	 * @returns {Array.<module:engine/view/node~Node>} The array containing removed nodes.
 	 */
-	removeChildren( element, index, howMany = 1 ) {
+	removeChildren( index, howMany, element ) {
 		return element._removeChildren( index, howMany );
 	}
 
@@ -79,10 +79,11 @@ export default class UpcastWriter {
 	 * Removes given element from the view structure. Will not have effect on detached elements.
 	 *
 	 * @param {module:engine/view/element~Element} element Element which will be removed.
-	 * @returns {Array.<module:engine/view/node~Node>} The array of removed nodes.
+	 * @returns {Array.<module:engine/view/node~Node>} The array containing removed nodes.
 	 */
 	remove( element ) {
 		const parent = element.parent;
+
 		if ( parent ) {
 			return this.removeChildren( parent, parent.getChildIndex( element ) );
 		}
@@ -94,7 +95,7 @@ export default class UpcastWriter {
 	 * Replaces given element with the new one in the view structure. Will not have effect on detached elements.
 	 *
 	 * @param {module:engine/view/element~Element} oldElement Element which will be replaced.
-	 * @param {module:engine/view/element~Element} newElement Element which will inserted in the place of the old element.
+	 * @param {module:engine/view/element~Element} newElement Element which will be inserted in the place of the old element.
 	 * @returns {Boolean} Whether old element was successfully replaced.
 	 */
 	replace( oldElement, newElement ) {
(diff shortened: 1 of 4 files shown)


pjasiun:
With @Reinmar and @scofalik we just talked and agreed that `UpcastWriter`  and `DowncastWriter` will be the best name. Also, I think that both will need to have improved documentation to explain cases where they should be used.

pjasiun:
Also, I disagree that the writer has nothing to do with downcasting, because if it used a lot during downcasting. It is as important for downcasting as converters, doing a big part of creating view during downcasting.

f1ames:
> With @Reinmar and @scofalik we just talked and agreed that UpcastWriter and DowncastWriter will be the best name.

I'm fine with it, just one clarification - so the existing `Writer` should be `UpcastWriter` and the one introduced in this PR should be `DowncastWriter`?

> Also, I disagree that the writer has nothing to do with downcasting, because if it used a lot during downcasting. It is as important for downcasting as converters, doing a big part of creating view during downcasting.

That's true for the current `Writer`. But when it comes to writer introduced in this PR it has nothing to do with model, it just manipulates the view - so basically it may work on a view instance which will be upcasted to model or which was downcasted from it, but it does not participate in any way in the process of upcasting/downcasting itself.

pjasiun:
> I'm fine with it, just one clarification - so the existing Writer should be UpcastWriter and the one introduced in this PR should be DowncastWriter?

Nope. The existing one  (`Writer`) should be `DowncastWriter`, the new one (`RawWriter`) should be `UpcastWriter`.

pjasiun:
> That's true for the current Writer. But when it comes to writer introduced in this PR it has nothing to do with model, it just manipulates the view - so basically it may work on a view instance which will be upcasted to model or which was downcasted from it, but it does not participate in any way in the process of upcasting/downcasting itself.

That's true, but we could bend the definition a little for this case. How, upcasting is View -> Model, but it could be Data -> View -> Model (downcasting Model -> View -> Data). Then it will fit well.

f1ames:
> That's true, but we could bend the definition a little for this case. How, upcasting is View -> Model, but it could be Data -> View -> Model (downcasting Model -> View -> Data). Then it will fit well.

Yes, it makes sense and makes it easier to understand with upcasting/downcasting defined in such way 👍

f1ames:
> Also, I think that both will need to have improved documentation to explain cases where they should be used.

@pjasiun I will update the naming and docs and ping you when it's ready.

f1ames:
@pjasiun  I have renamed `RawWriter` to `UpcastWriter`. Now I'm working on renaming `Writer` to `DowncastWriter`, however I will put it in a separate PR as there will be plenty of changes.

So this PR is ready for R.

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+
+	/**
+	 * Appends a child node or a list of child nodes at the end of this node
+	 * and sets the parent of these nodes to this element.
+	 *
+	 * @see module:engine/view/element~Element#_appendChild
+	 * @param {module:engine/view/element~Element|module:engine/view/documentfragment~DocumentFragment} element Element
+	 * to which items will be appended.
+	 * @param {module:engine/view/item~Item|Iterable.<module:engine/view/item~Item>} items Items to be inserted.
+	 * @fires module:engine/view/node~Node#change
```
scofalik: We (at least I ;)) usually write `event:change`. Does the line above create proper docs?

BTW., those change events on view tree are so internal that I am not sure if it is worth documenting. @pjasiun ?

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	/**
+	 * Inserts a child node or a list of child nodes on the given index and sets the parent of these nodes to
+	 * this element.
+	 *
+	 * @see module:engine/view/element~Element#_insertChild
+	 * @param {module:engine/view/element~Element|module:engine/view/documentfragment~DocumentFragment} element Element
+	 * to which items will be inserted.
+	 * @param {Number} index Position where nodes should be inserted.
```
scofalik: `Offset from which`?

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+		return element._insertChild( index, items );
+	}
+
+	/**
+	 * Removes number of child nodes starting at the given index and set the parent of these nodes to `null`.
+	 *
+	 * @see module:engine/view/element~Element#_removeChildren
+	 * @param {module:engine/view/element~Element|module:engine/view/documentfragment~DocumentFragment} element Element
+	 * from which children will be removed.
+	 * @param {Number} index Number of the first node to remove.
```
scofalik: `Offset from which`?

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+	 * @param {Number} [howMany=1] Number of nodes to remove.
+	 * @fires module:engine/view/node~Node#change
+	 * @returns {Array.<module:engine/view/node~Node>} The array of removed nodes.
```
scofalik: `containing`?

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+	 */
+	removeChildren( element, index, howMany = 1 ) {
+		return element._removeChildren( index, howMany );
+	}
+
+	/**
+	 * Removes given element from the view structure. Will not have effect on detached elements.
+	 *
+	 * @param {module:engine/view/element~Element} element Element which will be removed.
+	 * @returns {Array.<module:engine/view/node~Node>} The array of removed nodes.
```
scofalik: `with removed nodes`?

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+	 */
+	remove( element ) {
+		const parent = element.parent;
+		if ( parent ) {
```
scofalik: New line please? :)

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+		}
+
+		return [];
+	}
+
+	/**
+	 * Replaces given element with the new one in the view structure. Will not have effect on detached elements.
+	 *
+	 * @param {module:engine/view/element~Element} oldElement Element which will be replaced.
+	 * @param {module:engine/view/element~Element} newElement Element which will inserted in the place of the old element.
```
scofalik: `will be`

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+	 * was not replaced (happens for detached elements).
+	 */
+	rename( element, newName ) {
+		const newElement = new Element( newName, element.getAttributes(), element.getChildren() );
+
+		return this.replace( element, newElement ) ? newElement : null;
+	}
+
+	/**
+	 * Adds or overwrite element's attribute with a specified key and value.
```
scofalik: `overwrites` (I see that this is a typo also at the place you copied from :P)

src/view/upcastwriter.js
```diff
@@ -0,0 +1,243 @@
...
+	 *
+	 *		writer.setAttribute( linkElement, 'href', 'http://ckeditor.com' );
+	 *
+	 * @see module:engine/view/element~Element#_setAttribute
+	 * @param {module:engine/view/element~Element} element
```
scofalik: `element Element to set attribute on` ?

scofalik: Okay I see the laziness creep in next methods :)

scofalik:
I have one overall issue: the parameters in methods in this class have a different order than the parameters in `DowncastWriter`. There we follow the rule that `element` is the last parameter. Here it is first. Sorry.

pjasiun:
Note that `DOMConverter` should now use this writer. In can be changed as a part of this ticket or reported as a follow-up.

f1ames:
> I have one overall issue: the parameters in methods in this class have a different order than the parameters in DowncastWriter. There we follow the rule that element is the last parameter. Here it is first.

Yes... The reason I did it that way is that in a former `Writer` (`DowncastWriter`) class, the only methods which has `element` as a parameter are `set/remove` ones (the other usually have position or range, etc) and it is indeed the last parameter there.

However, in the `UpcastWriter` all methods have one parameter in common which is an `element` and the rest changes, e.g.:

* `clone( element, deep = false ) { ... }`
* `appendChild( element, items ) { ... }`
* `removeChildren( element, index, howMany = 1 ) { ... }`
* `rename( element, newName ) { ... }`

So:

1. It makes it consistent internally, since all methods needs `element` and it is always the first parameter.
1. To keep it consistent with `DowncastWriter`, `element` parameter should be the last parameter, but then for methods with default parameters it actually can't be the last keeping things inconsistent, like:
    * `clone( element, deep = false ) { ... }`
    * `rename( newName, element ) { ... }`
    * `removeChildren( index, element, howMany = 1 ) { ... }`
    * `insertChild( index, items, element ) { ... }`

From my perspective it is a choice between keeping it internally consistent or trying to make it consistent with some `DowncastWriter` methods by making it internally inconsistent. WDYT @scofalik ?

f1ames:
> Note that DOMConverter should now use this writer. In can be changed as a part of this ticket or reported as a follow-up.

I will propose a follow-up issue.

pjasiun:
Yep. DOM converter now is using protected elements API and it should use this writer instead.

scofalik:
@f1ames Correct me if I am wrong, but AFAICS the methods that are not available in `DowncastWriter` are:

* `clone( element, deep = false )`
* `appendChild( element, items )`
* `insertChild( element, index, items )`
* `removeChildren( element, index, howMany = 1 )`
* `remove( element )`
* `replace( oldElement, newElement )`
* `rename( element, newName )`

So if we would change `set/remove`s, those are which would be "inconsistent". Let's break them down.

* `clone()` - can't do much here, but since `deep` really is an optional parameter, I think it is fine. * `appendChild()` and `insertChild()` make sense with a different parameters order too. * `removeChildren` looks bad, TBH, but I'd drop default value there and make `element` the last parameter. This is the only method that IMO will look bad.
* `remove()` is inconsistent only because `downcastWriter` takes a `range` only. TBH, I'd change `downcastWriter` to accept both `range` and `element` and simply check if `element` is passed, if so, create a range on it (`Range.createOn()`). When this is done, the methods are consistent enough to get a thumbs up from me.
* `replace()` is fine as it is as it takes two elements, however, to be consistent, `oldElement` probably should be the second parameter, but I don't like it that way. I'd leave it as it is.
* `rename()` can be changed to be consistent and it will still look okay.

scofalik:
Sorry, `rename()` is available in `DowncastWriter` and has wrong order. I'd change the order there.

src/view/upcastwriter.js
f1ames: From what I see, both produce the same link `.../module_engine_view_node-Node.html#event:change` but I will update it to make it consistent.

However, I just discovered it doesn't work properly in our docs 😱 The generated links are e.g. [`...api/module_engine_conversion_upcastdispatcher-UpcastDispatcher.html#event:element`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_conversion_upcastdispatcher-UpcastDispatcher.html#event:element), while to working one is [`...api/module_engine_conversion_upcastdispatcher-UpcastDispatcher.html#event-element`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_conversion_upcastdispatcher-UpcastDispatcher.html#event-element) (so `-` instead of `:` in event name). I will report it.

f1ames:
As I applied all requested changes, it is ready another R.

PR #1514, opened 2018-08-24
Changed long name returned by Operation.className property to the short one

### LINKED ISSUE #1513: Change long Operation.className to short one.
We don't need this long names anymore. Besides, we are going to use this names to identify Operation and using such a long name will be inconvenient.

oskarwrobel:

Feature: Changed long name returned by `Operation.className` property to the short one. Closes #1513.

---

### Additional information

### DIFF
--- a/src/model/operation/operation.js
+++ b/src/model/operation/operation.js
@@ -112,7 +112,7 @@ export default class Operation {
 	 * @type {String}
 	 */
 	static get className() {
-		return 'engine.model.operation.Operation';
+		return 'Operation';
 	}
 
 	/**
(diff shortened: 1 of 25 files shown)


src/model/operation/operation.js
```diff
@@ -112,7 +112,7 @@ export default class Operation {
 	 * @type {String}
 	 */
 	static get className() {
-		return 'engine.model.operation.Operation';
+		return 'Operation';
 	}
```
pjasiun: I think you should never ask for this property of the base class. I think we should remove this method or replace it with `throw`.

oskarwrobel: Agree that this method should be never used (in source code) for the base class. But If you want to drop this method for `Operation` class let's do it in a separate ticket because it needs some fixes in `toJSON()` method and in tests.

PR #1535, opened 2018-09-07
Prevent rendering when one is in the `change` block.

### LINKED ISSUE #1528: Nested contenteditables + comment markers make page jumps
**How to reproduce?**
1. Create an image
2. Write some text in its caption
3. Create a comment to the caption
4. Scroll down the page. 
5. Select text in the caption
6. Select text in the main content editable element

**Current behavior:**
Scroll jumps to the top.

**Additional info**
Reproducible on Chrome, Safari
The same happens with table cells.

I've tried to debug it with @oskarwrobel and here are our thoughts:

Clicking the marker makes the comment's contenteditable active and triggers its rendering. Then Clicking outside of it, makes it inactive, rerenders the comment and rerenders the main content editable element. It triggers two times https://github.com/ckeditor/ckeditor5-engine/blob/5c0a34d6f277fed063d8ed40af5b8ebda4690e14/src/view/renderer.js#L756
which probably doesn't find the selection in the editor for the second time, thus it creates a selection for a while at the top of the editor and scrolls to it. 

Possible solutions:
* Switch these two lines - https://github.com/ckeditor/ckeditor5-engine/blob/5c0a34d6f277fed063d8ed40af5b8ebda4690e14/src/view/renderer.js#L231-L232 by @oskarwrobel. Not sure if it doesn't break something else.
* Add `setTimeout 0` in comments when they change state between active and inactive.
* Add `domElement.focus( { preventScroll: false } )`. Unfortunately, it works only in Chrome, doesn't work in Safari and it looks like a "postfixer".

WDYT?
cc @pjasiun @Reinmar

pjasiun:

Internal: Prevent rendering when one is in the `change` block. Closes #1528.

### DIFF
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -73,6 +73,15 @@ export default class EditingController {
 		// Whenever model document is changed, convert those changes to the view (using model.Document#differ).
 		// Do it on 'low' priority, so changes are converted after other listeners did their job.
 		// Also convert model selection.
+		this.listenTo( this.model, '_beforeChanges', () => {
+			this.view.disabledRendering = true;
+		}, { priority: 'highest' } );
+
+		this.listenTo( this.model, '_afterChanges', () => {
+			this.view.disabledRendering = false;
+			this.view.render();
+		}, { priority: 'lowest' } );
+
 		this.listenTo( doc, 'change', () => {
 			this.view.change( writer => {
 				this.downcastDispatcher.convertChanges( doc.differ, writer );
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -70,9 +70,10 @@ export default class EditingController {
 		const selection = doc.selection;
 		const markers = this.model.markers;
 
-		// Various plugins listen on model changes (on selection change, post fixers, etc.) and change the view as a
-		// result of the model change, what causes rendering in the middle of conversion, sometimes before the selection
-		// is converted. This is why we want to disable rendering as long as one is in the `model.change` block.
+		// When plugins listen on model changes (on selection change, post fixers, etc) and change the view as a result of
+		// model's change, they might trigger view rendering before the conversion is completed (e.g. before the selection
+		// is converted). We disable rendering for the length of the outermost model change() block to prevent that.
+		//
 		// See  https://github.com/ckeditor/ckeditor5-engine/issues/1528
 		this.listenTo( this.model, '_beforeChanges', () => {
 			this.view._disabledRendering = true;
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -70,7 +70,7 @@ export default class EditingController {
 		const selection = doc.selection;
 		const markers = this.model.markers;
 
-		// Various plugins listen on model changes (on selection change, post fixers, etc.) and changes the view as a
+		// Various plugins listen on model changes (on selection change, post fixers, etc.) and change the view as a
 		// result of the model change, what causes rendering in the middle of conversion, sometimes before the selection
 		// is converted. This is why we want to disable rendering as long as one is in the `model.change` block.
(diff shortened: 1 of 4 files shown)


src/controller/editingcontroller.js
```diff
@@ -70,6 +70,19 @@ export default class EditingController {
 		const selection = doc.selection;
 		const markers = this.model.markers;
 
+		// Various plugins listen on model changes (on selection change, post fixers, etc.) and change the view as a
```
Reinmar: When plugins listen on model changes (on selection change, post fixers, etc) and change the view as a result of model's change, they might trigger view rendering before the conversion is completed (e.g. before the selection is converted). We disable rendering for the length of the outermost model change() block to prevent that.

See https://github.com/ckeditor/ckeditor5-engine/issues/1528

src/model/model.js
```diff
@@ -487,6 +491,22 @@ export default class Model {
 	 * @param {module:engine/model/writer~Writer} writer `Writer` instance that has been used in the change block.
 	 */
 
+	/**
+	 * Fired when entering the first {@link module:engine/model/model~Model#enqueueChange} or
+	 * {@link module:engine/model/model~Model#change} block of the pending changes.
```
Reinmar: That "pending changes" thing is misleading. It sounds like this is the Nth+ block. A one that was scheduled but needed to be postponed so it's "pending". Isn't it just the "outermost enqueueChange/change block"?

src/model/model.js
```diff
@@ -487,6 +491,22 @@ export default class Model {
...
+	 *
+	 * @protected
+	 * @event _beforeChanges
+	 */
+
+	/**
+	 * Fired when leaving the last {@link module:engine/model/model~Model#enqueueChange} or
+	 * {@link module:engine/model/model~Model#change} block of the pending changes.
```
Reinmar: The same as above.

src/view/view.js
```diff
@@ -126,6 +126,14 @@ export default class View {
 		 */
 		this._postFixersInProgress = false;
 
+		/**
+		 * Internal flag to temporary disable rendering. See usage in the editing controller.
+		 *
+		 * @protected
+		 * @member {Boolean} module:engine/view/view~View#_disabledRendering
+		 */
+		this._disabledRendering = false;
```
Reinmar: `_renderingDisabled` would be more grammatical. Plus, you have `_renderingInProgress` above already.

tests/controller/editingcontroller.js
```diff
@@ -339,6 +339,112 @@ describe( 'EditingController', () => {
 			expect( getViewData( editing.view, { withoutSelection: true } ) )
 				.to.equal( '<p></p><p>f<span>oo</span></p><p>bar</p>' );
 		} );
+
+		describe( 'rendering preventing in the change block', () => {
```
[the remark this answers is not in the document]
Reinmar: or even:

preventing rendering while in the change() block

tests/controller/editingcontroller.js
```diff
@@ -339,6 +339,112 @@ describe( 'EditingController', () => {
...
+			let renderSpy;
+
+			beforeEach( () => {
+				renderSpy = sinon.spy();
+
+				editing.view.on( 'render', renderSpy );
+			} );
+
+			it( 'should not call render in the change block', () => {
```
Reinmar: model.change()

tests/controller/editingcontroller.js
```diff
@@ -339,6 +339,112 @@ describe( 'EditingController', () => {
...
+				model.change( writer => {
+					executeSomeModelChange( writer );
+
+					expect( renderSpy.called ).to.be.false;
+				} );
+
+				expect( renderSpy.called ).to.be.true;
+			} );
+
+			it( 'should not call render in the change block even if view change was called', () => {
```
Reinmar: view.change()

Reinmar:
Except for some wording and naming, this LGTM.

Reinmar:
Tests look good so I'm merging it.

PR #1544, opened 2018-09-12
Merge x merge becomes NoOperation if it is not going to be affected by undoing operation

### LINKED ISSUE #1543: New OT: Merge x merge special case not incorrect for redo
In `MergeOperation` x `MergeOperation` transformation, the merge may become a "do-nothing-merge" if both merge are same. That "do-nothing-merge" is a `MergeOperation` that moves `0` nodes from a node in graveyard (it merges a node that is already in graveyard). This has no impact on model tree but by not changing to `NoOperation` it is possible to "revert" this operation if the other `MergeOperation` got later undone.

Anyway, an example of such do-nothing-merge is: `MergeOperation $graveyard [ 0 ] -> main [ 1, 2 ] (howMany = 0)`. However, when this gets reversed, it becomes a `SplitOperation` with `howMany = 0` which is incorrect.

The scenario when the bug happens is this:

1. Client A and Client B merge at the same position.
2. Client A undoes and sends change to Client B.
3. Client B then undoes and generates "do-nothing-merge".
4. Client B redoes.
5. Editor throws.

scofalik:

Internal: `MergeOperation` transformed by an equal `MergeOperation` becomes a `NoOperation` if it is not going to be affected by undoing operation. Closes #1543.

### DIFF
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -1220,13 +1220,17 @@ setTransformation( MergeOperation, MergeOperation, ( a, b, context ) => {
 	// Doing this instead of returning `NoOperation` allows for a correct undo later.
 	//
 	if ( a.sourcePosition.isEqual( b.sourcePosition ) && a.targetPosition.isEqual( b.targetPosition ) ) {
-		const path = b.graveyardPosition.path.slice();
-		path.push( 0 );
+		if ( context.bWasUndone ) {
+			const path = b.graveyardPosition.path.slice();
+			path.push( 0 );
 
-		a.sourcePosition = new Position( b.graveyardPosition.root, path );
-		a.howMany = 0;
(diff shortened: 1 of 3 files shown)


src/model/operation/transform.js
```diff
@@ -1220,13 +1220,17 @@ setTransformation( MergeOperation, MergeOperation, ( a, b, context ) => {
...
-		a.sourcePosition = new Position( b.graveyardPosition.root, path );
-		a.howMany = 0;
+			a.sourcePosition = new Position( b.graveyardPosition.root, path );
+			a.howMany = 0;
 
-		return [ a ];
+			return [ a ];
+		} else {
+			return [ new NoOperation( 0 ) ];
+		}
```
pjasiun: A few words of comments are needed here. At least link to the ticket.

PR #1545, opened 2018-09-13
T/ckeditor5/1243: Unify widget insertion

jodator:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Other: The `model.insertContent()` accepts range and position. Closes ckeditor/ckeditor5#1243.

---

### Additional information

* Associated PRs:
    - https://github.com/ckeditor/ckeditor5-widget/pull/53
    - https://github.com/ckeditor/ckeditor5-table/pull/111
    - https://github.com/ckeditor/ckeditor5-image/pull/232
    - https://github.com/ckeditor/ckeditor5-media-embed/pull/34
- API updates-only changes on branches (no PRs):
    - https://github.com/ckeditor/ckeditor5-list/tree/t/ckeditor5/1243
    - https://github.com/ckeditor/ckeditor5-autosave/tree/t/ckeditor5/1243
    - https://github.com/ckeditor/ckeditor5-clipboard/tree/t/ckeditor5/1243
- suggested merge message for list, autosave & clipboard:
```md
Internal: Remove default parameter from `model.insertContent()` calls.
```
- Above branch constellation is commited on [t/1243](https://github.com/ckeditor/ckeditor5/tree/t/1243) branch on `ckeditor5` repo.

PR #1550, opened 2018-09-17
 Schema#schema.checkAttributeInSelection() checks attributes of text

pomek:

Fix: `Schema#schema.checkAttributeInSelection()` will check attributes of the text. Closes #1545.

### DIFF
--- a/tests/model/schema.js
+++ b/tests/model/schema.js
@@ -1047,6 +1047,26 @@ describe( 'Schema', () => {
 					return true;
 				}
 			} );
+
+			schema.addAttributeCheck( ( ctx, attributeName ) => {
+				// Disallow 'italic' on $text that has attribute 'bold'.
+				if ( inTextWithBold( ctx ) && attributeName == 'italic' ) {
+					return false;
+				}
+				// Allow 'italic' on p>$text.
+				if ( ctx.endsWith( 'p $text' ) && attributeName == 'italic' ) {
+					return true;
+				}
+
+				// Allow 'italic' on $root>p.
+				if ( ctx.endsWith( '$root p' ) && attributeName == 'italic' ) {
+					return true;
+				}
+
+				function inTextWithBold( context ) {
+					return context.endsWith( '$text' ) && context.last.getAttribute( 'bold' );
+				}
+			} );
 		} );
 
 		describe( 'when selection is collapsed', () => {
@@ -1061,6 +1081,11 @@ describe( 'Schema', () => {
 
 				setData( model, '[]' );
 				expect( schema.checkAttributeInSelection( doc.selection, attribute ) ).to.be.false;
+			} );
+
+			it( 'should check attributes of text', () => {
+				setData( model, '<p><$text bold="true">f[]oo</$text></p>' );
+				expect( schema.checkAttributeInSelection( doc.selection, 'italic' ) ).to.be.false;
 			} );
 		} );
 
@@ -1101,6 +1126,11 @@ describe( 'Schema', () => {
 			it( 'should return true when checking element when attribute is already present', () => {
 				setData( model, '[<figure name="figure" title="title"></figure>]' );
 				expect( schema.checkAttributeInSelection( doc.selection, 'title' ) ).to.be.true;
+			} );
+
+			it( 'should check attributes of text', () => {
+				setData( model, '<p><$text bold="true">f[o]o</$text></p>' );
+				expect( schema.checkAttributeInSelection( doc.selection, 'italic' ) ).to.be.false;
 			} );
 		} );
 	} );
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -488,8 +488,14 @@ export default class Schema {
 	 */
 	checkAttributeInSelection( selection, attribute ) {
 		if ( selection.isCollapsed ) {
+			const firstPosition = selection.getFirstPosition();
+			const context = [
+				...firstPosition.getAncestors(),
+				firstPosition.root.getNodeByPath( firstPosition.path )
+			].filter( Boolean );
+
 			// Check whether schema allows for a text with the attribute in the selection.
-			return this.checkAttribute( [ ...selection.getFirstPosition().getAncestors(), '$text' ], attribute );
+			return this.checkAttribute( context, attribute );
 		} else {
 			const ranges = selection.getRanges();
 


Reinmar:
Note: you can set selection attributes completely independently from the text it's in. That happens e.g. when you press ctrl+b on an empty selection – we set only the selection's attribute then. This action (e.g. ctrl+b) should already lead to `Schema#checkAttributeInSelection()` to consider this attribute when doing a check.

Which means that you need to include an artificial `$text` + attributes in the check. I think that `SchemaContext()` does not allow that (you can pass either a node or a string) so you'll need to extend it to start supporting objects.

pomek:
> This action (e.g. ctrl+b) should already lead to Schema#checkAttributeInSelection() to consider this attribute when doing a check.

May I understood anything incorrectly but it doesn't. Ctrl+b on collapsed selection just set its attribute.

I even could set attribute on the selection that wasn't allowed in the schema. I disabled this:

```js
editor.model.schema.extend( '$text', { allowAttributes: BOLD } );
```

and in `AttributeCommand#refresh()`:

```js
this.isEnabled = true;
```

After pressing CTRL+B, selection's attributes looks like:

```js
MapIterator {"bold" => true}
```

And after typing something in the editor:

```html
<p><i>This</i> <s>is</s> <code>an</code> editor <u>instance</u>.<strong>I should not be bold.</strong></p>
```

So the question is - why I'm able to typing bolded text even if the schema does not allow "bold" attribute on $text?

image

scofalik:
`Schema` is (and should be) used by a specific feature not by the engine itself. It means that writer, conversion or other associated parts of the engine do not check `Schema`. So if you already have an attribute set on the `Selection` you will be able to insert text with the given attribute. Same way as, for example, you could insert a `<paragraph>` into `<listItem>` using the writer directly.

So the responsibility to check the `Schema` is on the feature. The feature should if given action is possible and deny it if not.

Reinmar:
It's now documented: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/deep-dive/schema.html#implementing-additional-constraints (+ the next section)

scofalik:
> Which means that you need to include an artificial `$text` + attributes in the check. I think that `SchemaContext()` does not allow that (you can pass either a node or a string) so you'll need to extend it to start supporting objects.

What you write makes sense cause you can go: <kbd>Ctrl + B</kbd> then <kbd>Ctrl + I</kbd> and then start typing, which could make an incorrect text node. You can't really right now create an artificial text node inside `Schema#checkAttributeInSelection` cause you need a writer for that. So the only way to go is to add more code to`checkAttributeInSelection`.

I am not sure about `SchemaContext`. How will it make use of an object (I assume you want to pass something like `{ name, attribute }`)? I am not really experienced with the internals of new `Schema` so I'll just ask a stupid question - will `SchemaContext` have to build a node from that data (which I think is not possible/difficult/hacky). Also wouldn't it be backward incompatible for custom checks we already have?

I rather thought about extending `checkAttributeInSelection` itself. Maybe extract some of it to a private `_checkAttributeForCollapsedSelection` if it would grow too big. Or make it possible to pass an array instead of a string to `checkAttribute`. But that makes it backward incompatible too.

Anyway sorry for stupid questions or maybe getting it into wrong direction but `Schema` is a little complicated.

Reinmar:
> You can't really right now create an artificial text node inside `Schema#checkAttributeInSelection` cause you need a writer for that.

I don't want to create a real node. It's enough if we could do it like this:

```js
const context = [
    '$root',                                            // SUPPORTED
    paragraphElement,                                   // SUPPORTED
    { name: '$text', attributes: { bold: true } }       // NOT SUPPORTED YET
];

schema.checkAttribute( context, 'italic' );
```

The `context`'s type is defined in [`SchemaContextDefinition`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-SchemaContextDefinition.html) and it's pretty reach already. It's a matter of adding support for node definition.

In the next step, that `context` is passed to `SchemaContext( context )` and that constructor turns every item in `context` to [`SchemaContextItem`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-SchemaContextItem.html). That will be a no brainer as well.

Reinmar:
OTOH... maybe `Schema#checkAttributeInElement()` could create that text node on the fly... Why not, in fact?

The whole fix would boild down to this:

```js
const context = [
	...firstPosition.getAncestors(),
	new Text( '', selection.getAttributes() )
];

// Check whether schema allows for a text with the attribute in the selection.
return this.checkAttribute( context, attribute );
```

Yes, we can use `new Text()` cause it's a protected constructor.

Other developers would need to do this which is not that bad too:

```js
const text = model.change( writer => writer.createText( '', attributes ) );
```

And thanks to passing a real text node to Schema's methods we'll be using the existing `SchemaContext()` mechanisms for mapping them to `SchemaContextItem`. So... the fix may really be as simple as what I showed above. It's worth checking, IMO.

BTW, theoretically `new Text()` requires `attributes` to be an `Object`, but we use `toMap()` in `Node()` so I think you can pass an iterable there like `new Text( 'foo', selection.getAttributes() )`.

pomek:
After applying the patch that takes selection attributes to check, the case described in the basic-styles issue (https://github.com/ckeditor/ckeditor5-basic-styles/issues/77#issuecomment-421932652) is no longer reproducible.

Reinmar:
Please also add a test which checks `<paragraph>foo[]bar</paragraph>` but add attributes to the selection via `writer.setSelectionAttribute()`. We want to be sure that base on the selection's attrs, not on anything in the model structure.

tests/model/schema.js
```diff
@@ -1047,6 +1047,26 @@ describe( 'Schema', () => {
 					return true;
 				}
 			} );
+
+			schema.addAttributeCheck( ( ctx, attributeName ) => {
+				// Disallow 'italic' on $text that has attribute 'bold'.
+				if ( inTextWithBold( ctx ) && attributeName == 'italic' ) {
+					return false;
+				}
```
Reinmar: Missing blank line.

Reinmar:
BTW, I'm not sure why but tests are failing on one of CIs jobs. Maybe you need to merge `master` to your branch?

pomek:
> BTW, I'm not sure why but tests are failing on one of CIs jobs. Maybe you need to merge master to your branch?

It can be this. My branch is behind the master.

tests/model/schema.js
Reinmar: Right above there's another addAttributeCheck() call. Why haven't you extended it?

tests/model/schema.js
```diff
@@ -1062,6 +1083,33 @@ describe( 'Schema', () => {
 				setData( model, '[]' );
 				expect( schema.checkAttributeInSelection( doc.selection, attribute ) ).to.be.false;
 			} );
+
+			it( 'should check attributes of the selection (selection at the beginning of the text)', () => {
+				setData( model, '<p><$text bold="true">[]foo</$text></p>' );
+				expect( schema.checkAttributeInSelection( doc.selection, 'italic' ) ).to.be.false;
```
Reinmar: This is suspicious... why is it false? What gave the selection the bold attr? I guess `setData()` did, but I'd remove this test because it's confusing.

PR #1561, opened 2018-09-28
Mark attributes to be refreshed.

### LINKED ISSUE #1560: updateElementMappings does not refresh attributes
This code:

https://github.com/ckeditor/ckeditor5-engine/blob/0821d9098b76c25fef577d6ccfe0c1324d0bc8ed/src/view/renderer.js#L317-L318

mark children of the remap element to be refreshed but do not map its attributes. It means that the DOM element may have old attributes after remapping.

pjasiun:

Bug: `updateElementMappings` does not refresh attributes. Closes #1560.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -314,8 +314,9 @@ export default class Renderer {
 		this.domConverter.unbindDomElement( domElement );
 		this.domConverter.bindElements( domElement, viewElement );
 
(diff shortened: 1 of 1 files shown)


pjasiun:
Will be fixed by https://github.com/ckeditor/ckeditor5-engine/pull/1567.

PR #1567, opened 2018-10-01
Marked reused element attributes to be rendered if replacing element was also marked

### LINKED ISSUE #1560: updateElementMappings does not refresh attributes

f1ames:

Fix: Marked reused element attributes to be rendered if replacing element was also marked. Closes #1560. Closes #1561.

---

### Additional information

In-depth explanation of the bug - https://github.com/ckeditor/ckeditor5-engine/issues/1560#issuecomment-425948884 and proposed solution explanation - https://github.com/ckeditor/ckeditor5-engine/issues/1560#issuecomment-425950993.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -306,7 +306,9 @@ export default class Renderer {
 		// It may also happen that 'newViewChild' mapping is not present since its parent mapping
 		// was already removed (the 'domConverter.unbindDomElement()' method also unbinds children
 		// mappings) so we also check for '!newViewChild'.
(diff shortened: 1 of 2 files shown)


f1ames:
For the push, CI is failing because of some 3 wrap/unwrap tests:
image

PR #1569, opened 2018-10-02
T/1564

### LINKED ISSUE #1564: View post-fixers should be called only once just before "render" event
Currently the post fixer are called on every `view.change()` call but I think that they should be called after all view changes - before rendering.

jodator:

Other: View post-fixer should be called once while rendering model changes. Closes #1564.

BREAKING CHANGE: View post-fixers are now called once while rendering model changes.

---

### Additional information

* Required fix for bugs related to time of post-fixers are called after model changes.

### DIFF
--- a/src/view/view.js
+++ b/src/view/view.js
@@ -361,10 +361,9 @@ export default class View {
 		callback( this._writer );
 		this._ongoingChange = false;
 
-		// This lock is used by editing controller to render changes from outer most modelchange() once. As plugins might call
-		// view.change() inside model.change() block - this will ensures that postfixers are called once before rendering.
+		// This lock is used by editing controller to render changes from outer most model.change() once. As plugins might call
+		// view.change() inside model.change() block - this will ensures that postfixers and rendering are called once after all changes.
 		if ( !this._renderingDisabled ) {
-			// Execute all document post-fixers after the change.
 			this._postFixersInProgress = true;
 			this.document._callPostFixers( this._writer );
 			this._postFixersInProgress = false;
--- a/src/view/view.js
+++ b/src/view/view.js
@@ -361,11 +361,14 @@ export default class View {
 		callback( this._writer );
 		this._ongoingChange = false;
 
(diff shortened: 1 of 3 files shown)


src/view/view.js
```diff
@@ -361,12 +361,14 @@ export default class View {
 		callback( this._writer );
 		this._ongoingChange = false;
 
-		// Execute all document post-fixers after the change.
-		this._postFixersInProgress = true;
-		this.document._callPostFixers( this._writer );
-		this._postFixersInProgress = false;
-
+		// This lock is used by editing controller to render changes from outer most modelchange() once. As plugins might call
```
pjasiun: `model.change`

src/view/view.js
```diff
@@ -361,12 +361,14 @@ export default class View {
...
+		// view.change() inside model.change() block - this will ensures that postfixers are called once before rendering.
```
pjasiun: `ensures that postfixers are called once before rendering.` -> `ensures that postfixers and rendering are called once after all changes.`

src/view/view.js
```diff
@@ -361,12 +361,14 @@ export default class View {
...
 		if ( !this._renderingDisabled ) {
+			// Execute all document post-fixers after the change.
```
pjasiun: I believe this comment is not needed anymore.

tests/view/view/view.js
```diff
@@ -396,6 +396,35 @@ describe( 'view', () => {
...
+			const postFixerSpy = sinon.spy( () => false );
+			const changeSpy = sinon.spy();
+			const eventSpy = sinon.spy();
+
+			viewDocument.registerPostFixer( postFixerSpy );
+
+			view.on( 'render', eventSpy );
+
+			// This is set in editing controller on `model#_beforeChanges` event
+			view._renderingDisabled = true;
```
pjasiun: Can you use `model.change` here instead of `_renderingDisabled`? (and most probably move this test to editing controller then).

PR #1573, opened 2018-10-05
Range#_getTransformedBySplitOperation will expand the range if insertionPosition is equal to the range end

scofalik:

Internal: `Range#_getTransformedBySplitOperation` will expand the range if `insertionPosition` is equal to the range end. Modified transformations to align with that change. Closes https://github.com/ckeditor/ckeditor5/issues/1278.

### DIFF
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -497,6 +497,10 @@ class ContextFactory {
 							this._setRelation( opA, opB, 'mergeTargetNotMoved' );
 						}
 
+						if ( opA.sourcePosition.isEqual( opB.sourcePosition ) ) {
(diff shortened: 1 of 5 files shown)


scofalik:
Updated, sorry.

Reinmar:
image

The same scenario, but undo, redo:

oct-05-2018 11-00-06

scofalik:
Actually, this is a little different scenario. In the original one selection was collapsed. Please move it to a separate ticket, thanks.

Reinmar:
You're right, I'm reporting it to a separate ticket.

Reinmar:
Here it is: https://github.com/ckeditor/ckeditor5/issues/1287.

PR #1574, opened 2018-10-05
T/1554: Model.insertContent() should work with item & related changes

### LINKED ISSUE #1554: model.insertContentent( content, element ) does not work (and related issues)
It does not work because you cannot do this:

```js
const sel = new Selection( element );
```

Unfortunately, `Selection.setTo()` forces you to pass the `offset`. I think it's unnecessary, but we can keep this limitation for clarity purposes.

However, in both cases, the additional problem with `insertContent()` is that it does not allow you to specify the `offset`:

```js
model.insertContent( content, element, 0 );
model.insertContent( content, element, 'end' );
```

All this should be doable, but it's not.

jodator:

Fix: The `model.insertContent()` will accept offset parameter. Closes #1554.

BREAKING CHANGE: The `offset` parameter of model `Position.createAt()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of model `Range.createCollapsedAt()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of model `Selection#setFocus()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of model `writer#insert()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of model `writer#insertText()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of model `writer#insertElement()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of model `writer#move()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of model `writer#setSelectionFocus()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of view `writer#setSelectionFocus()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of view `Position.createAt()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of view `Range.createCollapsedAt()` is no longer optional when `itemOrPosition` parameter is an item.
BREAKING CHANGE: The `offset` parameter of view `Selection#setFocus()` is no longer optional when `itemOrPosition` parameter is an item.

---

### Additional information

* There are changes in other repos gathered in `mgit.json` on ckeditor5 repository: https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-engine/1554 which includes:
  - https://github.com/ckeditor/ckeditor5-clipboard/tree/t/ckeditor5-engine/1554
  - https://github.com/ckeditor/ckeditor5-image/tree/t/ckeditor5-engine/1554
  - https://github.com/ckeditor/ckeditor5-list/tree/t/ckeditor5-engine/1554
  - https://github.com/ckeditor/ckeditor5-media-embed/tree/t/ckeditor5-engine/1554
  - https://github.com/ckeditor/ckeditor5-paragraph/tree/t/ckeditor5-engine/1554
  - https://github.com/ckeditor/ckeditor5-table/tree/t/ckeditor5-engine/1554
  - https://github.com/ckeditor/ckeditor5-widget/tree/t/ckeditor5-engine/1554
* I've added the support for `Element` in the `model.insertContent()` method. I've also removed all the optional `offset` parameters from model & view methods (so the view methods will have the same API as they model counterparts - ie `Position.createAt()`.

### DIFF
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -816,7 +816,7 @@ export default class Position {
 	 * * {@link module:engine/model/position~Position.createFromPosition}.
 	 *
 	 * @param {module:engine/model/item~Item|module:engine/model/position~Position} itemOrPosition
-	 * @param {Number|'end'|'before'|'after'} offset Offset or one of the flags. Used only when
+	 * @param {Number|'end'|'before'|'after'} [offset] Offset or one of the flags. Used only when
 	 * first parameter is a {@link module:engine/model/item~Item model item}.
 	 */
 	static createAt( itemOrPosition, offset ) {
@@ -834,7 +834,7 @@ export default class Position {
 			} else if ( offset !== 0 && !offset ) {
 				throw new CKEditorError(
 					'model-position-createAt-required-second-parameter: ' +
-					'Position.createAt requires the second parameter offset.' );
+					'Position.createAt requires the second parameter offset when first parameter is a model item.' );
 			}
 
 			return this.createFromParentAndOffset( node, offset );
--- a/src/model/position.js
+++ b/src/model/position.js
@@ -816,7 +816,7 @@ export default class Position {
 	 * * {@link module:engine/model/position~Position.createFromPosition}.
 	 *
 	 * @param {module:engine/model/item~Item|module:engine/model/position~Position} itemOrPosition
-	 * @param {Number|'end'|'before'|'after'} [offset=0] Offset or one of the flags. Used only when
+	 * @param {Number|'end'|'before'|'after'} offset Offset or one of the flags. Used only when
 	 * first parameter is a {@link module:engine/model/item~Item model item}.
 	 */
 	static createAt( itemOrPosition, offset ) {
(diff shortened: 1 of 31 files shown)


src/model/position.js
```diff
@@ -831,8 +831,10 @@ export default class Position {
 				return this.createBefore( node );
 			} else if ( offset == 'after' ) {
 				return this.createAfter( node );
-			} else if ( !offset ) {
-				offset = 0;
+			} else if ( offset !== 0 && !offset ) {
+				throw new CKEditorError(
```
Reinmar: Requires a doc string.

src/model/writer.js
```diff
@@ -1043,7 +1043,7 @@ export default class Writer {
 	 * {@link module:engine/model/position~Position.createAt `Position.createAt()`} parameters.
 	 *
 	 * @param {module:engine/model/item~Item|module:engine/model/position~Position} itemOrPosition
-	 * @param {Number|'end'|'before'|'after'} [offset=0] Offset or one of the flags. Used only when
+	 * @param {Number|'end'|'before'|'after'} [of~fset=0] Offset or one of the flags. Used only when
```
Reinmar:
```suggestion
	 * @param {Number|'end'|'before'|'after'} [offset=0] Offset or one of the flags. Used only when
```

PR #1579, opened 2018-10-16
Upcast writing and View enhancements

### LINKED ISSUE #1549: Inconsistent "upcast writing"
We've just introduced `UpcastWriter` and when I've stumbled upon this piece of code (in the documentation) I thought that something's wrong:

```js
editor.plugins.get( 'Clipboard' ).on( 'inputTransformation', ( evt, data ) => {
	if ( data.content.childCount == 1 && isUrlText( data.content.getChild( 0 ) ) ) {
		const linkUrl = data.content.getChild( 0 ).data;

		data.content = new ViewDocumentFragment( [
			ViewElement(
				'a',
				{ href: linkUrl },
				[ new ViewText( linkUrl ) ]
			)
		] );
	}
} );
```

Why aren't we using `UpcastWriter` here? If, instead of creating a fresh new document fragment, we'd like to modify the one in `data.content`, we'd need to use the upcast writer. So I actually thought that this piece of code is wrong... and it is – constructors of these node classes are protected so we can't use them like this. But they are not exposed in `UpcastWriter`.

Also, `View#change()` doesn't return the return value of its callback just like `Model#change()` does which would be highly inconvenient.

f1ames:

Other: Introduced `createDocumentFragment()`, `createElement()` and `createText()` methods in `UpcastWriter`. The `View.change()` method now returns result of its callback. Closes #1549.

---

### Additional information

See #1549 for details.

### DIFF
--- a/src/view/documentfragment.js
+++ b/src/view/documentfragment.js
@@ -14,11 +14,11 @@ import isIterable from '@ckeditor/ckeditor5-utils/src/isiterable';
 import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
 
 /**
- * DocumentFragment class.
+ * Document fragment.
  *
- * The constructor of this class shouldn't be used directly. To create new DocumentFragment instance use the
+ * To create a new document fragment instance use the
  * {@link module:engine/view/upcastwriter~UpcastWriter#createDocumentFragment `UpcastWriter#createDocumentFragment()`}
- * method instead.
+ * method.
  */
 export default class DocumentFragment {
 	/**
--- a/src/view/documentfragment.js
+++ b/src/view/documentfragment.js
@@ -20,6 +20,9 @@ export default class DocumentFragment {
 	/**
 	 * Creates new DocumentFragment instance.
 	 *
+	 * **Note:** Constructor of this class shouldn't be used directly in the code. Use the
+	 * {@link module:engine/view/upcastwriter~UpcastWriter#createDocumentFragment} method instead.
+	 *
 	 * @protected
 	 * @param {module:engine/view/node~Node|Iterable.<module:engine/view/node~Node>} [children] List of nodes to be inserted into
 	 * created document fragment.
--- a/src/view/documentfragment.js
+++ b/src/view/documentfragment.js
@@ -15,18 +15,18 @@ import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
 
 /**
  * DocumentFragment class.
+ *
+ * The constructor of this class shouldn't be used directly. To create new DocumentFragment instance use the
+ * {@link module:engine/view/upcastwriter~UpcastWriter#createDocumentFragment `UpcastWriter#createDocumentFragment()`}
+ * method instead.
  */
 export default class DocumentFragment {
 	/**
 	 * Creates new DocumentFragment instance.
 	 *
-	 * **Note:** Constructor of this class shouldn't be used directly in the code. Use the
-	 * {@link module:engine/view/upcastwriter~UpcastWriter#createDocumentFragment `UpcastWriter#createDocumentFragment()`}
-	 * method instead.
-	 *
 	 * @protected
-	 * @param {module:engine/view/node~Node|Iterable.<module:engine/view/node~Node>} [children] List of nodes to be inserted into
-	 * created document fragment.
+	 * @param {module:engine/view/node~Node|Iterable.<module:engine/view/node~Node>} [children]
+	 * A list of nodes to be inserted into the created document fragment.
 	 */
 	constructor( children ) {
 		/**
--- a/src/view/documentfragment.js
+++ b/src/view/documentfragment.js
@@ -21,7 +21,8 @@ export default class DocumentFragment {
 	 * Creates new DocumentFragment instance.
 	 *
 	 * **Note:** Constructor of this class shouldn't be used directly in the code. Use the
-	 * {@link module:engine/view/upcastwriter~UpcastWriter#createDocumentFragment} method instead.
+	 * {@link module:engine/view/upcastwriter~UpcastWriter#createDocumentFragment `UpcastWriter#createDocumentFragment()`}
+	 * method instead.
 	 *
 	 * @protected
 	 * @param {module:engine/view/node~Node|Iterable.<module:engine/view/node~Node>} [children] List of nodes to be inserted into
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -53,8 +53,9 @@ export default class Element extends Node {
 	 * {@link module:engine/view/downcastwriter~DowncastWriter#createAttributeElement} for inline element,
 	 * {@link module:engine/view/downcastwriter~DowncastWriter#createContainerElement} for block element,
 	 * {@link module:engine/view/downcastwriter~DowncastWriter#createEditableElement} for editable element,
-	 * {@link module:engine/view/downcastwriter~DowncastWriter#createEmptyElement} for empty element or
-	 * {@link module:engine/view/downcastwriter~DowncastWriter#createUIElement} for UI element instead.
+	 * {@link module:engine/view/downcastwriter~DowncastWriter#createEmptyElement} for empty element,
+	 * {@link module:engine/view/downcastwriter~DowncastWriter#createUIElement} for UI element instead or
+	 * {@link module:engine/view/upcastwriter~UpcastWriter#createElement} for general element creation.
 	 *
 	 * @protected
 	 * @param {String} name Node name.
--- a/src/view/element.js
+++ b/src/view/element.js
@@ -22,20 +22,21 @@ import { isPlainObject } from 'lodash-es';
  * This is why the type of the {@link module:engine/view/element~Element} need to
  * be defined by the feature developer. When creating an element you should use one of the following methods:
  *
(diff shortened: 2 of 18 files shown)


f1ames:
There is related PR (change in docs) in `clipboard` plugin - https://github.com/ckeditor/ckeditor5-clipboard/pull/59.

src/view/documentfragment.js
```diff
@@ -20,6 +20,9 @@ export default class DocumentFragment {
 	/**
 	 * Creates new DocumentFragment instance.
 	 *
+	 * **Note:** Constructor of this class shouldn't be used directly in the code. Use the
+	 * {@link module:engine/view/upcastwriter~UpcastWriter#createDocumentFragment} method instead.
```
Reinmar:
```suggestion
	 * {@link module:engine/view/upcastwriter~UpcastWriter#createDocumentFragment `UpcastWriter#createDocumentFragment()`}
	 * method instead.
```

Reinmar: Please repeat that change in other places as well. Without the additional link text, such a link generates as a useless `createDocumentFragment`. Without clicking it you don't know where to find it. Plus, it does not render with `()` so without reading the text it does not look like a method.

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
  * see {@link module:engine/view/downcastwriter~DowncastWriter writer}.
  */
 export default class UpcastWriter {
+	/**
```
Reinmar: Missing blank line.

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
...
+	 * Creates new {@link module:engine/view/documentfragment~DocumentFragment}.
```
Reinmar:
```suggestion
	 * Creates a new {@link module:engine/view/documentfragment~DocumentFragment}.
```

We don't need to specify the link's text in this case because we'd just repeat the default one.

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
...
+	 *
+	 * @see module:engine/view/documentfragment~DocumentFragment#constructor
+	 * @param {module:engine/view/node~Node|Iterable.<module:engine/view/node~Node>} [children]
+	 * List of nodes to be inserted into created document fragment.
+	 * @returns {module:engine/view/documentfragment~DocumentFragment} Created document fragment.
```
Reinmar:
```suggestion
	 * @returns {module:engine/view/documentfragment~DocumentFragment} The created document fragment.
```

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
...
+	createDocumentFragment( children ) {
+		return new DocumentFragment( children );
+	}
+
+	/**
+	 * Creates new {@link module:engine/view/element~Element}.
+	 *
+	 * Attributes can be passed in various formats:
+	 *
+	 *		new Element( 'div', { 'class': 'editor', 'contentEditable': 'true' } ); // object
```
Reinmar: This code sample is incorrect. You shouldn't use `new Element()`. Plus, you don't need to wrap those key names in `''`.

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
...
+	 * @param {module:engine/view/node~Node|Iterable.<module:engine/view/node~Node>} [children]
+	 * List of nodes to be inserted into created document fragment.
+	 * @returns {module:engine/view/documentfragment~DocumentFragment} Created document fragment.
+	 */
+	createDocumentFragment( children ) {
+		return new DocumentFragment( children );
+	}
+
+	/**
+	 * Creates new {@link module:engine/view/element~Element}.
```
Reinmar:
```suggestion
	 * Creates a new {@link module:engine/view/element~Element} instance.
```

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
...
+	 * @param {module:engine/view/node~Node|Iterable.<module:engine/view/node~Node>} [children]
+	 * List of nodes to be inserted into created element.
+	 * @returns {module:engine/view/element~Element} Created element.
+	 */
+	createElement( name, attrs, children ) {
+		return new Element( name, attrs, children );
+	}
+
+	/**
+	 * Creates new {@link module:engine/view/text~Text}.
```
Reinmar:
```suggestion
	 * Creates a new {@link module:engine/view/text~Text} instance.
```

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
...
+	 *
+	 * @see module:engine/view/text~Text#constructor
+	 * @param {String} data Text
+	 * @returns {module:engine/view/text~Text} Created text.
```
Reinmar:
```suggestion
	 * @returns {module:engine/view/text~Text} The created text node.
```

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
...
+	 */
+	createElement( name, attrs, children ) {
+		return new Element( name, attrs, children );
+	}
+
+	/**
+	 * Creates new {@link module:engine/view/text~Text}.
+	 *
+	 * @see module:engine/view/text~Text#constructor
+	 * @param {String} data Text
```
Reinmar:
```suggestion
	 * @param {String} data The text's data.
```

src/view/upcastwriter.js
```diff
@@ -17,6 +19,49 @@ import { isPlainObject } from 'lodash-es';
...
+	 * @returns {module:engine/view/element~Element} Created element.
+	 */
+	createElement( name, attrs, children ) {
+		return new Element( name, attrs, children );
+	}
+
+	/**
+	 * Creates new {@link module:engine/view/text~Text}.
+	 *
+	 * @see module:engine/view/text~Text#constructor
```
Reinmar: Don't link to protected things from public things. If there's any useful documentation in that protected API, move it here.

src/view/view.js
```diff
@@ -308,16 +308,19 @@ export default class View {
 	 * Change method is the primary way of changing the view. You should use it to modify any node in the view tree.
 	 * It makes sure that after all changes are made view is rendered to DOM. It prevents situations when DOM is updated
 	 * when view state is not yet correct. It allows to nest calls one inside another and still perform single rendering
-	 * after all changes are applied.
+	 * after all changes are applied. It also returns the result of its callback.
 	 *
-	 *		view.change( writer => {
-	 *			writer.insert( position1, writer.createText( 'foo' ) );
+	 *		const text = view.change( writer => {
+	 *			const newText = writer.createText( 'foo' );
```
Reinmar: Missing blank line.

src/view/view.js
```diff
@@ -308,16 +308,19 @@ export default class View {
 	 * Change method is the primary way of changing the view. You should use it to modify any node in the view tree.
 	 * It makes sure that after all changes are made view is rendered to DOM. It prevents situations when DOM is updated
 	 * when view state is not yet correct. It allows to nest calls one inside another and still perform single rendering
-	 * after all changes are applied.
+	 * after all changes are applied. It also returns the result of its callback.
```
Reinmar:
```suggestion
	 * after all changes are applied. It also returns the return value of its callback.
```

src/view/element.js
```diff
@@ -53,8 +53,9 @@ export default class Element extends Node {
 	 * {@link module:engine/view/downcastwriter~DowncastWriter#createAttributeElement} for inline element,
 	 * {@link module:engine/view/downcastwriter~DowncastWriter#createContainerElement} for block element,
 	 * {@link module:engine/view/downcastwriter~DowncastWriter#createEditableElement} for editable element,
-	 * {@link module:engine/view/downcastwriter~DowncastWriter#createEmptyElement} for empty element or
-	 * {@link module:engine/view/downcastwriter~DowncastWriter#createUIElement} for UI element instead.
+	 * {@link module:engine/view/downcastwriter~DowncastWriter#createEmptyElement} for empty element,
+	 * {@link module:engine/view/downcastwriter~DowncastWriter#createUIElement} for UI element instead or
+	 * {@link module:engine/view/upcastwriter~UpcastWriter#createElement} for general element creation.
```
Reinmar: I'd remove this documentation from here (and move it to the class docs if that information is missing there). Let's keep the maximum in the class definition. This method is protected so not many people will see it. 

Please also review other element/text/df classes for the same issue.

Reinmar:
A control question – did you check for `new Element/DF/Text()` use in at least API docs? I'm sure there's a lot of that in tests, but that we can leave (it's also correct cause constructors are protected, not private).

f1ames:
When it comes to https://github.com/ckeditor/ckeditor5-engine/pull/1579/commits/b0aa93b01d1214b4faa9fc1f8f94ee706a611dbf, I have changed all API samples which used protected constructors (e.g. `new Element()`) to use `DowncastWriter`.

It could be also changed to use `UpcastWriter`, however as `UpcastWriter` doesn't provide methods to create specialized elements (like `downcastWriter.createAttributeElement()`) some samples would require using both downcast and upcast writer which may look confusing.

PR #1581, opened 2018-10-16
MoveOperation x SplitOperation transformation for a case when graveyard element is moved

### LINKED ISSUE #1580: MoveOperation x SplitOperation transformation when graveyard element is moved
In `MoveOperation` x `SplitOperation` transformations, a case when graveyard element is moved has not been handled. This leads to bugs.

scofalik:

Fix: `MoveOperation` x `SplitOperation` transformation for a case when graveyard element is moved. Closes #1580.

---

### Additional information

While working on that bug I found an additional unhandled case in `MoveOperation` x `MergeOperation` transformation. We already handled a situation when an element to merge was removed. However, we didn't handle the situation when the removed element was one of multiple removed elements. I also added fixes for that case.

### DIFF
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -1669,12 +1669,21 @@ setTransformation( MoveOperation, SplitOperation, ( a, b, context ) => {
 	// The default case.
 	//
 	const transformed = moveRange._getTransformedBySplitOperation( b );
+	const ranges = [ transformed ];
 
-	a.sourcePosition = transformed.start;
-	a.howMany = transformed.end.offset - transformed.start.offset;
-	a.targetPosition = newTargetPosition;
+	// Case 5:
+	//
+	// Moved range contains graveyard element used by split operation. Add extra move operation to the result.
+	//
+	if ( b.graveyardPosition ) {
+		const movesGraveyardElement = moveRange.start.isEqual( b.graveyardPosition ) || moveRange.containsPosition( b.graveyardPosition );
 
-	return [ a ];
+		if ( a.howMany > 1 && movesGraveyardElement ) {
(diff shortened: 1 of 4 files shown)


src/model/range.js
```diff
@@ -517,13 +517,16 @@ export default class Range {
...
 		if ( this.end.isEqual( operation.insertionPosition ) ) {
 			end = this.end.getShiftedBy( 1 );
-		} else {
-			end = this.end._getTransformedBySplitOperation( operation );
+		}
+
+		// Below may happen when range contains graveyard element used by split operation.
+		if ( start.root != end.root ) {
+			// End position was next to the moved graveyard element and was moved with it. Fix it.
+			end = this.end.getShiftedBy( -1 );
```
pjasiun: Are you sure it will always fix the situation?

src/model/operation/transform.js
```diff
@@ -1909,11 +1941,32 @@ setTransformation( SplitOperation, MergeOperation, ( a, b, context ) => {
 } );
 
 setTransformation( SplitOperation, MoveOperation, ( a, b, context ) => {
+	const rangeToMove = Range.createFromPositionAndShift( b.sourcePosition, b.howMany );
+
 	if ( a.graveyardPosition ) {
+		// Case 1:
+		//
+		// Split operation graveyard node was moved. In this case move operation is stronger. Since graveyard element
```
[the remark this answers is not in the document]
scofalik: Other than "node" -> "element" it is fine, but it is a shortcut. It means that the element from the graveyard described by `SplitOpertaion#graveyardPosition` was moved to other place.

src/model/range.js
scofalik: There were two additional `if`s there before and I just couldn't write a test to cover that code.

PR #1582, opened 2018-10-18
t/1439: Firefox should visually move the caret to the new line after a soft break

### LINKED ISSUE #1439: [FF] Caret is rendered before a <br> despite the model selection being after it
It's Firefox only ;/

1. Clear all content.
2. Type "foo".
3. Press Shift+Enter. New line appears but caret isn't moved anywhere.
4. Type sth. The text appears in the new line (as if the selection was there).

jun-19-2018 14-52-09

oleq:

Fix: Firefox should visually move the caret to the new line after a soft break. Closes #1439.

---

### Additional information

* [CI](https://github.com/ckeditor/ckeditor5/compare/t/ckeditor5-engine/1439?expand=1)
* Requires https://github.com/ckeditor/ckeditor5-utils/pull/254

In a F2F talk with @scofalik we decided to keep browser quirks in the engine rather in features (like `ShiftEnterCommand`).

The entire fix is just `domSelection.addRange( domSelection.getRangeAt( 0 ) );` BTW ;-)

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -3,6 +3,8 @@
  * For licensing, see LICENSE.md.
  */
 
+/* globals Node */
+
 /**
  * @module engine/view/renderer
  */
@@ -20,6 +22,7 @@ import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 import isText from '@ckeditor/ckeditor5-utils/src/dom/istext';
 import isNode from '@ckeditor/ckeditor5-utils/src/dom/isnode';
 import fastDiff from '@ckeditor/ckeditor5-utils/src/fastdiff';
+import env from '@ckeditor/ckeditor5-utils/src/env';
 
 /**
  * Renderer is responsible for updating the DOM structure and the DOM selection based on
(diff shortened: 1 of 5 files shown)


tests/view/renderer.js
```diff
@@ -1714,6 +1715,74 @@ describe( 'Renderer', () => {
 			expect( domRoot.innerHTML ).to.equal( '<ul><li>Foo<ul><li><b>Bar</b><i>Baz</i></li></ul></li></ul>' );
 		} );
 
+		// #1439
+		it( 'should force–refresh the selection in FF after a soft break to nudge the caret (non-gecko)', () => {
```
scofalik: Shouldn't the test name be something like "it should not force-refresh the selection after a soft break in non-gecko browsers" and the other one negated?

scofalik: Also, I am not that sure if we need such a test and if we can test it properly. I mean, the test works but checking if `addRange` was called is a bit dangerous. If we ever use it in the normal selection setting process the test will start failing and we will have to remove it anyway. I would be fine with manual test only, I guess.

src/view/renderer.js
```diff
@@ -752,6 +755,31 @@ export default class Renderer {
...
+			if ( parent.nodeType != Node.ELEMENT_NODE || focus.offset != parent.childNodes.length - 1 ) {
+				return;
+			}
+
+			const childAtOffset = parent.childNodes[ focus.offset ];
+
+			// To stay on the safe side, the fix being as specific as possible, it targets only the
+			// selection which is at the very end of the element and preceded by <br />.
+			if ( childAtOffset && childAtOffset.tagName == 'BR' ) {
+				domSelection.addRange( domSelection.getRangeAt( 0 ) );
```
f1ames: I had a concern about `domSelection.addRange` if adding another range will not result in selection having two ranges (as they are the same it doesn't makes sense probably) but it seems there is one range after this call 👍

Reinmar:
@oleq could you resolve CI and coverage issues?

src/view/renderer.js
```diff
@@ -752,6 +755,31 @@ export default class Renderer {
...
+
+		// The following is a Firefox–specific hack (https://github.com/ckeditor/ckeditor5-engine/issues/1439).
+		// When the native DOM selection is at the end of the block and preceded by <br /> e.g.
+		//
+		//		<p>foo<br/>[]</p>
+		//
+		// which happens a lot when using the soft line break, the browser fails to (visually) move the
+		// caret to the new line. A quick fix is as simple as force–refreshing the selection with the same range.
+		if ( env.isGecko ) {
+			const parent = focus.parent;
```
Reinmar: I'll move the entire body of this condition to another function because it looks bad that there's a `return` inside it. What if someone would like to add another such hack after this one? We'll need to refactor it anyway.

PR #1585, opened 2018-10-26
T/1555: WiP

jodator:

Type: Message. Closes #000.

BREAKING CHANGE: The model `Position.createAt()` was renamed to `._createAt()` and is now protected. Use `writer.createPositionAt()` instead.
BREAKING CHANGE: The model `Position.createAfter()` method was renamed to `_.createAfter()` and is now protected. Use `writer.createPositionAfter()` instead.
BREAKING CHANGE: The model `Position.createBefore()` method was renamed to `_.createBefore()` and is now protected. Use `writer.createPositionBefore()` instead.
BREAKING CHANGE: The model `Position.createFromPosition()` method was removed. Use `writer.createPositionAt( position )` to create a new position instance.
BREAKING CHANGE: The model `Position.createFromParentAndOffset()` method was removed. Use `writer.createPositionAt( parent, offset )` instead.

BREAKING CHANGE: The model `Range.createIn()` method was renamed to `._createIn()` and is now protected. Use `writer.createRangeIn()` instead.
BREAKING CHANGE: The model `Range.createOn()` method was renamed to `._createOn()` and is now protected. Use `writer.createRangeOn()` instead.
BREAKING CHANGE: The model `Range.createFromRange()` method was renamed to `_createFromRange()` and is now protected.
BREAKING CHANGE: The model `Range.createFromParentsAndOffsets()` method was renamed to `_createFromParentsAndOffsets()` and is now protected.
BREAKING CHANGE: The model `Range.createFromPositionAndShift()` - method was renamed to `._createFromPositionAndShift()` and is now protected.
BREAKING CHANGE: The model `Range.createCollapsedAt()` - removed method was removed. Use `writer.createRange( position )` to create collapsed range.

BREAKING CHANGE: The model `Range.createFromRanges()` - move to helper method?! Used in 2-3 places & tests method was renamed to `createFromRanges(` and is now protected. Use `writer.createFromRanges(` instead.

BREAKING CHANGE: The view `Position.createAt()` was renamed to `._createAt()` and is now protected. Use `writer.createPositionAt()` instead.
BREAKING CHANGE: The view `Position.createAfter()` method was renamed to `_.createAfter()` and is now protected. Use `writer.createPositionAfter()` instead.
BREAKING CHANGE: The view `Position.createBefore()` method was renamed to `_.createBefore()` and is now protected. Use `writer.createPositionBefore()` instead.
BREAKING CHANGE: The view `Position.createFromPosition()` method was removed. Use `writer.createPositionAt( position )` to create a new position instance.

BREAKING CHANGE: The view `Range.createIn()` method was renamed to `._createIn()` and is now protected. Use `writer.createRangeIn()` instead.
BREAKING CHANGE: The view `Range.createOn()` method was renamed to `._createOn()` and is now protected. Use `writer.createRangeOn()` instead.
BREAKING CHANGE: The view `Range.createFromRange()` method was renamed to `_createFromRange()` and is now protected.
BREAKING CHANGE: The view `Range.createFromPositionAndShift()` - method was renamed to `._createFromPositionAndShift()` and is now protected.
BREAKING CHANGE: The view `Range.createFromParentsAndOffsets()` method was renamed to `_createFromParentsAndOffsets()` and is now protected.
BREAKING CHANGE: The view `Range.createCollapsedAt()` - removed method was removed. Use `writer.createRange( position )` to create collapsed range.

---

### Additional information

### DIFF
--- a/src/model/model.js
+++ b/src/model/model.js
@@ -309,16 +309,16 @@ export default class Model {
 	 *
 	 *		// Insert text at given position - document selection will not be modified.
 	 *		editor.model.change( writer => {
-	 *			editor.model.insertContent( writer.createText( 'x' ), Position.createAt( doc.getRoot(), 2 ) );
+	 *			editor.model.insertContent( writer.createText( 'x' ), writer.createPositionAt( doc.getRoot(), 2 ) );
 	 *		} );
 	 *
 	 * If an instance of {@link module:engine/model/selection~Selection} is passed as `selectable`
 	 * it will be moved to the target position (where the document selection should be moved after the insertion).
 	 *
-	 *		// Insert text replacing given selection instance.
-	 *		const selection = new Selection( paragraph, 'in' );
-	 *
 	 *		editor.model.change( writer => {
+	 *			// Insert text replacing given selection instance.
+	 *			const selection = writer.createSelection( paragraph, 'in' );
+	 *
 	 *			editor.model.insertContent( writer.createText( 'x' ), selection );
 	 *
 	 *			// insertContent() modifies the passed selection instance so it can be used to set the document selection.
@@ -471,6 +471,22 @@ export default class Model {
 	}
 
 	/**
+	 * Creates a position.
+	 *
+	 * **Note:** This method is also available on `writer` instance as
+	 * {@link module:engine/model/writer~Writer#createPositionFromPath writer.createPositionFromPath()}:
+	 *
+	 * @param {module:engine/model/element~Element|module:engine/model/documentfragment~DocumentFragment} root Root of the position.
+	 * @param {Array.<Number>} path Position path. See {@link module:engine/model/position~Position#path}.
+	 * @param {module:engine/model/position~PositionStickiness} [stickiness='toNone'] Position stickiness.
+	 * See {@link module:engine/model/position~PositionStickiness}.
+	 * @returns {module:engine/model/position~Position}
+	 */
+	createPositionFromPath( root, path, stickiness ) {
+		return new ModelPosition( root, path, stickiness );
+	}
+
+	/**
 	 * Creates position at the given location. The location can be specified as:
 	 *
 	 * * a {@link module:engine/model/position~Position position},
@@ -478,10 +494,13 @@ export default class Model {
 	 * * parent element and `'end'` (sets position at the end of that element),
 	 * * {@link module:engine/model/item~Item model item} and `'before'` or `'after'` (sets position before or after given model item).
 	 *
-	 * This method is a shortcut to other constructors such as:
+	 * This method is a shortcut to other factory methods such as:
 	 *
-	 * * {@link module:engine/model/position~Position._createBefore},
-	 * * {@link module:engine/model/position~Position._createAfter},
+	 * * {@link module:engine/model/model~Model#createBefore},
+	 * * {@link module:engine/model/model~Model#createAfter}.
+	 *
+	 * **Note:** This method is also available on `writer` instance as
+	 * {@link module:engine/model/writer~Writer#createPositionAt writer.createPositionAt()}:
 	 *
 	 * @param {module:engine/model/item~Item|module:engine/model/position~Position} itemOrPosition
 	 * @param {Number|'end'|'before'|'after'} [offset] Offset or one of the flags. Used only when
@@ -494,6 +513,9 @@ export default class Model {
 	/**
 	 * Creates a new position, after given {@link module:engine/model/item~Item model item}.
 	 *
+	 * **Note:** This method is also available on `writer` instance as
+	 * {@link module:engine/model/writer~Writer#createPositionAfter writer.createPositionAfter()}:
+	 *
 	 * @param {module:engine/model/item~Item} item Item after which the position should be placed.
 	 * @returns {module:engine/model/position~Position}
 	 */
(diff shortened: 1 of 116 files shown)


jodator:
Remaining engine imports (source code - not tests) from other packages (exluding core & editors):

```js
import bindTwoStepCaretToAttribute from '@ckeditor/ckeditor5-engine/src/utils/bindtwostepcarettoattribute';

import HtmlDataProcessor from '@ckeditor/ckeditor5-engine/src/dataprocessor/htmldataprocessor';

import Observer from '@ckeditor/ckeditor5-engine/src/view/observer/observer';
import ClickObserver from '@ckeditor/ckeditor5-engine/src/view/observer/clickobserver';
import DomEventObserver from '@ckeditor/ckeditor5-engine/src/view/observer/domeventobserver';
import MouseObserver from '@ckeditor/ckeditor5-engine/src/view/observer/mouseobserver';

import DomEventData from '@ckeditor/ckeditor5-engine/src/view/observer/domeventdata';

import DomConverter from '@ckeditor/ckeditor5-engine/src/view/domconverter';

import { attachPlaceholder } from '@ckeditor/ckeditor5-engine/src/view/placeholder';

import { transformSets } from '@ckeditor/ckeditor5-engine/src/model/operation/transform';

import { downcastAttributeToElement } from '@ckeditor/ckeditor5-engine/src/conversion/downcast-converters';
import { downcastElementToElement } from '@ckeditor/ckeditor5-engine/src/conversion/downcast-converters';

import { upcastElementToAttribute } from '@ckeditor/ckeditor5-engine/src/conversion/upcast-converters';
import { upcastElementToElement } from '@ckeditor/ckeditor5-engine/src/conversion/upcast-converters';

import LivePosition from '@ckeditor/ckeditor5-engine/src/model/liveposition';
import LiveRange from '@ckeditor/ckeditor5-engine/src/model/liverange';
```

src/model/model.js
```diff
@@ -466,6 +469,208 @@ export default class Model {
...
+	 *		// Creates selection at the given ranges
+	 *		const ranges = [ writer.createRange( start1, end2 ), writer.createRange( star2, end2 ) ];
+	 *		const selection = writer.createSelection( ranges );
+	 *
+	 *		// Creates selection from the other selection.
+	 *		// Note: It doesn't copies selection attributes.
+	 *		const otherSelection = writer.createSelection();
+	 *		const selection = writer.createSelection( otherSelection );
+	 *
+	 *		// Creates selection from the given document selection.
```
jodator: This line bothers me - I'd wrote this as:
```js
const documentSelection = doc.selection;
const selection = writer.createSelection( documentSelection );
```

pjasiun:
I reviewed how plugins use this API (I will check the engine on Monday) and... wow, that's huge! Good job @jodator!

What I like is that usually `writer` or `model` is available and it needs to be passed to the function only in a few cases.

What I do not like is the way how the selection is created from a position, for instance:
```
writer.setSelection( writer.createRange(
	writer.createPositionAt( root.getNodeByPath( [ 0 ] ), 2 )
) );
```
https://github.com/ckeditor/ckeditor5-basic-styles/pull/79/files#diff-b26fa587bf35fdff179cbebd6e3dff8bR228

I think we should have something like `writer.setSelectionAt`:
```
writer.setSelectionAt( root.getNodeByPath( [ 0 ] ), 2 );
```

The construction `writer.setSelection( writer.createRange( writer.createPositionAt( ... ) ) );` appears often:
 - https://github.com/ckeditor/ckeditor5-basic-styles/blob/94640761257c27bb7613ce62383a94bc7ccef94c/tests/attributecommand.js#L228-L230
 - https://github.com/ckeditor/ckeditor5-font/blob/08485052857402098e5aafb9526cb8a43077ffcf/tests/fontcommand.js#L173-L176
 - https://github.com/ckeditor/ckeditor5-highlight/blob/d639db9125c7744f95684b9229c22ef33a44f990/tests/highlightcommand.js#L155
 - ...
so I believe such shortcut would be useful.

jodator:
> writer.setSelectionAt( root.getNodeByPath( [ 0 ] ), 2 );

Nah, just use `writer.setSelection( node, offset )` as this is already supported. I can review usage of such construct though.

jodator:
All engine tests passes locally - Right now engine requires parapraph which requires engine which requires paragraph .... hence the errors on Travis :disappointed:

src/view/documentselection.js
```diff
@@ -35,19 +35,19 @@ export default class DocumentSelection {
 	 *		const selection = new DocumentSelection();
 	 *
 	 *		// Creates selection at the given range.
-	 *		const range = new Range( start, end );
+	 *		const range = writer.createSelection( start, end );
```
pjasiun:
```
const range = writer.createSelection( start, end ); 
```
There is something wrong here.

pjasiun:
> Things to discuss:
> 1. I've failed with the `.clone()` and `LiveRange.fromPosition()` changes as in some cases methods could have both `Position` or `LivePosition` and I wasn't able to catch them all. But:
>  - I've removed `Position.createFromPosition()` - it is `LivePosition.fromPosition()` and `livePos.toPosition()` now - I've replaced it were it was obious. Also you can copy position by `Position.createAt( position )`.

I notice it reviewing engine code. And I need to say that it is a little bit messy now. We have `LivePosition.fromPosition` vs `Range._createFromRange` vs no cloning method for the position now. What we agreed, `clone` method for `Position` and `Range` and`(from|to)(Position|Range)
` methods in `LivePosition` and `LiveRange` would be great.

> 2. The `writer.createPositionFromPath()` is tooooo long IMHO. I'd call that `writer.createPosition()` - similarly to `writer.createRange()`.

I think it is good that this method is long. It should be used only in special cases, so the long name properly suggestion that this should not be your first choice.

> 3. Range copy - right now we do not have `writer` methods for:
>  - `Range._createFromRange()` - it could be exposed as `writer.createRange( range )` alternative.

As mentioned before, I prefer `range.clone()`.

>  - `Range._createFromRanges()` - this might be as well a utility function as it is only used in about 5 places in the engine.

That's a good idea, but we can keep it as internal engine method at the moment.

> 4. I had [minor conceptual issues](https://github.com/ckeditor/ckeditor5-engine/pull/1585/files#r228508104) with rewriting the `new DocumentSelection()` examples - mostly in the `createSeleciton()` docs. Does public API should/or will expose creator for this? Or we just remove lines about that?

We should not expose `DocumentSelection` constructor. There should be only a single document selection (`doc.selection`), so your suggestion to change docs is good.

> 5. I've created factory methods on `model` and then in the `writer` I use model instance. Obviously, I've linked it properly (hope so) in docs. Now, for the `view` the `DowncastWriter` does not have a reference to the view so I could not reuse it. Probably a minor change to pass `view` to the writer but I wanted to ask you guys about such change. Right now the code is duplicated for `view` and `downcastWriter`.

I think we can keep it the way it is now. Users are able to create view writers whenever when need it and I think that better docs are not a good enough reason to add `view` instance to the constructor.

Also, these methods should be added to `upcastWriter`.

jodator:
> I notice it reviewing engine code. And I need to say that it is a little bit messy now. We have `LivePosition.fromPosition` vs `Range._createFromRange` vs no cloning method for the position now. What we agreed, `clone` method for `Position` and `Range` and`(from|to)(Position|Range) ` methods in `LivePosition` and `LiveRange` would be great.

@pjasiun again - many methods in `Position` assumes instance of `Position` not `LivePosition`. The must must `Postion._createAt( positionOrLivePosition )` to not change `LivePosition` path - it will trigger events and some loops will occur (probably). So actual `position.clone()` to be used must be in context where you are 100% sure that you have position. So now I'm not 100% sure that we need `position.clone()` method.

jodator:
> @pjasiun again - many methods in `Position` assumes instance of `Position` not `LivePosition`. The must must `Postion._createAt( positionOrLivePosition )` to not change `LivePosition` path - it will trigger events and some loops will occur (probably). So actual `position.clone()` to be used must be in context where you are 100% sure that you have position. So now I'm not 100% sure that we need `position.clone()` method.

I think that I've done my best to overcome this - I've just put `new Range()` when the `Range` class is needed - special cases.

jodator:
@pjasiun  & @Reinmar I think I've included all the suggestions from the comments. Feel free to merge ;)

PR #1595, opened 2018-11-23
Do not run `AttributeToAttribute` downcast conversion on `TextProxy` nodes

### LINKED ISSUE #1587: Error during conversion of text inside element when specific attribute converters are defined for both elements
Follow up of https://github.com/ckeditor/ckeditor5/issues/1317.

Considering conversions defined like:

```js
editor.conversion.for( 'downcast' ).add( downcastAttributeToElement( {
    model: {
        key: 'test',
        name: '$text'
    },
    view: ( value, writer ) => {
        return writer.createAttributeElement( 'span', { test: value } );
    }
} ) );

editor.conversion.for( 'downcast' ).add( downcastAttributeToAttribute( {
    model: 'test',
    view: 'test'
} ) );
```

and setting model data like:

```js
const root = editor.model.document.getRoot();
const p = writer.createElement( 'paragraph' );
const text = writer.createText( 'FooBar', { 'test': 3 } );

writer.insert( text, p );
writer.insert( p, root );
```

throws an `Cannot read property '_setAttribute' of undefined` error:

image

Works fine if text is inserted into `root`, but when inserted as a child of `p` the error is thrown. The interesting thing is when generic conversion is removed from the example above:

```js
// editor.conversion.for( 'downcast' ).add( downcastAttributeToAttribute( {
//    model: 'test',
//    view: 'test'
// } ) );
```

everything starts to work fine.

f1ames:

Fix: Do not run `AttributeToAttribute` downcast conversion on `TextProxy` nodes. Closes #1587.

---

### Additional information

I wanted to elaborate a little on this fix, because it looks like "_Here we got a `null` exception, lets add an `if`, easy-peasy, done_". But it is a little more complex:

The cause of this issue is described in https://github.com/ckeditor/ckeditor5-engine/issues/1587#issuecomment-441062367 - in short attribute to attribute downcast conversion is run for model `TextProxy` which is downcasted to a view `Text` node which cannot have attributes directly (but converter assumes it can).

The reasonable approach will be to downcast (via attribute to attribute converters) only model elements which can have attributes. However, in the model, `TextProxy` is the element which allows attributes. So checking model element for attributes acceptance doesn't make sense. Also model doesn't have a direct knowledge to what view element the model element will be downcasted to (that is the responsibility of converters). So basically to check if element can be processed by attribute to attribute conversion, view element should be checked. And this is what exactly happens here. Since [`conversionApi.mapper.toViewElement( data.item )`](https://github.com/ckeditor/ckeditor5-engine/pull/1595/files#diff-81862f954090bd9f5e3d83507ed19287R685) is the first place where to be converted model element is mapped to view, it is also a proper place to check if element directly accepts attributes. In the view all `Element` nodes can have attributes and other types cannot (so basically `Text` and `TextProxy`). So any node different than `Element` should not be converted and since `conversionApi.mapper.toViewElement()` method works only on `Element` nodes, it returns `null` for other elements.

### DIFF
--- a/src/conversion/downcast-converters.js
+++ b/src/conversion/downcast-converters.js
@@ -10,6 +10,7 @@ import ModelElement from '../model/element';
 import ViewAttributeElement from '../view/attributeelement';
 import DocumentSelection from '../model/documentselection';
 
+import log from '@ckeditor/ckeditor5-utils/src/log';
 import { cloneDeep } from 'lodash-es';
 
 /**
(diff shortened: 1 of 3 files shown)


scofalik:
I discussed this issue with @f1ames.

The problem with this is that it works correctly, however it is easy to mess up. `attributeToAttribute`  (in short `aToA`) conversion is suited **only** for model elements. It does not and should not work for text nodes, so the crash is not surprising.

Basically, the situation when the crash occurs is when `aToA` is defined but the same attribute is used on a text. We already had two questions from the community about configuring converters to have the same attribute on elements and texts. The trick is to provide `attributeToElement` converter for texts ( `model: { key: 'key', name: '$text' }`) and have it fire before `aToA` converter (using `converterPriority` or proper code ordering). The problem is that this is not straightforward and is not explained anywhere.

For that reason, I am actually for having the fix as proposed in this PR. I'd just add a `log.warn()` informing that attribute to attribute downcast conversion was skipped because model attribute was set on model text node (or something similar that makes sense, I leave it up to you to format it in a short and informative way :P).

Also, please make it clear in API docs that `aToA` is for element-to-element conversion and is not suitable for text nodes and that `aToE` should be used to convert text attributes.

f1ames:
@scofalik Ready for another review.

src/conversion/downcast-converters.js
```diff
@@ -685,9 +686,44 @@ export function changeAttribute( attributeCreator ) {
...
+			 *				name: '$text'
+			 *			},
+			 *			view: ( value, writer ) => {
+			 *				return writer.createAttributeElement( 'span', { 'attribute-name': value } );
+			 *			}
+			 *		} ), { priority: 'high' } );
+			 *
+			 * @error conversion-attribute-to-attribute-on-text
+			 */
+			log.warn( 'conversion-attribute-to-attribute-on-text: Trying to convert text node with attribute to attribute converter.' );
```
scofalik: text node's attribute

attribute-to-attribute

src/conversion/downcast-converters.js
```diff
@@ -685,9 +686,44 @@ export function changeAttribute( attributeCreator ) {
 		const viewElement = conversionApi.mapper.toViewElement( data.item );
 		const viewWriter = conversionApi.writer;
 
-		// If model item cannot be mapped to view element, it means item is not an `Element` instance (it is a `TextProxy`).
-		// Only elements can have attributes in a view so do not proceed for anything else (see #1587).
+		// If model item cannot be mapped to a view element, it means item is not an `Element` instance but a `TextProxy` node.
+		// Only elements can have attributes in a view so do not proceed for anything else (#1587).
 		if ( !viewElement ) {
+			/**
+			 * This error occurs when a {@link module:engine/model/textproxy~TextProxy text node} is to be downcasted
```
scofalik: text node's attribute

PR #1603, opened 2018-12-05
T/edge18 selection test

### LINKED ISSUE #1607: CI is red because of Edge 18
Hi - we have problem with CI. Some test (in SelectionObserver) fails on Edge 18. I've created dumb fixes: https://github.com/ckeditor/ckeditor5-engine/pull/1603 but I have feeling that is not best method.

It looks like Edge 18 does note fire "selectionchange" in certain scenarios:

Setting (programatically) collapsed selection at the same node:
https://github.com/ckeditor/ckeditor5-engine/blob/5edb9e6831cd102d980dada67ae82807c7675491/tests/view/observer/selectionobserver.js#L165-L191

On Edge it will be only **one** "selectionchange" event for all 70 selection changes.

If it is changed to something more "fancy" (extending that selection):
https://github.com/ckeditor/ckeditor5-engine/pull/1603/commits/8ec291017fa8fd3f182d1f083c1ac1cf82d8b40c

It does fire "selectionchange" enents as expected.

Right now it is hard to debug it - ie through BrowserStack. My propositions is to envorce Edge 17 on Travis until MS releases the Edge 18 VMs.

jodator:

Tests: Fix failing tests on Edge 18. Closes #1607.

---

### Additional information

* Those "fixes" are straight forward and I'm not sure if it isn't Edge bug or largely changed behavior of 18. Also it is cumbersome to debug as [no official VMs](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/)  are available.


Reinmar:
There's very small chance that we'll be doing these changes due to Edge being replaced with Blink, so I'm closing this PR.

PR #1605, opened 2018-12-05
T/1551 Selectable type definition

### LINKED ISSUE #1551: Define Selectable interface (model/view).
In many places the:

```js
 @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection|
	 * module:engine/model/position~Position|module:engine/model/node~Node|
	 * Iterable.<module:engine/model/range~Range>|module:engine/model/range~Range|null} selectable
```
is repeated as a `selectable` so we can  introduce `model/selectable` and `view/selectable` so above would be shortened to:

```js
 @param {module:engine/model/selectable~Selectable|null} selectable
```

```js
/**
 * @module engine/model/selectable
 */

/**
 * @typedef {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection|
 * module:engine/model/position~Position|module:engine/model/element~Element|
 * Iterable.<module:engine/model/range~Range>|module:engine/model/range~Range} module:engine/model/selectable~Selectable
 */
```
Interface for this purpose - with classes:
- `module:engine/model/selection~Selection`
- `module:engine/modeldocumentselection~DocumentSelection`
-  `module:engine/model/position~Position`
- `module:engine/model/element~Element`
- `Iterable.<module:engine/model/range~Range>`
- `module:engine/model/range~Range`

And Similar for the view:

- `module:engine/view/selection~Selection`
- `module:engine/view/position~Position`
- `Iterable.<module:engine/view/range~Range>`
- `module:engine/view/range~Range`
- `module:engine/view/item~Item`

ps.: Some methods uses `Node` while others uses `Element` - I think that we can safely use `Node` as `Selectable`:
```
module:engine/model/position~Position|module:engine/model/node~Node
```
vs
```
module:engine/model/position~Position|module:engine/model/element~Element
```

jodator:

Docs: Define Selectable type. Closes #1551.

---

### Additional information

* As for [Node vs Element](https://github.com/ckeditor/ckeditor5-engine/issues/1551#issuecomment-421968316) in model's `selection.setTo()` - this `setTo()` method uses `Node` internally so it is safe to use `Node` not `Element`.


src/model/model.js
```diff
@@ -331,9 +331,7 @@ export default class Model {
 	 *
 	 * @fires insertContent
 	 * @param {module:engine/model/documentfragment~DocumentFragment|module:engine/model/item~Item} content The content to insert.
-	 * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection|
-	 * module:engine/model/position~Position|module:engine/model/item~Item|
```
f1ames: This is the only place which mentions `module:engine/model/item~Item` instead of `module:engine/model/Node~Node` (which means `TextProxy` is also supported) which makes me confused about `Item` vs `Node` in `Selectable` type. Have you checked if it should be `Node` or `Item`? Because now new `Selectable` is incompatible with this line.

jodator: Yeah - I remember that one. It boils down to that this method didn't check or call `TextNode`'s specific methods. The chain is:

* `model.insertContent()` calls `insertContent()` function
* the `insertContent()` function uses `selectable` to create a `Selection` instance (if not a `Selection` already) using `writer.createSelection()`
* `writer.createSelection()` calls `model.createSelection()` :trollface: 
* `model.createSelection()` creates new instance of `Selection`
* `Selection`'s constructors checks the `selectable` is one of `Selectable` types: (`Selection`, `DocumentSelection`, `Range`, `Position`, `Node` (!), iterable (`Array.<Range>`) or `null`.
https://github.com/ckeditor/ckeditor5-engine/blob/1649b5f1624e7e0adb43cd686d6a82b968885435/src/model/selection.js#L368-L405

So I've just unified all the usages so no `Item` nor `Element` there. Now it's all cleaned up :wastebasket: :recycle: :)

jodator: ps.: good catch with those docs :+1:

PR #1608, opened 2018-12-07
T/ckeditor5 table/99: Selection post-fixer improvements.

### LINKED ISSUE #1593: Impossible to select an object element in some cases
The case is https://github.com/ckeditor/ckeditor5/issues/1331.

It should be possible to make such a selection:

```html
<div>[<div></div>]</div>
```

because `div` is marked as `isObject` (all objects are selectable).

However, that does not work. The reason why it works in the demo I created in that ticket is because I accidentally added:

```js
	editor.model.schema.extend( '$text', {
		allowIn: 'div'
	} );
```

Which made this scenario fall into "selection is allowed in text" case.

jodator:

Add selection post-fixer improvements. Closes #1593.

---

### Additional information

* Required for table PR: https://github.com/ckeditor/ckeditor5-table/pull/155.
* Required for widget PR: https://github.com/ckeditor/ckeditor5-widget/pull/62.

### DIFF
--- a/src/model/utils/selection-post-fixer.js
+++ b/src/model/utils/selection-post-fixer.js
@@ -192,7 +192,7 @@ function tryFixingNonCollapsedRage( range, schema ) {
 			const isEndObject = end.nodeBefore && schema.isObject( end.nodeBefore );
  			const fixedEnd = isEndObject ? end : schema.getNearestSelectionRange( end, 'backward' ).end;
 
-			return new Range( fixedStart ? fixedStart.start : start, fixedEnd ? fixedEnd.start : end );
+ 			return new Range( fixedStart, fixedEnd );
 		}
 	}
 
--- a/src/model/utils/selection-post-fixer.js
+++ b/src/model/utils/selection-post-fixer.js
@@ -249,7 +249,7 @@ function findOutermostLimitAncestor( startingNode, schema ) {
 	return isLimitNode;
 }
 
-// Checks whether one of range ends is placed around non-limit elements.
+// Checks whether any of range boundaries is placed around non-limit elements.
 //
 // @param {module:engine/model/position~Position} start
 // @param {module:engine/model/position~Position} end
--- a/src/model/utils/selection-post-fixer.js
+++ b/src/model/utils/selection-post-fixer.js
@@ -187,12 +187,12 @@ function tryFixingNonCollapsedRage( range, schema ) {
 		// - [<block>foo</block><object></object>]    ->    <block>[foo</block><object></object>]
 		if ( checkSelectionOnNonLimitElements( start, end, schema ) ) {
 			const isStartObject = start.nodeAfter && schema.isObject( start.nodeAfter );
- 			const fixedStart = isStartObject ? start : schema.getNearestSelectionRange( start, 'forward' ).start;
+			const fixedStart = isStartObject ? start : schema.getNearestSelectionRange( start, 'forward' ).start;
 
 			const isEndObject = end.nodeBefore && schema.isObject( end.nodeBefore );
- 			const fixedEnd = isEndObject ? end : schema.getNearestSelectionRange( end, 'backward' ).end;
+			const fixedEnd = isEndObject ? end : schema.getNearestSelectionRange( end, 'backward' ).end;
 
- 			return new Range( fixedStart, fixedEnd );
+			return new Range( fixedStart, fixedEnd );
 		}
 	}
 
--- a/src/model/utils/selection-post-fixer.js
+++ b/src/model/utils/selection-post-fixer.js
@@ -187,7 +187,7 @@ function tryFixingNonCollapsedRage( range, schema ) {
 		// - [<block>foo</block><object></object>]    ->    <block>[foo</block><object></object>]
 		if ( checkSelectionOnNonLimitElements( start, end, schema ) ) {
 			const isStartObject = start.nodeAfter && schema.isObject( start.nodeAfter );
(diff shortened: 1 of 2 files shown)


src/model/utils/selection-post-fixer.js
```diff
@@ -159,6 +159,7 @@ function tryFixingCollapsedRange( range, schema ) {
 //
 // @param {module:engine/model/range~Range} range Expanded range to fix.
 // @param {module:engine/model/schema~Schema} schema
+
```
scofalik: 🤔

src/model/utils/selection-post-fixer.js
```diff
@@ -183,9 +184,13 @@ function tryFixingNonCollapsedRage( range, schema ) {
 		// - [<block>foo</block>]    ->    <block>[foo]</block>
 		// - [<block>foo]</block>    ->    <block>[foo]</block>
 		// - <block>f[oo</block>]    ->    <block>f[oo]</block>
+		// - [<block>foo</block><object></object>]    ->    <block>[foo</block><object></object>]
 		if ( checkSelectionOnNonLimitElements( start, end, schema ) ) {
-			const fixedStart = schema.getNearestSelectionRange( start, 'forward' );
-			const fixedEnd = schema.getNearestSelectionRange( end, 'backward' );
+			const isStartObject = start.nodeAfter && schema.isObject( start.nodeAfter );
+			const fixedStart = isStartObject ? null : schema.getNearestSelectionRange( start, 'forward' );
```
scofalik:
```suggestion
 			const fixedStart = isStartObject ? start : schema.getNearestSelectionRange( start, 'forward' ).start;
```

src/model/utils/selection-post-fixer.js
```diff
@@ -183,9 +184,13 @@ function tryFixingNonCollapsedRage( range, schema ) {
...
+
+			const isEndObject = end.nodeBefore && schema.isObject( end.nodeBefore );
+			const fixedEnd = isEndObject ? null : schema.getNearestSelectionRange( end, 'backward' );
```
scofalik:
```suggestion
 			const fixedEnd = isEndObject ? end : schema.getNearestSelectionRange( end, 'backward' ).end;
```

src/model/utils/selection-post-fixer.js
```diff
@@ -183,9 +184,13 @@ function tryFixingNonCollapsedRage( range, schema ) {
...
 
 			return new Range( fixedStart ? fixedStart.start : start, fixedEnd ? fixedEnd.start : end );
```
scofalik:
```suggestion
 			return new Range( fixedStart, fixedEnd );
```

src/model/utils/selection-post-fixer.js
```diff
@@ -197,13 +202,23 @@ function tryFixingNonCollapsedRage( range, schema ) {
 	// At this point we eliminated valid positions on text nodes so if one of range positions is placed inside a limit element
 	// then the range crossed limit element boundaries and needs to be fixed.
 	if ( isStartInLimit || isEndInLimit ) {
+		const bothInSameParent = ( !!start.nodeAfter && !!end.nodeBefore ) && start.nodeAfter === end.nodeBefore;
```
scofalik: I think we can omit `!!`.

scofalik: Also, I am not sure if the name is correct. It's more that the selection is on the same element, isn't it?

src/model/utils/selection-post-fixer.js
```diff
@@ -212,34 +227,45 @@ function tryFixingNonCollapsedRage( range, schema ) {
...
-// @param {module:engine/model/position~Position} position
+// @param {module:engine/model/node~Node} startingNode
 // @param {module:engine/model/schema~Schema} schema
 // @param {String} expandToDirection Direction of expansion - either 'start' or 'end' of the range.
-// @returns {module:engine/model/position~Position}
-function expandSelectionOnIsLimitNode( position, schema, expandToDirection ) {
-	let node = position.parent;
-	let parent = node;
+// @returns {module:engine/model/node~Node}
+function findOuterMostIsLimitAncestor( startingNode, schema ) {
```
scofalik: I'd call it `findOutermostLimitAncestor`. The `is` part makes the name confusing. I know that we are looking for an object that passes `isLimit()` in `Schema` but still. Also AFAIR outermost is a one word.

src/model/utils/selection-post-fixer.js
```diff
@@ -212,34 +227,45 @@ function tryFixingNonCollapsedRage( range, schema ) {
...
 		parent = parent.parent;
 	}
 
-	// Depending on direction of expanding selection return position before or after found node.
-	return expandToDirection === 'start' ? Position._createBefore( node ) : Position._createAfter( node );
+	return isLimitNode;
 }
 
-// Checks whether both range ends are placed around non-limit elements.
+// Checks whether one of range ends is placed around non-limit elements.
```
scofalik: ends => boundaries

scofalik: one => any

src/model/utils/selection-post-fixer.js
jodator: The `schema.getNearestSelectionRange()` might return null so this check.

I've updated the code to improve readability of this anyway.

src/model/utils/selection-post-fixer.js
jodator: The name is correct but the check is not. Updated to look for parent check.

Issue #843, opened 2017-02-24
Range.createFromRanges() could return an array of ranges and be named differently


Reinmar:
It now creates just one result, while it'd be very useful if it returned all groups od adjacent ranges. I had such use case in the blockquote where I had to find groups of adjacent blocks. Then, the function would be superbly helpful.

Also, the name is totally confusing. It should be something like `getCombinedRanges()`, `getGroupedRanges()` or similar.

scofalik:
It returns one range to be similar to other `Range` creators. It makes sense that it might return an array with combined ranges and we could change the name, so overall +1.

Reinmar:
Turns out this method is useful not only for us (https://github.com/ckeditor/ckeditor5-engine/pull/1610#pullrequestreview-198133441), so we could resolve these issues one day and publish it.

Issue #1555, opened 2018-09-20
Implement a builtin Position, Range and Selection factories


Reinmar:
Strongly related to https://github.com/ckeditor/ckeditor5/issues/667.

1. We need to check what do we import from `@ckeditor/ckeditor5-engine/src` in plugins (source, not tests). I used this regexp `import [\w, {},_]+ from '@ckeditor/ckeditor5-engine/src`. From my quick check, 90% of that are `Position`, `Range` and `Selection` from the model and the view. There are also some converters from `src/conversion/downcast-converters` and `src/conversion/upcast-converters`, but one thing at a time.
2. Design a set of factories like `createRange()`, `createRangeOn()`, etc.
3. Review this list based on the actual use. Decide which static methods of `Range` and `Position` may be removed since the factories will be created.
4. Decide where to expose those factories. On the `Document` classes or on `View`&`Model`?
5. Implement those methods.
6. Update all existing code. It may turn out that the dependency on `@ckeditor/ckeditor5-engine` can be removed in a couple of plugins.
7. Remove static methods if we'll decide in step 3. to remove them.

cc @jodator @pjasiun

Reinmar:
We've been planning this for a long time but I realised that it will significantly affect some of the documentation. This is currently one of the biggest blockers to write plugins without dependency on the engine.

pjasiun:
Wow, that's huge :/ However, I agree that this is important.

Reinmar:
I actually don't think that it's so big. It may even turn out that we'll simply move the static factory methods to some other place and implement a simple pass-through method for constructors. We've got most of the code and tests. I hope it's mostly about moving it around.

jodator:
> I hope it's mostly about moving it around.

I also think that is just that. Maybe huge in terms of lines changed but doable right now.

jodator:
@pjasiun with pleasure ;) Already on it. Some conclusion are already there, like `Position.createFromParentsAndOffset()`  might be private and not exposed as it is used only by table and might be replaced by other factory methods. Anyway I'll prepare full report on this.

jodator:
# Current usage of model imports

## Imports:

- `import Range from '@ckeditor/ckeditor5-engine/src/model/range';` (15)
- `import Position from '@ckeditor/ckeditor5-engine/src/model/position';` (12)
- `import Selection from '@ckeditor/ckeditor5-engine/src/model/selection';` (6)
[...]

PR #1610, opened 2018-12-07
Unbound fake selection when unbinding DOM elements

### LINKED ISSUE #1609: Fake Selection is Not Unbound when DOM Element is Unbound
I'm writing a plugin that allows encapsulating one widget within another. When the widget to be encapsulated is selected with a fake selection, the encapsulation would happen, but then the whole outer widget would disappear from the editor (though the elements were correctly arranged still in the model document).

What I believe is happening is that the renderer has recycled the `div` used to represent the fake selection to create the `div` of the outer widget, but `DomConverter._fakeSelectionMapping` never removed the `div` from its map. So when the engine goes to select the inner widget again after the inner has been moved to be a child of the inserted outer widget, it tries to use the recycled `div` -- now the outer widget -- to represent the fake selection again (assuming its first child is a text node, etc.), leaving things in an odd state. I _think_ the fix would be as simple as having the call to `DomConverter.unbindDomElement()` also remove the DOM element from `_fakeSelectionMapping` if it exists in the map, but I'm still learning my way around the code base, so I recognize there may be more to it than that. Thanks!

kobsy:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Fix: Unbound fake selection mapping when unbinding DOM elements. Closes #1609.

### DIFF
--- a/tests/manual/tickets/1609/1.js
+++ b/tests/manual/tickets/1609/1.js
@@ -1,0 +1,139 @@
+/* globals document, console, window */
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import Widget from '@ckeditor/ckeditor5-widget/src/widget';
+import Command from '@ckeditor/ckeditor5-core/src/command';
+import ModelRange from '../../../../src/model/range';
+import { toWidget, toWidgetEditable } from '@ckeditor/ckeditor5-widget/src/utils';
+import ButtonView from '@ckeditor/ckeditor5-ui/src/button/buttonview';
+import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor';
+import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
+import Essentials from '@ckeditor/ckeditor5-essentials/src/essentials';
+
+const OUTER_MODEL = 'speaker';
+const INNER_MODEL = 'special-char';
+
+class SpeakerCommand extends Command {
+	execute( { speaker } ) {
+		const model = this.editor.model;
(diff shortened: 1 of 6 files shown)


tests/manual/tickets/1609/1.js
```diff
@@ -0,0 +1,139 @@
...
+
+class SpeakerCommand extends Command {
+	execute( { speaker } ) {
+		const model = this.editor.model;
+		const document = model.document;
+
+		model.change( writer => {
+			const blocks = Array.from( document.selection.getSelectedBlocks() );
+			// NB: This is now private API, but a public equivalent doesn't currently exist
+			const range = ModelRange._createFromRanges( blocks.map( block => writer.createRangeOn( block ) ) );
```
Reinmar: @scofalik @pjasiun I remember you could not agree whether this method is needed in the public API.

pjasiun: I do not remember any reason to keep it protected.

Reinmar: We don't use it outside engine, so that's why we thought it can be made protected: https://github.com/ckeditor/ckeditor5-engine/issues/1555

Also, there are issues with its name and return value: https://github.com/ckeditor/ckeditor5-engine/issues/843

I guess we should first resolve https://github.com/ckeditor/ckeditor5-engine/issues/843 before bringing it back to the public API.

kobsy:
Closing since #1609 is no longer an issue on `master`. Thanks for the consideration, though!

PR #1611, opened 2018-12-12
Upcast element to attribute defaults to 'low' priority instead of 'normal'

f1ames:

Other: Upcast element to attribute defaults to `low` priority instead of `normal`. Closes ckeditor/ckeditor5#1399.

---

### Additional information

See https://github.com/ckeditor/ckeditor5/issues/1399 and https://github.com/ckeditor/ckeditor5/issues/1399#issuecomment-446567893.

I'm not sure, should it be marked as breaking change?

As for testing I have run unit tests from all packages and did some manual testing. However, @Mgsy could also take a quick look. This PR changes the order of conversions (so elements first and then its attributes) so any operation which involves both converting element and its attributes its affected.

Btw. The failing CI tests are just selection tests failing on Edge (#1607).

### DIFF
--- a/src/conversion/conversion.js
+++ b/src/conversion/conversion.js
@@ -410,7 +410,7 @@ export default class Conversion {
 				upcastElementToAttribute( {
 					view,
 					model,
-					priority: definition.priority
+					converterPriority: definition.priority
 				} )
 			);
 		}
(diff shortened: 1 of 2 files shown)


f1ames:
There is a related PR in `ckeditor5-list` which should be closed with this one - https://github.com/ckeditor/ckeditor5-list/pull/120.

PR #1613, opened 2018-12-18
T/1556: Expose conversion utilities

### LINKED ISSUE #1556: Expose all conversion utilities
A sibling ticket of https://github.com/ckeditor/ckeditor5-engine/issues/1555.

We exposed half of conversion utilities as [`editor.conversion`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_conversion_conversion-Conversion.html)'s methods. What we miss are the methods from:

* [`downcast-converters`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_conversion_downcast-converters.html)
* [`upcast-converters`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_conversion_upcast-converters.html)

Some of those functions are needed quite frequently which we experienced when answering to various questions. I think that in most questions about conversion we needed some of these lower-level functions rather than those big factories we already have exposed. 

I can see that there are quite a lot of `downcast-converters` but I'd expose all of them anyway, without a research which of them are used and how often (half of them are used in quite advanced cases with markers, but we'll want to make markers more popular anyway).

I'd expose this functions as e.g.:

```js
editor.conversion.downcastConverters.downcastElementToElement()
editor.conversion.upcastConverters.upcastElementToElement()
```

jodator:

Enhancement: Expose conversion utilities. Closes #1556.

BREAKING CHANGE: The `conversion.register()` method now accepts single options object as a parameter.
BREAKING CHANGE: The `downcastElementToElement()` helper was removed from public API. Use `conversion.for( 'downcast' ).elementToElement()` instead.
BREAKING CHANGE: The `downcastAttributeToElement()` helper was removed from public API. Use `conversion.for( 'downcast' ).attributeToElement()` instead.
BREAKING CHANGE: The `downcastAttributeToAttribute()` helper was removed from public API. Use `conversion.for( 'downcast' ).attributeToAttribute()` instead.
BREAKING CHANGE: The `downcastMarkerToElement()` helper was removed from public API. Use `conversion.for( 'downcast' ).markerToElement()` instead.
BREAKING CHANGE: The `downcastMarkerToHighlight()` helper was removed from public API. Use `conversion.for( 'downcast' ).markerToHighlight()` instead.
BREAKING CHANGE: The `upcastElementToElement()` helper was removed from public API. Use `conversion.for( 'upcast' ).elementToElement()` instead.
BREAKING CHANGE: The `upcastElementToAttribute()` helper was removed from public API. Use `conversion.for( 'upcast' ).elementToAttribute()` instead.
BREAKING CHANGE: The `upcastAttributeToAttribute()` helper was removed from public API. Use `conversion.for( 'upcast' ).attributeToAttribute()` instead.
BREAKING CHANGE: The `upcastElementToMarker()` helper was removed from public API. Use `conversion.for( 'upcast' ).elementToMarker()` instead.

---

### Additional information

* This PR comes with constelation of change gathered in [ckeditor5 branch](https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-engine/1556).
* I think that I've exposed all the required utilities:
	* For downcast (model-to-view conversion), these are:
		* `elementToElement`
		* `attributeToElement`
		* `attributeToAttribute`
		* `markerToElement`
		* `markerToHighlight`
	* For upcast (view-to-model conversion), these are:
		* `elementToElement`
		* `elementToAttribute`
		* `attributeToAttribute`
		* `elementToMarker`
* All above converter methods are marked as `@protected` and underscored.
* The API is still chainable (the helpers uses `this.add()` internally to add a protected converter):
	```js
	editor.conversion.for( 'upcast' )
		.add( customConverterA )
		.elementToElement( { model: 'foo', view: 'bar' } )
		.add( customConverterB )
		.elementToAttribute( { model: 'baz', view: 'bez' } );
	```

---
Some questions:
- There are other exported methods in those files - what to do with them? Chceck the code end remove exports if used internally? Methods like `convertText()`, `convertToModelFragment()` etc.
- The functions were left in place to not loose `git blame` immediate history in those functions. Also in this file export and private methods are mixed - should we sort them as `public export > protected export > internal/private`?

### DIFF
--- a/src/conversion/conversion.js
+++ b/src/conversion/conversion.js
@@ -87,7 +87,8 @@ export default class Conversion {
 	 * module:engine/conversion/upcastdispatcher~UpcastDispatcher|Array.<module:engine/conversion/downcastdispatcher~DowncastDispatcher|
 	 * module:engine/conversion/upcastdispatcher~UpcastDispatcher>} options.dispatcher Dispatcher or array of dispatchers to register
 	 * under the given name.
-	 * @param {module:engine/conversion/downcast-converters~DowncastHelpers} helpers
+	 * @param {module:engine/conversion/downcast-converters~DowncastHelpers|
+	 * module:engine/conversion/upcast-converters~UpcastHelpers} helpers
 	 */
 	register( options ) {
 		if ( this._dispatchersGroups.has( options.name ) ) {
@@ -151,7 +152,8 @@ export default class Conversion {
 	 *		conversion.for( 'downcast' ).add( conversion.customConverter( 'insert:paragraph', myConverter ) );
 	 *
 	 * @param {String} groupName The name of dispatchers group to add the converters to.
-	 * @returns {module:engine/conversion/conversion~ConversionHelpers|module:engine/conversion/downcast-converters~DowncastHelpers}
+	 * @returns {module:engine/conversion/conversion~ConversionHelpers|module:engine/conversion/downcast-converters~DowncastHelpers|
+	 * module:engine/conversion/upcast-converters~UpcastHelpers}
 	 * An object with the `.add()` method, providing a way to add converters.
 	 */
 	for( groupName ) {
@@ -599,7 +601,7 @@ export default class Conversion {
  * @property {String} name Group name
  * @property {Array.<module:engine/conversion/downcastdispatcher~DowncastDispatcher|
  * module:engine/conversion/upcastdispatcher~UpcastDispatcher>} dispatchers
- * @property {module:engine/conversion/downcast-converters~DowncastHelpers} helpers
+ * @property {module:engine/conversion/downcast-converters~DowncastHelpers|module:engine/conversion/upcast-converters~UpcastHelpers} helpers
  */
 
 /**
--- a/src/conversion/conversion.js
+++ b/src/conversion/conversion.js
@@ -97,13 +97,16 @@ export default class Conversion {
 		this._dispatchersGroups.set( options.name, group );
 	}
 
+	/* eslint-disable max-len */
 	/**
 	 * Provides chainable API to assign converters to dispatchers registered under a given group name. Converters are added
-	 * by calling the `.add()` method of an object returned by this function.
+	 * by calling the {@link module:engine/conversion/conversion~ConversionHelpers#add `.add()`} method of an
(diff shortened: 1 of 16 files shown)


jodator:
ps.: The [code works](https://travis-ci.org/ckeditor/ckeditor5/builds/469054714) with all repos checked out as in CKEditor5 (the core is crucial as it adds proper helpers to dispatcher groups).

src/conversion/conversion.js
```diff
@@ -402,17 +405,16 @@ export default class Conversion {
...
 
 		// Set up upcast converter.
 		for ( const { model, view } of _getAllUpcastDefinitions( definition ) ) {
-			this.for( 'upcast' ).add(
-				upcastElementToAttribute( {
+			this.for( 'upcast' )
+				.elementToAttribute( {
 					view,
 					model,
 					converterPriority: definition.priority
```
pjasiun: Since methods have the same names now, shouldn't they have the same parameters names too?

pjasiun: I mean `converterPriority` vs `priority`, in `conversion.for( 'downcast' ).attributeToElement` vs `conversion.attributeToElement`.

pjasiun:
I think we could rename `downcast-converters` to `downcasthelpers` with the `DowncastHelpers` class as the default export. Then all other `downcast-converters` and `downcast-selection-converters` which are needed by other classes make additional exports from this class (mostly default converters imported by controllers). Same for upcast.

jodator:
@pjasiun I like the `class` idea so to be 100% sure we're on the same page I'll describe how it will be refactored before doing it:

1. There will be base `ConversionHelpers` class with `.add()` method.
2. There will be to classed `UpcastHelpers` (curren `upcastHelpers`) and `DowncastHelpers` (current `downcastHelpers`)
3. Registering conversion pipeline:
```js
this.conversion.register(
    'downcast',
    new DowncastHelpers( [ 
        this.editing.downcastDispatcher,
        this.data.downcastDispatcher
    ]
);
this.conversion.register( 'upcast', new UpcastHelpers( this.data.upcastDispatcher ) );
```

pjasiun:
I agree with the first half, up to "Also" :) @jodator said that I should not care about the second half ;)

jodator:
@pjasiun done

PR #1624, opened 2019-01-02
T/1619: Expose protectely insertElement and createUIElement conversion helpers

jodator:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Internal: Expose `insertElement()` and `createUIElement()` conversion helpers as protected functions. Closes #1619.

---

### Additional information

* This change is needed due to "catch-all" converters used in `getData()` from `dev-utils/model.js` used in tests
* as those method are internal I've left integration tests for them as now - I plan to do the same for `wrap()` method removal from public API. It will need to stay as export but will be tested only by integration tests (#1616). The `wrap()` method is used by attribute change "catch-all" converter in `dev-utils#getData()`

PR #1625, opened 2019-01-02
T/1616: Hide wrap from public API

### LINKED ISSUE #1616: `wrap` should not be exporte from dowcast helper
Follow-up of: https://github.com/ckeditor/ckeditor5-engine/pull/1613#issuecomment-449055628.

The wrap function is now the only function which is exported and should not be. It should also have no direct test, but only integrational tests.

See also: https://github.com/ckeditor/ckeditor5-engine/issues/1306#issuecomment-378870980

jodator:

Other: Remove wrap from public API. Closes #1616.

BREAKING CHANGE: The `wrap()` conversion helper was removed from public API.

---

### Additional information

* After #1624.


jodator:
Tests fails because of #1624.

PR #1627, opened 2019-01-02
Support for handling data in multiple roots in `DataController` added

### LINKED ISSUE #1626: Support for initializing data on multiple roots at once in Data Controller
Extracted from https://github.com/ckeditor/ckeditor5-editor-decoupled/issues/21.

When working with multiple root editor, `DataController#init()` method cannot be used to set data on multiple roots, because initializing data on the first root changes document version and any consequent call throws an `datacontroller-init-document-not-empty` error.

The first idea may be to use `DataController#set()` instead of `init()`, however it does not trigger `init` event which may affect code parts which relies on this event. 

The `DataController#init()` should support initializing data on multiple roots at once.

f1ames:

Other: Added support for handling data in multiple roots in `DataController`. Closes #1626.

BREAKING CHANGE: The second parameter (`rootName`) from `DataController#init()` method has been removed. To initialize data on a root different than default one an object with `rootName` - `data` pair should be passed.

BREAKING CHANGE: The second parameter (`rootName`) from `DataController#set()` method has been removed. To set data on a root different than default one an object with `rootName` - `data` pair should be passed.

---

### Additional information

See #1626.

This PR causes two unit tests from `paragraph` and `table` plugin to fail. After merging ths one, corresponding PRs should be merged to - https://github.com/ckeditor/ckeditor5-paragraph/pull/38 and https://github.com/ckeditor/ckeditor5-table/pull/158.


f1ames:
This PR is needed for https://github.com/ckeditor/ckeditor5-editor-decoupled/issues/21.

src/controller/datacontroller.js
```diff
@@ -182,8 +182,9 @@ export default class DataController {
...
 	 *
 	 * @fires init
-	 * @param {String} data Input data.
-	 * @param {String} [rootName='main'] Root name.
+	 * @param {String|Object.<String,String>} data Input data as a string or an object containing rootName - initialData pairs to
+	 * initialize data on multiple roots at once.
+	 * @param {String} [rootName='main'] Root name. Takes effect only if the first argument is provided as a string.
 	 * @returns {Promise} Promise that is resolved after the data is set on the editor.
 	 */
 	init( data, rootName = 'main' ) {
```
pjasiun: I think `rootName = 'main'` does not make much sense now for this method. You can use the object if you want to init a different root.

pjasiun: Yep, I was thinking about using object always in as the `data` parameter, but I think it would be too big breaking change. However, removing `rootName` should be fine because, as you said, most probably, no one uses it.

pjasiun: Yep. See https://github.com/ckeditor/ckeditor5-editor-decoupled/issues/21#issuecomment-451898238.

pjasiun: Actually https://github.com/ckeditor/ckeditor5-editor-decoupled/issues/21#issuecomment-451904627:)

f1ames:
@pjasiun I think I suffer from NCCD (non-consistent code disorder:D). I have moved the `setData` logic from decoupled editor to `DataController` as mentioned in https://github.com/ckeditor/ckeditor5-editor-decoupled/pull/22#pullrequestreview-189735548.

It looks good and `DataController#init()` and `DataController#set()` methods are now consistent (almost). However, in `DataController#set()` there is now an `if` throwing an error when setting data on non-existent root. Shouldn't we have the same at least for `DataController#init()` method? From what I see `DataController#get()` also will break when non-existent root will be passes so maybe we should have this check in all 3 methods for manipulating data? P.S. I know the `init` is usually used in a way that it is harder to generate such error (because editor is created and then its data initiated from the same object with the same root names usually).

pjasiun:
Agree, all of them should throw when the root is not defined.

NCCD is useful, sometimes ;)

f1ames:
Since some code was moved here from , these comments - https://github.com/ckeditor/ckeditor5-editor-decoupled/pull/22#pullrequestreview-189969644 and https://github.com/ckeditor/ckeditor5-editor-decoupled/pull/22#pullrequestreview-189969763 now relates to this PR (it's about rewording error messages):

> `editor-getData-non-existent-root
>
> It's an id, not a sentence – it should look like a namespace – from general to specific information.`

src/controller/datacontroller.js
```diff
@@ -182,11 +197,11 @@ export default class DataController {
 	 * used by e.g. collaborative editing plugin that syncs remote data on init.
 	 *
 	 * @fires init
-	 * @param {String} data Input data.
-	 * @param {String} [rootName='main'] Root name.
+	 * @param {String|Object.<String,String>} data Input data as a string or an object containing rootName - initialData pairs to
+	 * initialize data on multiple roots at once.
 	 * @returns {Promise} Promise that is resolved after the data is set on the editor.
```
pjasiun: Usage examples in docs would be great.

src/controller/datacontroller.js
```diff
@@ -113,6 +113,21 @@ export default class DataController {
 	 * @returns {String} Output data.
 	 */
 	get( rootName = 'main' ) {
+		if ( !_checkIfRootsExists( [ rootName ], this.model.document.getRootNames() ) ) {
```
pjasiun:
```suggestion
		if ( !this._checkIfRootsExists( [ rootName ] ) ) {
```

src/controller/datacontroller.js
```diff
@@ -216,19 +262,51 @@ export default class DataController {
...
+	 *		dataController.set( { main: '<p>Foo</p>', title: '<h1>Bar</h1>' } ); // Sets data on the `main` and `title` roots.
+	 *
+	 * @param {String|Object.<String,String>} data Input data as a string or an object containing `rootName` - `data`
+	 * pairs to set data on multiple roots at once.
 	 */
-	set( data, rootName = 'main' ) {
-		// Save to model.
-		const modelRoot = this.model.document.getRoot( rootName );
+	set( data ) {
+		let initialData = {};
```
pjasiun: This is not `initialData` in this case. Maybe you can just use `data`?

f1ames: `data` is a param name already so I can go with something like:

```js
if ( typeof data === 'string' ) {
    data = { main: data }; // Default root is 'main'. To set data on a different root, object should be passed.
}
```

but it is a little bit ugly or try to come up with a new name.

src/controller/datacontroller.js
f1ames: Moved `_checkIfRootsExists` function to class.

src/controller/datacontroller.js
f1ames: I changed it to `newData`.

src/controller/datacontroller.js
```diff
@@ -216,19 +262,51 @@ export default class DataController {
...
 
 		this.model.enqueueChange( 'transparent', writer => {
-			writer.setSelection( null );
-			writer.removeSelectionAttribute( this.model.document.selection.getAttributeKeys() );
+			for ( const rootName of Object.keys( newData ) ) {
+				// Save to model.
+				const modelRoot = this.model.document.getRoot( rootName );
+
+				writer.setSelection( null );
+				writer.removeSelectionAttribute( this.model.document.selection.getAttributeKeys() );
```
pjasiun: You can remove the selection reset from the loop.

f1ames:
The CI is red, but I have checked locally and it's green (apart from two tests mentioned in PR description). Isn't it related to the recent `yarn` migration?

f1ames:
Yep, I see same tests failing on master - https://travis-ci.org/ckeditor/ckeditor5-engine/builds/477750134.

src/controller/datacontroller.js
pjasiun: `removeSelectionAttribute` is also part of the selection reset.

f1ames: 🤦‍♂️ 👍

PR #1628, opened 2019-01-02
T/1618: Merge *cast-selection-converters into *casthelpers

### LINKED ISSUE #1618: Merge *cast-selection-converters into *casthelpers
We have too many files in the `conversion` directory. For both upcast and downcast we have a dispatcher, helpers, and selection converters.

Also, when we had `*cast-converters` and `*cast-selection-converters` we had a nice symmetry. Now, `*cast-selection-converters` looks just strange there.

This is why I think that they should be merged into `*casthelpers`, as additional exports. It makes sense now since we managed to limit the number of exports from these files.

jodator:

Other: Merge *cast-selection-converters into *casthelpers. Closes #1618.

BREAKING CHANGE: The `clearAttributes()` downcast conversion helper was moved to `module:engine/conversion/downcasthelpers`.
BREAKING CHANGE: The `convertCollapsedSelection()` downcast conversion helper was moved to `module:engine/conversion/downcasthelpers`.
BREAKING CHANGE: The `convertRangeSelection()` downcast conversion helper was moved to `module:engine/conversion/downcasthelpers`.
BREAKING CHANGE: The `convertSelectionChange()` upcast conversion helper  was moved to `module:engine/conversion/upcasthelpers`.

---


pjasiun:
I think they should be also marked as protected exports. Note that it is fine to use the protected code in tests and related files in the same module.

pjasiun:
Ach, I see that `insertText`, `remove`, etc. are not protected. Thet let me rethink it.

pjasiun:
I was thinking about them for a while (and I even started writing a follow-up) and let's keep it the way it is. These helpers are not exposed in the `editor` API, you need an external import to use them. Also, I can imagine one who wants to build a tool similar to `dev-utils` based on that and this is fine.

PR #1632, opened 2019-01-03
Fix: Prevent `model.Writer` from inserting empty text nodes

### LINKED ISSUE #1320: model.Writer should not be letting creating empty text nodes

scofalik:

Fix: Prevent `model.Writer` from inserting empty text nodes. Closes #1320.


scofalik:
This is a new version of PR #1322.

I removed array tests because even though they pass it is incorrect to pass `Array` to `model.Writer#insert`.

OTOH, I've seen that in some places `Array` are in fact used. So we may need to enable `Array` as a correct input for `model.writer#insert` but this can and should be done in a separate issue.

PR #1644, opened 2019-01-23
T/ckeditor5/479: Made it possible to use addPlaceholder() on the root editable

### LINKED ISSUE #899: Rename attachPlaceholder to something more humanly
We're not attaching the placeholder. You attach a poster to a wall. We could either use a simple `addPlaceholder()` or `enablePlaceholder()`.

oleq:

Feature: Moved the root element DOM attributes management from the UI to the engine. Made it possible to use `addPlaceholder()` (now `enablePlaceholder()`) on the root editable. Introduced the `View.detachDomRoot()` method. Implemented additional placeholder helpers (`showPlaceholder()`, `hidePlaceholder()`, `needsPlaceholder()`) (see ckeditor/ckeditor5#479). Closes #899.

BREAKING CHANGE: The `attachPlaceholder()` has been renamed to `enablePlaceholder()`.
BREAKING CHANGE: `enablePlaceholder()` accepts a configuration object instead of separate parameters.
BREAKING CHANGE: The `detachPlaceholder()` has been renamed to `disablePlaceholder()`.

---

### Additional information

Part of the [constellation](https://github.com/ckeditor/ckeditor5/tree/t/479).

### DIFF
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -1,5 +1,5 @@
 /**
- * @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
+ * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
  * For licensing, see LICENSE.md.
  */
 
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -82,14 +82,11 @@ export default class EditingController {
 		//
 		// See  https://github.com/ckeditor/ckeditor5-engine/issues/1528
 		this.listenTo( this.model, '_beforeChanges', () => {
-			this.view._renderingDisabled = true;
+			this.view._disableRendering( true );
 		}, { priority: 'highest' } );
 
-		this.listenTo( this.model, '_afterChanges', ( evt, { hasModelDocumentChanged } ) => {
-			this.view._renderingDisabled = false;
-			if ( hasModelDocumentChanged ) {
-				this.view.render();
-			}
+		this.listenTo( this.model, '_afterChanges', () => {
+			this.view._disableRendering( false );
 		}, { priority: 'lowest' } );
 
 		// Whenever model document is changed, convert those changes to the view (using model.Document#differ).
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -69,6 +69,8 @@ export default class EditingController {
 		const selection = doc.selection;
 		const markers = this.model.markers;
 
+		this.view._renderingDisabled = true;
+
 		// When plugins listen on model changes (on selection change, post fixers, etc) and change the view as a result of
 		// model's change, they might trigger view rendering before the conversion is completed (e.g. before the selection
 		// is converted). We disable rendering for the length of the outermost model change() block to prevent that.
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -85,9 +85,11 @@ export default class EditingController {
 			this.view._renderingDisabled = true;
 		}, { priority: 'highest' } );
 
-		this.listenTo( this.model, '_afterChanges', () => {
+		this.listenTo( this.model, '_afterChanges', ( evt, { hasModelDocumentChanged } ) => {
 			this.view._renderingDisabled = false;
-			this.view.render();
+			if ( hasModelDocumentChanged ) {
+				this.view.render();
+			}
 		}, { priority: 'lowest' } );
 
 		// Whenever model document is changed, convert those changes to the view (using model.Document#differ).
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -69,13 +69,6 @@ export default class EditingController {
 		const selection = doc.selection;
 		const markers = this.model.markers;
 
-		// When taking over an editing view, disable the rendering until the first
-		// changes were made to the model. This allows a free manipulation of the view root
-		// attributes without any risk of breaking anything, for instance, the initial content
-		// of the DOM root element. If the rendering was allowed, any change of attributes would
-		// trigger rendering and synchronize the DOM root with a (still) empty model.
-		this.view._renderingDisabled = true;
-
 		// When plugins listen on model changes (on selection change, post fixers, etc) and change the view as a result of
 		// model's change, they might trigger view rendering before the conversion is completed (e.g. before the selection
 		// is converted). We disable rendering for the length of the outermost model change() block to prevent that.
--- a/src/controller/editingcontroller.js
+++ b/src/controller/editingcontroller.js
@@ -69,6 +69,11 @@ export default class EditingController {
 		const selection = doc.selection;
 		const markers = this.model.markers;
 
+		// When taking over an editing view, disable the rendering until the first
+		// changes were made to the model. This allows a free manipulation of the view root
+		// attributes without any risk of breaking anything, for instance, the initial content
+		// of the DOM root element. If the rendering was allowed, any change of attributes would
+		// trigger rendering and synchronize the DOM root with a (still) empty model.
 		this.view._renderingDisabled = true;
 
 		// When plugins listen on model changes (on selection change, post fixers, etc) and change the view as a result of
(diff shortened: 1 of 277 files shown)


src/view/rooteditableelement.js
```diff
@@ -35,6 +35,17 @@ export default class RootEditableElement extends EditableElement {
...
+		/**
+		 * A DOM attributes cache. It saves the initial values of DOM attributes before the DOM element
+		 * is {@link module:engine/view/view~View#attachDomRoot attached} to the view so later on, when
+		 * the view is destroyed, they can be easily restored. This way, the DOM element can go back to
+		 * the (clean) state as if the editing view never used it.
+		 *
+		 * @protected
+		 * @member {Object.<String,String>}
+		 */
+		this._initialDomAttributes = {};
```
pjasiun: I think it is a wrong place to store this. This is `attachDomRoot` internal logic and should be stored in a `WeakMap` inside `view/view.js`, not as a `RootEditableElement` property. This is a property which is not used by `RootEditableElement` class at all.

src/view/placeholder.js
```diff
@@ -44,12 +46,16 @@ export function attachPlaceholder( view, element, placeholderText, checkFunction
...
 
 /**
- * Removes placeholder functionality from given element.
+ * Removes the placeholder functionality from a given element.
  *
  * @param {module:engine/view/view~View} view
  * @param {module:engine/view/element~Element} element
  */
-export function detachPlaceholder( view, element ) {
+export function removePlaceholder( view, element ) {
```
pjasiun: I was looking for usage of this function but I found none (only tests). If it is not needed maybe the change in this function is not needed too.

src/view/placeholder.js
```diff
@@ -145,3 +159,31 @@ function updateSinglePlaceholder( writer, element, info ) {
...
+ *
+ * Instead, the placeholder must be added to that empty element and this helper returns it, if there
+ * is one.
+ *
+ * @param {module:engine/view/rooteditableelement~RootEditableElement} root The root editable view that
+ * is to have a placeholder.
+ * @returns {module:engine/view/element~Element|null} An element the placeholder can be added to.
+ */
+export function getRootPlaceholderElement( root ) {
+	return () => {
```
pjasiun: It is a little too magic for me. `get...Element` function does not return an element, but a function which returns element. In fact, it also decides if the placeholder should be added based on the child could, so can be used instead of `checkFunction`. Even if it let us integrate placeholders with a few lines of code, the time one will need to spent to understand how it works is not worth it. I think this logic should be some outside in the `EditorUI` class.

pjasiun: I think such code in the editor UI would be more readable:
```js
let elementWithPlaceholder;
const viewRoot = view.document.getRoot();

model.document.on( 'change', () => {
	const firstRootChild = viewRoot.getChild( 1 );

	// Remove placeholder when is not needed.
	if ( viewRoot.childCount != 1 || !firstRootChild || firstRootChild.is( 'element' ) ) {
		if ( elementWithPlaceholder ) {
			removePlaceholder( view, elementWithPlaceholder );
			elementWithPlaceholder = undefined;
		}

		return;
	}

	// Attach placeholder only if firstRootChild element changed.
	if ( elementWithPlaceholder !== firstRootChild ) {
		addPlaceholder( view, firstRootChild, label );
		elementWithPlaceholder = firstRootChild;
	}
}, { priority: 'low' } );
```

Then it can be change to a utils.

src/view/placeholder.js
oleq: AFAIR it is used in Letters, isn't it?

src/view/placeholder.js
oleq: I have mixed feelings about this piece of code you proposed. Where would you like to keep it?

src/view/view.js
```diff
@@ -183,20 +185,64 @@ export default class View {
...
+	 *
+	 * @param {String} name Name of the root to detach.
+	 */
+	detachDomRoot( name ) {
+		const domRoot = this.domRoots.get( name );
+		const viewRoot = this.document.getRoot( name );
+
+		this.change( writer => {
+			// Clean-up the changes made by the change:isReadOnly listener.
+			writer.removeAttribute( 'contenteditable', viewRoot );
```
pjasiun: Not only `contenteditable` attribute should be removed by all attributes added by the engine.

Reinmar:
During a F2F talk we discussed such a simplified API:

```js
export function hidePlaceholder() {}

export function showPlaceholder() {}

export function needsPlaceholder() {}

export function enablePlaceholder( view, element, isElementParagraphLike=true ) {
  const whereNeeded = isElementParagraph ? element : element.getChild( 0 );

  view.registerPostFixer( () =>
    if ( needsPlaceholder( whereNeeded ) ) {
      showPlaceholder( view, whereNeeded );
    } else {
      hidePlaceholder( view, whereNeeded );
    }
  } );
}
```

That's a pseudo code of how it could look. The idea is that the "whether the placeholder should be added" is limited to two supported cases – directly in the passed element or in its first child *if it's a ContainerElement*.

If anyone needs to support more tricky cases like displaying the placeholder in the second element in the root, they can use `needsPlaceholder()` and `show/hidePlaceholder()` functions. Ideally, they shouldn't need to write much more code.

Reinmar:
One more thing – we discussed moving this code to `src/utils`. I'm not sure this is necessary now, because it won't use `model` for anything (we considered an option where it works on the model), but it's a possibility.

pjasiun:
> That's a pseudo code of how it could look.

It is a very PSEUDO code ;) It is not as simple as:

```js
const whereNeeded = isParagraphLikeElement ? element : element.getChild( 0 );
```

If it is not a paragraph-like element we need to follow changes in the parent. We need to remove the placeholder if the element split and there is more than one child, it might happen that we will have to remove the placeholder from a different than the first element (for instant some content is added at the beginning of the document, before the child with the placeholder).

Also, `needsPlaceholder` most probably also will need `isParagraphLikeElement` flag, because it should use a very different check if it checks paragraph and if it checks root element. Or we could limit `needsPlaceholder` to paragraph-like elements.

> One more thing – we discussed moving this code to `src/utils`. I'm not sure this is necessary now, because it won't use `model` for anything (we considered an option where it works on the model), but it's a possibility.

Agree, it could stay in the view, where it is now.

src/controller/editingcontroller.js
```diff
@@ -69,6 +69,13 @@ export default class EditingController {
 		const selection = doc.selection;
 		const markers = this.model.markers;
 
+		// When taking over an editing view, disable the rendering until the first
+		// changes were made to the model. This allows a free manipulation of the view root
+		// attributes without any risk of breaking anything, for instance, the initial content
+		// of the DOM root element. If the rendering was allowed, any change of attributes would
+		// trigger rendering and synchronize the DOM root with a (still) empty model.
+		this.view._renderingDisabled = true;
```
pjasiun: Unfortunately, it does not help much. It is enough to have any model change, even a one which does not change anything in the model document (like [data parsing](https://github.com/ckeditor/ckeditor5-engine/blob/05e4e200a378f99d94afbe03d84baf6b2170165a/src/controller/datacontroller.js#L329)) to cause change block which [sets this flag to `true` and runs rendering](https://github.com/ckeditor/ckeditor5-engine/blob/v12.0.0/src/controller/editingcontroller.js#L83-L84). However, @ma2ciek is [working on it](https://github.com/ckeditor/ckeditor5-engine/issues/1653).

oleq: I reverted those changes in favor of #1661.

src/view/view.js
```diff
@@ -200,14 +213,48 @@ export default class View {
 		// Set view root name the same as DOM root tag name.
 		viewRoot._name = domRoot.tagName.toLowerCase();
 
+		const initialDomRootAttributes = {};
```
pjasiun: I think it should be `WeakMap` `root => Object ( attributeName => attributeValue )` The way it now means that you put attributes for all roots together and when you restore them they can be mixed. It will work fine only for a single root.

oleq: It is like you say. Check out https://github.com/ckeditor/ckeditor5-engine/pull/1644/files#diff-0dfbdba2b2bbf34988f67a22d584575dR239.

pjasiun: Ach, you're right. My bad.

PR #1654, opened 2019-01-31
T/ckeditor5/1096: Introduce schema.isInline definition.

jodator:

Feature: Introduce inline object type to schema. Closes ckeditor/ckeditor5#1049. Closes ckeditor/ckeditor5#1426.

---

### Additional information

* Part of changes required for inline-widgets. [Master PR](https://github.com/ckeditor/ckeditor5-widget/pull/70).


src/conversion/mapper-helpers.js
```diff
@@ -0,0 +1,113 @@
...
+		}
+	};
+}
+
+/**
+ * Maps model position to view position if the position set after/before inline element (ie. inline widget).
+ *
+ *		editor.editing.mapper.on( 'modelToView', mapModelPositionOnInlineElement( model, view ) );
+ *
+ * This methods ensures that selection set on inline element is set outside surrounding text nodes as by default the position would be set
```
ma2ciek:
```suggestion
 * This method ensures that selection set on an inline element is set outside surrounding text nodes as by default the position would be set
```

src/conversion/mapper-helpers.js
```diff
@@ -0,0 +1,113 @@
...
+ * at the end or begging of a text node that is previous/next to the inline element:
```
ma2ciek:
```suggestion
 * at the end or beginning of a text node that is previous/next to the inline element:
```

Reinmar:
> Feature: Introduce inline object type to schema. Closes [ckeditor/ckeditor5#1049](https://github.com/ckeditor/ckeditor5/issues/1049). Closes [ckeditor/ckeditor5#1426](https://github.com/ckeditor/ckeditor5/issues/1426).

U sure that this description covers the changes? ;)

jodator:
Well it does introduces the `isInline` thingy to the schema with additional changes needed to make that work. It also changes `$text` schema definition to include it (dunno if BREAKING CHANGE though).

> Feature: Introduce inline object type to schema.

But I don't know how to name it nicely without writing an elaboration about that.

Feature: Introduce inline object type to the editor model schema, handling positions around inline object elements and add isInline=ture to $text schema definition.

Would probably cover that better but is too long...

Reinmar:
What about?

image

jodator:
This is outcome of introducing the inline elements. But i get the point that it also needs a mention. I'll try to came with something better.

Reinmar:
I think we can do better with mapping positions than registering custom converters. @jodator explained to me that those converters were added by him because view position that was located inside an inline element was not correctly converted by the mapper which was crashing the editor. The converters proposed by @jodator handle those cases as well as try to do what selection post-fixer would normally do.

I think we should fix mapper so it's able to convert positions which end up inside inline model elements. Those positions are still fine and, ideally, should work and not crash the editor. In the future, this will handle well inline widgets with inline nested editables.

Then, we need to figure out whether the selection post-fixer is correcting those selections that start outside an inline limit element and end inside it – just like in the block widget cases.

scofalik:
I've pushed two changes: removed mapping callbacks and fixed `view.Selection#getSelectedElement()`. I hope this covers what we discussed F2F. I will also publish the removed mapping callbacks as utils in `ckeditor5-widget` repo.

PR #1656, opened 2019-01-31
Introduced `model.isEmpty()` method. `DataController.get()` method can now trim empty data

f1ames:

Feature: Introduced `model.isEmpty()` method. `DataController.get()` method can now trim empty data. Closes ckeditor/ckeditor5#401.

BREAKING CHANGE: `DataController.get()` method now returns an empty string by default when editor content is empty.

---

### Additional information

See ckeditor/ckeditor5#401. I have also created https://github.com/ckeditor/ckeditor5/tree/t/401 branch for easier testing.

There are a few related PRs which should be closed together with this one:

* https://github.com/ckeditor/ckeditor5-core/pull/161
* https://github.com/ckeditor/ckeditor5-block-quote/pull/32
* https://github.com/ckeditor/ckeditor5-enter/pull/61
* https://github.com/ckeditor/ckeditor5-paragraph/pull/41
* https://github.com/ckeditor/ckeditor5-table/pull/169
* https://github.com/ckeditor/ckeditor5-typing/pull/180

### DIFF
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -118,8 +118,8 @@ export default class DataController {
 	 * @param {Object} [options]
 	 * @param {String} [options.rootName='main'] Root name.
 	 * @param {String} [options.trim='empty'] Whether returned data should be trimmed. This option is set to `empty` by default,
-	 * which means whenever editor content is considered empty, the empty string will be returned. To turn off trimming completely
-	 * use `none`. In such cases exact content will be returned (for example `<p>&nbsp;</p>` for empty editor).
+	 * which means whenever editor content is considered empty, an empty string will be returned. To turn off trimming completely
+	 * use `'none'`. In such cases exact content will be returned (for example `<p>&nbsp;</p>` for an empty editor).
 	 * @returns {String} Output data.
 	 */
 	get( options ) {
@@ -131,7 +131,7 @@ export default class DataController {
 			 * is called with non-existent root name. For example, if there is an editor instance with only `main` root,
 			 * calling {@link #get} like:
 			 *
-			 * 		data.get( 'root2' );
+			 *		data.get( 'root2' );
 			 *
 			 * will throw this error.
 			 *
@@ -142,8 +142,11 @@ export default class DataController {
 
 		const root = this.model.document.getRoot( rootName );
 
-		// Get model range.
-		return trim === 'empty' && !this.model.hasContent( root, { trimWhitespaces: true } ) ? '' : this.stringify( root );
+		if ( trim === 'empty' && !this.model.hasContent( root, { ignoreWhitespaces: true } ) ) {
+			return '';
+		}
+
+		return this.stringify( root );
 	}
 
 	/**
--- a/src/controller/datacontroller.js
+++ b/src/controller/datacontroller.js
@@ -123,10 +123,7 @@ export default class DataController {
 	 * @returns {String} Output data.
 	 */
 	get( options ) {
-		options = options || {};
-
-		const rootName = options.rootName || 'main';
-		const trim = options.trim || 'empty';
+		const { rootName = 'main', trim = 'empty' } = options || {};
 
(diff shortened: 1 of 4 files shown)


src/controller/datacontroller.js
```diff
@@ -145,14 +154,21 @@ export default class DataController {
 	 *
 	 * @param {module:engine/model/element~Element|module:engine/model/documentfragment~DocumentFragment} modelElementOrFragment
 	 * Element whose content will be stringified.
+	 * @param {Boolean} [skipEmpty=false] Whether content considered empty should be skipped. The method
```
Reinmar: I think that this logic should stay in `get()`. Less API changes is better.

src/controller/datacontroller.js
```diff
@@ -115,10 +115,19 @@ export default class DataController {
...
-	 * @param {String} [rootName='main'] Root name.
+	 * @param {Object} [options]
+	 * @param {String} [options.rootName='main'] Root name.
+	 * @param {String} [options.trim='empty'] Whether returned data should be trimmed. This option is set to `empty` by default,
+	 * which means whenever editor content is considered empty, the empty string will be returned. To turn off trimming completely
+	 * use `none`. In such cases exact content will be returned (for example `<p>&nbsp;</p>` for empty editor).
 	 * @returns {String} Output data.
 	 */
-	get( rootName = 'main' ) {
+	get( options ) {
```
Reinmar: This can be done with destructuring.

src/model/model.js
```diff
@@ -472,6 +474,49 @@ export default class Model {
 		return false;
 	}
 
+	/**
+	 * Checks whether the given {@link module:engine/model/range~Range range} or
+	 * {@link module:engine/model/element~Element element} is considered empty.
+	 *
+	 * The range or element is considered non-empty if it contains any:
+	 * 		* EmptyElement
```
Reinmar: Can a model range contain a view empty element? ;)

src/model/model.js
```diff
@@ -472,6 +474,49 @@ export default class Model {
...
+	 * 		* Non-plain `ContainerElement` (for example widget)
+	 * 		* Non-plain `AttributeElement` (for example comment)
+	 *
+	 * 	This method should be used to check if the element/range/editor contains any printable/meaningful content.
+	 * 	It is also the correct method to check if editor is empty.
+	 *
+	 * @param {module:engine/model/range~Range|module:engine/model/element~Element} rangeOrElement Range or element to check.
+	 * @returns {Boolean}
+	 */
+	isEmpty( rangeOrElement ) {
```
Reinmar: The relation between `isEmpty()` and `hasContent()` needs to be clarified. Both these methods are defined on `Model` but they have different semantics. I wonder if we shouldn't rename `hasContent()` or `isEmpty()` to somehow differentiate between these two.

cc @pjasiun

Reinmar: Let's merge `isEmpty()` into `hasContent()` (i.e. we will have a single method for that called `hasContent()`).

Right now there are 2 differences between them – check for markers and `/\S+/`. The former can be done automatically by `hasContent()`. The latter could be enabled by an `option.trimWhitespaces`.

f1ames: 👍

src/controller/datacontroller.js
```diff
@@ -134,8 +140,10 @@ export default class DataController {
 			throw new CKEditorError( 'datacontroller-get-non-existent-root: Attempting to get data from a non-existing root.' );
 		}
 
+		const root = this.model.document.getRoot( rootName );
+
 		// Get model range.
-		return this.stringify( this.model.document.getRoot( rootName ) );
+		return trim === 'empty' && !this.model.hasContent( root, { trimWhitespaces: true } ) ? '' : this.stringify( root );
```
Reinmar: This comment... ;)

Espumisan:
I think the probelm in here is that there are no samples how to implement this in my project

PR #1657, opened 2019-01-31
Stop invoking `view.render()` by `EditingController` when the model document isn't changed

### LINKED ISSUE #1653: Prevent UI from reacting to the conversion in some situations
This issue is a continuation of the https://github.com/ckeditor/ckeditor5-table/issues/168.

In short:

Editors initialize plugins, UI, and then the data controller in that order. It might happen, that the data controller initialization is replaced with a custom implementation, which uses the conversion API to parse the initial data.

During such conversion, the view and the model are still `empty`. Among others, the view document selection is unset. The conversion triggers a cascade of events 
```
TypeError: Cannot read property 'parent' of null
    at findAncestor (utils.js:18)
    at Object.getTableWidgetAncestor [as getRelatedElement] (utils.js:64)
    at WidgetToolbarRepository._updateToolbarsVisibility (widgettoolbarrepository.js:152)
    at BalloonEditorUI.listenTo (widgettoolbarrepository.js:87)
    at BalloonEditorUI.fire (emittermixin.js:196)
    at BalloonEditorUI.update (editorui.js:92)
    at Document.EditorUI.listenTo (editorui.js:63)
    at View.on (view.js:165)
    at View.change (view.js:376)
    at View.render (view.js:390)
    at Model.EditingController.listenTo (editingcontroller.js:83)
    at Model._runPendingChanges (model.js:705)
    at Model.change (model.js:156)
    at DataController.toModel (datacontroller.js:352)
    at DataController.parse (datacontroller.js:334)
```

After latest changes in the `ckeditor5-table` and `ckeditor5-widget`, the `WidgetToolbar` invokes the visibility callbacks for the table's `tableContent` toolbar, which operates on the position (which is equal to `null` as the selection isn't set at this point.

Such quick fix as:

https://github.com/ckeditor/ckeditor5-engine/blob/fd2005d2f84db05242d689ce3a31208f287e04c5/src/view/view.js#L161-L166

```
this.on( 'render', () => {
	this._render();

	if ( this.document.selection.rangeCount !== 0 ) {
		// Informs that layout has changed after render.
		this.document.fire( 'layoutChanged' );
	}
} );
```
...works and doesn't break editor and existing tests. I'm not sure if it's perfect (but editor's UI shouldn't respond if the selection is empty).

PS: IMO, the conversion should be more `pure`, in that case, parsing invokes UI actions. So the fix could be even more generic, preventing any UI action during the parsing.

ma2ciek:

Fix: Stopped invoking `view.render()` by `EditingController` when the model document isn't changed. Closes #1653.

---

### Additional information

* `Model#_change` event marked as deprecated.
* Introduced protected method: `_handleChangeBlock()`.
* Introduced protected method: `_hasDocumentChangedFromTheLastChangeBlock()`.

### DIFF
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -151,6 +151,7 @@ export default class Document {
 			}
 		}, { priority: 'low' } );
 
+		// Listen to selection changes. If selection changed, mark it.
 		this.listenTo( this.selection, 'change', () => {
 			this._hasSelectionChangedFromTheLastChangeBlock = true;
 		} );
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -303,8 +303,8 @@ export default class Document {
 	 * @fires change:data
 	 * @param {module:engine/model/writer~Writer writer} writer The writer on which post-fixers will be called.
 	 */
-	_runPostFixersAndResetDiffer( writer ) {
-		if ( this._hasDocumentChanged() ) {
+	_handleChangeBlock( writer ) {
+		if ( this._hasDocumentChangedFromTheLastChangeBlock() ) {
 			this._callPostFixers( writer );
 
 			if ( this.differ.hasDataChanges() ) {
@@ -325,9 +325,9 @@ export default class Document {
 	 * or {@link module:engine/model/model~Model#change `change()` block}.
 	 *
 	 * @protected
-	 * @returns {Boolean} Returns `true` if document has changed from the differ's reset.
+	 * @returns {Boolean} Returns `true` if document has changed from the last `change()` or `enqueueChange()` block.
 	 */
-	_hasDocumentChanged() {
+	_hasDocumentChangedFromTheLastChangeBlock() {
 		return !this.differ.isEmpty || this._hasSelectionChangedFromTheLastChangeBlock;
 	}
 
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -279,6 +279,21 @@ export default class Document {
 	}
 
 	/**
+	 * A custom `toJSON()` method to solve child-parent circular dependencies.
+	 *
+	 * @returns {Object} A clone of this object with the document property changed to a string.
+	 */
+	toJSON() {
+		const json = clone( this );
+
+		// Due to circular references we need to remove parent reference.
+		json.selection = '[engine.model.DocumentSelection]';
+		json.model = '[engine.model.Model]';
+
+		return json;
+	}
+
+	/**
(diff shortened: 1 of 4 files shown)


src/model/document.js
```diff
@@ -306,6 +293,44 @@ export default class Document {
...
+
+	/**
+	 * Returns whether there is a buffered change or if the selection has changed from the last
+	 * {@link module:engine/model/model~Model#enqueueChange `enqueueChange()` block}
+	 * or {@link module:engine/model/model~Model#change `change()` block}.
+	 *
+	 * @protected
+	 * @returns {Boolean} Returns `true` if document has changed from the differ's reset.
+	 */
+	_hasDocumentChanged() {
```
pjasiun:
```suggestion
	_hasDocumentChangedFromTheLastChangeBlock() {
```

src/model/document.js
```diff
@@ -306,6 +293,44 @@ export default class Document {
...
+	 * Check if there were any changes done on document, and if so, call post-fixers,
+	 * fire `change` event for features and conversion and then reset the differ.
+	 * Fire `change:data` event when at least one operation or buffered marker changes the data.
+	 *
+	 * @protected
+	 * @fires change
+	 * @fires change:data
+	 * @param {module:engine/model/writer~Writer writer} writer The writer on which post-fixers will be called.
+	 */
+	_runPostFixersAndResetDiffer( writer ) {
```
pjasiun:
```suggestion
	_handleChange( writer ) {
```

pjasiun: I do not like the current name. The most important thing this method does is firing `document#change`. However, `_runPostFixersAndFireChangeAndResetDiffer` sounds ridiculous, so I am for shorter `_handleChangeBlock `

src/model/model.js
```diff
@@ -695,14 +696,19 @@ export default class Model {
 			const callbackReturnValue = this._pendingChanges[ 0 ].callback( this._currentWriter );
 			ret.push( callbackReturnValue );
 
-			// Fire internal `_change` event.
+			// Collect an information whether the model document has changed during from the last pending change.
+			hasModelDocumentChanged = hasModelDocumentChanged || this.document._hasDocumentChanged();
+
+			// Fire '_change' event before resetting differ.
 			this.fire( '_change', this._currentWriter );
```
pjasiun: It should be marked as deprecated.

ma2ciek: This event is marked as deprecated on top of that file.

PR #1659, opened 2019-01-31
Introduced `Differ#getChangedMarkers`.

### LINKED ISSUE #1658: Introduce `Differ#getChangedMarkers`
At the moment, in `Differ`, we have a possibility to get markers to remove and markers to add, which is tailored for conversion dispatcher. However, there is no way to actually get a list of markers which changed.

scofalik:

Feature: Introduced `Differ#getChangedMarkers`. Closes #1658.


src/model/differ.js
```diff
@@ -292,6 +292,23 @@ export default class Differ {
 		return result;
 	}
 
+	/**
+	 * Returns all markers which changed.
+	 *
+	 * @returns {Array.<Object>}
```
oskarwrobel:
```suggestion
	 * @returns {Array.<module:engine/model/markercollection~Marker>}
```

PR #1661, opened 2019-02-05
Prevented `View` from firing the `render` event if there were no changes since the last `view.render()` call

### LINKED ISSUE #1653: Prevent UI from reacting to the conversion in some situations

### LINKED ISSUE #1660: Some changes in view are not rendered.
Ater this change https://github.com/ckeditor/ckeditor5-engine/pull/1657 when you change class on the marker element directly in the view (now it is the only way to do it, see https://github.com/ckeditor/ckeditor5-engine/issues/1649) this change is not rendered in the DOM.

ma2ciek:

Fix: Prevented `View` from firing the `render` event if there were no changes since the last rendering. Closes #1653. Closes #1660.

---

### Additional information

* BREAKING CHANGE: The `editing.view.render()` method was renamed to `editing.view.forceRender()`. It should be used with caution as it will re-render editing view and repaint the UI.

### DIFF
--- a/src/view/view.js
+++ b/src/view/view.js
@@ -66,7 +66,7 @@ export default class View {
 		 * Instance of the {@link module:engine/view/document~Document} associated with this view controller.
 		 *
 		 * @readonly
-		 * @member {module:engine/view/document~Document} module:engine/view/view~View#document
+		 * @type {module:engine/view/document~Document}
 		 */
 		this.document = new Document();
 
@@ -76,7 +76,7 @@ export default class View {
 		 * and {@link module:engine/view/observer/observer~Observer observers}.
 		 *
 		 * @readonly
-		 * @member {module:engine/view/domconverter~DomConverter} module:engine/view/view~View#domConverter
+		 * @type {module:engine/view/domconverter~DomConverter}
 		 */
 		this.domConverter = new DomConverter();
 
@@ -84,7 +84,7 @@ export default class View {
 		 * Instance of the {@link module:engine/view/renderer~Renderer renderer}.
 		 *
 		 * @protected
-		 * @member {module:engine/view/renderer~Renderer} module:engine/view/view~View#renderer
+		 * @type {module:engine/view/renderer~Renderer}
 		 */
 		this._renderer = new Renderer( this.domConverter, this.document.selection );
 		this._renderer.bind( 'isFocused' ).to( this.document );
@@ -93,7 +93,7 @@ export default class View {
 		 * Roots of the DOM tree. Map on the `HTMLElement`s with roots names as keys.
 		 *
 		 * @readonly
-		 * @member {Map} module:engine/view/view~View#domRoots
+		 * @type {Map.<String, HTMLElement>}
 		 */
 		this.domRoots = new Map();
 
@@ -101,7 +101,7 @@ export default class View {
 		 * Map of registered {@link module:engine/view/observer/observer~Observer observers}.
 		 *
 		 * @private
-		 * @member {Map.<Function, module:engine/view/observer/observer~Observer>} module:engine/view/view~View#_observers
+		 * @type {Map.<Function, module:engine/view/observer/observer~Observer>}
 		 */
 		this._observers = new Map();
 
@@ -109,7 +109,7 @@ export default class View {
 		 * Is set to `true` when {@link #change view changes} are currently in progress.
 		 *
 		 * @private
-		 * @member {Boolean} module:engine/view/view~View#_ongoingChange
+		 * @type {Boolean}
 		 */
 		this._ongoingChange = false;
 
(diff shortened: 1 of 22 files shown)


ma2ciek:
I see that there's one test failing in the ckeditor5-image (image caption placeholder). I'll need to fix it, but it shouldn't block reviewing.

pjasiun:
Looks promising.

I checked all usage of the `render` method and all of them uses render with the force flag (except the one in the editing controller). It makes sense, `render` with no force flag should not be needed since all view changes cause rendering. This is why I propose such additional change: https://github.com/ckeditor/ckeditor5-engine/compare/t/1653-b...t/1653-b2?expand=1

It makes a huge number of tests fail, but mostly because of the missing `view.render` method. When I've added alias:
```
	render() {
		this.forceRender();
	}
```
Only 8 tests are failing (4 in `FocusObserver`, 2 integration test, 1 `SelectionObserver` and one `view.focus()`). Could you check such change?

ma2ciek:
Could you clarify then how the `view.change()` / `view.render()` should be called in the `EditingController`?

oleq:
I checked this branch in the context of https://github.com/ckeditor/ckeditor5/issues/479 constellation and it seems to be working.

pjasiun:
> Could you clarify then how the `view.change()` / `view.render()` should be called in the `EditingController`?

I pushed one more commit to my POC.

ma2ciek:
I've just merged your proposal @pjasiun to that branch.

src/view/view.js
```diff
@@ -309,9 +330,10 @@ export default class View {
 
 	/**
 	 * The `change()` method is the primary way of changing the view. You should use it to modify any node in the view tree.
-	 * It makes sure that after all changes are made the view is rendered to the DOM. It prevents situations when the DOM is updated
-	 * when the view state is not yet correct. It allows to nest calls one inside another and still performs a single rendering
-	 * after all those changes are made. It also returns the return value of its callback.
+	 * It makes sure that after all changes are made the view is rendered to the DOM (assuming that the view will be changed
+	 * inside the callback). It prevents situations when the DOM is updated when the view state is not yet correct.It allows
```
pjasiun:
```suggestion
	 * inside the callback). It prevents situations when the DOM is updated when the view state is not yet correct. It allows
```

src/view/view.js
```diff
@@ -386,7 +409,8 @@ export default class View {
 	 * Throws {@link module:utils/ckeditorerror~CKEditorError CKEditorError} `applying-view-changes-on-rendering` when
 	 * trying to re-render when rendering to DOM has already started.
 	 */
```
pjasiun: I think you should add to this docs that this method does not need to be used after changes in the view - they will cause rendering anyway. This method is dedicated to special cases when view does not change be need to be rendered anyway (for instance for some observers which overwrites DOM changes if they were not handled).

pjasiun:
Only a view docs improvements needed.

ma2ciek:
I've fixed the placeholder implementation (wrapped initial changes in the `view.change()` block) and its tests so all tests (especially the one failing in the `ckeditor5-image`) are green now.

PR #1666, opened 2019-02-07
Fix: Moving to the same position is not handled by the Differ as a change

### LINKED ISSUE #1664: Moving node to the same position is handled by differ as a change
When node or range is moved to the same position (it's doable through API or as a result of OT) then differ marks this move as a change and fires change event after that. It causes some issues with markers that are inside moved nodes. Markers are not refreshed because of the live range mechanism does not see any change (range is the same, content is the same). In the result, markers are not rendered in the view.

oskarwrobel:

Fix: Moving to the same position is not handled by the Differ as a change. Closes #1664.

---

### Additional information

### DIFF
--- a/src/model/differ.js
+++ b/src/model/differ.js
@@ -151,7 +151,10 @@ export default class Differ {
 			case 'reinsert': {
 				// When range is moved to the same position then not mark it as a change.
 				// See: https://github.com/ckeditor/ckeditor5-engine/issues/1664.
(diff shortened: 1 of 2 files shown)


scofalik:
There's a similar situation that could happen: moving to a position equal to the end of the moved range:
`ab[cd]|ef` move `[cd]` to `|`. It should be handled too. This `]` position is equal to `sourcePosition.getShiftedBy( operation.howMany )`.

PR #1674, opened 2019-02-11
Update model selection data before the post fixers are called.

oskarwrobel:

Fix: Update model selection data after each change that affects the selection. Closes https://github.com/ckeditor/ckeditor5-engine/issues/1673.

---

### Additional information


pjasiun:
I think that we should keep selection attributes and markers up-to-date, so it should be updated after each operation applied.

PR #1678, opened 2019-02-14
Undo and redo no longer crashes in scenarios featuring pasting content into earlier pasted content

scofalik:

Fix: Undo and redo no longer crashes in scenarios featuring pasting content into earlier pasted content. Closes https://github.com/ckeditor/ckeditor5/issues/1385.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-undo/pull/96/files

I had to re-introduce `forceWeakRemove` flag. This mechanism/flag was used before when we had deltas. I figured it was no longer needed after the refactorings we had. It wasn't needed for long, however, unfortunately, those cases need it. They can't be solved in a way other than explicitly saying: "this is undo, do (not) do this or that".

@Mgsy please test *extensively* undo after this PR as it touches multiple scenarios. Please test scenarios with merges, splits and removes. All unit tests were fine, though.

### DIFF
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -306,7 +306,7 @@ export function transformSets( operationsA, operationsB, options ) {
 		originalOperationsBCount: operationsB.length
 	};
 
-	const contextFactory = new ContextFactory( options.document, options.useRelations );
+	const contextFactory = new ContextFactory( options.document, options.useRelations, options.forceWeakRemove );
 	contextFactory.setOriginalOperations( operationsA );
 	contextFactory.setOriginalOperations( operationsB );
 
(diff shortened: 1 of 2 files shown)


Reinmar:
I pasted the same content twice and undid that twice and additional list item was left:

feb-18-2019 09-59-41

Reinmar:
ok, I can see that this is an unrelated issue. I'll extract it to a separate ticket. @Mgsy, unless you know about similar ticket already?

Mgsy:
>@Mgsy, unless you know about similar ticket already?

AFAIR, we don't have it reported.

Reinmar:
Reported: https://github.com/ckeditor/ckeditor5/issues/1540

PR #1679, opened 2019-02-18
Fake selection container was not appended to the new fake selection editable when fake selection changes

scofalik:

Fix: Fake selection container was not appended to the new fake selection editable when the fake selection changes. Closes https://github.com/ckeditor/ckeditor5/issues/1523.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -707,7 +707,10 @@ export default class Renderer {
 			container.appendChild( domDocument.createTextNode( '\u00A0' ) );
 		}
 
-		// Add fake container if not already added.
+		if ( container.parentElement && container.parentElement != domRoot ) {
+			container.remove( container );
+		}
+
 		if ( !container.parentElement ) {
 			domRoot.appendChild( container );
 		}
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -707,11 +707,7 @@ export default class Renderer {
 			container.appendChild( domDocument.createTextNode( '\u00A0' ) );
 		}
 
-		if ( container.parentElement && container.parentElement != domRoot ) {
-			container.remove( container );
-		}
-
-		if ( !container.parentElement ) {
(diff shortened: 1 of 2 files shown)


src/view/renderer.js
```diff
@@ -707,7 +707,10 @@ export default class Renderer {
 			container.appendChild( domDocument.createTextNode( '\u00A0' ) );
 		}
 
-		// Add fake container if not already added.
+		if ( container.parentElement && container.parentElement != domRoot ) {
```
scofalik: This (and below) could simply be

```js
if ( container.parentElement ) {
    container.remove();
}

domRoot.appendChild( container );
```

So simpler, but I am not sure if we want to remove and append dom element (`container`) if there's no need for it. Can it break something? If not, we can go with simpler solution.

Reinmar:
@jodator @Mgsy could you take a look at it? Let's make sure it doesn't blow up something.

src/view/renderer.js
Reinmar: I'd want to avoid moving it around. That may do some strange things to focus. I even considered leaving just one line: `domRoot.appendChild( container );` (and no conditions), but I'm afraid what it may do. It depends on how it's implemented internally in the browsers. 

However, if you feel braver than I, then you can test it :D

scofalik: I may feel brave but I also feel lazy.

jodator: AFAICS:

```js
domRoot.appendChild( container );
```

1. It works
2. Tests pass
3. Memory tests for `editor-classic` passes
4. No increase in detached DOM nodes (I was worrying that it may happen).

Tested on Chrome only though.

ps.: to be clear:
selection_201

Reinmar: > Tested on Chrome only though.

Which means nothing ;) It has to be tested thoroughly in all browsers. That's why I'd choose the safe path myself.

PR #1683, opened 2019-02-18
All content is properly removed after undoing paste in some scenarios

scofalik:

Fix: All content is properly removed after undoing paste in some scenarios. Closes https://github.com/ckeditor/ckeditor5/issues/1540.

### DIFF
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -2304,9 +2304,7 @@ function _makeMoveOperationsFromRanges( ranges, targetPosition ) {
 		const op = new MoveOperation(
 			range.start,
 			range.end.offset - range.start.offset,
-			// If the target is the end of the move range this operation doesn't really move anything.
-			// In this case, it is better for OT to use range start instead of range end.
-			targetPosition.isEqual( range.end ) ? range.start : targetPosition,
+			targetPosition,
 			0
 		);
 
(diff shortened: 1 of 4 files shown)


src/model/operation/transform.js
```diff
@@ -2304,9 +2304,7 @@ function _makeMoveOperationsFromRanges( ranges, targetPosition ) {
 		const op = new MoveOperation(
 			range.start,
 			range.end.offset - range.start.offset,
-			// If the target is the end of the move range this operation doesn't really move anything.
-			// In this case, it is better for OT to use range start instead of range end.
-			targetPosition.isEqual( range.end ) ? range.start : targetPosition,
+			targetPosition,
```
scofalik: I assume that maybe at some point it was important to change the target position, but then we might have added some cases and this got fixed in a different place. It didn't cause any test to fail and in this scenario, it looked like an obvious error (to use `range.start` instead of `targetPosition`).

scofalik:
It seems that along the way (not with this fix) we also fixed "by accident" two of tests that were skipped 🎉. I unskipped them and added them to this PR.

scofalik:
Ugh, you already merged it but there is a coverage drop (one branch) after the changes I've introduced. I will fix it straight on master then.

PR #1686, opened 2019-02-25
Remove fake selection container before inspecting changed child nodes in Renderer.render().

jodator:

Fix: Remove fake selection container before inspecting changed child nodes in Renderer.render().. Closes ckeditor/ckeditor5#1578.

---

### Additional information

* WiP: working solution - beautifying it

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -171,6 +171,10 @@ export default class Renderer {
 	render() {
 		let inlineFillerPosition;
 
+		// Remove unneeded fake selection container from the DOM so it will not affect the DOM nodes diffing.
+		// It will be re-added if needed when rendering the selection.
+		this._removeFakeSelection();
+
 		// Refresh mappings.
 		for ( const element of this.markedChildren ) {
 			this._updateChildrenMappings( element );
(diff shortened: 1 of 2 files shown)


Reinmar:
I'm afraid we shouldn't remove the FSC if we don't need to change the selection. By removing it we force the browser to think where to move the selection and it will be somewhere at the end of the content so it can lead to really strange behaviours.

src/view/renderer.js
```diff
@@ -171,6 +171,10 @@ export default class Renderer {
 	render() {
 		let inlineFillerPosition;
 
+		// Remove unneeded fake selection container from the DOM so it will not affect the DOM nodes diffing.
```
Reinmar: You say it's unneeded, but the changes might be completely unrelated to the selection (e.g. collab changes).

jodator: It's a draft :stuck_out_tongue: 

I'll update the PR with previous solution as I thought on it over night.

jodator: The other change I'm going to add & test is to exclude the FSC from the differ because it doesn't belong to the view. Additionally the FSC rendering is done afterwards after the differ. The FSC shouldn't affect the differ.

ps.: AFAICS the differ always marked the FSC as "removed" node.

jodator:
Closing in favor of: https://github.com/ckeditor/ckeditor5-engine/pull/1687.

PR #1693, opened 2019-03-06
Model#insertContent should return range affected by the insertion

### LINKED ISSUE #1688: Make Model#insertContent return affected range
I propose that `Model#insertContent` could return a range affected by the insertion. It's not strictly a range over the inserted elements, because that range also should include a possible element split. 

At the moment `Model#insertContent` can modify passed selection but in many cases it is simply different than the "affected range". The affected range should be a range which, if passed to `deleteContent`, would "undo" the `insertContent` call.

Examples:

```html
<paragraph>Foo[]bar</paragraph>
```

Insert `<widget></widget>`:

```html
<paragraph>Foo[</paragraph><widget></widget></paragraph>]bar</paragraph>
```

Note that if the range would be set on `<widget>` only, removing it won't set the content back to the original state.

Another example:

```html
<paragraph>Foo[]bar</paragraph>
```

Insert `<paragraph>X</paragraph><table>...</table>`:

```html
<paragraph>Foo[X</paragraph><table>...</table><paragraph>]bar</paragraph>
```

### LINKED ISSUE #1689: Help LivePosition._createAt get out of the undocumented zone
As noted by @jodator, `LivePosition._createAt` "lives in the undocumented zone". It would be nice to document it, like the other static creators (`_createAfter`, `_createBefore`).

scofalik:

Feature: `Model#insertContent()` will return a range affected by the insertion.

In `Model#deleteContent()`, added `doNotAutoparagraph` flag to `options`.

`Position` and `LivePosition` static creators should handle `stickiness` param.

Closes #1689. Closes #1688.

---

### Additional information

### DIFF
--- a/src/model/utils/insertcontent.js
+++ b/src/model/utils/insertcontent.js
@@ -22,6 +22,8 @@ import Selection from '../selection';
  * If an instance of {@link module:engine/model/selection~Selection} is passed as `selectable` it will be modified
  * to the insertion selection (equal to a range to be selected after insertion).
  *
+ * If `selectable` is not passed, the content will be inserted using the current selection of the model document.
+ *
  * **Note:** Use {@link module:engine/model/model~Model#insertContent} instead of this function.
  * This function is only exposed to be reusable in algorithms which change the {@link module:engine/model/model~Model#insertContent}
  * method's behavior.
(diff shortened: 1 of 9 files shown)


src/model/utils/insertcontent.js
```diff
@@ -312,8 +397,32 @@ class Insertion {
...
+				this._affectedStart = LivePosition._createAt( mergePosLeft.nodeBefore, 'end', 'toPrevious' );
+			}
+
 			this.writer.merge( mergePosLeft );
 
+			// If only one element (the merged one) is in the "affected range", also move the affected range end appropriately.
+			//
+			// <paragraph>Foo</paragraph>]<paragraph>Abc</paragraph>[<paragraph>Bar</paragraph>   -->
+			// <paragraph>Foo]Abc</paragraph>[<paragraph>Bar</paragraph>   -->
+			// <paragraph>Foo]Abc[</paragraph><paragraph>Bar</paragraph>
```
pjasiun: I don't get it :/

src/model/utils/insertcontent.js
```diff
@@ -424,19 +545,25 @@ class Insertion {
...
 					this.writer.remove( parent );
 				}
 			} else if ( this.position.isAtEnd ) {
+				// If insertion position is at the end of the parent, move it out instead of splitting.
+				// <p>Foo^</p> -> <p>Foo</p>^
 				this.position = this.writer.createPositionAfter( this.position.parent );
 			} else {
 				const tempPos = this.writer.createPositionAfter( this.position.parent );
 
+				this._setAffectedBoundaries( this.position );
```
pjasiun: And do not get this one :/

scofalik:
CI failed due to rendering time tests.

PR #1694, opened 2019-03-11
Feature: Introduced `Writer#refreshMarker` method.

### LINKED ISSUE #1649: Trigger marker refresh
It is a common case that we want to change the color of the marker whenever the selection is inside it. The problem is that we have no easy way to tell marker to re-render. We can do direct changes in the view, but it caused already a number of issues.

This is why, I propose to have:
```
writer.updateMarker( 'name', { refresh: true } );
```
Or make sure that simple
```
writer.updateMarker( 'name' ); // with no options or with any option
```
causes marker rerender.

---
Originally, this ticket had to be called markers attributes. However, we realized that the state of the marker should not be sent to other collaboration clients and it would be strange to keep it in the marker which is synchronized. Also, the visual state of the marker is not the data model, it is more like a view state. This is why the information if the marker is active or not should be stored in the plugin which handles it, not in the model. All we need to tell `writer` that the marker should be refreshed.

oskarwrobel:

Feature: Added possibility to refresh the marker with no changes through `Writer#updateMarker()` method. Closes #1649.

---

### Additional information

### DIFF
--- a/src/model/markercollection.js
+++ b/src/model/markercollection.js
@@ -167,7 +167,7 @@ export default class MarkerCollection {
 		const marker = this._markers.get( markerName );
 
 		if ( !marker ) {
-			throw new CKEditorError( 'markers-refresh-marker-not-exists: Marker with provided name does not exists.' );
+			throw new CKEditorError( 'markercollection-refresh-marker-not-exists: Marker with provided name does not exists.' );
 		}
 
 		const range = marker.getRange();
--- a/src/model/markercollection.js
+++ b/src/model/markercollection.js
@@ -154,6 +154,27 @@ export default class MarkerCollection {
 	}
 
 	/**
+	 * Fires an {@link module:engine/model/markercollection~MarkerCollection#event:update} event for the given {@link ~Marker marker}
+	 * but does not change the marker. Useful to force {@link module:engine/conversion/downcastdispatcher~DowncastDispatcher downcast
+	 * conversion} for the marker.
+	 *
+	 * @protected
+	 * @fires module:engine/model/markercollection~MarkerCollection#event:update
+	 * @param {String} markerOrName Marker or name of a marker to refresh.
+	 */
+	_refresh( markerOrName ) {
+		const markerName = markerOrName instanceof Marker ? markerOrName.name : markerOrName;
+		const marker = this._markers.get( markerName );
+
+		if ( !marker ) {
+			throw new CKEditorError( 'markers-refresh-marker-not-exists: Marker with provided name does not exists.' );
+		}
+
+		const range = marker.getRange();
+		this.fire( 'update:' + markerName, marker, range, range, marker.managedUsingOperations, marker.affectsData );
+	}
+
+	/**
 	 * Returns iterator that iterates over all markers, which ranges contain given {@link module:engine/model/position~Position position}.
 	 *
 	 * @param {module:engine/model/position~Position} position
--- a/src/model/markercollection.js
+++ b/src/model/markercollection.js
@@ -171,6 +171,7 @@ export default class MarkerCollection {
 		}
 
 		const range = marker.getRange();
+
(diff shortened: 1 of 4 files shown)


pjasiun:
We agreed to use `writer.updateMarker( name )`.

src/model/markercollection.js
```diff
@@ -153,6 +153,27 @@ export default class MarkerCollection {
...
+	 * @protected
+	 * @fires module:engine/model/markercollection~MarkerCollection#event:update
+	 * @param {String} markerOrName Marker or name of a marker to refresh.
+	 */
+	_refresh( markerOrName ) {
+		const markerName = markerOrName instanceof Marker ? markerOrName.name : markerOrName;
+		const marker = this._markers.get( markerName );
+
+		if ( !marker ) {
+			throw new CKEditorError( 'markers-refresh-marker-not-exists: Marker with provided name does not exists.' );
```
scofalik:
```suggestion
			throw new CKEditorError( 'markercollection-refresh-marker-not-exists: Marker with provided name does not exists.' );
```

src/model/markercollection.js
```diff
@@ -153,6 +153,27 @@ export default class MarkerCollection {
...
+		}
+
+		const range = marker.getRange();
```
scofalik:
```suggestion
		const range = marker.getRange();
		
```

src/model/writer.js
```diff
@@ -939,13 +939,38 @@ export default class Writer {
 	}
 
 	/**
-	 * Adds or updates a {@link module:engine/model/markercollection~Marker marker}. Marker is a named range, which tracks
+	 * Adds updates or refreshes a {@link module:engine/model/markercollection~Marker marker}. Marker is a named range, which tracks
```
scofalik:
```suggestion
	 * Adds, updates or refreshes a {@link module:engine/model/markercollection~Marker marker}. Marker is a named range, which tracks
```

src/model/writer.js
```diff
@@ -939,13 +939,38 @@ export default class Writer {
...
 	 * changes in the document and updates its range automatically, when model tree changes. Still, it is possible to change the
 	 * marker's range directly using this method.
 	 *
 	 * As the first parameter you can set marker name or instance. If none of them is provided, new marker, with a unique
 	 * name is created and returned.
 	 *
+	 * As the second parameter you can set the new marker data or leave this parameter as empty what will only refresh
```
scofalik:
```suggestion
	 * As the second parameter you can set the new marker data or leave this parameter as empty which will just refresh
```

tests/model/markercollection.js
```diff
@@ -195,6 +195,24 @@ describe( 'MarkerCollection', () => {
...
+
+			markers._refresh( 'name' );
+
+			sinon.assert.calledWithExactly( markers.fire, 'update:name', marker, range, range, false, false );
+		} );
+
+		it( 'should throw if marker does not exist', () => {
+			expect( () => {
+				markers._refresh( 'name' );
+			} ).to.throw( CKEditorError, 'markers-refresh-marker-not-exists: Marker with provided name does not exists.' );
```
scofalik:
```suggestion
			} ).to.throw( CKEditorError, 'markercollection-refresh-marker-not-exists: Marker with provided name does not exists.' );
```

src/model/writer.js
oskarwrobel: Fixed.

src/model/writer.js
oskarwrobel: Fixed.

src/model/markercollection.js
oskarwrobel: Fixed.

src/model/markercollection.js
oskarwrobel: Fixed.

PR #1698, opened 2019-03-15
Internal: Allowed $marker element in any parent element.

### LINKED ISSUE #1697: $marker element should be allowed everywhere
Model `$marker` is a temporary element created in the conversion process that helps setting markers from the input data to the `Model#markers` collection. Currently, this element is allowed only inside a `$block` and `$root` elements. It causes an issue with disappearing marker while setting data to the editor with a marker inside a table cell. Since `$marker` element is only a temporary element it should be allowed everywhere because a marker range does not have any limitations as long as is inside a `$root` bounds.

oskarwrobel:

Internal: Allowed `$marker` element in any parent element. Closes #1697.

---

### Additional information

### DIFF
--- a/src/model/model.js
+++ b/src/model/model.js
@@ -104,7 +104,7 @@ export default class Model {
 		this.schema.extend( '$text', { allowIn: '$clipboardHolder' } );
 
 		// An element needed by the `upcastElementToMarker` converter.
-		// This element temporarily represents marker bound during the conversion process and is removed
+		// This element temporarily represents a marker boundary during the conversion process and is removed
 		// at the end of the conversion. `UpcastDispatcher` or at least `Conversion` class looks like a
 		// better place for this registration but both know nothing about `Schema`.
 		this.schema.register( '$marker' );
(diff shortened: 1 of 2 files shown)


src/model/model.js
```diff
@@ -103,12 +103,15 @@ export default class Model {
...
 		this.schema.extend( '$text', { allowIn: '$clipboardHolder' } );
 
-		// Element needed by `upcastElementToMarker` converter.
-		// This element temporarily represents marker bound during conversion process and is removed
-		// at the end of conversion. `UpcastDispatcher` or at least `Conversion` class looks like a better for this
-		// registration but both know nothing about Schema.
-		this.schema.register( '$marker', {
-			allowIn: [ '$root', '$block' ]
+		// An element needed by the `upcastElementToMarker` converter.
+		// This element temporarily represents marker bound during the conversion process and is removed
```
scofalik:
```suggestion
		// This element temporarily represents marker boundary during the conversion process and is removed
```

src/model/model.js
```diff
@@ -103,12 +103,15 @@ export default class Model {
...
+		// at the end of the conversion. `UpcastDispatcher` or at least `Conversion` class looks like a
+		// better place for this registration but both know nothing about the Schema.
```
scofalik:
```suggestion
		// better place for this registration but both know nothing about `Schema`.
```

PR #1700, opened 2019-03-15
Removed Model#event:_change

### LINKED ISSUE #1699: Remove Model#event:_change and its uses
As the title says: we decided to remove that internal event. We don't need it anymore internally.

scofalik:

Internal: Removed `Model#event:_change`. Closes #1699.

### DIFF
--- a/src/model/model.js
+++ b/src/model/model.js
@@ -749,18 +749,6 @@ export default class Model {
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


pjasiun:
You should also remove docs: https://github.com/ckeditor/ckeditor5-engine/blob/f36bc2669b73db9b531cd139b988070d7a474320/src/model/model.js#L751-L761

PR #1704, opened 2019-03-25
Introduced UpcastConversionApi#getSplitParts

scofalik:

Feature: Introduced `UpcastConversionApi#getSplitParts`. Also, provided a way to set upcast conversion helper fired for every view element. Closes https://github.com/ckeditor/ckeditor5/issues/1580. Closes https://github.com/ckeditor/ckeditor5/issues/1581.

------

### Additional information

I added a possibility to specify conversion from any view element to model element by not setting `view` property in upcast helper configuration:

```js
conversion.for( 'upcast' ).elementToElement( { model: 'paragraph' } );
```

Will convert any non-converted element to a `paragraph` model element.

### DIFF
--- a/src/conversion/upcastdispatcher.js
+++ b/src/conversion/upcastdispatcher.js
@@ -573,16 +573,12 @@ function createContextTree( contextDefinition, writer ) {
  * Example of a usage in a converter code:
  *
  *		const myElement = conversionApi.writer.createElement( 'myElement' );
- *		// ...
+ *
+ *		// Children conversion may split `myElement`.
  *		conversionApi.convertChildren( myElement, modelCursor );
- *		// ...
+ *
  *		const splitParts = conversionApi.getSplitParts( myElement );
  *		const lastSplitPart = splitParts[ splitParts.length - 1 ];
- *
- *		// Additional processing on all the parts after the children are converted and the original element might be split.
- *		for ( const element of splitParts ) {
- *			// ...
- *		}
  *
  *		// Setting `data.modelRange` basing on split parts:
  *		data.modelRange = conversionApi.writer.createRange(
--- a/src/conversion/upcastdispatcher.js
+++ b/src/conversion/upcastdispatcher.js
@@ -283,14 +283,15 @@ export default class UpcastDispatcher {
 		// Split element to allowed parent.
 		const splitResult = this.conversionApi.writer.split( modelCursor, allowedParent );
 
-		// Using the range returned by `model.Writer#split`, pair original elements with their split parts.
-		// The range returned from the writer spans "between" the split or precisely saying, from the end of the split element
-		// to the beginning of the other part of the split element:
+		// Using the range returned by `model.Writer#split`, we will pair original elements with their split parts.
+		//
+		// The range returned from the writer spans "over the split" or, precisely saying, from the end of the original (split) element
+		// to the beginning of the other (new) part of that element:
 		//
 		// <limit><a><b><c>X[]Y</c></b><a></limit> ->
 		// <limit><a><b><c>X[</c></b></a><a><b><c>]Y</c></b></a>
 		//
-		// After the split no meaningful content can be between the positions in `splitRange` - they have to be touching.
+		// After the split there cannot be any full node between the positions in `splitRange`. The positions are touching.
 		// Also, because of how splitting works, it is easy to notice, that "closing tags" are in the reverse order than "opening tags".
 		// Also, since we split all those elements, each of them has to have the other part.
 		//
--- a/src/conversion/upcastdispatcher.js
+++ b/src/conversion/upcastdispatcher.js
@@ -108,15 +108,14 @@ export default class UpcastDispatcher {
 	 */
 	constructor( conversionApi = {} ) {
 		/**
-		 * List of elements that will be checked after conversion process and if element in the list will be empty it
-		 * will be removed from conversion result.
+		 * List of the elements that were created during splitting.
 		 *
-		 * After conversion process list is cleared.
+		 * After conversion process the list is cleared.
 		 *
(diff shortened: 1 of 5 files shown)


scofalik:
I've also added integration tests for the issue: https://github.com/ckeditor/ckeditor5-block-quote/pull/34. I wasn't sure where really to put them though :S. I hope this place is fine with you.

src/conversion/upcastdispatcher.js
```diff
@@ -283,14 +283,17 @@ export default class UpcastDispatcher {
...
 
-		// Remember all elements that are created as a result of split.
-		// This is important because at the end of conversion we want to remove all empty split elements.
-		//
-		// Loop through positions between elements in range (except split result position) and collect parents.
-		// <notSplit><split1><split2>[pos]</split2>[pos]</split1>[omit]<split1>[pos]<split2>[pos]</split2></split1></notSplit>
-		for ( const position of splitResult.range.getPositions() ) {
-			if ( !position.isEqual( splitResult.position ) ) {
-				this._removeIfEmpty.add( position.parent );
+		const stack = [];
```
pjasiun: I am an old man. I need an image to understand this code.

src/conversion/upcastdispatcher.js
```diff
@@ -301,25 +304,64 @@ export default class UpcastDispatcher {
...
-	 * We need to continue iterating over split elements as long as any element will be removed.
+	 * Registers that `split` element is a split part of the `original` element.
+	 *
+	 * Data set by this method is then used by {@link #_getSplitParts} and {@link #_removeEmptyElements}
+	 *
+	 * @private
+	 * @param {module:engine/model/element~Element} original The original element.
+	 * @param {module:engine/model/element~Element} split The split part element.
+	 */
+	_registerSplitPair( original, split ) {
```
pjasiun:
```suggestion
	_registerSplitPair( original, copy ) {
```

src/conversion/upcastdispatcher.js
```diff
@@ -301,25 +304,64 @@ export default class UpcastDispatcher {
...
-	 * We need to do it smart because there could be elements that are not empty because contains
-	 * other empty elements and after removing its children they become available to remove.
-	 * We need to continue iterating over split elements as long as any element will be removed.
+	 * Registers that `split` element is a split part of the `original` element.
+	 *
+	 * Data set by this method is then used by {@link #_getSplitParts} and {@link #_removeEmptyElements}
+	 *
+	 * @private
+	 * @param {module:engine/model/element~Element} original The original element.
+	 * @param {module:engine/model/element~Element} split The split part element.
```
pjasiun:
```suggestion
	 * @param {module:engine/model/element~Element} copy The copy of the original element which was created as the result of split.
```

src/conversion/upcastdispatcher.js
```diff
@@ -283,14 +283,17 @@ export default class UpcastDispatcher {
...
-				this._removeIfEmpty.add( position.parent );
+		const stack = [];
+
+		for ( const treeWalkerValue of splitResult.range.getWalker() ) {
+			if ( treeWalkerValue.type == 'elementEnd' ) {
+				stack.push( treeWalkerValue.item );
+			} else {
+				// There should not be any text nodes after the element is split, so the only other value is `elementStart`.
+				const originalElement = stack.pop();
+				const splitElement = treeWalkerValue.item;
```
pjasiun: I am not sure about the make `splitElement`. I think `copyElement` is much better.

pjasiun: `partElement`, `partialElement` or `splitPartElement`?

src/conversion/upcastdispatcher.js
```diff
@@ -301,25 +304,64 @@ export default class UpcastDispatcher {
...
+	_getSplitParts( element ) {
+		let parts;
+
+		if ( !this._splitParts.has( element ) ) {
+			parts = [ element ];
+		} else {
+			parts = this._splitParts.get( element );
+		}
+
+		parts.last = parts[ parts.length - 1 ];
```
pjasiun: Actually, I am not sure if we should do it. We do not do it for any other array, even considering that in many cases it could be useful. If we will do it now, we may do it more often and end up in the situation where half of arrays has property `last` and the other half do not.

scofalik: That was your idea :P But, yeah, I am not sure about it either.

src/conversion/upcastdispatcher.js
scofalik: Sorry, I forgot to add it.

src/conversion/upcastdispatcher.js
scofalik: I am not a huge fan of `split` name but I think it is better than `copy`. This is not a copy of that element, they are both split parts. If not `split`, I'd go with `originalPart` + `newPart`.

src/conversion/upcastdispatcher.js
scofalik: See above.

src/conversion/upcastdispatcher.js
scofalik: As above, I can agree for `originalPart` and `splitPart` or `newPart`.

src/conversion/upcastdispatcher.js
scofalik: Or `originalPart` and `splitPart`.

src/conversion/upcastdispatcher.js
ma2ciek: `if (!Array.prototype.last) { Object.defineProperty...` 😈 😄

pjasiun: :trollface:

PR #1707, opened 2019-03-27
Model#deleteContent will not throw anymore if the passed selection is in the graveyard root

### LINKED ISSUE #1706: `deleteContent` in graveyard fails
If `deleteContent()` is executed on a selection which is in the graveyard root it will fail. It's because `startPos` is created as a static position but since it is in the graveyard, it should be updated:

```js
const startPos = selRange.start;
// ...
writer.remove( selRange );
// ...
mergeBranches( writer, startPos, endPos ); // <-- Incorrect `startPos` is used.
```

There are two possible solutions for this:

1. Make `startPos` a `LivePosition`.
2. Cancel `deleteContent` if `selRange` is in the graveyard.

We chose the second option because it is less prone to an error and generates no operations.

scofalik:

Fix: `Model#deleteContent` will not throw anymore if the passed selection is in the graveyard root. Closes #1706.

### DIFF
--- a/src/model/utils/deletecontent.js
+++ b/src/model/utils/deletecontent.js
@@ -63,6 +63,13 @@ export default function deleteContent( model, selection, options = {} ) {
 	const schema = model.schema;
 
 	model.change( writer => {
+		const selRange = selection.getFirstRange();
+
+		// If the selection is already removed, don't do anything.
+		if ( selRange.root.rootName == '$graveyard' ) {
+			return;
+		}
+
 		// 1. Replace the entire content with paragraph.
 		// See: https://github.com/ckeditor/ckeditor5-engine/issues/1012#issuecomment-315017594.
 		if ( !options.doNotResetEntireContent && shouldEntireContentBeReplacedWithParagraph( schema, selection ) ) {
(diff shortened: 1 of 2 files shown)


src/model/utils/deletecontent.js
```diff
@@ -63,6 +63,13 @@ export default function deleteContent( model, selection, options = {} ) {
 	const schema = model.schema;
 
 	model.change( writer => {
+		const selRange = selection.getFirstRange();
+
+		// If the selection is already removed, don't do anything.
+		if ( selRange.root.rootName == '$graveyard' ) {
```
Reinmar: Why not outside, right after checking `selection.isCollapsed`? It seems more natural to have there all the early returns. Since you've put it here, in the `model.change()` block I had to scroll the code down to see if there's anything after `model.change()` that might need (or perhaps should not) be executed.

scofalik: I did it like this to be sure that the selection doesn't get into the graveyard before `model.change` actually executes, but I forgot that it executes right away in all circumstances (as opposed to `enqueueChange`). Will fix.

PR #1711, opened 2019-03-28
Introduced schema#setAttributeProperties() and schema#getAttributeProperties() methods

mlewand:

Feature: Introduced `schema#setAttributeProperties()` and `schema#getAttributeProperties()` methods. Closes ckeditor/ckeditor5#1659.

---

## Followups

- [ ] - implement it in our formatting plugins
	- [ ] - bold, italic, underline, strikethrough, code, subscript, superscript
	- [ ] - fontSize, fontFamily
	- [ ] - alignment
	- [ ] - highlight
- [ ] - integrate it with remove formatting plugin
- [ ] - integrate it in CKE5 inspector

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -498,7 +498,8 @@ export default class Schema {
 	 *			isFormatting: false
 	 *		} );
 	 *
-	 * You can also use custom properties:
+	 * Properties are not limited to members defined in the
+	 * {@link module:engine/model/schema~AttributeProperties `AttributeProperties` type} and you can also use custom properties:
 	 *
 	 *		schema.setAttributeProperties( 'blockQuote', {
 	 *			customProperty: 'value'
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -149,70 +149,6 @@ export default class Schema {
 	}
 
 	/**
-	 * Registers custom properties to a given attribute.
-	 *
-	 * It can be used to mark the attributes relation and handle them in a common way.
-	 *
-	 *		// Mark blockQuote as a formatting attribute.
-	 *		schema.setAttributeProperties( 'blockQuote', {
-	 *			isFormatting: true
-	 *		} );
-	 *
-	 *		// Override code not to be considered a formatting markup.
-	 *		schema.setAttributeProperties( 'code', {
-	 *			isFormatting: false
-	 *		} );
-	 *
-	 * You can also use custom attributes:
-	 *
-	 *		schema.setAttributeProperties( 'blockQuote', {
-	 *			customAttribute: 'value'
-	 *		} );
-	 *
-	 * Subsequent calls to the same attributes will add up the value:
-	 *
-	 *		schema.setAttributeProperties( 'blockQuote', {
-	 *			one: 1
-	 *		} );
-	 *
-	 *		schema.setAttributeProperties( 'blockQuote', {
-	 *			two: 2
-	 *		} );
-	 *
-	 *		console.log( schema.getAttributeProperties( 'blockQuote' ) );
-	 *		// Logs: {one: 1, two: 2}
-	 *
-	 * @param {String} attributeName Name of the attribute to receive properties.
-	 * @param {module:engine/model/schema~AttributeProperties} properties A dictionary of properties.
-	 */
-	setAttributeProperties( attributeName, properties ) {
-		this._attributeProperties[ attributeName ] = Object.assign( this._attributeProperties[ attributeName ] || {}, properties );
-	}
-
-	/**
-	 * Returns properties assigned to a given attribute.
-	 *
-	 * @param {String} attributeName Name of the attribute.
-	 * @returns {module:engine/model/schema~AttributeProperties}
-	 */
-	getAttributeProperties( attributeName ) {
-		return this._attributeProperties[ attributeName ];
-	}
-
-	/**
-	 * Returns all registered items.
-	 *
-	 * @returns {Object.<String,module:engine/model/schema~SchemaCompiledItemDefinition>}
-	 */
-	getDefinitions() {
-		if ( !this._compiledDefinitions ) {
-			this._compile();
-		}
-
-		return this._compiledDefinitions;
-	}
-
-	/**
 	 * Returns a definition of the given item or `undefined` if item is not registered.
 	 *
 	 * @param {module:engine/model/item~Item|module:engine/model/schema~SchemaContextItem|String} item
@@ -532,6 +468,70 @@ export default class Schema {
 				evt.return = retValue;
 			}
 		}, { priority: 'high' } );
+	}
+
+	/**
+	 * Registers custom properties to a given attribute.
+	 *
+	 * It can be used to mark the attributes relation and handle them in a common way.
+	 *
+	 *		// Mark blockQuote as a formatting attribute.
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			isFormatting: true
+	 *		} );
+	 *
+	 *		// Override code not to be considered a formatting markup.
+	 *		schema.setAttributeProperties( 'code', {
+	 *			isFormatting: false
+	 *		} );
+	 *
+	 * You can also use custom attributes:
+	 *
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			customAttribute: 'value'
+	 *		} );
+	 *
+	 * Subsequent calls to the same attributes will add up the value:
+	 *
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			one: 1
+	 *		} );
+	 *
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			two: 2
+	 *		} );
+	 *
+	 *		console.log( schema.getAttributeProperties( 'blockQuote' ) );
+	 *		// Logs: {one: 1, two: 2}
+	 *
+	 * @param {String} attributeName Name of the attribute to receive properties.
+	 * @param {module:engine/model/schema~AttributeProperties} properties A dictionary of properties.
+	 */
+	setAttributeProperties( attributeName, properties ) {
+		this._attributeProperties[ attributeName ] = Object.assign( this._attributeProperties[ attributeName ] || {}, properties );
+	}
+
+	/**
+	 * Returns properties assigned to a given attribute.
+	 *
+	 * @param {String} attributeName Name of the attribute.
+	 * @returns {module:engine/model/schema~AttributeProperties}
+	 */
+	getAttributeProperties( attributeName ) {
+		return this._attributeProperties[ attributeName ];
+	}
+
+	/**
+	 * Returns all registered items.
+	 *
+	 * @returns {Object.<String,module:engine/model/schema~SchemaCompiledItemDefinition>}
+	 */
+	getDefinitions() {
+		if ( !this._compiledDefinitions ) {
+			this._compile();
+		}
+
+		return this._compiledDefinitions;
 	}
 
 	/**
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -149,6 +149,19 @@ export default class Schema {
 	}
 
 	/**
+	 * Returns all registered items.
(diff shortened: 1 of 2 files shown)


src/model/schema.js
```diff
@@ -39,6 +39,14 @@ export default class Schema {
 	constructor() {
 		this._sourceDefinitions = {};
 
+		/**
+		 * A map containing attribute's properties.
+		 *
+		 * @private
+		 * @member {Map.<String,String>}
```
Reinmar: This is a plain object, not a map. So `Object.<String,String>`

src/model/schema.js
```diff
@@ -140,19 +148,6 @@ export default class Schema {
 		this._clearCache();
 	}
 
-	/**
-	 * Returns all registered items.
-	 *
-	 * @returns {Object.<String,module:engine/model/schema~SchemaCompiledItemDefinition>}
-	 */
-	getDefinitions() {
```
Reinmar: Why this change? This method was here on purpose (next to `getDefinition()`).

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
 		}, { priority: 'high' } );
 	}
 
+	/**
+	 * Registers custom properties to a given attribute.
```
Reinmar: Sets? Registering means saying "I exist!". We register nodes (`Schema#register()`). You can call `register( foo )` only once too (cause you report your existence once).

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
...
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			one: 1
+	 *		} );
+	 *
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			two: 2
+	 *		} );
+	 *
+	 *		console.log( schema.getAttributeProperties( 'blockQuote' ) );
+	 *		// Logs: {one: 1, two: 2}
```
Reinmar: `{ one: 1, two: 2 }`

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
...
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			isFormatting: true
+	 *		} );
+	 *
+	 *		// Override code not to be considered a formatting markup.
+	 *		schema.setAttributeProperties( 'code', {
+	 *			isFormatting: false
+	 *		} );
+	 *
+	 * You can also use custom attributes:
```
Reinmar: custom properties

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
...
+	 *
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			customAttribute: 'value'
```
Reinmar: customProperty

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
 		}, { priority: 'high' } );
 	}
 
+	/**
+	 * Registers custom properties to a given attribute.
+	 *
+	 * It can be used to mark the attributes relation and handle them in a common way.
+	 *
+	 *		// Mark blockQuote as a formatting attribute.
```
Reinmar: blockQuote is not an attribute – it's an element

Let's use something realistic – `bold`, `italic`, `fontFamily`, etc.

src/model/schema.js
Reinmar: Or rather:

> Defines properties of the given attribute. 
>
> This method allows assigning additional information to model attributes. Currently, it is only used to mark formatting attributes (like `bold`, `italic`) which allows differentiating them from semantic attributes (such as `src`, `listType`, etc.). This knowledge is then used by features like "Remove format".

And that's it. Good documentation should back all big words by giving real-life examples.

Reinmar: I see you used "metadata" below. That's a nice word too. You may try to use it here too. It may be better than "information" or "properties".

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
...
+	 *
+	 * @param {String} attributeName Name of the attribute to receive properties.
+	 * @param {module:engine/model/schema~AttributeProperties} properties A dictionary of properties.
+	 */
+	setAttributeProperties( attributeName, properties ) {
+		this._attributeProperties[ attributeName ] = Object.assign( this._attributeProperties[ attributeName ] || {}, properties );
+	}
+
+	/**
+	 * Returns properties assigned to a given attribute.
```
Reinmar: > Returns properties of the given attribute.

src/model/schema.js
```diff
@@ -1285,6 +1344,15 @@ export class SchemaContext {
  * @typedef {Object} module:engine/model/schema~SchemaContextItem
  */
 
+/**
+ * A structure containing additional metadata describing the attribute.
+ *
+ * See {@link module:engine/model/schema~Schema#setAttributeProperties `Schema#setAttributeProperties()`} for usage examples.
+ *
+ * @typedef {Object} module:engine/model/schema~AttributeProperties
+ * @property {Boolean} [isFormatting] Indicates that the attribute should be considered as a visual formatting.
```
Reinmar: Missing examples of what it means. Show what common attrs are considered "formatting" and which not.

src/model/schema.js
mlewand: That was actually the output copied straight from the browser. I'll reformat this.

src/model/schema.js
```diff
@@ -39,6 +39,14 @@ export default class Schema {
 	constructor() {
 		this._sourceDefinitions = {};
 
+		/**
+		 * A map containing attribute's properties.
```
oleq:
```suggestion
		 * A map containing attribute properties.
```

src/model/schema.js
oleq:
```suggestion
		 * @member {Object.<String,String>}
```

src/model/schema.js
oleq:
```suggestion
	 * Adds custom properties to an attribute.
```

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
 		}, { priority: 'high' } );
 	}
 
+	/**
+	 * Registers custom properties to a given attribute.
+	 *
+	 * It can be used to mark the attributes relation and handle them in a common way.
```
oleq:
```suggestion
	 * Properties allow categorization of model attributes. Thanks to that, certain groups of attributes can be recognized and handled in the same way by some of the editor features.
```

src/model/schema.js
oleq: This is very confusing. We're talking about properties and all of the sudden there's this line about custom attributes of... attributes? This needs rewording. I guess the intention was that one can use any property name they want.

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
...
+	 *			isFormatting: false
+	 *		} );
+	 *
+	 * You can also use custom attributes:
+	 *
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			customAttribute: 'value'
+	 *		} );
+	 *
+	 * Subsequent calls to the same attributes will add up the value:
```
oleq:
```suggestion
	 * Subsequent calls with the same attribute extend its the custom properties:
```

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
...
+	 *		} );
+	 *
+	 *		schema.setAttributeProperties( 'blockQuote', {
+	 *			two: 2
+	 *		} );
+	 *
+	 *		console.log( schema.getAttributeProperties( 'blockQuote' ) );
+	 *		// Logs: {one: 1, two: 2}
+	 *
+	 * @param {String} attributeName Name of the attribute to receive properties.
```
oleq:
```suggestion
	 * @param {String} attributeName A name of the attribute to gain the properties.
```

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
...
+	 * @param {module:engine/model/schema~AttributeProperties} properties A dictionary of properties.
```
oleq: This is confusing. First, it is stated that

> You can also use custom attributes: 

Then we enforce a type, which clearly declares which properties are allowed.

src/model/schema.js
```diff
@@ -475,6 +470,70 @@ export default class Schema {
...
+	 */
+	setAttributeProperties( attributeName, properties ) {
+		this._attributeProperties[ attributeName ] = Object.assign( this._attributeProperties[ attributeName ] || {}, properties );
```
oleq:
```suggestion
		this._attributeProperties[ attributeName ] = Object.assign( this.getAttributeProperties( attributeName ) || {}, properties );
```

src/model/schema.js
oleq:
```suggestion
	 * Returns properties associated with a given model attribute. See {@link #setAttributeProperties `setAttributeProperties()`}.
```

src/model/schema.js
```diff
@@ -1285,6 +1344,15 @@ export class SchemaContext {
  * @typedef {Object} module:engine/model/schema~SchemaContextItem
  */
 
+/**
+ * A structure containing additional metadata describing the attribute.
```
oleq: As I wrote earlier, `setAttributeProperties()` allows any property names while this type clearly narrows possible options to `isFormatting` only. We should figure that out.

src/model/schema.js
mlewand: I have used dedicated `@see` tag for the "see" part.

src/model/schema.js
mlewand: I'm not sure about receive -> gain change. I semantically both words are very similar, while receive is used more commonly.

I'll keep receive - but ofc feel free to follow this up.

src/model/schema.js
mlewand: This was addressed by fix for one issues raised in previous review.

src/model/schema.js
mlewand: Just a slight modification: I have also removed (now) stray the in "extend its the custom properties" and added will between attribute and extend.

Feel free to follow this up.

src/model/schema.js
mlewand: This was addressed by fix for one issues raised in previous review.

mlewand: The introduction was modified to shortly mention cases and link `AttributeProperties#isFormatting` example for further reading.

Unfortunately I could not link to isFormatting property directly due to ckeditor/ckeditor5-dev#504.

PTAL

src/model/schema.js
mlewand: As funny as it might sound I have grabbed one method too much where moving `setAttributeProperties` and `getAttributeProperties`. Thanks for catching that.

Restored to the original position.

src/model/schema.js
mlewand: Yes it was a typo, should be custom properties.

mlewand:
All the requested changes were addressed. There's one thing that requires some comments:

> This is confusing. First, it is stated that
> 
> You can also use custom attributes:
> 
> Then we enforce a type, which clearly declares which properties are allowed.

This is interesting point @oleq and I can see how it can be an issue.

I can see two ways to resolve it:

* Stick to a dedicated typedef and be explicit about the fact that it could be extended/patched (as basically everything in js world).
	* Pros:
		* Reduced DRY factor.
		* Allows for a more detailed property documentation.
	* Cons:
* Describe the structure using inline object notation. (pushed an example to the `t/ckeditor5/1659b` branch)
	* Pros:
		* The method description is closer to (g/s)etAttributeProperties.
	* Cons:
		* I'm not even sure if our docs properly handle `@returns` tag with inline object. As it renders pretty 💩y.
		* Method API docs length increases, containing details that are not necessary to understand this method.


Taking above into consideration I'm strongly for going with a dedicated typedef, but explicitly say that it can be extended.

Reinmar:
Perfect! ❤️

PR #1719, opened 2019-04-02
view.DowncastWriter#wrap() and #unwrap() fixes for nested attribute elements

### LINKED ISSUE #1716: view.DowncastWriter does not handle partial unwrapping in nested attribute elements
For given view and selection:

```html
<p>
    [<i>
        <span class="foo bar">Foo</span>
    </i>]
</p>
```

Unwrapping `<span class="foo">` will not do anything.

`DowncastWriter#unwrap()` is incorrectly implemented in this regard. When it enters a child, it only looks for full matches, there's no partial unwrapping there.

EDIT: AFAICS the selection doesn't need to be exactly on the `<i>` element, it can be somewhere inside the `<span>`.

scofalik:

Fix: `view.DowncastWriter` will now correctly wrap and unwrap nested attribute elements. Closes #1716. Closes ckeditor/ckeditor5-font#30.

### DIFF
--- a/src/view/downcastwriter.js
+++ b/src/view/downcastwriter.js
@@ -1064,16 +1064,16 @@ export default class DowncastWriter {
 	}
 
 	/**
-	 * Wraps children with provided `attribute`. Only children contained in `parent` element between
+	 * Wraps children with provided `wrapElement`. Only children contained in `parent` element between
 	 * `startOffset` and `endOffset` will be wrapped.
 	 *
 	 * @private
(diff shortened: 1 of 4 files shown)


scofalik:
Rebased with newest master

mlewand:
Before we'll carry out the code review, I'll ask you @Mgsy to manual check whether this fix addresses issue reported ckediotr/ckeditor5-font#30. Also do some random smoke testing with the editor.

PR #1730, opened 2019-04-21
Fix: Prevented from losing selection attributes between operations - IME.

oskarwrobel:

Fix: Prevented from losing selection attributes between operations - IME. Closes https://github.com/ckeditor/ckeditor5-typing/issues/188.

---

### Additional information

Related https://github.com/ckeditor/ckeditor5-typing/pull/189.

### DIFF
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -396,6 +396,10 @@ export default class Document {
 			for ( const callback of this._postFixers ) {
 				// Ensure selection attributes are up to date before each post-fixer.
 				// https://github.com/ckeditor/ckeditor5-engine/issues/1673.
+				//
+				// It might be good to refresh the selection after each operation but at the moment it leads
+				// to losing attributes for composition or and spell checking
+				// https://github.com/ckeditor/ckeditor5-typing/issues/188
 				this.selection.refresh();
 
 				wasFixed = callback( writer );
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -307,14 +307,14 @@ export default class Document {
 		if ( this._hasDocumentChangedFromTheLastChangeBlock() ) {
 			this._callPostFixers( writer );
 
+			// Refresh selection attributes according to the final position in the model after the change.
+			this.selection.refreshAttributes();
+
 			if ( this.differ.hasDataChanges() ) {
 				this.fire( 'change:data', writer.batch );
 			} else {
 				this.fire( 'change', writer.batch );
 			}
-
-			// Refresh selection attributes according to the final position in the model after the change.
-			this.selection.refreshAttributes();
 
 			this.differ.reset();
 		}
--- a/src/model/document.js
+++ b/src/model/document.js
@@ -313,6 +313,9 @@ export default class Document {
 				this.fire( 'change', writer.batch );
 			}
 
+			// Refresh selection attributes according to the final position in the model after the change.
+			this.selection.refreshAttributes();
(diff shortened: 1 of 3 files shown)


Reinmar:
Let's be extremely cautious here. I assigned also @scofalik who worked the most on selection attributes.

src/model/document.js
```diff
@@ -391,6 +394,10 @@ export default class Document {
 
 		do {
 			for ( const callback of this._postFixers ) {
+				// Ensure selection attributes are up to date before each post-fixer.
+				// https://github.com/ckeditor/ckeditor5-engine/issues/1673.
```
scofalik: I'd add here additional comment on why it is done like this and not after every operation so we won't be re-thinking this problem again in a year.

src/model/documentselection.js
```diff
@@ -359,6 +359,14 @@ export default class DocumentSelection {
 		return this._selection.hasAttribute( key );
 	}
 
+	/**
+	 * Refreshes selection attributes and markers according to the current position in the model.
+	 */
+	refreshAttributes() {
```
scofalik: This should be named `refresh()`.

scofalik: Or... different. It doesn't look good when `refreshAttributes` refreshes more than attributes. Or simply let's not wrap it in a function.

oskarwrobel: `refresh()` sounds ok. I was considering it.

src/model/document.js
```diff
@@ -313,6 +313,9 @@ export default class Document {
 				this.fire( 'change', writer.batch );
 			}
 
+			// Refresh selection attributes according to the final position in the model after the change.
```
scofalik: As I understand, the selection attributes are now not refreshed on `change` at all. Does it mean that the selection might have wrong attributes after last post-fixer changes it?

scofalik: Maybe this should be added as a `change` event callback. First because of that reason, second because I think it is better if there's less selection-related code directly in `Document` class. We already have a lot of listeners in `DocumentSelection`. WDYT?

src/model/document.js
```diff
@@ -391,6 +394,10 @@ export default class Document {
 
 		do {
 			for ( const callback of this._postFixers ) {
+				// Ensure selection attributes are up to date before each post-fixer.
+				// https://github.com/ckeditor/ckeditor5-engine/issues/1673.
+				this.selection.refreshAttributes();
```
scofalik: I don't like hard-coding it like this here but I can't find a better idea :(.

src/model/document.js
oskarwrobel: I moved it to Document class to not split the refreshes between two classes.

src/model/document.js
```diff
@@ -307,15 +307,15 @@ export default class Document {
 		if ( this._hasDocumentChangedFromTheLastChangeBlock() ) {
 			this._callPostFixers( writer );
 
+			// Refresh selection attributes according to the final position in the model after the change.
+			this.selection.refreshAttributes();
```
scofalik: Shouldn't it be both before and after `change` event?

oskarwrobel: Theoretically, it should be not necessary. Changing the model on `change` event is not recommended, post fixers are the last who should change the model.

oskarwrobel: Or I missed something?

PR #1731, opened 2019-04-26
Update view.Document#registerPostFixer() docs.

### LINKED ISSUE #1726: view.Document#registerPostFixer() API doc issues
1. Some example is missing.
2. This seems to be incorrect: " If the view should change after registering the post-fixer then it should be done manually calling view.forceRender();" because I'd guess I have to call this method when I changed something in the view, not when I added a post fixer.

pjasiun:

Docs: Update view.Document#registerPostFixer() docs. Closes #1726.


pjasiun:
We should also add to the view post fixer information that one may break model-view mapping if execute structural change in the view postfixer (what recently happens in the table plugin). Plus examples of what can and what should not be changed in the view postfixer. Plus information that one may call `differ#refreshItem` in view postfixer, if needed.

Also, I am not sure if now my code snippet is fine. It can be not safe.

pjasiun:
Docs are updated.

src/view/document.js
```diff
@@ -109,14 +109,47 @@ export default class Document {
...
-	 * to elements based on the view structure or selection. However, is you need DOM elements to be already updated, use
+	 * to elements based on the view structure or selection. However, if you need DOM elements to be already updated, use
 	 * {@link module:engine/view/view~View#event:render render event}.
 	 *
 	 * As a parameter, a post-fixer callback receives a {@link module:engine/view/downcastwriter~DowncastWriter downcast writer}
 	 * instance connected with the executed changes block.
 	 *
-	 * Note that registering a post-fixer won't re-render the editor's view. If the view should change after registering the post-fixer then
-	 * it should be done manually calling `view.forceRender();`.
+	 * Here is an example which wraps all spaces into spans to add a gray background to all of them:
```
Reinmar: Is this correct? I remember you said recently that view post-fixers should not change the structure? Or does it apply only to containers?

pjasiun: Komentujesz jakąś starą wersję. Tego już tam nie ma.

Reinmar: LOL :D I kept this tab open for just too long :P

src/view/document.js
```diff
@@ -108,15 +108,49 @@ export default class Document {
...
+	 * If you need DOM elements to be already updated, use {@link module:engine/view/view~View#event:render render event}.
 	 *
 	 * As a parameter, a post-fixer callback receives a {@link module:engine/view/downcastwriter~DowncastWriter downcast writer}
 	 * instance connected with the executed changes block.
 	 *
-	 * Note that registering a post-fixer won't re-render the editor's view. If the view should change after registering the post-fixer then
-	 * it should be done manually calling `view.forceRender();`.
+	 * Here is an example from the link plugin which adds highlight class to the link if the selection is in that link:
+	 *
+	 *		editor.editing.view.document.registerPostFixer( writer => {
```
Reinmar: This code snippet is incomplete. What is `findLinkRange()`? And what does `highlightedLinks` do?

Reinmar: It's fine to have some pseudo-code-like example, but it must be clear what it does. It's actually not clear to me how this one works.

pjasiun: I removed `highlightedLinks`, it should not be there.

However, I believe we should not put the code of `findLinkRange` here. It is pretty complicated and would make the whole snippet less readable.

It is not easy to find a good example of the `registerPostFixer` function. Usually, it is part of a bigger picture.

pjasiun: I do not know how I should describe what does `findLinkRange` do. For me, the name tells pretty well what this function does.

Reinmar: I'll try to figure something out regarding `findLinkRange()`. Probably link to its doc.

Reinmar:
This snippet doesn't work. I tried to fix a couple of things (e.g. pass the right params to `findLinkRange()`, check whether the class isn't applied yet, etc). But, in general,  this is all highly incomplete.

If we post a code snippet, it should be logically correct. This one isn't. It misses the logic which controls that the class isn't applied infinitely (which happens right now – it freezes the browser). 

We need a better or more complete example.

pjasiun:
So I think this documentation should not have a snippet since it is not possible to have a one which will be both simple enough (we do not want to have a whole plugin in the snippet) and correct. Unfortunately, it is not that easy to create a good view of post-fixer and maybe it should be kind of a more advanced API.

PR #1736, opened 2019-05-10
Implementing 'is' methods to classes related to view and to model

### LINKED ISSUE #1667: Add is() method to other model/view objects
After https://github.com/ckeditor/ckeditor5-engine/issues/1663 Range, Position, Marker and perhaps something else will still miss this method. Let's add it.

Also, lets remove as many `instanceof` checks from packages other than engine.

One remaining thing to check would be how often we use `instanceof` to distinguish between model and view things. Perhaps we could also have `is( 'view:element' )` and secure our code a bit with this.

msamsel:

Other: Implementing 'is' method to classes related to view and model. Closes #1667 .


src/model/rootelement.js
```diff
@@ -58,10 +58,11 @@ export default class RootElement extends Element {
 	 * @inheritDoc
 	 */
 	is( type, name ) {
+		const cutType = type.replace( 'model:', '' );
```
Reinmar: Missing blank line after var def.

Reinmar: Also, what's "cut type"?

Reinmar:
I'm unsure whether we need this method on classes such as `Batch`, `Model`, `Differ`, `Schema`. I think that the rule of thumb could be that if there's no `instanceof X` yet (anywhere), it's unnecessary.

WDYT, @scofalik @pjasiun?

src/model/rootelement.js
msamsel: It's type without `module:` at the beginning to simplify logic in further lines.
Statement like:
```
( ( type == 'rootElement' || type == 'module:rootElement'  ) && name == this.name ) || super.is( type, name );
```
seems to be a little to complex in my opinion.

msamsel:
> I'm unsure whether we need this method on classes such as Batch, Model, Differ, Schema. I think that the rule of thumb could be that if there's no instanceof X yet (anywhere), it's unnecessary.

@Reinmar, should I remove `is()` method from mentioned classes and put this PR on review again?

pjasiun:
> I'm unsure whether we need this method on classes such as Batch, Model, Differ, Schema. 

It does not make sense to have `is` method there.

scofalik:
As a rule of thumb, I'd say that we need those methods where we need those methods.

Reinmar:
> @Reinmar, should I remove `is()` method from mentioned classes and put this PR on review again?

Can you check which classes are used with `instanceof` and which not? We need this method for objects that we know we that we needed or may need to check the type of.

scofalik:
Basically, stuff like `Differ` or `Schema` aren't similar to other classes and is not returned / expected as a parameter next to (together) with other classes, so we don't need `.is()` for them cause we always know what are we dealing with. There's nothing like `function foo( modelOrDocument )`.

msamsel:
> Can you check which classes are used with instanceof and which not? We need this method for objects that we know we that we needed or may need to check the type of.

I take only set from `engine` as there is no other place where we check for model/view classes. 
```
instanceof Array
instanceof AttributeElement
instanceof CKEditorError
instanceof DocumentFragment
instanceof DocumentSelection
instanceof DropdownButtonView
instanceof EditableElement
instanceof Element
instanceof EventInfo
instanceof FileLoader
instanceof Function
instanceof HTMLElement
instanceof HTMLTextAreaElement
instanceof Marker
instanceof Model
instanceof ModelDocumentFragment
instanceof ModelElement
instanceof ModelPosition
instanceof ModelRange
instanceof ModelSelection
instanceof Node
instanceof NodeList
instanceof Position
instanceof Promise
instanceof ProxyEmitter
instanceof Range
instanceof RegExp
instanceof RootElement
instanceof SchemaContext
instanceof Selection
instanceof SwitchButtonView
instanceof Template
instanceof TemplateBinding
instanceof TemplateIfBinding
instanceof Text
instanceof TextProxy
instanceof View
instanceof ViewCollection
instanceof ViewText
```

Reinmar:
`instanceof TemplateBinding`

This is `ckeditor5-ui`. Could you limit this list to the things from `ckeditor5-engine/src` only? I mean – `instanceof` calls used in code located in `ckeditor5-engine/src` but also `instanceof` checks outside the engine which check stuff from the engine.

Reinmar:
Also, please exclude any code from `src/dev-utils` from this check.

msamsel:
```
instanceof AttributeElement
instanceof DocumentFragment
instanceof DocumentSelection
instanceof EditableElement
instanceof Element
instanceof Marker
instanceof ModelElement
instanceof ModelRange
instanceof ModelSelection
instanceof Node
instanceof NodeList
instanceof Position
instanceof Range
instanceof SchemaContext
instanceof Selection
instanceof Text
instanceof TextProxy
instanceof ViewText
```

scofalik:
I'd guess that `SchemaContext` and `NodeList` are used together with `Array`s. I wonder if we need `.is()` there if that's the case.

msamsel:
@scofalik only those entries exists for those two:
https://github.com/ckeditor/ckeditor5-engine/blob/1fbee48d9ca942f6457c4e6a9a7f701cc002ca3c/src/model/schema.js#L1073
https://github.com/ckeditor/ckeditor5-engine/blob/694ee89033ff69c39210a293ace12c8a93ba6d7e/src/model/writer.js#L195

Reinmar:
Agree with @scofalik .`SchemaContext` and `NodeList` are mostly internal. So we can exclude those.

Reinmar:
So, to conclude:

* all types of view and model nodes,
* selection, doc selection, range, position classes,
* marker,
* text proxy,
* doc frags

And that's all for now.

msamsel:
I've adjusted scope of this PR to cover only mentioned classes.

msamsel:
Rebase to current master to remove merge conflicts.

src/model/element.js
```diff
@@ -108,10 +113,11 @@ export default class Element extends Node {
 	 * @returns {Boolean}
 	 */
 	is( type, name = null ) {
+		const cutType = type.replace( 'model:', '' );
```
Reinmar:
```suggestion
		const cutType = type.replace( 'model:', '' );

```

src/model/liverange.js
```diff
@@ -40,6 +40,13 @@ export default class LiveRange extends Range {
 		this.stopListening();
 	}
 
+	/**
+	 * @inheritDoc
```
Reinmar: You send people to `Range#is()` which documents that:

> Checks whether given object is of `range` type.

Which means that you still don't know how to use this method to check for a live range

src/model/position.js
```diff
@@ -502,6 +502,18 @@ export default class Position {
 		}
 	}
 
+	/**
+	 * Checks whether given object is of `position` type.
+	 *
+	 * Read more at {@link module:engine/model/node~Node#is `Node#is()`}.
```
Reinmar: Sending people to `Node#is()` from `Position` (which isn't a node) seems confusing.

src/view/attributeelement.js
```diff
@@ -128,10 +128,11 @@ export default class AttributeElement extends Element {
 	 * @inheritDoc
 	 */
 	is( type, name = null ) {
+		const cutType = type && type.replace( 'view:', '' );
```
Reinmar:
```suggestion
		const cutType = type && type.replace( /^view:/, '' );
```

src/view/downcastwriter.js
```diff
@@ -788,26 +788,26 @@ export default class DowncastWriter {
...
-     * an instance of {@link module:engine/view/attributeelement~AttributeElement AttributeElement}.
-     *
-     * Throws {@link module:utils/ckeditorerror~CKEditorError} `view-writer-wrap-nonselection-collapsed-range` when passed range
-     * is collapsed and different than view selection.
-     *
-     * @param {module:engine/view/range~Range} range Range to wrap.
-     * @param {module:engine/view/attributeelement~AttributeElement} attribute Attribute element to use as wrapper.
-     * @returns {module:engine/view/range~Range} range Range after wrapping, spanning over wrapping attribute element.
-    */
+	 * Wraps elements within range with provided {@link module:engine/view/attributeelement~AttributeElement AttributeElement}.
```
Reinmar: Why touching this file in this PR?

src/view/range.js
```diff
@@ -394,6 +394,17 @@ export default class Range {
 		}
 	}
 
+	/**
+	 * Checks whether given object is of `range` type following the convention set by
```
Reinmar: This wording is a lot better. I complained in the other place that you link to `Node#is()` without much of an explanation. Saying that "following the convention set by" is a good way out.

src/model/position.js
Reinmar: But see my other comment below about mentioning that this is a convention.

src/view/downcastwriter.js
msamsel: At the first stage when I add `is()` method, my editor fix those spaces by replacing them with tabs. So it sees to be leftover. I'll revert it.

msamsel:
I've rewritten the description. I hope it's better now. :)

Reinmar:
Yes, it's definitely better. But there's now a lot of repetition of things which can be centralised. It's hard to find the place where this method could be described in a bit deeper way, but Node#is sounds acceptable. In all other places, I'd keep the documentation as short as possible, but still very easy to parse thanks to examples. Because of them you don't have to necessarily go to any other place to understand how to use it. But you can if you want more details.

I applied changes to a couple of model objects. Could you update all the other model objects and the view ones accordingly?

PR #1745, opened 2019-06-10
editor.setData() will no longer crash when the data to set contains markers that are already in the editor content

### LINKED ISSUE #1721: DataController#set() throws when adding content with an existing marker
Assume editor data with a marker, for example:

```html
<p>
  A
  <comment id="e64dce95ad23165bb6aca6c5efb543636" type="start"></comment>
  bcd
  <comment id="e64dce95ad23165bb6aca6c5efb543636" type="end"></comment>
  ef
</p>
```

Calling `editor.setData( editor.getData() )` with throw an error. Why?

The whole data is removed, but the marker isn't removed it is actually moved to the graveyard. Then `.setData()` adds the marker with the same name second time which throws an error.

It feels incorrect that under any circumstances `editor.setData( editor.getData() )` fails. Also, I think that it feels natural, that if the specified marker already exists in the editor, it should be moved to a new place. OTOH, we rather discourage using `editor.setData()` as it is rarely correct to do so.

What exactly happens is that `DataController#set()` uses `model.Writer#insert()` and the writer handles the markers insertions. We need to think about two things:

1. Maybe a marker should be automatically removed when it is moved to the graveyard. I already thought about it earlier and now @Reinmar also suggested that. As I think of it, it makes sense for all marker usages I can think of. We already need to write post-fixers which removes those markers after they end up in the graveyard! Mind, that auto-removing will not break undo.

2. If not, we should think whether this issue should be fixed in `model.Writer#insert()` or in `DataController#set()`. We need to think if the writer, on inserting, should handle existing markers or if it is strictly `DataController#set()` scenario and responsibility.

scofalik:

Fix: `editor.setData()` will no longer crash when the data to set contains markers that are already in the editor content. Closes #1721.

### DIFF
--- a/src/model/writer.js
+++ b/src/model/writer.js
@@ -216,7 +216,13 @@ export default class Writer {
 					markerRange.end._getCombined( rangeRootPosition, position )
 				);
 
-				this.addMarker( markerName, { range, usingOperation: true, affectsData: true } );
+				const options = { range, usingOperation: true, affectsData: true };
+
+				if ( this.model.markers.has( markerName ) ) {
+					this.updateMarker( markerName, options );
+				} else {
+					this.addMarker( markerName, options );
+				}
 			}
 		}
 	}
--- a/tests/controller/datacontroller.js
+++ b/tests/controller/datacontroller.js
@@ -336,6 +336,24 @@ describe( 'DataController', () => {
 
 			expect( getData( model, { withoutSelection: true } ) ).to.equal( 'foo' );
 		} );
+
+		// https://github.com/ckeditor/ckeditor5-engine/issues/1721.
+		it( 'should not throw when setting the data with markers that already exist in the editor', () => {
+			schema.extend( '$text', { allowIn: '$root' } );
+
+			data.set( 'foo' );
+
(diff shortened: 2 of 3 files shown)


scofalik:
Go home @coveralls, you are drunk.

src/controller/datacontroller.js
```diff
@@ -328,7 +328,14 @@ export default class DataController {
 				const modelRoot = this.model.document.getRoot( rootName );
 
 				writer.remove( writer.createRangeIn( modelRoot ) );
-				writer.insert( this.parse( newData[ rootName ], modelRoot ), modelRoot, 0 );
+
+				const modelDocumentFragment = this.parse( newData[ rootName ], modelRoot );
+
+				for ( const markerName of modelDocumentFragment.markers.keys() ) {
+					this.model.markers._remove( markerName );
```
pjasiun: Why do you use the protected method instead of public `writer.removeMarker`?

pjasiun:
Also, I think that this fix should land inside `writer.insert`.

scofalik:
Fixed

Issue #1017, opened 2017-07-17
Improve placeholder behaviour when ui element is present


scofalik:
At the moment, placeholder is removed when there is a `view.UIElement` present in the element that has the placeholder. This is not really correct, as `UIElement` is not data. Most of time it will be some tiny graphical indicator like a dot, arrow, line, etc. Placeholder should be preserved in such situation.

Reinmar:
Should it? Where will be those UIElements rendered then? My guess is on the right hand side of the placeholder because it's rendered with `::before` style.

Couldn't you hide you UIElements using CSS when content has the `.ck-placeholder` class?

scofalik:
Yes, of course, we have to change `::before` to `::after`.

Reinmar:
OK, I get it now. You wouldn't be able to hide those markers anyway because the `.ck-placeholder` class is removed in the first place.

But once it will stop being removed if there are any markers in the view I'd propose to hide them anyway.

scofalik:
This is more difficult than I initially thought.

Things to consider:

1) If we want to correctly display `UIElement`s, placeholder has to be applied on `::after`.
2) But then, bogus br cannot be used as a filler element, because it moves the placeholder to the next line.
3) We can set `view.ContainerElement#getFillerOffset()` to return `null` if it has `data-placeholder` attribute.
4) Then, inline filler is created instead of bogus br.
5) Then, after clicking a placeholder, inline filler is created and then selection is moved to the beginning of element, that is before an inline filler (which is an error).
6) This means that we need to dig deeper, in selection rendering mechanism. We have this whole "similar selection" mechanism which prevents re-rendering selection when DOM and view selections are similar. Unfortunately this causes a bug in this case, cause we would like to reposition selection after the inline filler. I changed `view.Renderer#_updateDomSelection` to do it, but then new bugs appeared. And also tests started to fail.

I also tried another route with:
```css
.ck-placeholder [data-cke-filler] {
	display: none;
}
```
But it also didn't seem like a good solution.

scofalik:
We can take easy path with what you suggested, @Reinmar, but it is not really correct. I can imagine a need to use (and display) UI element in an empty element (for example some kind of button).

jodator:
> Most of time it will be some tiny graphical indicator like a dot, arrow, line, etc. Placeholder should be preserved in such situation.

@scofalik & @Reinmar: and the first non-empty element with UI-only child is the Media Embed widget.

[...]

PR #1749, opened 2019-06-19
Fix placeholder over media embed.

jodator:

Other: Emptiness check for placeholder containing elements will check number of its children. Closes ckeditor/ckeditor5#1684.

BREAKING CHANGE: The placeholder no longer considers elements with only `UIElement` children as empty for displaying the placeholder. 

---

### Additional information

I've been fiddling with the placeholder for some time to check what caused the ckeditor/ckeditor5#1684 and it turned out that the placeholder was enabled for "emptish" elements. The "emptish" element is such that has no children or all of the children are UI Elements. 

I'm not sure why this was created this way (maybe some collaboration features? but I doubt it) but this check would give false positive for widget with only `UIElement` children - this is what happens with MediaEmbed.

So taking this into consideration I don't see that `UIElement` should dictate that element is empty in regards for showing a placeholder.

The other change that we might consider is to tie emptiness check to the model. I didn't go this route as currently the placeholder operates on the view solely.

ps.: The isWidget check was also discarded.


oleq:
The `isEmptyish` check has been introduced by @scofalik  in https://github.com/ckeditor/ckeditor5-engine/commit/6a66f6e5e354792e197227bf88f3c8c1b6d692e8 and solved https://github.com/ckeditor/ckeditor5-engine/issues/1018 which, in turn, was a piece of https://github.com/ckeditor/ckeditor5-engine/issues/1017. Sounds like there was a solid reason for that check (something about markers, which makes sense on second thought). 

Can you investigate that? Because it looks like your PR could break some core functionality.

jodator:
> We cannot merge this without the research as pointed out in [#1749 (comment)](https://github.com/ckeditor/ckeditor5-engine/pull/1749#issuecomment-504986200).

Thanks bisect & git log master :D I feel so bad for not digging it properly :( - you're right the check should be better then. I'll check on this and I'll update the PR.

jodator:
@scofalik & @oleq:
OK so after some fiddling I can see that this check is done for markers case mostly. The placeholder must be visible when there are markers in empty paragraph or empty caption.

The solution for this could done by:
1. CSS override for media embed - do not show placeholder there.
2. Check if element is a widget and consider it always not empty either by:
    a. `isWidget()` util
    b. checking custom property `element.getCustomProperty( 'widget' )`

As the markers are hidden in the UI similarly (by CSS) with a placeholder active I think that we might go with CSS solution. But OTOH such solution is not clean enough - it will add the `ck-placeholder` class to the widget which is not totally OK. So I'd rather go with `2b` solution as the engine should not import `widget` repository.

oleq:
I'm for 2b. 2a would add a `ckeditor5-widget` dependency to `ckeditor5-engine` which we don't need (want?).

src/view/placeholder.js
```diff
@@ -153,6 +153,12 @@ export function needsPlaceholder( element ) {
 		return false;
 	}
 
+	// If the element is a Widget, always consider it a non-empty and thus not needing a placeholder.
+	// https://github.com/ckeditor/ckeditor5/issues/1684
+	if ( element.getCustomProperty( 'widget' ) ) {
```
oleq: On the second thought, that would be the only place in the codebase other than `isWidget()` from `ckeditor5-widget` that calls `getCustomProperty( 'widget' )` 😛

Maybe we could use `ckeditor5-widget` as an engine dependency after all. I'm not sure about consequences, though. cc @Reinmar.

jodator:
After F2F talk with @Reinmar I'll updat this PR with checks in the schema so the `needsPlaceholder()` method will also get `editingController`.

jodator:
@Reinmar I'm working on the rewrite so the `needsPlaceholder()` function would check model, not the view. 

This looks like a bigger change because the `enablePlaceholder()` function and some other functions in this feature must also accept the `editing` instead of the `view`.

 It is not a big issue in most places but it would require some more work at least in the `ImageCaptionEditing` because in the downcast conversion it first enables the placeholder and as the next step it binds view and model elements so the `needsPlacesholder()` fails on checking the schema.

TL;DR: I don't think we can squish the rewrite in the 25 milestone. My suggestion is that we could use the current state (checking for the widget) in 25 and rewrite it in 26.

src/view/placeholder.js
oleq: TBH I have nothing against the current state (`element.getCustomProperty( 'widget' )`) and some improvements in i26 as long as we don't change dependencies in i25.

scofalik:
To be honest, I think that the problem is using `UIElement` where we shouldn't. Media embed element is not a `UIElement`. We use `UIElement` only because we don't have a better option as far as view element types go. So maybe we should go that way?

Reinmar:
WDYM? I remember you were proposing introducing new element types. Would that (IIRC) `RawElement` help here?

scofalik:
I am not sure why exactly `UIElement` was used here because I haven't been developing the feature so I don't know what problems `EmptyElement` or `ContainerElement` had. All I am saying is that using `UIElement` for non-UI (read: content meaningful) elements is not the indented use case for `UIElement`. So saying that ["old placeholder solution is wrong because `UIElement` is used in media embed"](https://github.com/ckeditor/ckeditor5-engine/issues/1017#issuecomment-505029372) might be approaching the problem from an incorrect side. Same with introducing additional checks or extending schema. Although this is something I like because it might be useful for plugins creators.

Reinmar:
> I am not sure why exactly `UIElement` was used here because I haven't been developing the feature so I don't know what problems `EmptyElement` or `ContainerElement` had.

We need custom rendering, without going through our rendering engine.

jodator:
Guys, so the `RawElement` would play nice here (https://github.com/ckeditor/ckeditor5-engine/issues/1643). Am I right? From what I can see it would act similarly as `UIElement` - it should allow creating custom HTML code inside base elment - just like `UIElement` does but from model POV it will act like any other element.

scofalik:
I think so.

Reinmar:
Yep, I think so too. I wasn't sure initially about the semantics of this type of element and the difference between UI and Raw, but I think https://github.com/ckeditor/ckeditor5-engine/issues/1643#issuecomment-514653753 makes it clear and in that case, it would solve this case.

PR #1760, opened 2019-07-10
Removed usage of logger.

ma2ciek:

Internal: Removed usage of the CKEditor 5 logger. Part of ckeditor/ckeditor5#383.

---

### Additional information


src/conversion/downcasthelpers.js
```diff
@@ -824,8 +826,10 @@ function changeAttribute( attributeCreator ) {
 			 *
 			 * @error conversion-attribute-to-attribute-on-text
 			 */
-			log.warn( 'conversion-attribute-to-attribute-on-text: ' +
-				'Trying to convert text node\'s attribute with attribute-to-attribute converter.' );
+			console.warn( attachLinkToDocumentation(
```
pjasiun: This should be `CKEditorError` exception.

src/model/utils/insertcontent.js
```diff
@@ -322,8 +318,8 @@ class Insertion {
 		if ( !this.schema.checkChild( this.position, node ) ) {
 			// Algorithm's correctness check. We should never end up here but it's good to know that we did.
 			// Note that it would often be a silent issue if we insert node in a place where it's not allowed.
-			log.error(
-				'insertcontent-wrong-position: The node cannot be inserted on the given position.',
+			console.error(
+				attachLinkToDocumentation( 'insertcontent-wrong-position: The node cannot be inserted on the given position.' ),
```
pjasiun: This should be `CKEditorError` exception.

src/model/utils/insertcontent.js
```diff
@@ -442,7 +438,7 @@ class Insertion {
 				// Algorithm's correctness check. We should never end up here but it's good to know that we did.
 				// At this point the insertion position should be after the node we'll merge. If it isn't,
 				// it should need to be secured as in the left merge case.
-				log.error( 'insertcontent-wrong-position-on-merge: The insertion position should equal the merge position' );
```
pjasiun: This should be `CKEditorError` exception.

src/model/utils/insertcontent.js
```diff
@@ -442,7 +438,7 @@ class Insertion {
...
+				// @if CK_DEBUG // console.error( 'The insertion position should equal the merge position' );
```
pjasiun: This should be `CKEditorError` exception.

src/view/view.js
```diff
@@ -405,7 +406,9 @@ export default class View {
 				 *
 				 * @error view-focus-no-selection
 				 */
-				log.warn( 'view-focus-no-selection: There is no selection in any editable to focus.' );
+				console.warn( attachLinkToDocumentation(
```
pjasiun: This should be `// @if CK_DEBUG //`

src/conversion/downcasthelpers.js
ma2ciek: OTH it doesn't break the editor.

pjasiun: But this means that someone used our API in an incorrect way. I think there are 2 ways:  we agreed that using an attribute to attribute converter on text is fine and do not warn about it, or we agreed that it is not fine, and does not allow people to use it (throw `CKEditorError`). It does not make much sense to be in between.

pjasiun: I think we should throw now. I will ping @scofalik, if he will know about some cases when it is correct, we will change it later.

src/conversion/downcasthelpers.js
```diff
@@ -3,6 +3,14 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
+/**
+ * Contains downcast (model-to-view) converters for {@link module:engine/conversion/downcastdispatcher~DowncastDispatcher}.
+ *
+ * @module engine/conversion/downcasthelpers
+ */
+
+/* globals */
```
pjasiun: Is it needed?

scofalik:
```suggestion
```

pjasiun:
```suggestion
```

src/conversion/downcasthelpers.js
```diff
@@ -3,6 +3,13 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
+/**
+ * Contains downcast (model-to-view) converters for {@link module:engine/conversion/downcastdispatcher~DowncastDispatcher}.
+ *
+ * @module engine/conversion/downcasthelpers
+ */
+
+
```
scofalik:
```suggestion
```

PR #1768, opened 2019-07-29
Return original operations map in transformSets

### LINKED ISSUE #1767: Return original operations map in transformSets
In `engine/model/operation/transform~transformSets`, at the moment, we return two sets of the transformed operations. The problem is that the returned operations aren't linked in any way with the input operations. They are different instances, they could be transformed to different operations or `NoOperation`s, they could be broken into two or more operations, etc.

This means that if we have some information associated with the input operations, it will be lost.

To prevent that, it would be useful if `transformSets` would also return a map, that would link the new operations to the old ones.

Check the example:

```js
// Assume that there is some information stored in `myDataSet`, where keys
// are operations from `setA` and `setB`.
const result = transformSets( setA, setB, options );

for ( const operation of result.operationsA ) {
    const originalOperation = result.originalOperation.get( operation );

    // Now you can access data stored for operation from `setA` through
    // transformed operations `operationsA`.
    myDataSet.get( originalOperation );
}
```

elszczepano:

Feature: Return original operations map in transformSets. Closes #1767 .

---

### Additional information

Related tickets: 
https://github.com/cksource/cloud-features/issues/2753
https://github.com/cksource/ckeditor5-real-time-collaboration/pull/363

### DIFF
--- a/src/model/operation/transform.js
+++ b/src/model/operation/transform.js
@@ -390,6 +390,13 @@ class ContextFactory {
 	// @param {Boolean} [forceWeakRemove=false] If set to `false`, remove operation will be always stronger than move operation,
 	// so the removed nodes won't end up back in the document root. When set to `true`, context data will be used.
 	constructor( document, useRelations, forceWeakRemove = false ) {
+		// For each operation that is created during transformation process, we keep a reference to the original operation
+		// which it comes from. The original operation works as a kind of "identifier". Every contextual information
+		// gathered during transformation that we want to save for given operation, is actually saved for the original operation.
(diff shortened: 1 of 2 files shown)


src/model/operation/transform.js
```diff
@@ -401,7 +403,7 @@ class ContextFactory {
 		// gathered during transformation that we want to save for given operation, is actually saved for the original operation.
 		// This way no matter if operation `a` is cloned, then transformed, even breaks, we still have access to the previously
 		// gathered data through original operation reference.
-		this._originalOperations = new Map();
```
ma2ciek: Public properties should be moved above the private ones.

elszczepano: Done.

PR #1770, opened 2019-07-31
T/ckeditor5 widget/95: Remove the getTopMostBlocks() method

jodator:

Other: Remove the `selection.getTopMostBlocks()`. Closes ckeditor/ckeditor5-widget#95. Closes ckeditor/ckeditor5-table#199.

BREAKING CHANGE: The `selection.getTopMostBlocks()` was removed from the public API. Use `selection.getSelectedBlocks()` instead.

BREAKING CHANGE: The `selection.getSelectedBlocks()` does not return blocks nested in other blocks now.

---

### Additional information

Changes in other repositories:
- indent: https://github.com/ckeditor/ckeditor5-indent/pull/17
- block-quote: https://github.com/ckeditor/ckeditor5-block-quote/pull/39

* After some tests & digging I came to mind that `getSelectedBlocks()` should work as `getTopMostBlock()` so it does not return blocks nested in other blocks but it will search for blocks if they are nested in other non-block elements.

* most of the tests were refactored from the old ones  - I've used the `element#contents` notation in order to properly check the cases.

I've requested you both SC & PK for reviews since I'd like to have double feedback :)

### DIFF
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -233,14 +233,14 @@ export default class DocumentSelection {
 	}
 
 	/**
-	 * Gets elements of type "block" touched by the selection.
+	 * Gets elements of type {@link module:engine/model/schema~Schema#isBlock "block"} touched by the selection.
 	 *
 	 * This method's result can be used for example to apply block styling to all blocks covered by this selection.
 	 *
 	 * **Note:** `getSelectedBlocks()` returns blocks that are nested in other non-block elements
 	 * but will not return blocks nested in other blocks.
 	 *
-	 * In this case the function will return exactly all 3 paragraphs:
+	 * In this case the function will return exactly all 3 paragraphs (note: `<blockQuote>` is not a block itself):
 	 *
 	 *		<paragraph>[a</paragraph>
 	 *		<blockQuote>
@@ -252,7 +252,7 @@ export default class DocumentSelection {
 	 *
 	 *		<paragraph>[]a</paragraph>
 	 *
-	 *	In such scenario however, only blocks A, B & E will be returned as blocks C & D are nested in block B:
+	 * In such a scenario, however, only blocks A, B & E will be returned as blocks C & D are nested in block B:
 	 *
 	 *		[<blockA></blockA>
 	 *		<blockB>
@@ -261,7 +261,7 @@ export default class DocumentSelection {
 	 *		</blockB>
 	 *		<blockE></blockE>]
 	 *
-	 *	If the selection is inside a block all the inner blocks (A & B) are returned:
+	 * If the selection is inside a block all the inner blocks (A & B) are returned:
 	 *
 	 * 		<block>
 	 *			<blockA>[a</blockA>
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -233,9 +233,26 @@ export default class DocumentSelection {
 	}
 
 	/**
-	 * Returns blocks that aren't nested in other selected blocks.
+	 * Gets elements of type "block" touched by the selection.
 	 *
-	 * In this case the method will return blocks A, B and E because C & D are children of block B:
+	 * This method's result can be used for example to apply block styling to all blocks covered by this selection.
+	 *
+	 * **Note:** `getSelectedBlocks()` returns blocks that are nested in other non-block elements
+	 * but will not return blocks nested in other blocks.
+	 *
+	 * In this case the function will return exactly all 3 paragraphs:
+	 *
+	 *		<paragraph>[a</paragraph>
+	 *		<quote>
+	 *			<paragraph>b</paragraph>
+	 *		</quote>
+	 *		<paragraph>c]d</paragraph>
+	 *
+	 * In this case the paragraph will also be returned, despite the collapsed selection:
+	 *
+	 *		<paragraph>[]a</paragraph>
+	 *
+	 *	In such scenario however, only blocks A, B & E will be returned:
 	 *
 	 *		[<blockA></blockA>
 	 *		<blockB>
@@ -244,7 +261,12 @@ export default class DocumentSelection {
 	 *		</blockB>
 	 *		<blockE></blockE>]
 	 *
-	 * **Note:** To get all selected blocks use {@link #getSelectedBlocks `getSelectedBlocks()`}.
+	 * **Special case**: If a selection ends at the beginning of a block, that block is not returned as from user perspective
+	 * this block wasn't selected. See [#984](https://github.com/ckeditor/ckeditor5-engine/issues/984) for more details.
+	 *
+	 *		<paragraph>[a</paragraph>
+	 *		<paragraph>b</paragraph>
+	 *		<paragraph>]c</paragraph> // this block will not be returned
 	 *
 	 * @returns {Iterable.<module:engine/model/element~Element>}
 	 */
--- a/src/model/documentselection.js
+++ b/src/model/documentselection.js
@@ -243,16 +243,16 @@ export default class DocumentSelection {
 	 * In this case the function will return exactly all 3 paragraphs:
 	 *
 	 *		<paragraph>[a</paragraph>
-	 *		<quote>
+	 *		<blockQuote>
 	 *			<paragraph>b</paragraph>
-	 *		</quote>
+	 *		</blockQuote>
 	 *		<paragraph>c]d</paragraph>
 	 *
 	 * In this case the paragraph will also be returned, despite the collapsed selection:
 	 *
 	 *		<paragraph>[]a</paragraph>
 	 *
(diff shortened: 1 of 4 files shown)


jodator:
Oh.. I've check what is wrong with the build.

jodator:
OK.. the block-quote have to be checked-out also :/ Tests run locally just fine.

src/model/documentselection.js
```diff
@@ -237,7 +237,8 @@ export default class DocumentSelection {
 	 *
 	 * This method's result can be used for example to apply block styling to all blocks covered by this selection.
 	 *
-	 * **Note:** `getSelectedBlocks()` always returns the deepest block.
+	 * **Note:** `getSelectedBlocks()` returns blocks that are nested in other non-block elements
+	 * but will not return blocks nested in other blocks.
 	 *
 	 * In this case the function will return exactly all 3 paragraphs:
```
scofalik: Could you please change `quote` in the below example to `blockQuote`? AFAIR the element in the model is called`blockQuote`.

src/model/documentselection.js
```diff
@@ -276,12 +261,17 @@ export default class DocumentSelection {
...
+	 *
+	 *		<paragraph>[a</paragraph>
+	 *		<paragraph>b</paragraph>
+	 *		<paragraph>]c</paragraph> // this block will not be returned
 	 *
 	 * @returns {Iterable.<module:engine/model/element~Element>}
 	 */
-	getTopMostBlocks() {
-		return this._selection.getTopMostBlocks();
+	getSelectedBlocks( config ) {
```
scofalik: Mmmm, do we need `config` here? It is not described in docs (or I can't see it).

jodator: Good catch - it's a left-over. I didn't implement the `config.deep` after checking how blockquote is treated with blocks.

scofalik: Okay, AFAICS `Selection#getSelectedBlocks` does not take any arguments so this is probably an artifact after the previous solution.

src/model/selection.js
```diff
@@ -673,54 +683,27 @@ export default class Selection {
 		for ( const range of this.getRanges() ) {
 			const startBlock = getParentBlock( range.start, visited );
 
-			if ( startBlock ) {
+			if ( startBlock && isTopBlockInRange( startBlock, range ) ) {
```
scofalik: Isn't this also covered by the `for` below?

src/model/selection.js
```diff
@@ -673,54 +683,27 @@ export default class Selection {
...
 				yield startBlock;
 			}
 
 			for ( const value of range.getWalker() ) {
```
scofalik: As I understand the code below, what happens is that you, at some point, get a block element. But then, the tree walker steps into it and you traverse through more block elements and they might not be the topmost blocks. But you already visited that one, so everything inside can be skipped. This part of code can probably be optimized but the gain wouldn't be big so I won't be pushing on this.

scofalik: Probably the biggest efficiency loss will happen when table is selected. There are a lot of elements in tables and you will have to go through all of them, for nothing.

If you'd like to optimize this, you'd need to change `getWalker()` and use actual `TreeWalker` instance and then use `.skip()` method on it (which is a little difficult to use, so check existing code for examples).

As I said, I won't force it, but it would be nice if the algorithm was a little smarter :P

src/model/documentselection.js
```diff
@@ -276,12 +261,17 @@ export default class DocumentSelection {
 	 *		</blockB>
 	 *		<blockE></blockE>]
 	 *
-	 * **Note:** To get all selected blocks use {@link #getSelectedBlocks `getSelectedBlocks()`}.
+	 * **Special case**: If a selection ends at the beginning of a block, that block is not returned as from user perspective
+	 * this block wasn't selected. See [#984](https://github.com/ckeditor/ckeditor5-engine/issues/984) for more details.
+	 *
+	 *		<paragraph>[a</paragraph>
+	 *		<paragraph>b</paragraph>
+	 *		<paragraph>]c</paragraph> // this block will not be returned
```
scofalik: If I understood the code correctly, in following example:

```
<block>
  <blockA>[a</blockA>
  <blockB>b]</blockB>
</block>
```

Blocks A and B will be returned? It makes obvious sense, so I'd add it to examples. However what is returned in this scenario:

```
<blockX>
  <blockA>[a</blockA>
  <blockB>b</blockB>
</blockX>
<blockY>y]</blockY>
```

A, B, Y?

src/model/selection.js
jodator: I have the same thought as you (if I read this correctly) that for a table structure the `getSelectedBlocks()` should return: `table, p, p...` but in fact it returns `p, p, table` (children first). So when traversing the tree with a tree walker I don't have that information yet and I must extract it from somewhere).

Previously the `getTopMostBlocks()` would just get all the blocks and then yield only that that aren't inside other blocks which wasn't super great.

jodator: So no, I havn't visited `table` while visiting paragraph inside table cell.

src/model/documentselection.js
jodator: probably the `a,b,y` but I haven't tested this case. At the moment we do not have such structures. Taking tables as an example of a block that nests other blocks - the `blockX` would be a limit also so that selection wouldn't be possible and be fixed to `[<blockX>...y]</blockY>`

src/model/selection.js
scofalik: Oh, because you check `elementEnd`, not `elementStart`.

scofalik: Still, the same applies. When you have `<p>Fo[o</p><table>...</table><p>B]ar</p>` selected blocks are `p`, `table`, `p`. So there is no sense to traverse things in the table. But as I said this probably is a minor thing. If you don't want to tinker with it, we can fix it if it proves to be a problem for big tables.

scofalik:
There's one thing that should be noted about the changes.

Previously, when you made selection like this: `<p>Fo[o</p><table>...</table><p>B]ar</p>` all the paragraphs in cells would be returned too. This means that, if you, for example, set a heading with such selection, all paragraphs in the table would become headings.

Now, it won't happen because always the top blocks are returned. `<table>` in this case.

I think that's actually for good. I didn't like the previous behavior even though you could argue that it is technically correct (those cells were selected, sooo... 🤷‍♂they got changed to heading 🤷‍♂).

However, this can be discussed if we are not missing something with this change of behavior. Other than that R+ from me.

src/model/selection.js
jodator: OK - we listen there for `elementEnd` only. Maybe because of something of that linked bug in endElement. So we might listen to `elementStart` there and the order will be as expected?

Right now it is `p, p in table, p in table, table, p` because of examing this on `elementEnd` not the `elementStart`. I'll check it briefly.

src/model/selection.js
jodator: OK, I see why this construct was - to cover collapsed selection and multi-range collapsed selection also (by coincidence I think).

src/model/selection.js
jodator: The checks are done on `elementEnd` AFAICS to get only fully selected blocks and to properly check the case of #984 ticket.

jodator:
@Reinmar so the main work is done. I have only one doubt with @scofalik comments on this case:

> If I understood the code correctly, in following example:
> 
> ```
> <block>
>   <blockA>[a</blockA>
>   <blockB>b]</blockB>
> </block>
> ```
> 
> Blocks A and B will be returned? It makes obvious sense, so I'd add it to examples. However what is returned in this scenario:
> 
> ```
> <blockX>
>   <blockA>[a</blockA>
>   <blockB>b</blockB>
> </blockX>
> <blockY>y]</blockY>
> ```
> 
> A, B, Y?

As at this moment it will return `A, B, Y` and also:
```
<blockX>
   <blockA>[a</blockA>
   <blockB>b</blockB>
</blockX>
<blockY>
   <blockC>c]<blockC>
</blockY>
 ```
it will return A, B, C (so will omit the X & Y).

Now I have no opinion on that - is it OK or not? We do not encurage such constructs but only the `isLimit` or `isObject` elements would cause to post-fix such selection (like in tables). But nothing prevents bare blocks to be seleted like this AFAIR.

Sooo maybe the `getSelectedBlocks()` should check for `isLimit` rather then `block` in `block` case? (I' almost certain that this is how this should work and not the way I've implemented this ATM).

src/model/documentselection.js
jodator: I've moved the conversion on this below.

Reinmar:
I'm not 100% sure what should be the expected behaviour here regarding nested blocks. Since it's so hard to understand this, let's simply wait for use cases.

PR #1772, opened 2019-08-02
Add unwrapElement method to UpcastWriter

msamsel:

Other: Add unwrapElement method to UpcastWriter.

---

### Additional information

PR extracted from ckeditor/ckeditor5-paste-from-office#69`

### DIFF
--- a/src/view/upcastwriter.js
+++ b/src/view/upcastwriter.js
@@ -173,6 +173,27 @@ export default class UpcastWriter {
 	}
 
 	/**
+	 * Removes given element from view structure and places its children in its position.
+	 *
+	 * @param {module:engine/view/element~Element} element Element which will be unwrapped
+	 * @returns {Booolean} Whether element was successfully unwrapped.
+	 */
+	unwrapElement( element ) {
+		const parent = element.parent;
+
+		if ( parent ) {
+			const index = parent.getChildIndex( element );
+
+			this.remove( element );
+			this.insertChild( index, element.getChildren(), parent );
+
+			return true;
+		}
+
+		return false;
+	}
+
+	/**
 	 * Renames element by creating a copy of a given element but with its name changed and then moving contents of the
 	 * old element to the new one.
 	 *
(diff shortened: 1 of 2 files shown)


src/view/upcastwriter.js
```diff
@@ -172,6 +172,27 @@ export default class UpcastWriter {
...
+	unwrapElement( element ) {
+		const parent = element.parent;
+
+		if ( parent ) {
+			const index = parent.getChildIndex( element );
+
+			this.remove( element );
+			this.insertChild( index, element.getChildren(), parent );
+
+			return true;
```
jodator: The method should not return true/false - we don't need that.

Check the model's `writer.unwrap()` how we handle this: https://github.com/ckeditor/ckeditor5-engine/blob/ad9159ca794096cbed1ab44180df0c74181d2f31/src/model/writer.js#L851-L861

jodator:
After some clarification from @msamsel I see that some methods of upcast writer return true/false in similar scenarios but TBH I don't see why. So I'm for not returning anything but adding a `if( element.parent )` check to not perform unwrap on detached elements.

src/view/upcastwriter.js
```diff
@@ -172,6 +172,24 @@ export default class UpcastWriter {
...
 	}
 
+	/**
+	 * Removes given element from view structure and places its children in its position.
+	 * It does nothing if element has no parent.
+	 *
+	 * @param {module:engine/view/element~Element} element Element which will be unwrapped
+	 * @returns {Booolean} Whether element was successfully unwrapped.
+	 */
+	unwrapElement( element ) {
```
oskarwrobel: Just for my curiosity. Why not just `unwrap`? Other methods from this class (like `remove`, `replace`, `rename`) don't have `Element` suffix.

msamsel: I've followed byt this explanaition:
https://github.com/ckeditor/ckeditor5-paste-from-office/pull/72#discussion_r310022092

PR #1774, opened 2019-08-06
Docs: Added example to clarify the case explained in the ticket.

### LINKED ISSUE #1773: Upcast converter for element:* gets data.modelRange == null
When talking about upcasting, if you try to convert an element's attribute you need to listen to `element:*` event (there's no dedicated event for attribute conversion). Unfortunately, if you'd try this way:

```js
editor.conversion.for( 'upcast' ).add( dispatcher => {
	dispatcher.on( 'element:p', ( evt, data, conversionApi ) => {
		console.log( data.modelRange );
		console.log( Array.from( data.modelCursor.root.getChildren() ) );
	} );
} );
```

You can get `data.modelRange == null` and the model tree to which you convert (second `console.log()`) may not have the element that's being converted yet. In this specific case above this will happen if you add this converter before the `Paragraph` and `Heading` plugins are loaded. That's because in this scenario your listener will be executed before the `Paragraph`'s element converter.

The solution is simple and is actually what all our converters added e.g. with `attributeToAttribute()` do – they use `priority: 'low'`:

```js
editor.conversion.for( 'upcast' ).add( dispatcher => {
	dispatcher.on( 'element:p', ( evt, data, conversionApi ) => {
		console.log( data.modelRange );
		console.log( Array.from( data.modelCursor.root.getChildren() ) );
	}, { priority: 'low' } );
} );
```

This way, no matter how early you'll add your callback, it will always be executed after the default priority element converter from e.g. `Paragraph` feature.

The above needs to be documented.

Reinmar:

Docs: Added an example to clarify how to implement custom attribute upcast converters. Closes #1773.

---

### Additional information

I also fixed the documentation for default priorities.

### DIFF
--- a/src/conversion/upcastdispatcher.js
+++ b/src/conversion/upcastdispatcher.js
@@ -55,6 +55,27 @@ import mix from '@ckeditor/ckeditor5-utils/src/mix';
  *			}
  *		} );
  *
+ *		// Convert <p>'s font-size style.
+ *		// Note: You should use a low-priority observer in order to ensure that
+ *		// it's executed after the element-to-element converter.
+ *		editor.data.upcastDispatcher.on( 'element', ( evt, data, conversionApi ) => {
+ *			const { consumable, schema, writer } = conversionApi;
+ *
+ *			if ( !consumable.consume( data.viewItem, { style: 'font-size' } ) ) {
+ *				return;
+ *			}
+ *
+ *			const fontSize = data.viewItem.getStyle( 'font-size' );
+ *
+ *			// Don't go for the model element after data.modelCursor because it might happen
+ *			// that a single view element was converted to multiple model elements. Get all of them.
+ *			for ( const item of data.modelRange.getItems( { shallow: true } ) ) {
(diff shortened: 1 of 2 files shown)


src/conversion/upcastdispatcher.js
```diff
@@ -55,6 +55,27 @@ import mix from '@ckeditor/ckeditor5-utils/src/mix';
  *			}
  *		} );
  *
+ *		// Convert <p>'s font-size style.
+ *		// Note: You should use a low-priority observer in order to ensure that
+ *		// it's executed after the element-to-element converter.
+ *		editor.data.upcastDispatcher.on( 'element', ( evt, data, conversionApi ) => {
```
scofalik: `element` or `element:p`? Comment says `<p>`.

Reinmar: Yup, my bad. I've been testing too many things.

Reinmar:
```suggestion
 *		editor.data.upcastDispatcher.on( 'element:p', ( evt, data, conversionApi ) => {
```

PR #1777, opened 2019-08-07
Added validation for Position's getters

### LINKED ISSUE #1776: Validate position.parent's output
Positions and the tree to which they point are independent. It is possible to create and work with positions which point to non-existent places. This is mainly required by OT which works on a virtual states.

However, there's a set of position's properties/methods which do access the model structure. This is mainly done via reading `position.parent` (which is a getter) which, if the position is incorrect will either crash or return a text node. The first scenario is fairly easy to track and understand. The latter is harder because it fails later, when a text node reaches a code which expects an element.

Therefore, I propose to add validation for positions:

* for `position.parent` that the returned object is really an element and not a text node,
* (optionally) for all incorrect cases in `position.parent` – e.g. when you have offsets out of boundaries
* (optionally) for other position's props/methods which don't use `position.parent` and may throw anyway.

cc @scofalik @pjasiun

Reinmar:

Other: Position getters (such as `#parent` or `#index`) will throw when position points at an incorrect place in its root. Closes #1776.

### DIFF
--- a/src/model/operation/mergeoperation.js
+++ b/src/model/operation/mergeoperation.js
@@ -141,16 +141,16 @@ export default class MergeOperation extends Operation {
 		const targetElement = this.targetPosition.parent;
 
 		// Validate whether merge operation has correct parameters.
-		if ( !sourceElement || !sourceElement.is( 'element' ) || !sourceElement.parent ) {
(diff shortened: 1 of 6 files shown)


Reinmar:
Gotta fix CC.

PR #1779, opened 2019-08-08
T/404: Better handling of &nbsp; block filler.

### LINKED ISSUE #404: Only real block fillers should be removed on DOM to view conversion
For the data pipeline text node with `&nbsp;` is a block filler.

When user call set data them DOM data:

```
<p>&nbsp;</p>
```

Is converted to view:

```
<p></p>
```

What is fine.

But, unfortunately, all `&nbsp;` text node will be threaded as block fillers and ignored so:

```
<p>foo<b>&nbsp;</b></p>
```

Will be converted to:

```
<p>foo<b></b></p>
```

What is not fine anymore.

We should remove fillers only on positions which need fillers. But that is easier said than done. During the DOM to view conversion we do not have elements types, we do not know that `<p>` is a block element and `<b>` is not. We do not even have any DTD to check the type of the element to guess if it is the filler or not.

We could introduce some hardcoded (or extensible) DTD for such cases. Alternatively, if we will meet more problems like this, we could have DTD build by plugins and introduce types of all elements converted from DOM, but then we expect plugins to define one more think.

jodator:

Fix: Remove only real block fillers on DOM to view conversion. Closes #404. 

---

### Additional information

ATM this PR fixes all the issues addressed with handling `&nbsp;` in the data. I'm not however 100% satisfied with this solution because of a need to handle `<br data-cke-filler="true">` in editing differently than `&nbsp;` in data pipeline.

For editing pipeline the `isBlockFiller()` is pretty straight forward as we check `node.isEqualNode()` wchich compares two DOM nodes by theirs attributes. However for `&nbsp;` we compare `Text` elements by string equality. In `&nbsp;` case we need to detect if the block filler belongs to data or is it negligible in given context.

The other solution that came to my mind is to allow passing a `isBlockFiller()` check to  `DOMConverter` along with `blockFiller` option thus explicitly changing the `DOMConverter` for editing and data pipelines. The drawback is that you have to remember to pass this method when not using deafult block filler (`<br>`).

* After some testing (including PFO compatibility issues) I think that the current solution is OK.
* I've also fixed some tests that were incorrectly implemented (didn't test what they claimed in the description)
- I've hardcoded a short list of inline elements that are recognized in `<b>&npbsp</b>` and `<b>foo</b>&nbsp;<i>bar</i>` cases but this is to be extended before merging.

ps.: ~Some tests in PFO are failing - I'll be investigating this.~

I've added a sub-pr for PFO: https://github.com/ckeditor/ckeditor5-paste-from-office/pull/81 (branch `t/ckeditor5-engine/404`.

Requires: https://github.com/ckeditor/ckeditor5-utils/pull/302.

### DIFF
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -387,15 +387,7 @@ export default class DomConverter {
 	 */
 	domToView( domNode, options = {} ) {
 		if ( isBlockFiller( domNode, this.blockFillerMode ) ) {
-			const isSingle = domNode.parentNode && domNode.parentNode.childNodes.length <= 1;
-
-			if ( isText( domNode ) ) {
-				if ( isSingle && _hasDomParentOfType( domNode, this.blockElements ) ) {
-					return null;
-				}
-			} else {
-				return null;
-			}
+			return null;
 		}
 
 		// When node is inside UIElement return that UIElement as it's view representation.
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -7,7 +7,7 @@
  * @module engine/view/domconverter
  */
 
-/* globals document, Node, NodeFilter, Text */
+/* globals window, document, Node, NodeFilter, Text */
 
 import ViewText from './text';
 import ViewElement from './element';
@@ -16,15 +16,7 @@ import ViewRange from './range';
 import ViewSelection from './selection';
 import ViewDocumentFragment from './documentfragment';
 import ViewTreeWalker from './treewalker';
-import {
-	BR_FILLER,
-	NBSP_FILLER,
-	INLINE_FILLER_LENGTH,
-	isBlockFiller,
-	isInlineFiller,
-	startsWithFiller,
-	getDataWithoutFiller
-} from './filler';
+import { BR_FILLER, getDataWithoutFiller, INLINE_FILLER_LENGTH, isInlineFiller, NBSP_FILLER, startsWithFiller } from './filler';
 
 import global from '@ckeditor/ckeditor5-utils/src/dom/global';
 import indexOf from '@ckeditor/ckeditor5-utils/src/dom/indexof';
@@ -32,6 +24,9 @@ import getAncestors from '@ckeditor/ckeditor5-utils/src/dom/getancestors';
 import getCommonAncestor from '@ckeditor/ckeditor5-utils/src/dom/getcommonancestor';
 import isText from '@ckeditor/ckeditor5-utils/src/dom/istext';
 import { isElement } from 'lodash-es';
+
+// eslint-disable-next-line new-cap
+const BR_FILLER_REF = BR_FILLER( window.document );
 
 /**
  * DomConverter is a set of tools to do transformations between DOM nodes and view nodes. It also handles
@@ -387,7 +382,7 @@ export default class DomConverter {
 	 * or `null` if DOM node is a {@link module:engine/view/filler filler} or the given node is an empty text node.
 	 */
 	domToView( domNode, options = {} ) {
-		if ( isBlockFiller( domNode, this.blockFillerMode ) ) {
+		if ( this.isBlockFiller( domNode, this.blockFillerMode ) ) {
 			return null;
 		}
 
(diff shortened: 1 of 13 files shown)


jodator:
So for PFO some tests fail because:
```html
<p>
<span><img src="..."></span><span style='mso-spacerun:yes'>&nbsp;</span>foo 
</p>
```
is expected to be:
```html
<image src="foo.jpg"></image>
<paragraph>foo</paragraph>
```

which is IMO wrong as this should have space inserted before `"foo"` and with previous logic, the `&nbsp;` was cut off as it was considered a block filler.

jodator:
@Reinmar OK the code is fixed and we should be able to:

```js
editor.setData( '<p><strong>a</strong>&nbsp;<i>f</i></p>' );
```

The PR in PFO was also updated to reflect latest findings about tests and the PFO internals. In one test I had to add space between blocks (`<table>` and `<paragraph>`) because this is what DOM->View conversion will produce and since we allow spaces between blocks in clipboard holder then the `model` fixture must have this set as it represents model data produced by the PFO.

src/view/domconverter.js
```diff
@@ -371,7 +371,15 @@ export default class DomConverter {
 	 */
 	domToView( domNode, options = {} ) {
 		if ( isBlockFiller( domNode, this.blockFiller ) ) {
-			return null;
+			const isSingle = domNode.parentNode && domNode.parentNode.childNodes.length <= 1;
```
Reinmar: I think we should have this entire logic in `isBlockFiller()`. This is exactly what it's supposed to check.

Reinmar:
I think we can get rid of this fille creators and just have two modes – `'br'` and `'nbsp'`. Then we'll have this set of params: `isBlockFiller( node, mode )`. 

The biggest change, though, will be due to the fact that now `isBlockFiller()` will have to be called with nodes which are in some context. Calling it on detached nodes will have little sense.

src/view/filler.js
```diff
@@ -168,3 +164,39 @@ function jumpOverInlineFiller( evt, data ) {
...
+// @param {Node} domNode DOM node.
+// @returns {Boolean}
+function isNbspFillerFiller( domNode ) {
+	const isNBSP = isText( domNode ) && domNode.data == '\u00A0';
+
+	return isNBSP && hasBlockParent( domNode ) && domNode.parentNode.childNodes.length === 1;
+}
+
+// Name of DOM nodes that are considered a block.
+const blockElements = [ 'p', 'div', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ];
```
jodator: @Reinmar I don't like that this duplicates the code from a `DomConverter` - maybe we should intorduce `isBlockElement()` in ckeditor5-utisl/dom? That this could be unified there.

jodator:
@Reinmar I've rewritten the checks as we talked.

Requires:
- changes in PFO: https://github.com/ckeditor/ckeditor5-paste-from-office/pull/81 (branch `t/ckeditor5-engine/404`).
- Util `assertEqualMarkup()` it is used in PFO tests to nicely format error messages: https://github.com/ckeditor/ckeditor5-utils/pull/302.

src/view/filler.js
Reinmar: We should use the same block elements array which is defined in `DomConverter`. The idea is that it can be extended by someone else if needed. So, we need to find a way to make `filler.js` us that property.

Reinmar: One option would be to have a module here (in the engine) which exports a table.

This module would be imported by both `domconverter.js` and `filler.js`. The instance of this array would be shared by those modules. So, if you'd extend `DomConverter.blockElementNames` that'd also extend the value used by the `filler.js` module.

The downside of this would be that this would become a singleton. All editor instances would share this table. In fact, that'd mean that if you'd like to extend this array in your plugin's code (cause it adds some new element support to the editor), if you'd run 10 editors on one page, this element would be added 10 times. That's quite ugly.

So, I'd rather go in a different direction – e.g. move `isBlockFiller()` to `DomConverter`. Perhaps it doesn't make sense to have this as contextless util. Maybe it needs to be a part of bigger context.

jodator:
It makes sense to tie the `isBlockFiller()` to the `DomConverter` instance as it was done anyway in this proposal.

I've refactored the code so now you can check if a node is a block filler on the `converter` instance. Similarly, other utility methods are there (`isElement`, `isComment`, etc). So instead of `isBlockFiller( domNode, domConverter.blockFillerMode )` you have `domConverter.isBlockFiller( domNode )` which makes more sense.

Reinmar:
FYI: I've just removed entire ` hasBlockParent( domNode, blockElements )` from code and all the tests still pass. I also realised that you didn't add any tests for `isBlockFiller()` which would cover those scenarios.

Reinmar:
Added them in https://github.com/ckeditor/ckeditor5-engine/pull/1808.

PR #1781, opened 2019-08-13
Automatic marker re-rendering during conversion for markers which view element was unbound

### LINKED ISSUE #1780: Re-render markers for which view element has been unbound during conversion
There is a buggy scenario connected with markers, mapper and renaming elements:

1. Create a marker and map it to a view container element (for example table cell).
2. Rename that view container element (for example, set/unset table header).

Since the marker range has not changed, it won't be converted. But the renaming will remove the old view container element and insert a new one. This will lead to a situation when view marker representation is lost.

scofalik:

Other: Introduced automatic marker re-rendering during conversion for markers which view element was unbound. Closes #1780.

BREAKING CHANGE: New parameter introduced in `DowncastDispatcher#convertChanges()`. Now it is `convertChanges( differ, markers, writer )`.
BREAKING CHANGE: Although it was rather impossible to use `DowncastDispatcher` without specifying any conversion API in the constructor, now it is a required parameter.

### DIFF
--- a/src/conversion/downcastdispatcher.js
+++ b/src/conversion/downcastdispatcher.js
@@ -143,14 +143,12 @@ export default class DowncastDispatcher {
 			}
 		}
 
-		for ( const markerName of this.conversionApi.mapper._unboundMarkers ) {
+		for ( const markerName of this.conversionApi.mapper.flushUnboundMarkerNames() ) {
 			const markerRange = markers.get( markerName ).getRange();
 
 			this.convertMarkerRemove( markerName, markerRange, writer );
 			this.convertMarkerAdd( markerName, markerRange, writer );
 		}
-
-		this.conversionApi.mapper._unboundMarkers.clear();
 
 		// After the view is updated, convert markers which have changed.
 		for ( const change of differ.getMarkersToAdd() ) {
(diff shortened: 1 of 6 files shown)


src/conversion/downcastdispatcher.js
```diff
@@ -109,7 +109,7 @@ export default class DowncastDispatcher {
 	 * @param {Object} [conversionApi] Additional properties for interface that will be passed to events fired
```
pjasiun: You should update docs.

scofalik:
I've updated docs for `conversionApi` parameter but please hold on with this PR as I am not entirely sure how much do we actually need this change. I was under impression that this is a big and needed change and it looks like that on paper but I am struggling a bit with real test scenarios that are actually important.

I mean, I have a scenario that crashed without this fix but maybe actually it is possible (and makes sense) to provide a smaller fix than this.

src/conversion/downcastdispatcher.js
```diff
@@ -142,6 +143,15 @@ export default class DowncastDispatcher {
...
 		}
 
+		for ( const markerName of this.conversionApi.mapper._unboundMarkers ) {
+			const markerRange = markers.get( markerName ).getRange();
+
+			this.convertMarkerRemove( markerName, markerRange, writer );
+			this.convertMarkerAdd( markerName, markerRange, writer );
+		}
+
+		this.conversionApi.mapper._unboundMarkers.clear();
```
pjasiun: Maybe we could pack it into `this.conversionApi.mapper.flushUnboundMarkers()`?

Issue #1782, opened 2019-08-14
[IME][Safari] IME composistion broken in empty paragraph in Safari


farthinker:
The original issue was described in https://github.com/ckeditor/ckeditor5/issues/1333.

I've tried to fix it in https://github.com/ckeditor/ckeditor5-typing, but [failed](https://github.com/ckeditor/ckeditor5-typing/pull/218). Some IME cannot bear any DOM selection update.

So I thought maybe we should prevent DOM selection change while composing in the render process(like `return false` in [`_domSelectionNeedsUpdate`](https://github.com/ckeditor/ckeditor5-engine/blob/23a4c028fe/src/view/renderer.js#L773)). I'll submit a PR later.

PR #1783, opened 2019-08-14
[IME][Safari] Fix IME composistion broken issue in empty paragraph in Safari.

farthinker:
See https://github.com/ckeditor/ckeditor5-engine/issues/1782, https://github.com/ckeditor/ckeditor5-typing/pull/218 and https://github.com/ckeditor/ckeditor5/issues/1333 for details about the issue.

This fix make sure DOM selection untouched while composing. I've tested it with several IMEs, and it worked. But I need some help to confirm it won't break something else. @Reinmar @jodator @Mgsy

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -102,6 +102,13 @@ export default class Renderer {
 		this.isFocused = false;
 
 		/**
+		 * Indicates if the view document is composing.
+		 *
+		 * @member {Boolean}
+		 */
+		this.isComposing = false;
+
+		/**
 		 * The text node in which the inline filler was rendered.
 		 *
 		 * @private
@@ -771,6 +778,11 @@ export default class Renderer {
 	 * @returns {Boolean}
 	 */
 	_domSelectionNeedsUpdate( domSelection ) {
+		// Remain DOM selection untouched while composing (#1782)
+		if ( this.isComposing ) {
+			return false;
+		}
+
 		if ( !this.domConverter.isDomSelectionCorrect( domSelection ) ) {
 			// Current DOM selection is in incorrect position. We need to update it.
 			return true;
--- a/src/view/view.js
+++ b/src/view/view.js
@@ -105,6 +105,7 @@ export default class View {
 		 */
 		this._renderer = new Renderer( this.domConverter, this.document.selection );
 		this._renderer.bind( 'isFocused' ).to( this.document );
+		this._renderer.bind( 'isComposing' ).to( this.document );
 
 		/**
 		 * A DOM root attributes cache. It saves the initial values of DOM root attributes before the DOM element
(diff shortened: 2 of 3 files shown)


jodator:
Hi, thanks for the updated solution. We will try to check this as soon as possible but it can take a few days due to the holiday season and upcoming release.

The trick with pausing rendering in some cases recurs on various occasions so I'd ping @pjasiun to look at this briefly. (About pausing the renderer during text composing) Do you see any problems with that? Collaboration maybe?

jodator:
I mark this PR as a sub-PR for: https://github.com/ckeditor/ckeditor5-engine/pull/1783.

pjasiun:
I see it only touches the selection, so it could be fine. But since renderer is very tricky this change needs to be tested deeply.

jodator:
OK so please take another round of testing of this @Mgsy with both PRs (this & https://github.com/ckeditor/ckeditor5-typing/pull/218)

Reinmar:
Actually, we already did approach it in the past. It's crystal clear that we must not render the selection during composition. There's even a PR in which, from its title, was meant to try the same (stop selection rendering): https://github.com/ckeditor/ckeditor5-engine/pull/861. But when I checked the code it actually went much deeper and tried to disable all rendering (which we cannot do). Most likely that's why we never merged it.

So, I agree with @pjasiun – stoping rendering selection should be fine, but it requires some more testing. However, I'd actually prefer if we tested first this change alone to see what it fixes/breaks. Then, when we have a better understanding about it, we can see if we need more changes.

@farthinker, did you work on these changes (https://github.com/ckeditor/ckeditor5-engine/pull/1783 and https://github.com/ckeditor/ckeditor5-typing/pull/218) separately? I mean – do you know perhaps if the current PR is fixing itself some issues? It's much simpler than the other one and it makes it much easier to close it faster. The one in ckeditor5-typing may take more time because I can see it touches a lot more code.

farthinker:
@Reinmar I think https://github.com/ckeditor/ckeditor5-typing/pull/218 should be closed in favor of this PR. I've tried to fix the issue in ckeditor5-typing, but it seems impossible.

jodator:
@f1ames Sorry for the delay with the review but the topic is not an easy one. So we had to be sure that it will be OK. We'll be merging this soon.

@Reinmar I've check the tests and I don't see that we could do any better tests for composition. I'll add some small stylistic changes to the comments. But It's OK for me so I'll be merging this.

src/view/renderer.js
```diff
@@ -101,6 +101,13 @@ export default class Renderer {
 		 */
 		this.isFocused = false;
 
+		/**
+		 * Indicates if the view document is composing.
```
jodator:
```suggestion
		 * Indicates if the composition is in progress inside the view document view.
```

PR #1790, opened 2019-09-04
Improved filtering out disallowed attributes on child elements.

oskarwrobel:

Fix: Improved filtering out disallowed attributes on child elements. #1789.

---

### Additional information

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -732,15 +732,34 @@ export default class Schema {
 	 * @param {module:engine/model/writer~Writer} writer
 	 */
 	removeDisallowedAttributes( nodes, writer ) {
-		for ( const node of Array.from( nodes ) ) {
-			for ( const attribute of node.getAttributeKeys() ) {
-				if ( !this.checkAttribute( node, attribute ) ) {
-					writer.removeAttribute( attribute, node );
-				}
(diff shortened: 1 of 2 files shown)


Reinmar:
We need to be creative and write a couple more tests where text nodes would be merged during the process. I fear that there's a range of potentially broken implementations that we need to avoid.

I think that the implementation may look like this:

1. Scan the list of nodes for incorrect attributes. **Create a range for each of those places.**
2. Iterate over the list of those ranges and pass them to `removeAttribute()`.

This will be safe because merging text nodes does not change positions.

Reinmar:
BTW, I don't remember why we're checking whether the right type of operation is used. I think that it is enough if one or two tests check that. The other ones don't have to.

Reinmar:
That was a really cool idea with `range.getPositions()`! Iterating over the range and taking `value.item`s would not work because it'd hit the same problem. So the only alternative would be iterating over the range and using `nodeBefore || parent` as you did, so I actually prefer your solution.

Reinmar:
I only removed `continue;` because it makes reading the code harder. Because you actually need to understand that what's later in this loop will not happen when the former happened. if-else represents that semantics more directly.

PR #1792, opened 2019-09-06
Skip native DOM selection change in case of subsequent requests for the same fake selection

### LINKED ISSUE #1791: Fake selection is updated regardless of whether it needs to be updated
During image resizing the renderer is executed on every change of the image (I'm not sure who triggers the renderer, but it happens, and it just may happen, so let's accept this fact).

Unfortunately, the renderer doesn't check whether fake selection really needs to be updated so on every call it removes the native selection and sets it back in the same place.

This causes serious deoptimization – Chrome blocks rendering thread for a long periods, stacking some selection changes. It only renders (probably) when for a certain period of time it doesn't get any selection change.

mlewand:

Fix: Improved performance when working with fake selections. Closes #1791.

---

### Additional information

This approach introduces one change. If there are two subsequent calls to create fake selection with exactly the same label, then there will be no `HTMLDocument.selectionchange` event (as this case will also be optimized). This might be a case when there are for instance two different widgets, but both have the same label (e.g. both have simply "Image widget").

In case the above case should be covered, the implementation should be more complex.

### DIFF
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -803,8 +803,20 @@ export default class Renderer {
 	 */
 	_fakeSelectionNeedsUpdate( domRoot ) {
 		const container = this._fakeSelectionContainer;
+		const domSelection = domRoot.ownerDocument.getSelection();
 
-		return !container || container.parentElement != domRoot || container.textContent !== this.selection.fakeSelectionLabel;
+		// Selection fakeselection needs to be updated if there's no fake selection container, or container
+		// currently sits in different root.
+		if ( !container || container.parentElement != domRoot ) {
+			return false;
+		}
+
+		// Make sure that the selection actually is within the fake selection.
+		if ( !domSelection.anchorNode.isSameNode( container ) ) {
+			return true;
+		}
+
+		return container.textContent !== this.selection.fakeSelectionLabel;
 	}
 
 	/**
--- a/src/view/renderer.js
+++ b/src/view/renderer.js
@@ -691,29 +691,14 @@ export default class Renderer {
 	 */
 	_updateFakeSelection( domRoot ) {
 		const domDocument = domRoot.ownerDocument;
-		let container = this._fakeSelectionContainer;
+		// Create fake selection container if one does not exist.
+		const container = this._fakeSelectionContainer = this._fakeSelectionContainer || createFakeSelectionContainer( domDocument );
+
+		// Bind fake selection container with current selection.
+		this.domConverter.bindFakeSelection( container, this.selection );
 
 		if ( !this._fakeSelectionNeedsUpdate( domRoot ) ) {
-			// Container did not change, but the selection might point a different element with same fake selection.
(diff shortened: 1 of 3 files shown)


mlewand:
After some more testing I found an issue where focusing a widget, blurring the editor and focusing widget again would not mark the focus properly.

That's fixed with 1bc5df6.

src/view/renderer.js
```diff
@@ -711,6 +711,8 @@ export default class Renderer {
 
 		if ( !container.parentElement || container.parentElement != domRoot ) {
 			domRoot.appendChild( container );
+		} else if ( !this._fakeSelectionNeedsUpdate( domRoot ) ) {
```
Reinmar: This logic is a bit odd. First, we add the container and then we check whether it had to be updated? 

Also, inside `fakeSelecitonNeedsUpdate()` you check again whether `_fakeSelectionContainer` is defined... but we just did it here in `_updateFakeSelection()`

The logic should be reversed:

1. check whether we need to update anything; abort if not
2. if the container does not exist, create it
3. if the container isn't appended to the right root, append it

Reinmar:
Please correct the logic of this function because the flow seems incorrect.

tests/view/renderer.js
```diff
@@ -1804,6 +1804,57 @@ describe( 'Renderer', () => {
...
+					document.getSelection().removeAllRanges();
+					document.getSelection().addRange( interruptingRange );
+
+					selection._setTo( selection.getRanges(), { fake: true, label } );
+					renderer.render();
+
+					expect( createRangeSpy.callCount ).to.be.equal( 2 );
+				} );
+
+				it( '_fakeSelectionNeedsUpdate returns proper result when container is uninitialized', () => {
```
Reinmar: You shouldn't need to write a test which checks this function like this. Either a case exists in a real-life and can be tested in a more integration way, or it's not needed.

mlewand:
Notes provided above were addressed. Further testing shows that there is still a regression after this change, occurring when:

1. Focusing a widget.
1. Blurring the editor.
1. Focusing *another* widget.

Focus goes to the first (previously focused) widget instead of clicked one. I'll dig deeper for that.

mlewand:
All right I was able to address the regression described in https://github.com/ckeditor/ckeditor5-engine/pull/1792#issuecomment-529814641.

The issue was caused by the fact that in initial patch domConverter fake selection binding was not set (due to early return). Thus old selection was returned.

Reinmar:
> All right I was able to address the regression described in [#1792 (comment)](https://github.com/ckeditor/ckeditor5-engine/pull/1792#issuecomment-529814641).
> 
> The issue was caused by the fact that in initial patch domConverter fake selection binding was not set (due to early return). Thus old selection was returned.

Please also add a test for this scenario. It's something easy to miss. Especially that `bindFakeSelection()`'s behaviour is confusing.

Reinmar:
@Mgsy, could you take a look at this? The code looks fine, but it's hard to predict implications and they may be browser specific. I'm mostly worried about cases where editor or its nested editable gets focus together with a widget being selected.

PR #1798, opened 2019-10-07
Allow creating namespaced domelements

### LINKED ISSUE #1797: Support creation of namespaced elements
This relates to https://github.com/ckeditor/ckeditor5/issues/1842.

Currently it's not possible to use createElement to create working namespaced elements like SVGs.

PR is on its way.

Sebobo:
Fix: Create valid namespaced elements. Closes #1797.

---

### Additional information

* Relates to https://github.com/ckeditor/ckeditor5/issues/1842

### DIFF
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -224,7 +224,12 @@ export default class DomConverter {
 				return domElement;
 			} else {
 				// Create DOM element.
-				domElement = domDocument.createElement( viewNode.name );
+				const namespace = viewNode.getAttribute( 'xmlns' );
+				if ( namespace !== undefined ) {
+					domElement = domDocument.createElementNS( namespace, viewNode.name );
+				} else {
+					domElement = domDocument.createElement( viewNode.name );
+				}
 
 				if ( options.bind ) {
 					this.bindElements( domElement, viewNode );
--- a/src/view/domconverter.js
+++ b/src/view/domconverter.js
@@ -224,9 +224,8 @@ export default class DomConverter {
 				return domElement;
 			} else {
 				// Create DOM element.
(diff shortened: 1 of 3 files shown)


src/view/domconverter.js
```diff
@@ -224,7 +224,12 @@ export default class DomConverter {
 				return domElement;
 			} else {
 				// Create DOM element.
-				domElement = domDocument.createElement( viewNode.name );
+				const namespace = viewNode.getAttribute( 'xmlns' );
```
Reinmar: I think this would be enough:

```js
if ( viewNode.hasAttribute( 'xmlns' ) )
```

tests/view/domconverter/dom-to-view.js
```diff
@@ -175,6 +175,15 @@ describe( 'DomConverter', () => {
 			expect( converter.domToView( comment ) ).to.be.null;
 		} );
 
+		it( 'should create namespaced elements', () => {
```
Reinmar: AFAICS this test would pass even without this change. We need a test that would fail on master.

Reinmar:
I can see that CC dropped too.

src/view/domconverter.js
Sebobo: How do you mean? I need the namespace for the constructor of the element.

Sebobo: Ah understood. Will change.

tests/view/domconverter/dom-to-view.js
Sebobo: And the test was in the wrong place. Fixed now and fails without the change :)

Sebobo:
CC is now fine again too.

Reinmar:
Thanks!

PR #1801, opened 2019-10-11
I/5569:  Docs: Fix custom figure attributes snippet.

jodator:

Docs: Fix custom figure attributes snippet. Closes ckeditor/ckeditor5#5569.

---

### Additional information

* Schema extending must be done in `afterIniti()`.


Reinmar:
Just like I commented in https://github.com/ckeditor/ckeditor5/issues/5569#issuecomment-541069707. It seems that something else is wrong there.

jodator:
@Reinmar as in the issue: https://github.com/ckeditor/ckeditor5/issues/5569#issuecomment-541591630. The `extraPlugins` doesn't matter if you extend schema in plugin's constructor because in most cases feature plugins defines schema in `init()` which is called after all constructors.

Reinmar:
OK, I got the problem. [https://github.com/ckeditor/ckeditor5/issues/5569#issuecomment-541591630](https://github.com/ckeditor/ckeditor5/issues/5569#issuecomment-541591630) is clear to me.

I don't like the solution in this PR still, though :smile:. If we'll tell people to use the `Plugin` class they will not be able to do this without building the entire editor. Which significantly break this guide cause so far it was only using simple functions.

Two solutions:

*   Allow function plugin return object with `init` and `afterInit` callbacks.
*   Treat all function plugins like plugin constructors (so you don't have to inherit directly from `Plugin` and you can define a simple class with `init` and `afterInit`).

jodator:
> Two solutions:

Third options which works right now:
```js
plugins: [ ..., function( editor ) {
    this.intit = () => {
    }
}
```
which I don't really know if is OK or not but it does implement the `Plugin` interface ;)

Reinmar:
LOL, this works as well:

```plaintext
function MyPlugin( editor ) {
	return {
		init() {
			console.log( 'init' );
		},

		afterInit() {
			console.log( 'afterInit' );
		}
	};
}

ClassicEditor
	.create( document.querySelector( '#editor' ), {
		plugins: [ ArticlePluginSet, MyPlugin ],
		...
```

And this as well:

```plaintext
class MyPlugin {
	constructor( editor ) {
		console.log( 'constructor', editor );
	}
	init() {
		console.log( 'init' );
	}
	afterInit() {
		console.log( 'afterInit' );
	}
}
```

I think the last form is best so I'd recommend using exactly this in this guide.

jodator:
> LOL, this works as well:

A tale of the magic of JavaScript and a hundred ways to make a mess...

Sure the `MyPlugin` just implementing the iterface is also OK

PR #1806, opened 2019-10-24
Cannot read property 'toLowerCase' of undefined Bugfix

tikudev:
The function areSimilar examines whether two nodes are similar by checking their tag names. Textnodes are already excluded because they have no tagNames. This commit also excludes CommentNodes because they also have no tagNames, which leads to a Javascript Error, when HTMLCommentNodes are currently parsed by the CKEditor.


oleq:
Hi, @tiku91! 

Thanks for your contribution. Could you please tell us more about your problem? A test or a valid reproduction path (like editor data you used, your plugins and customization) would certainly help us understand it better.

Reinmar:
Curious whether it doesn't fix https://github.com/ckeditor/ckeditor5/issues/5734.

For now, I'm marking as R-.

tiku91:
@oleq @Reinmar 
Hello and thank your for your patience.
You can find a more detailed description of the error and a valid reproduction path in https://github.com/ckeditor/ckeditor5/issues/5734 
As I have stated in my commit message, the error in the code is pretty straight forward, there is a node.tagName.toLowerCase() call, in which it is not checked whether the node has a tagName. In comment nodes those tagNames are undefined and therefore the error: "cannot read property toLowerCase() of undefined".
I do not feel comfortable to disclose personal information in order to sign the CLA. If that means you can not accept the PR, then I hope that you can fix the Problem by writing the one line of code yourself ;) (Check whether the node is a comment-node before calling toLowerCase)

tomalec:
@tiku91 thank you for your support.
Given, lack of CLA from you, I created a new PR, and I'm closing this one in favor of https://github.com/ckeditor/ckeditor5/pull/7403.

PR #1807, opened 2019-10-28
I/3287: Styles normalization (part of table styles)

jodator:

Feature: Introduce CSS style normalization for conversion. Closes ckeditor/ckeditor5#6047.

---
This is a sub PR for table styles: https://github.com/ckeditor/ckeditor5-table/pull/224.

## PR constallation

- `font`: https://github.com/ckeditor/ckeditor5-font/pull/57
- `indent`: https://github.com/ckeditor/ckeditor5-indent/pull/21
- `utils`: https://github.com/ckeditor/ckeditor5-utils/pull/312

## Description

This PR comes as an outcome for works to enable table styles (mainly `border` shorthand) management. What we have lacked before was a better way of converting from `style` attribute entries because of CSS shorthand values.

Proposed solution.

1. Introduce `Styles` class that handles style normalization on write and on reading.
    1. write styles as in CSS inline style attribute, `styles.setStyle( 'border:1px solid blue;margin: 1px' )`
    2. write, remove, check single CSS property `styles.insertProperty( 'border', '1px solid blue' )` by writing shorthand or longhand values
    3. obtain normalized style or property (either single entry or an object - eg. whole style object)
    4. obtain inline style string - either whole `style` attribute or singe value. The output is standardized and predictable.
2. Use `Style` instance internally in `view/Element` instead of simple `Map`.
3. Introduce `StylesConverter` that provides a single entry point for defining normalization, extraction and inlining the styles.
    - 👎 right now it is a singleton instance in the module - I don't have a better idea
    - 👍 extensible via events - similar to other APIs, mostly to position mapping API (event data modification) - CSS property is searchable in code
        - `normalize:css-property-name`, to normalize a single entry
        - `reduce:css-property-name`, to get "reduced" form to use as inline style (other idea `inline:css-property-name`
        - I don't have idea to get rid of this: `extract:css-property-name` this event is used to extract some form of normalized value - again `border` here:
        Border stored as `{ color: { top, right, bottom, left }, style: { ... }, width: { ... } }` but for inline style is extracted as `border-top`, `border-left`, etc - which combines all of them. Another case is extracting `borer-top-color` longhand property. This is and edge-case (I didn't analyze other CSS properties) as most of the implemented properties are two-level values.
4. Upcast conversion requires a solution for automatic styles consuming (implmenented as `consumeAlso` mapping (should be moved somewhere from `ViewConsumable`). It solves the problem of consuming single entry from a shorthand. The first example is indented block converter which consumes only `margin-left` (in LTR, I'll skip all the RTL variants) so IMO the `margin` style should be consumed also but `margin-right`, `margin-top` and `margin-bottom` should be available for other converters to process. Similarly, if some converter consumes `margin` all `margin-` longhands should be consumed.
5. Next steps, possible enhancements: `font` shorthand (... )

 What this solves?

The main idea is that inline style can be written in many, many ways — especially the `border` shorthand sets 12 longhand properties (`border-top-color`, `border-left-width`, etc). Also, at least for the border, there are another 7 shorthands:  `border-top`, `border-right`, ... and  `border-color`, `border-style`, `border-width` which overlaps each other. Writing converters for each possible combination would result in around 20 converters (12 longhands, 8 shorthands) that could interfere with each other. This isn't a problem for styles coming from CKEditor as we always control this. The problem occurs if the styles come from other editors, pages, etc - there are various ways of writing the same and we can solve most of them by normalizing incoming styles.

Also it should parse shorthand properly as, ie for border shorthand the order of values doesn't matter so: `border: 1px solid blue` === `border:solid blue 1px` === `border: blue 1px solid`.

Anather thing is to simplify converters - especially the partial ones (`margin-right`).


src/view/styles.js
```diff
@@ -0,0 +1,490 @@
...
+	 * Returns a string containing normalized styles string or undefined if no style properties are set.
+	 *
+	 * @returns {String|undefined}
+	 */
+	getInlineStyle() {
+		const entries = this._getStylesEntries();
+
+		// Return undefined for empty styles map.
+		if ( !entries.length ) {
+			return;
```
Reinmar: I'd rather expect an empty string.

Reinmar: Also, `StylesMap.isEmpty` getter to be used at the very beginning. Methods of this class will be used hundreds of time and all of those normalization methods do quite a lot of things to return... an empty thing (which will be 99% of times). So, I'd make sure the "return empty thing" path is quick, just to feel safe.

src/view/styles.js
```diff
@@ -0,0 +1,490 @@
...
+		return entries.map( arr => arr.join( ':' ) ).join( ';' ) + ';';
+	}
+
+	/**
+	 * Returns property value string.
+	 *
+	 * @param {String} propertyName
+	 * @returns {String|undefined}
+	 */
+	getInlineProperty( propertyName ) {
```
Reinmar: `getPropertyAsString()`?

src/view/styles.js
```diff
@@ -0,0 +1,490 @@
...
+		// Return undefined for empty styles map.
+		if ( !entries.length ) {
+			return;
+		}
+
+		return entries.map( arr => arr.join( ':' ) ).join( ';' ) + ';';
+	}
+
+	/**
+	 * Returns property value string.
```
Reinmar: Does it handle normalized styles or not?

src/view/styles.js
```diff
@@ -0,0 +1,490 @@
...
+	 *
+	 * @param {String} propertyName
+	 * @returns {String|undefined}
```
Reinmar: For some reason, I'd also see here a string to be always returned.

src/view/styles.js
```diff
@@ -0,0 +1,490 @@
...
+	 *
+	 * @param name
+	 */
+	removeProperty( name ) {
+		unset( this._styles, toPath( name ) );
+		delete this._styles[ name ];
+	}
+
+	/**
+	 * Return normalized style object;
```
Reinmar:
```suggestion
	 * Return a normalized style object.
```

src/view/styles.js
```diff
@@ -0,0 +1,490 @@
...
+	 *
+	 *		const styles = new Styles();
+	 *		styles.setStyle( 'margin:1px 2px 3em;' );
+	 *
+	 *		console.log( styles.getNormalized( 'margin' ) );
```
Reinmar: What will `margin-left` return?

src/view/styles.js
```diff
@@ -0,0 +1,490 @@
...
+
+/**
+ * @module engine/view/styles
+ */
+
+import { get, isObject, merge, set, unset } from 'lodash-es';
+import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
+import mix from '@ckeditor/ckeditor5-utils/src/mix';
+
+import BorderStyles from './styles/borderstyles';
```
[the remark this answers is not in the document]
Reinmar: Also, I'd do that via `editor.editing.view.document.addStyleProcessor()` which will be a shorthand to call `StylesMap.addProcessor()`. This way, if we'll ever decide to bind the processor with a particular document instance, it will not change the API.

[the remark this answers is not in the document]
Reinmar: (☝️ I used a new naming above – see next comments)

src/view/styles/borderstyles.js
```diff
@@ -0,0 +1,211 @@
...
+ * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import { getShorthandValues, getTopRightBottomLeftValueReducer, getTopRightBottomLeftValues, isLength, isLineStyle } from './utils';
+
+/**
+ * @module engine/view/styles
+ */
+export default class BorderStyles {
```
Reinmar: Why a class? Wouldn't a single function be enough? `addBorderStylesProcessor()`

src/view/styles.js
```diff
@@ -0,0 +1,490 @@
...
+	/**
+	 * Returns normalized styles entries for further processing.
+	 *
+	 * @private
+	 * @returns {module:engine/view/styles~PropertyEntry}
+	 */
+	_getStylesEntries() {
+		const parsed = [];
+
+		const keys = Object.keys( this._styles ).sort();
```
Reinmar: I don't see a need for sorting here. Isn't sorting only needed for getting things as string?

Reinmar: (also, performance)

src/view/styles/borderstyles.js
```diff
@@ -0,0 +1,211 @@
+/**
```
Reinmar: This is awfully long :D But I realised that the only thing I care about is – what will happen with a style value which is not recognized by a processor? E.g., if they'll introduce a more complex `border` notation or some other ways to write a color or a possibilit to set multiple borders at once (I think it's even possible), then what will happen?

The thing is – those things must be preserved.

Reinmar:
Just to clarify my review comment – proposed changes and the questions I left in the comments are mostly cosmetic. There are many of them due to the size of this API, but the changes will not need to be dramatic. There just will be a lot of them.

src/view/element.js
```diff
@@ -342,8 +334,11 @@ export default class Element extends Node {
 		}
 
 		// Check if styles are the same.
-		for ( const [ property, value ] of this._styles ) {
-			if ( !otherElement._styles.has( property ) || otherElement._styles.get( property ) !== value ) {
+		for ( const property of this._styles.getStyleNames() ) {
```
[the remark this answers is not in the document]
jodator: This method works OK (the number of styles is checked earlier) - I've refactored the tests so we have that described a bit clearer: https://github.com/ckeditor/ckeditor5-engine/pull/1809.

jodator:
I've gone thought all the comments and I think that I'll start with adding the most needed changes - example usage and other cleanup to make it digestible.

jodator:
Problems:
- Enabling some styles from a feature.
- The code is hackish and hard to extend.

Constrains:
- Style converter should be one per document.
- Style conversion should be done outside element.


Solution 1: Style modification must be done by `writer` (AFAICS its done already) as writer has access to the document.
Solution 2: Style should be stored as Object (normalized form).
Solution 3: Style should be stored as before (Map but already normalized by writer).

New problems:
- some Style getters (getIdentity, hasStyle, getStyleNames) are not suitable with above solutions as Element has no access to the document. (no normalized names).

```js
element.setStyle( 'margin', '2px', converter ); //by writer
element.setStyle( 'margin-top', '4px', converter ); //by writer
// this will set margin = { top: 4px, left, right, bottom: 2px }
element.hasStyle( 'margin-left' ) /// should true?
element.hasStyle( 'margin' ) /// should true?
```

If however the element would have reference to the docuemnt then it would be easy but OTOH memory leaks...

jodator:
After F2F talk with @Reinmar:
- we're OK with the possibility of two editors enabling something for other editors (because of imports/singletons)
- the `ViewElement.styles` must be smart as proposed (as in DOM)
- in the futrue we might add document to element - then all the processors will be attached to view.document.

src/view/element.js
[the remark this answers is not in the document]
jodator: Tracked here: https://github.com/ckeditor/ckeditor5/issues/6068

jodator:
So after ~58 commits this PR has:

- most of the @Reinmar's comments included
- documentation added in various places but it still lacks some general guide
- I've refactored the code from events to maps because events will be slower then map lookup
- maybe some concepts (reducer, normalizer) requires better names but I'm out of creativity...

Reinmar:
Welcome back, i/3297! 🤚

Reinmar:
```plaintext
JSDoc started...
Incorrect link: module:engine/view/styles/background~addMarginRules
	at /Users/p/Workspace/ckeditor5/packages/ckeditor5-engine/src/view/document.js:175

Invalid return type: Array.<Array.<String, String>>.
	at /Users/p/Workspace/ckeditor5/packages/ckeditor5-engine/src/view/stylesmap.js:498
```

jodator:
@Reinmar done.

Reinmar:
```    conversion
      using offset
        ✔ should not convert class to indent attribute
        left–to–right content
          ✔ should convert margin-left to indent attribute (known offset)
          ✔ should convert margin-left to indent attribute (any offset)
          ✖ should convert margin shortcut to indent attribute (one entry)
          ✖ should convert margin shortcut to indent attribute (two entries)
          ✖ should convert margin shortcut to indent attribute (three entries)
          ✖ should convert margin shortcut to indent attribute (four entries)
        right–to–left content
          ✔ should convert margin-right to indent attribute (known offset)
          ✔ should convert margin-right to indent attribute (any offset)
          ✖ should convert margin shortcut to indent attribute (one entry)
          ✖ should convert margin shortcut to indent attribute (two entries)
          ✖ should convert margin shortcut to indent attribute (three entries)
          ✖ should convert margin shortcut to indent attribute (four entries)
```

I get the above red tests in ckeditor5-indent for this setup:

```
│ editor-inline            │ master   │ 02763b0 │        │
│ engine                   │ ! i/3287 │ 709bbb2 │ M1     │
│ enter                    │ master   │ 4848bcb │        │
│ essentials               │ master   │ 792e693 │        │
│ font                     │ ! i/3287 │ 4d1b9fb │ M2     │
│ heading                  │ master   │ 94fbca7 │        │
│ highlight                │ master   │ 9589cc3 │        │
│ horizontal-line          │ master   │ 1391e25 │        │
│ image                    │ master   │ 8f06626 │        │
│ indent                   │ ! i/3287 │ 38e2948 │        │
│ link                     │ master   │ 73d5596 │        │
│ list                     │ master   │ cfe7f8e │        │
│ markdown-gfm             │ master   │ fb5cee3 │        │
│ media-embed              │ master   │ c8ba8bf │        │
│ mention                  │ master   │ 3c207d0 │        │
│ page-break               │ master   │ b2b57e1 │        │
│ paragraph                │ master   │ 92b1a8d │        │
│ paste-from-office        │ master   │ 958662e │        │
│ remove-format            │ master   │ 0a448ac │        │
│ restricted-editing       │ master   │ 5e437b7 │        │
│ special-characters       │ master   │ 58f7cac │        │
│ table                    │ ! i/3287 │ a737502 │        │
│ theme-lark               │ master   │ 5252417 │        │
│ typing                   │ master   │ 139989c │        │
│ ui                       │ ! i/3287 │ 1763666 │        │
│ undo                     │ master   │ 6ce1985 │        │
│ upload                   │ master   │ 2657a60 │        │
│ utils                    │ ! i/3287 │ 3fff7ec │        │
│ watchdog                 │ master   │ 72628be │        │
│ widget                   │ master   │ 096a9f9 │        │
│ word-count               │ master   │ 9270325 │        │
└──────────────────────────┴──────────┴─────────┴────────┘
```

jodator:
@Reinmar That's right I didn't check the Indent package. After digging up I've fixed their test on `i/3287` by: https://github.com/ckeditor/ckeditor5-indent/pull/21/commits/9f8d9d56ed15482d2d675c02af31f094072ce6be - you see how the code get uglier. The other thing to do is to leave previous **two** converters for `margin` and `margin-left`. That would, however, feel meh as we introduced those fancy normalizations.

Now back to [your comment about consumables](https://github.com/ckeditor/ckeditor5-engine/pull/1807#discussion_r347553163) - when fixing this PR I've forgotten about that case so I've just removed that code which did 2 things:

1. Updated `ViewConsumable` class logic with mappings for `margin` => `margin-top`, `margin-right``, etc.
2. Added logic for consuming longhand value if one of shorthand was consumed (ie when consuming `margin-left` the `margin` was automatically consumed.

The point 2. does look like too much but point 1 looks useful for conversion helpers.

How to repair this?

1. Make `getStyleNames()` -> more clever and return more values (with a switch) (ie all "consumable"/"possible" so `margin,margin-top,margin-right,...` for `margin:1px` (all 4 values set).
2. Make `ViewConsumable` smarter **if** [style is enabled](https://github.com/ckeditor/ckeditor5-engine/pull/1807#discussion_r347553519).

Right now the logic for creating view consumable names gets the style names from view element and uses only them. In other words, testing for consuming `margin-top` will fail if a style is `margin: 1px` because `viewElement.getStyleNames()` will return only `margin` and only this value will be created to create consumable entries.

ps.: this _might_ be a follow-up as `indent` PR is optional - master works OK. I've used Indent to validate the whole idea of normalizing styles and how we could use that in other places. In this scenario writing a converter for longhand value (`margin-top`) should be enough (assuming style normalization) for any `margin` style definition. This also could be used with borders and the shitload of shorthands for writing a bit nicer converters for `border-style`, `border-color` and `border-width` as current converter is not ideal: https://github.com/ckeditor/ckeditor5-table/blob/966bd7d271eba4ff6a0507a9cc5fc075c78b2749/src/converters/tableproperties.js#L40.

Reinmar:
We've had a F2F discussion with @jodator and @scofalik and decided that:

* All shorthand and longhand property names will become consumable. It makes sense to have `Element#getStyleNames( { includeAllCombinations } )` to return those names.
* Consuming a longhand property (e.g. `margin-left`) will consume `margin` too. 
* The goal is to keep converters as simple as possible (ideally, so they don't need to be aware of anything that we're doing here).

jodator:
>     * All shorthand and longhand property names will become consumable. It makes sense to have

@Reinmar I've gone with a previous logic but re-implemented with an ability to "enable" it in view consumable.

The idea of consuming "also" works the same (I've restored previous tests).

The only different thing which I've done in my commit is that I've restored the previous "consumeAlso" logic without `element.getStylename( { includeAllCombinations } )`. Let me know what you think of it. Eventual change to move that logic to `getStyleNames( { inludeAllCombinations: true } )` will be quick as the API for defining this mappings is registered by `editing.view.document` as other style processing rules.

Reinmar:
I created a followup for the border processor as I don't like how it works: [https://github.com/ckeditor/ckeditor5/issues/6090](https://github.com/ckeditor/ckeditor5/issues/6090). Not necessarily it29 material though.

Reinmar:
I found a quite big problem with our style processor singleton: [https://github.com/ckeditor/ckeditor5-utils/pull/312#issuecomment-576417306](https://github.com/ckeditor/ckeditor5-utils/pull/312#issuecomment-576417306) and I'm having troubles figuring out what we could do with it.

Reinmar:
> I found a quite big problem with our style processor singleton: [ckeditor/ckeditor5-utils#312 (comment)](https://github.com/ckeditor/ckeditor5-utils/pull/312#issuecomment-576417306) and I'm having troubles figuring out what we could do with it.

Extracted that to [https://github.com/ckeditor/ckeditor5/issues/6091](https://github.com/ckeditor/ckeditor5/issues/6091).

Other than that, I think we're good to merge these 4 PRs.

PR #1811, opened 2019-12-11
Mutations in UIElements should not be propagated / force render

mlewand:

Fix: Change in UIElement view will no longer force a view render nor will it trigger mutations event on the document. Closes ckeditor/ckeditor5#5600.

---

### Additional information

The tests are passing, but this change needs to be carefully tested manually.

* image resize doesn't play well with it (toolbar is stuck in place during resize) - that's fine as we're going to hide it during resize (ckeditor/ckeditor5#5964)

I suggest few testing scenario:

* image style (aligning them, making sure that the toolbar moves properly)
* inline toolbar for table
* Cloud Features that use inline toolbar
* not sure if it might get with toolbar rotator?

Also it's good that we'll be able to merge it at the beginning of iteration, so there's a bigger chance that we'll catch nuances if any appear.

### DIFF
--- a/src/view/observer/mutationobserver.js
+++ b/src/view/observer/mutationobserver.js
@@ -244,7 +244,7 @@ export default class MutationObserver extends Observer {
 			}
 		}
 
-		// In case only non-relevant mutations were recorded it skips the event & force render (#5600).
+		// In case only non-relevant mutations were recorded it skips the event and force render (#5600).
 		if ( viewMutations.length ) {
 			this.document.fire( 'mutations', viewMutations, viewSelection );
 
--- a/src/view/observer/mutationobserver.js
+++ b/src/view/observer/mutationobserver.js
@@ -244,11 +244,14 @@ export default class MutationObserver extends Observer {
 			}
 		}
 
-		this.document.fire( 'mutations', viewMutations, viewSelection );
+		// In case only non-relevant mutations were recorded it skip event / force render for the sake of performance (#5600).
+		if ( viewMutations.length ) {
+			this.document.fire( 'mutations', viewMutations, viewSelection );
 
-		// If nothing changes on `mutations` event, at this point we have "dirty DOM" (changed) and de-synched
-		// view (which has not been changed). In order to "reset DOM" we render the view again.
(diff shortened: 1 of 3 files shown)


src/view/observer/mutationobserver.js
```diff
@@ -244,11 +244,14 @@ export default class MutationObserver extends Observer {
 			}
 		}
 
-		this.document.fire( 'mutations', viewMutations, viewSelection );
+		// In case only non-relevant mutations were recorded it skip event / force render for the sake of performance (#5600).
```
Reinmar:
```suggestion
		// In case only non-relevant mutations were recorded it skips the event & force render (#5600).
```

This is only logic to not do those two things. Improved performance is a side effect.

Reinmar:
I think it's much more complicated to test than what you described, Marek. This PR is about changes that are not part of the editing pipeline, but instead come from the external logic and go directly to UIElements or changes that do not translate into mutations on the view tree. The problem is – it can be pretty much everything and it can happen at random moments because it may be done by the browser.

Because of that, I'd merge this ASAP and just live with it for some time and see.

Reinmar:
Of course, Kacper and Filip, you can play with a couple of things but I think there are many more things that can go wrong so keep your eyes open :)

Reinmar:
One thing's missing – a note in `UIElement#render()` that when you want to update the UI after re-rendering your element, you should call `editor.ui.update()`.

mlewand:
Pushed the changes. I'm not entirely sure though if the wording is right so please double check this.

src/view/uielement.js
```diff
@@ -123,6 +123,9 @@ export default class UIElement extends Element {
 	 *			return domElement;
 	 *		};
 	 *
+	 * If you wish to trigger render of entire editor UI you should call {@link module:core/editor/editorui~EditorUI#update} method
```
Reinmar:
```suggestion
	 * If you wish to trigger render of entire editor UI you should call
	 * the {@link module:core/editor/editorui~EditorUI#update `editor.ui.update()`} method
```

PR #1815, opened 2020-01-23
Introduced Model#createOperationFromJSON

scofalik:

Feature: Introduced `Model#createOperationFromJSON` which is an alias for `OperationFactory#fromJSON`. Closes ckeditor/ckeditor5#6094.

---

### Additional information

I actually used `OperationFactory` inside `Model#createOperationFromJSON` -- less hassle and more backward compatible. If you don't like this solution, I can remove `OperationFactory` and fix related places (I think it was used in three places in our codebase).

### DIFF
--- a/src/model/model.js
+++ b/src/model/model.js
@@ -18,6 +18,7 @@ import ModelElement from './element';
 import ModelRange from './range';
 import ModelPosition from './position';
 import ModelSelection from './selection';
+import OperationFactory from './operation/operationfactory';
 
 import insertContent from './utils/insertcontent';
 import deleteContent from './utils/deletecontent';
@@ -753,6 +754,18 @@ export default class Model {
 	 */
 	createBatch( type ) {
 		return new Batch( type );
+	}
+
+	/**
+	 * Creates an `Operation` instance from a JSON object with the operation data (e.g. taken from a parsed JSON string).
+	 *
+	 * This is an alias for {@link module:engine/model/operation/operationfactory~OperationFactory#fromJSON}.
+	 *
+	 * @param {Object} json Deserialized JSON object.
+	 * @returns {module:engine/model/operation/operation~Operation}
+	 */
+	createOperationFromJSON( json ) {
+		return OperationFactory.fromJSON( json, this.document );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


scofalik:
This changes was already tested and accepted by Cloud Services team.

src/model/operation/operationfactory.js
```diff
@@ -37,7 +37,7 @@ operations[ MergeOperation.className ] = MergeOperation;
  */
 export default class OperationFactory {
 	/**
-	 * Creates concrete `Operation` object from deserialized object, i.e. from parsed JSON string.
+	 * Creates an `Operation` instance from a JSON object with the operation data (e.g. taken from a parsed JSON string).
```
Reinmar:
```suggestion
	 * Creates an operation instance from a JSON object (parsed JSON string).
```

src/model/model.js
```diff
@@ -755,6 +756,18 @@ export default class Model {
 		return new Batch( type );
 	}
 
+	/**
+	 * Creates an `Operation` instance from a JSON object with the operation data (e.g. taken from a parsed JSON string).
```
Reinmar:
```suggestion
	 * Creates an operation instance from a JSON object (parsed JSON string).
```

Reinmar:
> I actually used `OperationFactory` inside `Model#createOperationFromJSON` -- less hassle and more backward compatible. If you don't like this solution, I can remove `OperationFactory` and fix related places (I think it was used in three places in our codebase).

What do you think I might not like in using operation factory here? Is there some other option?

Reinmar:
OMG, I realised that it should be `fromJson` everywhere... Similarly to `toHtml/fromHtml`.

Reinmar:
image

🤣

Reinmar:
Fixed in c102fc3c. I fixed the same thing in the merge message before but I didn't notice the same mistake in the docs. Static methods need to be referenced ad `.foo`, not `#foo`.

scofalik:
I used `createOperationFromJSON` because 1) that's how it was used in `OperationFactory`, two, AFAIR, `toJSON` is the name of the method for objects serialization.

I wanted to go with `Json` initially but then I realized it is upper-case in `OperationFactory` already.

scofalik:
> What do you think I might not like in using operation factory here? Is there some other option?

Another option would be to move the whole functionallity to the `Model` class. And not keep separate `OperationFactory`.

PR #1816, opened 2020-01-23
Introduce BoxEdges typedef.

jodator:

Docs: Introduce `BoxEdges` type definition. Closes ckeditor/ckeditor5#6132.

ps.: I've also removed the `TopRifghtBottomLeft` nomenclature from methods and used `BoxEdges` instead. I think that this is way better then previous.

---

### Additional information

* I was annoyed to describe this object over and over again in https://github.com/ckeditor/ckeditor5-table/pull/224.

### DIFF
--- a/src/view/stylesmap.js
+++ b/src/view/stylesmap.js
@@ -887,7 +887,7 @@ function appendStyleValue( stylesObject, nameOrPath, valueOrObject ) {
  */
 
 /**
- * An object describing box edges values.
+ * An object describing box sides values.
  *
  *		const margin = {
  *			top: '1px',
@@ -896,10 +896,10 @@ function appendStyleValue( stylesObject, nameOrPath, valueOrObject ) {
  *			left: '7px'
  *		}
  *
- * @typedef {Object} module:engine/view/stylesmap~BoxEdges
+ * @typedef {Object} module:engine/view/stylesmap~BoxSides
  *
- * @property {String} top Top edge value.
- * @property {String} right Right edge value.
- * @property {String} bottom Bottom edge value.
- * @property {String} left Left edge value.
+ * @property {String} top Top side value.
+ * @property {String} right Right side value.
+ * @property {String} bottom Bottom side value.
+ * @property {String} left Left side value.
  */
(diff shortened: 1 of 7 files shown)


src/view/stylesmap.js
```diff
@@ -885,3 +885,21 @@ function appendStyleValue( stylesObject, nameOrPath, valueOrObject ) {
...
+ * An object describing box edges values.
+ *
+ *		const margin = {
+ *			top: '1px',
+ *			right: '3px',
+ *			bottom: '3px',
+ *			left: '7px'
+ *		}
+ *
+ * @typedef {Object} module:engine/view/stylesmap~BoxEdges
```
oleq: I think it should rather be box **sides** instead. Check out how w3 uses both terms https://www.w3.org/TR/CSS2/box.html (search for both):

### These are for edges:

> The perimeter of each of the four areas (content, padding, border, and margin) is called an "**edge**", so each box has four edges:
> * content **edge** or inner **edge**
> * padding **edge**
> * border **edge**
> ...

image

### These are for **sides** (and IMO this is what we want to say):

> The 'padding' shorthand property sets the padding for all four **sides** while the other padding properties only set their respective side.

> If there is only one component value, it applies to all **sides**. If there are two values, the top and bottom paddings are set to the first value and the right and left paddings are set to the second.

> The 'border-color' property can have from one to four component values, and the values are set on the different sides as for 'border-width'.

### Summary

So **edges** are more like contours **around** the box (containing or not containing the padding, border, etc.) and the box has always 4 **sides** (top, left, bottom, right).

jodator: Thanks! I've looked at MDN docs and saw the edges used mostly. But your explanaition & wording is better. I'll fix that :)

oleq:
What do we do about https://ckeditor.com/docs/ckeditor5/latest/api/module_utils_dom_getborderwidths.html#static-function-getBorderWidths? It should use the same typedef. 

Is it OK if it used the typedef from the engine? Or maybe the typedef should be defined somewhere else?

jodator:
> Is it OK if it used the typedef from the engine? Or maybe the typedef should be defined somewhere else?

IDK to be fair... I'll just write what comes to my mind:

1. it would be nice to have one typedef for that
2. OTOH we cannot introduce circular dependency engine -> utils -> engine
3. I can live with no typedef for utils stuff
4. I've just check and we already use `module:engine/view/observer` in keystrokehandler docs

...

I'd go with a follow up to iron this out (as the keystroke handler looks like in wrong place)

PR #1818, opened 2020-01-24
Introduced DocumentSelection#event:change:marker

scofalik:

Feature: Introduced `DocumentSelection#event:change:marker`. Closes ckeditor/ckeditor5#6133.

### DIFF
--- a/tests/model/documentselection.js
+++ b/tests/model/documentselection.js
@@ -579,7 +579,7 @@ describe( 'DocumentSelection', () => {
 				} );
 			} );
 
-			it( 'selection ranges change does not change selection markers - no markers', () => {
+			it( 'selection ranges change does not change selection markers (no markers)', () => {
 				const spy = sinon.spy();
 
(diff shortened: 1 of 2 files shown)


tests/model/documentselection.js
```diff
@@ -442,6 +442,50 @@ describe( 'DocumentSelection', () => {
 
 			expect( selection.markers.map( marker => marker.name ) ).to.have.members( [ 'marker' ] );
 		} );
+
+		it( 'should fire change:marker event when selection markers change', () => {
```
Reinmar: I don't like that this is a single test. In the code I can see at least 4 or 5 scenarios. `setTo`, `setFocus`, `changed` not being set to `true`, being set to `true` by the first loop and by the second loop. 5 cases.

scofalik: I've split this test into multiple unit tests and added a few more. I hope you will be satisfied ;).

PR #1820, opened 2020-01-28
I/6154: Allow selection on object element (ie. on table cells)

jodator:

Enhancement: Allow selection on object elements. Closes ckeditor/ckeditor5#6154.

---

### Additional information

* This PR introduces required selection post-fixer changes for table selection.

* Those are a minimal engine changes required to allow multi-range selection on object elements (table cells).

* Two tests in [image](https://github.com/ckeditor/ckeditor5-image/pull/341) and [table](https://github.com/ckeditor/ckeditor5-table/pull/229) were required update because a test case:

```html
[]<object></object>
```

was post-fixed by proposed post-fixer to:

```html
[<object></object>]
```

fixed to:

```html
<paragraph>[]</paragraph><object></object>
```

This case might need to be checked wheter it is OK or not.

Sub PRs:

- https://github.com/ckeditor/ckeditor5-image/pull/341
- https://github.com/ckeditor/ckeditor5-table/pull/229


tests/model/utils/selection-post-fixer.js
```diff
@@ -593,13 +596,57 @@ describe( 'Selection post-fixer', () => {
...
+						'</tableRow>' +
+						'<tableRow>' +
+							'[<tableCell><paragraph>c</paragraph></tableCell>]' +
+							'<tableCell><paragraph>d</paragraph></tableCell>' +
+					'</tableRow>' +
+					'</table>'
+				);
+			} );
+
+			it( 'should allow selection on block - limit element', () => {
```
Reinmar: Allow or not? Cause I can see this selection is fixed.

tests/model/utils/selection-post-fixer.js
```diff
@@ -907,7 +954,7 @@ describe( 'Selection post-fixer', () => {
 				expect( getModelData( model ) ).to.equal( '<paragraph>foob[a]r</paragraph>' );
 			} );
 
-			it( 'should not fix #2', () => {
+			it( 'should not fix #2 (inline widget selected)', () => {
```
Reinmar: Consistency: you use parenthesis here. In other tests you use a dash.

tests/model/utils/selection-post-fixer.js
```diff
@@ -1034,31 +1081,52 @@ describe( 'Selection post-fixer', () => {
...
-				// []<paragraph>foo</paragraph>[]<table>...
-				model.change( writer => {
-					writer.setSelection(
-						[
-							writer.createRange( writer.createPositionAt( modelRoot, 0 ) ),
-							writer.createRange( writer.createPositionAt( modelRoot, 1 ) )
-						]
-					);
-				} );
+			it( 'should fix #3 - selection in limit&object element & before object element', () => {
```
Reinmar: That's not a selection before an object element.

src/model/utils/selection-post-fixer.js
```diff
@@ -263,6 +249,31 @@ function checkSelectionOnNonLimitElements( start, end, schema ) {
...
 }
 
+// Returns a minimal non-intersecting array of ranges.
+//
+// @param {Array.<module:engine/model/range~Range>} ranges
+// @returns {Array.<module:engine/model/range~Range>}
+function mergeIntersectingRanges( ranges ) {
+	const nonIntersectingRanges = [];
+
+	// First range will be always de-intersected.
```
Reinmar:
```suggestion
	// First range will always be fine.
```

Reinmar:
From what I can see you mostly removed outdated code and added this single function for fixing intersecting ranges. That sounds fine. I guess we could have some more tests for some less and more ugly scenarios. E.g.:

* a couple of cells selected, but not like now when one's directly next to another,
* some selection around rows – like `[<td>x</td>]<td>x</td>[<td>x</td></tr>]`,

Also, I thought about a case when one range is inside a cell and the other outside. Maybe it should be a general rule that if a single range is an object range, all ranges must be object ranges?

tests/model/utils/selection-post-fixer.js
jodator: I've used parenthesis everywhere as it was used more times then dash.

jodator:
@Reinmar I've added _some_ scenarios and when thinking about other ones it looked like they're covered elsewhere (mostly by the `limit` boundary-crossing tests). One test already spotted some error in the code so that one got fixed.

src/model/utils/selection-post-fixer.js
```diff
@@ -263,6 +249,35 @@ function checkSelectionOnNonLimitElements( start, end, schema ) {
 	return startIsOnBlock || endIsOnBlock;
 }
 
+// Returns a minimal non-intersecting array of ranges.
+//
+// @param {Array.<module:engine/model/range~Range>} ranges
+// @returns {Array.<module:engine/model/range~Range>}
+function mergeIntersectingRanges( ranges ) {
```
jodator: Needs a fix for adjacent ranges (valid ones).

jodator: That must have been some false negative :relieved: - probably due to refactoring in table selection PR and the way manual tests are built.

The two tests that I've added are for the case that "stoped" working for me during the aforementioned refactoring.

It is OK after all :)

jodator:
Self R+ :+1: 

It turned out to be some problems with manual tests during refactoring.

**Review** ;)

Reinmar:
BTW, there's no "Enhancement" type of change yet. I usually use "Other".

PR #1829, opened 2020-03-06
i/6364: Implemented the Range#getContainedElement() method.

oleq:

Feature: Implemented the model `Range#getContainedElement()` method. Closes ckeditor/ckeditor5#6364.

---

### Additional information

**Next up**: A PR to merge that takes advantage of the changes (https://github.com/ckeditor/ckeditor5-table/pull/264)

### DIFF
--- a/src/model/range.js
+++ b/src/model/range.js
@@ -502,8 +502,9 @@ export default class Range {
 	}
 
 	/**
-	 * Returns an element contained by a range. The element will be returned when it is the **only** node
-	 * within the range and **fully–contained** at the same time.
+	 * Returns an {@link module:engine/model/element~Element Element} contained by a range.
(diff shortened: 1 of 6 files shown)


Reinmar:
I'd rather have a symmetry here, between the model and the view. Even if we won't use this method in the view for now, there's a great chance that we or someone else will need it.

src/model/range.js
```diff
@@ -501,6 +501,27 @@ export default class Range {
 		return this.start.getCommonAncestor( this.end );
 	}
 
+	/**
+	 * Returns an element contained by a range. The element will be returned when it is the **only** node
```
Reinmar:
```suggestion
	 * Returns an element contained by the range. The element will be returned when it is the **only** node
```

oleq:
All done and ready for another round of review.

PR #1830, opened 2020-03-09
I/6116

Reinmar:

Other: Introduced support for multi-range selections. Closes ckeditor/ckeditor5#6116.

---

### Additional information

Accompanied with:

* https://github.com/ckeditor/ckeditor5-typing/pull/227
* https://github.com/ckeditor/ckeditor5-table/pull/267

By far the most tricky part of this task. I had to find the right balance between the `deleteContent()` (and hence `insertContent()`) doing reasonable things and overcomplicating them on the cost of maintenance issues and affecting some other cases (mostly: track changes).

By the occam's razor's call I chose the simplest solution – removing code. I removed some quite unnecessary behaviour from `deleteContent()` and hence was able to use the selection's real position in `insertContent()`. 

During the course of action I also stumbled upon a case where the schema's method was leaking out of the limit element and putting selection on a table cell instead. It was visible when pasting a word over multi-cell selection in an *empty* table. I fixed this too.

### DIFF
--- a/src/model/schema.js
+++ b/src/model/schema.js
@@ -672,12 +672,22 @@ export default class Schema {
 
 		let backwardWalker, forwardWalker;
 
+		// Never leave a limit element.
+		const limitElement = position.getAncestors().reverse().find( item => this.isLimit( item ) ) || position.root;
+
 		if ( direction == 'both' || direction == 'backward' ) {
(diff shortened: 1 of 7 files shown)


Reinmar:
Those changes were reviewed already by @scofalik and @oleq. Merging.

PR #1832, opened 2020-03-13
Fixed a crash that was happening in some scenarios when undoing table background change

scofalik:

Fix: Fixed a crash that was happening in some scenarios when undoing table background change. Closes ckeditor/ckeditor5#6265.


Reinmar:
I added integration test in ckeditor/ckeditor5-table@ebbebe4.

PR #1834, opened 2020-03-31
I/6501: Handle intersecting ranges when fixing graveyard selection

jodator:

Fix: Intersecting ranges resulting when fixing graveyard selection no longer breaks the editor. Closes ckeditor/ckeditor5#6501. Closes ckeditor/ckeditor5#6382.

---

### Additional information

PR to merge together: ckeditor/ckeditor5-table#293.

Well, it looks like intersecting ranges will be created on such modifications as in tables: multi-range selection set on table cells and their parent is removed. This would result in many ranges being fixed to the same range (thus intersecting with any other fixed range).

As with other ranges/positions fixes, I am not sure if I have a broad view of this problem. The proposed solution fixes this case and does not break other CKEditor 5 tests buuuut we do not have every case tested.

### DIFF
--- a/tests/model/documentselection.js
+++ b/tests/model/documentselection.js
@@ -1779,7 +1779,7 @@ describe( 'DocumentSelection', () => {
 				expect( selection.getFirstPosition().path ).to.deep.equal( [ 0, 0 ] );
 			} );
 
-			it( 'does not break if multi-range selection is inside text nodes', () => {
+			it( 'handles multi-range selection in a text node by merging it into one range (resulting in collapsed ranges)', () => {
 				const ranges = [
 					new Range( new Position( root, [ 1, 1 ] ), new Position( root, [ 1, 2 ] ) ),
 					new Range( new Position( root, [ 1, 3 ] ), new Position( root, [ 1, 4 ] ) )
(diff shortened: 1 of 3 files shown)


tests/model/documentselection.js
```diff
@@ -1778,6 +1778,63 @@ describe( 'DocumentSelection', () => {
 				// Now it's clear that it's the default range.
 				expect( selection.getFirstPosition().path ).to.deep.equal( [ 0, 0 ] );
 			} );
+
+			it( 'does not break if multi-range selection is inside text nodes', () => {
```
Reinmar, line 1782: "Does not break" does not explain what should actually happen. I prefer explaining what's the expected result.

```suggestion
			it( 'handles multi-range selection in a text node by merging it into one range', () => {
```

Reinmar: BTW, as mentioned in the test description – I'd actually expect the ranges to be merged if they are identical. Would that require a lot of work?

tests/model/documentselection.js
```diff
@@ -1778,6 +1778,63 @@ describe( 'DocumentSelection', () => {
...
+						doc.version
+					)
+				);
+
+				expect( selection.rangeCount ).to.equal( 2 );
+				expect( selection.getFirstPosition().path ).to.deep.equal( [ 1, 1 ] );
+				expect( selection.getLastPosition().path ).to.deep.equal( [ 1, 1 ] );
+			} );
+
+			it( 'does not break if multi-range selection is set on object nodes and resulting ranges will intersect', () => {
```
Reinmar, line 1804: Again – explain what's expected.

Reinmar, line 1804: Again – explain what's expected.

tests/model/documentselection.js
jodator: > I'd actually expect the ranges to be merged if they are identical.

I don't see a need to merging two identical ranges. With intersecting ones I would  have to check.

PR #1836, opened 2020-04-03
I/6529: Inline `is()` checks

jodator:

Other: Inline calls to parent class in models and views `is()` checks to improve editor performance. Closes ckeditor/ckeditor5#6529.

---

### Additional information

That's a lot of changes but fortunately tests were very helpful here.

The API remained the same.

### DIFF
--- a/src/model/element.js
+++ b/src/model/element.js
@@ -117,10 +117,13 @@ export default class Element extends Node {
 	 */
 	is( type, name = null ) {
 		if ( !name ) {
-			return type == 'element' || type == this.name || type == 'node';
-		} else {
-			return type == 'element' && name == this.name;
+			return type === 'element' || type === 'model:element' ||
+				type === this.name || type === 'model:' + this.name ||
+				// From super.is(). This is highly utilised method and cannot call super. See ckeditor/ckeditor5#6529.
+				type === 'node' || type === 'model:node';
 		}
+
+		return name === this.name && ( type === 'element' || type === 'model:element' );
 	}
 
 	/**
--- a/src/model/element.js
+++ b/src/model/element.js
@@ -116,12 +116,10 @@ export default class Element extends Node {
 	 * @returns {Boolean}
 	 */
 	is( type, name = null ) {
-		const cutType = type.replace( /^model:/, '' );
-
 		if ( !name ) {
-			return cutType == 'element' || cutType == this.name || super.is( type );
+			return type == 'element' || type == this.name || type == 'node';
 		} else {
-			return cutType == 'element' && name == this.name;
+			return type == 'element' && name == this.name;
 		}
 	}
 
--- a/src/model/liveposition.js
+++ b/src/model/liveposition.js
@@ -80,7 +80,9 @@ export default class LivePosition extends Position {
 	 * @returns {Boolean}
 	 */
 	is( type ) {
-		return type == 'livePosition' || type == 'model:livePosition' || type == 'position' || type == 'model:position';
+		return type == 'livePosition' || type == 'model:livePosition' ||
+			// From super.is(). This is highly utilised method and cannot call super. See ckeditor/ckeditor5#6529.
+			type == 'position' || type == 'model:position';
 	}
 
 	/**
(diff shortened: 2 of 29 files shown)


src/model/liveposition.js
```diff
@@ -80,7 +80,9 @@ export default class LivePosition extends Position {
 	 * @returns {Boolean}
 	 */
 	is( type ) {
-		return type == 'livePosition' || type == 'model:livePosition' || super.is( type );
+		return type == 'livePosition' || type == 'model:livePosition' ||
```
Reinmar, line 83: I can see that you were replacing `==` with `===` in some places, so let's be consistent and have it in all `is()` methods.

src/model/element.js
```diff
@@ -116,13 +116,14 @@ export default class Element extends Node {
 	 * @returns {Boolean}
 	 */
 	is( type, name = null ) {
-		const cutType = type.replace( /^model:/, '' );
-
 		if ( !name ) {
-			return cutType == 'element' || cutType == this.name || super.is( type );
-		} else {
-			return cutType == 'element' && name == this.name;
+			return type === 'element' || type === 'model:element' ||
```
Reinmar, line 120: One more optimization that you could consider is ordering those conditions in the likelihood of them being executed. You may need to do some stats and it's a subject of change, so we may need to repeat this. But if it yields nice results, it may be worth doing.

src/model/element.js
```diff
@@ -116,13 +116,14 @@ export default class Element extends Node {
 	 * @returns {Boolean}
 	 */
 	is( type, name = null ) {
-		const cutType = type.replace( /^model:/, '' );
-
 		if ( !name ) {
```
Reinmar, line 119: When we talk about micro optimizations `name === null` may be faster.

Reinmar:
:+1:  for the changes. I left two more ideas what could be checked, since we're reviewing this code anyway.

src/model/element.js
jodator: OK, I'll check this.  
  
Side note I've done some quick (I don't have the results saved though) check for the fastest 'false' path. I've counted the checks in the `RootElement` checks and one of the most used was `is( 'text' )` . **But** adding `if( type === 'text' ) return false` (fail fast) turned out to yield be slower than proposed solution in PR. It might be some browser optimization or a simple error in my methodology but this proves that intuition must be followed with tests.

Reinmar:
BTW, another change that I wanted to do for some time: ckeditor/ckeditor5#6547. It can actually be tuned with the order of conditions in `is()` methods.

src/model/element.js
jodator: I don't think that this will yield any significant result. I've done some tests and below are number of executions of some `is` checks in the long semanting + undo scenario:



As you can see there are milions of checks using `is()` methods and most of them checks if element is `text` or `element`. But if you check how much time is spent (self-time) in `is` methods then in the above scenario (~20 seconds of undo execution) they add up to maybe 150 - 200ms:



I don't see how we could optimize those paths reliably. At this stage I would try to

src/model/element.js
jodator: Something doesn't work (maybe `undefined` is passed sometimes).  Tests breaks after suggested change. Looking that this might be to micro optimization to care (looking on above times) I think we shouldn't dig into this more at this time.

jodator:
@Reinmar IMO it is ready to be merged now.

jodator:
> BTW, another change that I wanted to do for some time: [ckeditor/ckeditor5#6547](https://github.com/ckeditor/ckeditor5/issues/6547). It can actually be tuned with the order of conditions in `is()` methods.

Let's do this in ckeditor/ckeditor5#6547 (the order check). Maybe other problems will arise and other fixes will be needed. IE simple flags or direct 'type' check.

src/model/liveposition.js
Reinmar: Still not resolved.

Reinmar: I'll handle that, though.

